PR #6, opened 2023-06-14
split: move packages from internal to pkg

achilleas-k:
Move the packages that we intend to import in osbuild-composer out of `internal/` and into `pkg/` to make them public.
Prune any modules, documentation, and test configs that are only relevant to osbuild-composer.

There are still a lot of things that need to be cleaned up, but the changes here are enough to make this work as a dependency of osbuild-composer.  The biggest priority after this is figuring out the testing.  As it is now, all the integration tests will run as before, so only osbuild-composer can run the tests.

EDIT: I updated the `main` branch of this repository to match the current state in osbuild-composer.


supakeen (dismissed):
I've read through this before and really like this finally happening. Let's get it green and merged so we can update the `osbuild-composer`-side and find the real bugs :)

thozza (approved):
Everything looks reasonable from what I can tell...

thozza:
> I removed the gitlab-ci trigger from the workflows but I think it's being picked up from `main`.

I think so as well. Happy to admin-merge this...

achilleas-k:
I opened https://github.com/osbuild/osbuild-composer/pull/3551 on the composer side which vendors the branch from this PR.  Once this is merged, I'll drop the `replace` directive and it should be the same.  We can use that PR to verify that the interface after the split is working.

supakeen (approved):

Issue #10, opened 2023-07-10
Post-split PR migration


achilleas-k:
The following Pull Requests were open in osbuild-composer at the time of the split and modify image definitions.  These PRs should be rebased here.

- [ ] https://github.com/osbuild/osbuild-composer/pull/2830
- [ ] https://github.com/osbuild/osbuild-composer/pull/2840
- [ ] https://github.com/osbuild/osbuild-composer/pull/3030
- [ ] https://github.com/osbuild/osbuild-composer/pull/3146 (partially)
- [ ] https://github.com/osbuild/osbuild-composer/pull/3222
- [x] https://github.com/osbuild/osbuild-composer/pull/3282
- [ ] https://github.com/osbuild/osbuild-composer/pull/3332 (`dnf-json` rewrite)
- [ ] https://github.com/osbuild/osbuild-composer/pull/3447
- [x] https://github.com/osbuild/osbuild-composer/pull/3473
- [ ] https://github.com/osbuild/osbuild-composer/pull/3484
- [x] https://github.com/osbuild/osbuild-composer/pull/3488
- [x] https://github.com/osbuild/osbuild-composer/pull/3518
- [x] https://github.com/osbuild/osbuild-composer/pull/3539
- [x] https://github.com/osbuild/osbuild-composer/pull/3545

(Some of these are old and might not be moved, but I'm listing them anyway so we can decide on them explicitly)

supakeen:
I'll be closing https://github.com/osbuild/osbuild-composer/pull/3545 and re-opening later.

7flying:
Edge vsphere PR moved to this repo: https://github.com/osbuild/images/pull/24

achilleas-k:
iot-simplified-installer and iot-qcow2-image: https://github.com/osbuild/images/pull/791

achilleas-k:
Fedora IoT remote configuration package: #95

achilleas-k:
All done.

PR #11, opened 2023-07-10
Add wsl image type

croissanne:
Adds a wsl image type based on the UBI package set.

The UBI package set isn't entirely identical to the current ubi images (2 or 3 packages differ), mostly because we don't include CRB.

---

Also see https://github.com/osbuild/osbuild-composer/pull/3473 and #10


achilleas-k (commented):
Looks good.  I built the image locally and ran it as a container and it all worked great.
I see partition tables in the image definition though.  Copy-paste leftovers?

achilleas-k (approved):
Nice!  LGTM <3

PR #14, opened 2023-07-11
distro/fedora_ng: add a `fedora ng` distro

supakeen:
This PR is the continuation of https://github.com/osbuild/osbuild-composer/pull/3545. Instead of ripping out `distro/fedora` it introduces a new distro called `fedora_ng` (for **n**ext **g**eneration) which has no packages configured, all configuration is to happen in blueprints.

This is to support building Fedora spins and editions as defined in https://github.com/osbuild/fedora-blueprints
 
Necessary image types for `fedora_ng`:

- [x] Raw (`disk-raw`)
- [x] Live ISO (`iso-live`)

Sidequests:

- [x] Rename `image/live` to `image/disk` and all relevant `liveImage` -> `diskImage` to better describe what it actually does.

Documentation:

- [ ] Write down what the `corePackageSet` is.
- [ ] Write down what each image type adds to the `corePackageSet`.
- [ ] Write down what each image type does to the tree.

Questions:

- Do we need `syslinux-nonlinux` in the package set for `iso-live`? It might be a build package only as we seem to only copy it.


achilleas-k:
I'm very much on board with the idea of having it separate (at least initially), but I wonder if the name would be confusing as it's the _old_ Fedora name?  It's a good name and I don't have a better alternative though.

teg:
What's the plan for when this lands? Will we keep two fedora distros or will one be built on top of the other?

If the forever, why is this done as a separate distro rather than options on the existing one?

supakeen:
> What's the plan for when this lands? Will we keep two fedora distros or will one be built on top of the other?
> 


> What's the plan for when this lands? Will we keep two fedora distros or will one be built on top of the other?
> 
> If the forever, why is this done as a separate distro rather than options on the existing one?

The plan is to eventually have this become `fedora`.

PR #28, opened 2023-07-14
depsolving: mark packages as `user`

supakeen:
When DNF installs packages it marks the reason a package was installed. The available reasons are:

`user`: directly requested by the user (e.g: `dnf install vim`, marks `vim` as `user` and all dependencies as `dependency`).
`group`: top packages in a group directly requested by the user (e.g: `dnf install @core` marks `vim`, `foo`, and `bar` as `group`).

There's also `dependency` and `weak-dependency` which are more self-explanatory.

This PR introduces this concept to `dnf-json` so we can propagate the information to `osbuild` which can then mark packages with the correct reasons after installing them as `rpm`s.

Not marking packages correctly can lead to issues; the default behavior by `dnf` is to have packages unknown to it be marked as `dependency`. This means that when the user installs a package on one of our images (marked as `user`) then removes that package that dependencies of that package might be cleaned up as there are no leaf nodes with `user` or `group` anymore; alternatively packages might *not* be removed when they should.

The elephant in the room are `group` markings. When a package is marked as `group` it also marks for *which* group. This information is currently unavailable in both `dnf` and `dnf5`. I've got an open report; currently I've chosen to mark packages that are `group` as `user`. This isn't ideal as one can't do a `dnf remove @core` and have it remove that comps packages however this will let us put this in end-to-end. I've written this PR in such a way that if/when we have a way to get this information only `dnf-json` needs to be adjusted.

A fun complication is that a package can be marked as `dependency` in the first resolve of a transaction chain; however it can then be directly selected (`user`) in the second transaction chain. This conditional reason-upgrade is allowed (and should also be applied to `group`). Similarly, we also allow upgrading `weak-dependency` to `dependency`.


supakeen:
In the current state, this would add a stage like so for (for example) `fedora-39` with `vmdk` image type:

```
+          {
+            "type": "org.osbuild.dnf.mark",
+            "options": {
+              "packages": [
+                {
+                  "name": "dracut-config-generic",
+                  "mark": "user"
+                },
+                {
+                  "name": "grub2-pc",
+                  "mark": "user"
+                },
+                {
+                  "name": "dracut-config-generic",
+                  "mark": "user"
+                },
+                {
+                  "name": "efibootmgr",
+                  "mark": "user"
+                },
+                {
+                  "name": "grub2-efi-x64",
+                  "mark": "user"
+                },
+                {
+                  "name": "shim-x64",
+                  "mark": "user"
+                },
+                {
+                  "name": "kernel",
+                  "mark": "user"
+                },
+                {
+                  "name": "dosfstools",
+                  "mark": "user"
+                },
+                {
+                  "name": "e2fsprogs",
+                  "mark": "user"
+                },
+                {
+                  "name": "selinux-policy-targeted",
+                  "mark": "user"
+                },
+                {
+                  "name": "@Fedora Cloud Server",
+                  "mark": "user"
+                },
+                {
+                  "name": "chrony",
+                  "mark": "user"
+                },
+                {
+                  "name": "systemd-udev",
+                  "mark": "user"
+                },
+                {
+                  "name": "langpacks-en",
+                  "mark": "user"
+                },
+                {
+                  "name": "open-vm-tools",
+                  "mark": "user"
+                }
+              ]
+            }
+          },
```

supakeen:
This PR has been updated to pull the information from `libdnf` in `dnf-json` and to then propagate it down into `rpmmd` which then ends up in the `os` manifest.

Do we want to drop the mark `dependency`? It's the default but I *do* like to be explicit.

supakeen:
This PR has been updated to work around problems with chain-depsolving marking packages as `user` installed when selected from the previous transaction.

supakeen:
I've marked this PR ready for review as I'd really like some input. This will initially generate a dnf marking stage for the disk image types (which all use `liveImage`) for Fedora rawhide.

Note that this PR cannot be merged before the actual `org.osbuild.dnf.mark` stage has been merged into `osbuild`: https://github.com/osbuild/osbuild/pull/1333 which is my next step.

supakeen:
I realized that I forgot a thing. A package *can* be conditionally upgraded to `user`-selected if it was selected directly in a newer transaction. Added some code for that.

`group`-selected can also upgrade package reasons; but we don't have that information yet.

internal/dnfjson/dnfjson_test.go
```diff
@@ -383,113 +383,113 @@ func TestMakeDepsolveRequest(t *testing.T) {
...
+		rpmmd.PackageSpec{Name: "xz-libs", Epoch: 0x0, Version: "5.2.5", Release: "7.el9", Arch: "x86_64", RemoteLocation: "%s/Packages/xz-libs-5.2.5-7.el9.x86_64.rpm", Checksum: "sha256:770819da28cce56e2e2b141b0eee1694d7f3dcf78a5700e1469436461399f001", Secrets: "", CheckGPG: false, IgnoreSSL: true, Reason: "dependency", ReasonGroup: ""},
+		rpmmd.PackageSpec{Name: "zlib", Epoch: 0x0, Version: "1.2.11", Release: "31.el9", Arch: "x86_64", RemoteLocation: "%s/Packages/zlib-1.2.11-31.el9.x86_64.rpm", Checksum: "sha256:1c59b113fda8863e9066cc5d01c6d00bd9c50c4650e1c5b932082c8886e185d1", Secrets: "", CheckGPG: false, IgnoreSSL: true, Reason: "dependency", ReasonGroup: ""},
+		rpmmd.PackageSpec{Name: "zsh", Epoch: 0x0, Version: "5.8", Release: "7.el9", Arch: "x86_64", RemoteLocation: "%s/Packages/zsh-5.8-7.el9.x86_64.rpm", Checksum: "sha256:133da157fbd2b43e4a41af3ba7bb5267cf9ebed0aaf8124a76e5eca948c37572", Secrets: "", CheckGPG: false, IgnoreSSL: true, Reason: "user", ReasonGroup: ""},
```
achilleas-k, line 492: Why did the epoch change to hex for the test data?  Did you dump the result and copy it back in?
Not an issue, though.  Safe to ignore.

pkg/manifest/os.go
```diff
@@ -317,7 +319,7 @@ func (p *OS) getContainerSpecs() []container.Spec {
 }
 
 func (p *OS) serializeStart(packages []rpmmd.PackageSpec, containers []container.Spec, commits []ostree.CommitSpec) {
-	if len(p.packageSpecs) > 0 {
+	if len(p.getPackageSpecs()) > 0 {
```
achilleas-k, line 322: Why did this change?  Probably fine or even better this way, but curious.

achilleas-k (commented):
This is all great.  Thanks for this.

Couple of things:
- I'd like to see this broken down into multiple commits to make it easier to re-review and track down in the future.  The addition of the osbuild stage could be separate for example.  The change in `serialize()` from `p.packageSpecs` to `p.getPackageSpecs()` could also be separate with its own commit explaining the change.  The changes to `dnf-json` can also be separate with most of the current commit message being applied there.
- What's the status of the osbuild stage?  The stage in the open PR is just a schema so far.  Let's get that in so we can start testing the bits from here.

achilleas-k:
I've been cancelling the builds for this PR in CI because somehow it's triggering a rebuild of everything.  It shouldn't because even though the rpm metadata in the test cases changed, it shouldn't affect the manifests themselves.  The pipeline generator is only looking at the manifest data (not the metadata), so I'm not sure what's happening exactly.

pkg/distro/fedora/images.go
```diff
@@ -223,6 +223,13 @@ func liveImage(workload workload.Workload,
 
 	img.Filename = t.Filename()
 
+	distro := t.Arch().Distro()
+
+	// We start marking packages in the dnf state database in Fedora 39 and up
+	if !common.VersionLessThan(distro.Releasever(), "39") {
+		img.MarkPackages = true
+	}
```
thozza, line 231: Do we really need this? I think that we can consider not marking packages being a bug, no? I mean, if this can't break users, let's do it for all versions :thinking:

internal/dnfjson/dnfjson_test.go
```diff
@@ -383,113 +383,113 @@ func TestMakeDepsolveRequest(t *testing.T) {
...
-		{Name: "xz-libs", Epoch: 0, Version: "5.2.5", Release: "7.el9", Arch: "x86_64", RemoteLocation: "%s/Packages/xz-libs-5.2.5-7.el9.x86_64.rpm", Checksum: "sha256:770819da28cce56e2e2b141b0eee1694d7f3dcf78a5700e1469436461399f001", Secrets: "", CheckGPG: false, IgnoreSSL: true},
-		{Name: "zlib", Epoch: 0, Version: "1.2.11", Release: "31.el9", Arch: "x86_64", RemoteLocation: "%s/Packages/zlib-1.2.11-31.el9.x86_64.rpm", Checksum: "sha256:1c59b113fda8863e9066cc5d01c6d00bd9c50c4650e1c5b932082c8886e185d1", Secrets: "", CheckGPG: false, IgnoreSSL: true},
-		{Name: "zsh", Epoch: 0, Version: "5.8", Release: "7.el9", Arch: "x86_64", RemoteLocation: "%s/Packages/zsh-5.8-7.el9.x86_64.rpm", Checksum: "sha256:133da157fbd2b43e4a41af3ba7bb5267cf9ebed0aaf8124a76e5eca948c37572", Secrets: "", CheckGPG: false, IgnoreSSL: true},
+		rpmmd.PackageSpec{Name: "acl", Epoch: 0x0, Version: "2.3.1", Release: "3.el9", Arch: "x86_64", RemoteLocation: "%s/Packages/acl-2.3.1-3.el9.x86_64.rpm", Checksum: "sha256:986044c3837eddbc9231d7be5e5fc517e245296978b988a803bc9f9172fe84ea", Secrets: "", CheckGPG: false, IgnoreSSL: true, Reason: "dependency", ReasonGroup: ""},
```
thozza, line 386: Any reason why the `Epoch` is now a hexadecimal number? :thinking: I think it should be kept decimal (for consistency), since we IIRC use it in some places to construct package NEVRA.

thozza (commented):
Nice work, I'm amazed how clean is the final solution!

I'm wondering if there is some good reason to not mark packages in all images on all distros by default? I think that we can consider this to be a bug, although an intentional one... But I can't imagine any situation when this would break users, but maybe I just don't have good imagination :smile:

internal/dnfjson/dnfjson_test.go
supakeen: Oh, it's because I copy/pasted the 'expected' result (initially all `Reason` and `ReasonGroup` fields were empty). I think that's how Go represents it. I'll fix that!

thozza: I just noticed that each slice entry has `rpmmd.PackageSpec` type, which is not needed and should be removed.

supakeen: I've addressed the integer literals and the implicit `rpmmd.PackageSpec`.

pkg/distro/fedora/images.go
supakeen: Since it changes behavior of future images I'd prefer not to do it for current versions but I also don't hold that opinion very strongly. Enabling it on everything is fine by me.

pkg/manifest/os.go
supakeen: It changed because we were being inconsistent; some methods accessed `.packageSpecs`, some called `.getPackageSpecs()`.

internal/dnfjson/dnfjson_test.go
supakeen: I did. @thozza pointed it out as well and I've converted it back to decimal for the literal.

supakeen:
@achilleas-k @thozza I've split up this PR into 3 commits.

1. fix packageSpec
2. introduce dnf marking stage
3. dnf-json + images code

pkg/distro/fedora/images.go
thozza: @ondrejbudai WDYT?

thozza (dismissed):
The PR looks good from my PoV.

Enabling this for all images can be done in a follow-up. In addition, please make sure to bump the required osbuild version with the support for `dnf_mark` in the `osbuild-composer.spec`...

achilleas-k (changes requested):
The osbuild stage isn't ready yet.

pkg/osbuild/dnf_mark_stage.go
```diff
@@ -0,0 +1,56 @@
...
+	return nil
+}
+
+func NewDNFMarkStageOptions(packages []DNFMarkStagePackageOptions) *DNFMarkStageOptions {
+	return &DNFMarkStageOptions{
+		Packages: packages,
+	}
+}
+
+type DNFMarkStage struct {
```
ondrejbudai, line 29: I have a feeling that this isn't used anywhere.

pkg/osbuild/dnf_mark_stage.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func NewDNFMarkStage(options *DNFMarkStageOptions) *Stage {
+	if err := options.validate(); err != nil {
+		panic(err)
+	}
+
+	return &Stage{
+		Type:    "org.osbuild.dnf.mark",
```
ondrejbudai, line 38: The corresponding osbuild's PR introduces `dnf3` and `dnf5` stages, but no `dnf` stage, so this needs a change.

supakeen: Yea, this PR will need to be adjusted again to implement both osbuild stages and logic to select the correct one after we decided to split the osbuild stage :)

supakeen:
More changes will be required:
- [ ] Selecting dnf4 or dnf5 based on the imagetype (or packageset, or requested packages, not sure yet)
- [ ] In the osbuild PR I decided to follow the dnf command line interface more closely, that means that the dnf4 stage accepts `install`, `group` as marks and the dnf5 stage accepts `user`, `weak`, `dependency`, and `group`. I'm considering having `dnf-json` output the `dnf5`-terms and then translating it in stage options.

supakeen:
Ok, there is now a testcase that verifies a package is marked correctly. I've removed the `dnf5` parts as well, only `dnf4` here.

teg:
> More changes will be required:
> - [ ] Selecting dnf4 or dnf5 based on the imagetype (or packageset, or requested packages, not sure yet)
> - [ ] In the osbuild PR I decided to follow the dnf command line interface more closely, that means that the dnf4 stage accepts `install`, `group` as marks and the dnf5 stage accepts `user`, `weak`, `dependency`, and `group`. I'm considering having `dnf-json` output the `dnf5`-terms and then translating it in stage options here.

As dnf5 has been delayed until F41, we might make things easier for ourselves by doing dnf4 only at first.

supakeen:
> > More changes will be required:
> > 
> > * [ ]  Selecting dnf4 or dnf5 based on the imagetype (or packageset, or requested packages, not sure yet)
> > * [ ]  In the osbuild PR I decided to follow the dnf command line interface more closely, that means that the dnf4 stage accepts `install`, `group` as marks and the dnf5 stage accepts `user`, `weak`, `dependency`, and `group`. I'm considering having `dnf-json` output the `dnf5`-terms and then translating it in stage options here.
> 
> As dnf5 has been delayed until F41, we might make things easier for ourselves by doing dnf4 only at first.

Funny how that goes. I had just removed dnf5 from the equation in the osbuild-PR and doing the same here.

This PR now generates a valid manifest, understood by osbuild with the `org.osbuild.dnf4.mark` stage; output seems to be correct (this is the `minimal-raw` image for Fedora 38).

Note that I had to build Fedora 38 because Fedora 39 in our snapshots still has `dnf5` as the default.


```
user@muja images € dnf --installroot=/opt/build/os repoquery --installed --qf='%{name},%{reason}' | grep -v unknown
NetworkManager,user
audit,user
basesystem,user
bash,user
coreutils,user
curl,user
dhcp-client,user
dnf,user
dnf-plugins-core,user
dosfstools,user
dracut-config-generic,user
dracut-config-rescue,user
e2fsprogs,user
efibootmgr,user
fedora-repos-modular,user
filesystem,user
firewalld,user
fwupd,user
glibc,user
grub2-efi-x64,user
grubby,user
hostname,user
iproute,user
iputils,user
kbd,user
kernel,user
less,user
man-db,user
ncurses,user
openssh-clients,user
openssh-server,user
parted,user
passwd,user
plymouth,user
policycoreutils,user
procps-ng,user
rootfiles,user
rpm,user
selinux-policy-targeted,user
setup,user
shadow-utils,user
shim-x64,user
sssd-kcm,user
sudo,user
systemd,user
systemd-resolved,user
util-linux,user
vim-minimal,user
yum,user
zram-generator-defaults,user
```

teg:
Love it!

pkg/osbuild/dnf_mark_stage.go
```diff
@@ -0,0 +1,56 @@
+package osbuild
+
+import (
+	"github.com/osbuild/images/pkg/rpmmd"
+)
+
+type DNFMarkStagePackageOptions struct {
+	Name  string `json:"name"`
+	Mark  string `json:"mark"`
+	Group string `json:"group,omitempty"`
```
ondrejbudai, line 10: I think that the latest version of the `osbuild` PR doesn't have `group` in the stage options.

pkg/runner/fedora.go
```diff
@@ -15,5 +15,6 @@ func (p *Fedora) GetBuildPackages() []string {
 		"glibc",   // ldconfig
 		"systemd", // systemd-tmpfiles and systemd-sysusers
 		"python3", // osbuild
+		"dnf",     // temporary
```
ondrejbudai, line 18: temporary? What does this mean?

pkg/distro/fedora/images.go
ondrejbudai: I'm happy to treat it as a regression, so enable it also for older Fedoras.

pkg/runner/fedora.go
supakeen: Ah yes, good that this was caught. I am wondering if it's correct to add `dnf` to the build packages here.

ondrejbudai: @achilleas-k wdyt?

achilleas-k: If we're going to be marking packages for all fedora builds, then we'll need it in the build root, yes.

supakeen:
PR rebased and updated both concerns addressed and the comment changed to clarify that the package is added to the build root for marking.

Is it possible to have a stage conditionally add something to the build root?

thozza:
> Is it possible to have a stage conditionally add something to the build root?

Not stage, but the pipeline implementation can do it based on its configuration which also affects if a stage is added to the pipeline or not.

See https://github.com/osbuild/images/blob/main/pkg/manifest/os.go#L262

achilleas-k:
> > Is it possible to have a stage conditionally add something to the build root?
> 
> Not stage, but the pipeline implementation can do it based on its configuration which also affects if a stage is added to the pipeline or not.
> 
> See https://github.com/osbuild/images/blob/main/pkg/manifest/os.go#L262

Since we have the option set on the `os`, and the stage isn't added for all versions, best to add the package to the build root conditionally like we do for everything else.

supakeen:
Nice; then I've moved DNF from the runner into the getBuildPackages for the os pipeline :)

pkg/distro/fedora/images.go
```diff
@@ -252,6 +252,10 @@ func diskImage(workload workload.Workload,
 
 	img.Filename = t.Filename()
 
+	distro := t.Arch().Distro()
+
+	img.MarkPackages = true
```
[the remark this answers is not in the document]
thozza: In addition, package marking is enabled for all Fedora version AFAICT, which contradicts the commit message which states that it should be enabled only for rawhide.

thozza (commented):
Overall nice work. I added a few inline comments. Most notably, I don't think that putting the option into `DiskImage` is the right place. Instead, it should be part of the `ImageConfig` which then translates to `OSCustomizations`... Distro / image type - specific `ImageConfig` can be then used in Fedora to enable this feature where needed.

pkg/osbuild/dnf_mark_stage.go
```diff
@@ -0,0 +1,64 @@
+package osbuild
```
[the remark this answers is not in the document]
supakeen: Oh, I wanted to put both the dnf4 and dnf5 stage in `dnf_mark_stage.go`. I'll rename it to be specific since it only has 4 for now.

pkg/image/disk.go
```diff
@@ -25,6 +25,7 @@ type DiskImage struct {
 	Workload         workload.Workload
 	Filename         string
 	Compression      string
+	MarkPackages     bool
```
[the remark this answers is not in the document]
supakeen: I've moved this into `ImageConfig` and then set it to true by default in `fedora/images.go`.

pkg/distro/fedora/images.go
```diff
@@ -50,6 +50,10 @@ func osCustomizations(
 	osc.ExcludeBasePackages = osPackageSet.Exclude
 	osc.ExtraBaseRepos = osPackageSet.Repositories
 
+	if !t.rpmOstree {
+		osc.MarkPackages = true
+	}
```
thozza, line 55: I think that the correct way would be to unconditionally set `osc.MarkPackages = imageConfig.MarkPackages`.

Then set to `true` in https://github.com/osbuild/images/blob/d9e892f0d7f7063a71862f6a20ca152a9d68442f/pkg/distro/fedora/distro.go#L404-L407 and overridden for edge images. There may be probably a more elegant way, but the thing is that now, the value can't be actually configured via `ImageConfig`.

In addition, it may make sense to extend the `osCustomizations()` in each distro even if not used there, to prevent any surprises in the future (since `ImageConfig` and OS pipeline implementation are used globally).

PR #37, opened 2023-07-20
Build ostree images that depend on commits

achilleas-k:
This PR adds image building to CI for image types that require pulling an ostree commit.  The whole process of how this works is described in the update test README.

Copying the description of the relevant stage here:
The new stage of the workflow runs the `./test/generate-ostree-build-config` script. It has the same purpose as the config generator for the regular builds, but it sets up ostree containers to serve commits to generate manifests for the image types that depend on them.

The config generator:
- Generates all the manifests for build config dependencies for a given architecture using the `./cmd/gen-manifests` tool.
  - Build config dependencies are image type and config pairings that appear in the `depends` part of a build config .
  - For example `iot-ostree-pull-empty` will cause a manifest to be generated for `iot-container` with the `empty` config for all distros.
- Determines the container name and tag from the build name and manifest ID and pulls each container from the registry.
- Runs each container with a unique port mapped to the internal web service port.
- For each build config that defines a dependency and for each image that config applies to, creates build configs and a config map that defines the URL, port, and ref for the ostree commit source.
  - For example, the config `iot-ostree-pull-empty` is mapped in the `config-map` to the image types `iot-ami`, `iot-installer`, `iot-raw-image`, and `iot-vsphere`. This will create four configs for each distro, one for each image type, that will all have ostree options to pull an ostree commit from an `iot-container` of the same distro.
- Generates all the manifests defined in the config map that was generated in the previous step.
  - Note that this manifest generation step uses the `-skip-noconfig` flag, which means that any image type not defined in the map is skipped.
- Downloads the test build cache.
- Filters out any manifest with an ID that exists in the build cache.
- For each remaining manifest, creates a build job which runs the ostree container that was used to generate the manifest and runs `./test/cases/build-image.sh` script for a given distro, image type, and config file.
  - If no builds are needed, it generates a `NullConfig`, which is a simple shell runner that exits successfully. This is required because the child pipeline config cannot be empty.


test/generate-build-config
```diff
@@ -35,7 +32,7 @@ def generate_manifests(outputdir):
     Generate all manifest using the default config map and return a dictionary mapping each manifest file to the
     manifest data and its ID.
     """
-    arches_arg = ",".join(ARCHITECTURES)
+    arches_arg = os.uname().machine
```
thozza, line 35: Not sure if a person would run this script locally in some valid case, but doing this will make it impossible to generate the config for other architecture locally.

test/README.md
```diff
@@ -10,4 +10,96 @@ will build a Fedora 38 qcow2 image using the configuration specified in the file
...
+
+for example:
+```json
+{
+  "distro": "fedora-38",
+  "arch": "x86_64",
+  "image-type": "qcow2",
+  "config": "all-customizations",
+  "manifest-checksum": "8c0ce3987d78fe6f3307494cd57ceed861de61c3b04786d6a7f570faacbdb5df",
+  "obuild-version": "osbuild 89",
```
thozza, line 74: Just noticed that the version contains `osbuild`. Why is that so? Wouldn't just the actual version be enough?

test/README.md
```diff
@@ -18,6 +18,12 @@ The config map is also used in CI to dynamically generate test builds using the
...
 
+
+### Dynamic pipelines
+
+Jobs are created dynamically using GitLab CI's [Dynamic child pipelines](https://docs.gitlab.com/ee/ci/pipelines/downstream_pipelines.html#dynamic-child-pipelines) feature. A simple example of how this works that mimics the setup of the pipeline generation in this project (but with very simple bash scripts) can be found in the [achkts/ci-shenanigans](https://gitlab.com/achkts/ci-shenanigans) project on GitLab. The project contains an [annotated `.gitlab-ci.yml` file](https://gitlab.com/achkts/ci-shenanigans/-/blob/5914c7432eaa810cfea7ca35ffb9f01700197b02/.gitlab-ci.yml) and [a couple of bash scripts](https://gitlab.com/achkts/ci-shenanigans/-/tree/5914c7432eaa810cfea7ca35ffb9f01700197b02/scripts) that generate pipeline configurations dynamically.
```
thozza, line 24: Not sure if it is viable to refer to your personal Gitlab account and project. May make more sense to move this somewhere under our organization instead.

test/README.md
achilleas-k: It takes the output of `osbuild --version` directly.  I didn't want to be too smart about it and parse or cut the output since it's just there for troubleshooting purposes (what did we use to build this).  Ideally, we'd have the osbuild commit SHA but I haven't gotten around to that yet (with osbuild pinning).

test/README.md
achilleas-k: Moved it and updated the URLs.

achilleas-k:
Found a bug that only generates one ostree build job for each distro.  Fixing now.

achilleas-k:
> Found a bug that only generates one ostree build job for each distro. Fixing now.

Fixed and amended.
Diff from rebase: https://github.com/osbuild/images/compare/f03bb326e8356ecc1b3f1fcb492594131c359f0f..bc31f9312f5877e5f037300e9db45486780fa801 (in case it gets lost in other force pushes or rebases).

test/README.md
thozza: Thanks for the explanation. It makes sense to keep it as is if this is the output from osbuild itself...

test/generate-build-config
thozza: Sure. I had the impression that this script is not intended to be run by a user, just wanted to double-check... A followup sounds OK to me, but not adding the option if not needed is also OK from my PoV.

thozza (approved):
LGTM FAICT...

PR #39, opened 2023-07-20
depsolving: weak-deps selection

supakeen:
We want to be able to select if weak dependencies should be installed; currently `dnf-json` always enables weak dependencies for the first transaction in its depsolve; let's make this configurable on the packageset(s) (and perhaps even in blueprints).


supakeen:
This initial commit moves the logic of marking the first branch as 'with weak deps' from `dnf-json` into the Go code.

supakeen:
Initially applying this only to Fedora; I expect RHEL image builds to fail on this but the Fedora ones to come out correctly.

internal/dnfjson/dnfjson.go
```diff
@@ -538,6 +539,9 @@ type transactionArgs struct {
 
 	// IDs of repositories to use for this depsolve
 	RepoIDs []string `json:"repo-ids"`
+
+	// If we want weak deps for this depsolve
+	WithWeakDeps bool `json:"with_weak_deps"`
```
ondrejbudai, line 544: I wonder if we should call this `InstallWeakDeps` to match the dnf configuration name. :thinking:

ondrejbudai (approved):
This is pretty amazing, tbh! :)

internal/dnfjson/dnfjson.go
supakeen: Done :)

supakeen:
I've moved setting the `InstallWeakDeps` property into the `os` pipeline. This should ensure that no images changed.

achilleas-k:
Good idea.  It'll be nice to be able to control this from the image definitions.

We need to add `InstallWeakDeps = true` to all static package sets as well now right (the ones defined in `pkg/distro/*/package_sets.go`)?
And the dynamic build package sets (`getPackageSetChain()` for the build pipelines).

supakeen:
> Good idea. It'll be nice to be able to control this from the image definitions.
> 
> We need to add `InstallWeakDeps = true` to all static package sets as well now right (the ones defined in `pkg/distro/*/package_sets.go`)? And the dynamic build package sets (`getPackageSetChain()` for the build pipelines).

So that was the route I initially took but it gets really murky really quickly as PackageSets are often appended to eachother and you might want PackageSet A to not have weak deps enabled and PackageSet B to *have* them enabled.

For that reason I made it so you *don't* do that and the `os` pipeline which puts together all the architecture/etc-dependent packages instead marks the first merged packageset as 'InstallWeakDeps'.

achilleas-k:
We discussed this outside the comments so I'll summarise here for the record.

> > Good idea. It'll be nice to be able to control this from the image definitions.
> > We need to add `InstallWeakDeps = true` to all static package sets as well now right (the ones defined in `pkg/distro/*/package_sets.go`)? And the dynamic build package sets (`getPackageSetChain()` for the build pipelines).
> 
> So that was the route I initially took but it gets really murky really quickly as PackageSets are often appended to eachother and you might want PackageSet A to not have weak deps enabled and PackageSet B to _have_ them enabled.
> 

In that case, these would be two package sets in the same chain, which is what we have already when we depsolve the base OS packages and then the user-selected ones.

> For that reason I made it so you _don't_ do that and the `os` pipeline which puts together all the architecture/etc-dependent packages instead marks the first merged packageset as 'InstallWeakDeps'.

The current behaviour makes the most sense.  Say we start with a package set for a given purpose (build package set) and we know we want weak deps, so we set it to true.  Anything we append to it after that can come from anywhere, but the primary purpose remains.  So we should keep the logic of the `Append()` method the way it is now and document it in the method docstring.
Any caller that wants to make sure the package set has the right setting can always set the flag to whatever's needed after the append (or at the end of multiple appends), so I think it's okay to define one behaviour in the docstring and let callers flip it if they need to.

supakeen:
Wooo, it's green. Note the important bit: no images were built which means that there were no manifest changes with the code.

pkg/distro/fedora/distro.go
```diff
@@ -124,7 +124,7 @@ var (
 		packageSets: map[string]packageSetFunc{
 			osPkgsKey: iotCommitPackageSet,
 			containerPkgsKey: func(t *imageType) rpmmd.PackageSet {
-				return rpmmd.PackageSet{}
+				return rpmmd.PackageSet{InstallWeakDeps: false}
```
thozza, line 127: nitpick: you don't need to set the option to `false`, since it is the default value for the type...

thozza (dismissed):
LGTM.

there is one cosmetic thing, but it can be changed in a follow up.

pkg/distro/fedora/distro.go
supakeen: Done :)

thozza (approved):
:+1:

achilleas-k (approved):
Thanks!

PR #41, opened 2023-07-21
github: auto-approve and auto-merge PRs from dependabot

ondrejbudai:
If the tests pass, I don't see a reason why not to just merge them.


achilleas-k (approved):
Would've thought there would be a dependabot option for this.
Good idea! ❤️

PR #43, opened 2023-07-21
OpenSCAP autotailoring stage

kingsleyzissou:
This pull request adds autotailoring support for openscap customizations.
These customizations allow users to override specific OpenSCAP rules which
are passed to the autotailor stage which is responsible for creating the tailoring
file.

This pull request includes:

- [ ] adequate testing for the new functionality or fixed issue
- [ ] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

### DIFF
--- a/internal/oscap/oscap.go
+++ b/internal/oscap/oscap.go
@@ -1,7 +1,11 @@
 package oscap
 
 import (
+	"fmt"
+	"path/filepath"
 	"strings"
+
+	"github.com/osbuild/images/internal/fsnode"
 )
 
 type Profile string
@@ -35,6 +39,9 @@ const (
 	defaultCentos9Datastream string = "/usr/share/xml/scap/ssg/content/ssg-cs9-ds.xml"
 	defaultRHEL8Datastream   string = "/usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml"
 	defaultRHEL9Datastream   string = "/usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml"
+
+	// tailoring directory path
+	tailoringDirPath string = "/usr/share/xml/osbuild-openscap-data"
 )
 
 func DefaultFedoraDatastream() string {
@@ -69,4 +76,16 @@ func IsProfileAllowed(profile string, allowlist []Profile) bool {
 	}
 
 	return false
+}
+
+func GetTailoringFile(profile string) (string, string, *fsnode.Directory, error) {
+	newProfile := fmt.Sprintf("%s_osbuild_tailoring", profile)
+	path := filepath.Join(tailoringDirPath, "tailoring.xml")
+
+	tailoringDir, err := fsnode.NewDirectory(tailoringDirPath, nil, nil, nil, true)
+	if err != nil {
+		return "", "", nil, err
+	}
+
+	return newProfile, path, tailoringDir, nil
 }
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -118,22 +118,6 @@ func osCustomizations(
 		osc.SElinux = "targeted"
 	}
 
-	if oscapConfig := c.GetOpenSCAP(); oscapConfig != nil {
-		if t.rpmOstree {
-			panic("unexpected oscap options for ostree image type")
-		}
-		var datastream = oscapConfig.DataStream
-		if datastream == "" {
-			datastream = oscap.DefaultFedoraDatastream()
-		}
-		osc.OpenSCAPConfig = osbuild.NewOscapRemediationStageOptions(
-			osbuild.OscapConfig{
-				Datastream: datastream,
-				ProfileID:  oscapConfig.ProfileID,
-			},
-		)
-	}
-
 	var err error
 	osc.Directories, err = blueprint.DirectoryCustomizationsToFsNodeDirectories(c.GetDirectories())
 	if err != nil {
@@ -172,6 +156,49 @@ func osCustomizations(
 
 	for filename, repos := range yumRepos {
 		osc.YUMRepos = append(osc.YUMRepos, osbuild.NewYumReposStageOptions(filename, repos))
+	}
+
+	if oscapConfig := c.GetOpenSCAP(); oscapConfig != nil {
+		if t.rpmOstree {
+			panic("unexpected oscap options for ostree image type")
+		}
+		var datastream = oscapConfig.DataStream
+		if datastream == "" {
+			datastream = oscap.DefaultFedoraDatastream()
+		}
+
+		oscapStageOptions := osbuild.OscapConfig{
+			Datastream: datastream,
+			ProfileID:  oscapConfig.ProfileID,
+		}
+
+		if oscapConfig.Tailoring != nil {
+			newProfile, tailoringFilepath, tailoringDir, err := oscap.GetTailoringFile(oscapConfig.ProfileID)
+			if err != nil {
+				panic(fmt.Sprintf("unexpected error creating tailoring file options: %v", err))
+			}
+
+			tailoringOptions := osbuild.OscapAutotailorConfig{
+				Selected:   oscapConfig.Tailoring.Selected,
+				Unselected: oscapConfig.Tailoring.Unselected,
+				NewProfile: newProfile,
+			}
+
+			osc.OpenSCAPTailorConfig = osbuild.NewOscapAutotailorStageOptions(
+				tailoringFilepath,
+				oscapStageOptions,
+				tailoringOptions,
+			)
+
+			// overwrite the profile id with the new tailoring id
+			oscapStageOptions.ProfileID = newProfile
+			oscapStageOptions.Tailoring = tailoringFilepath
+
+			// add the parent directory for the tailoring file
+			osc.Directories = append(osc.Directories, tailoringDir)
+		}
+
+		osc.OpenSCAPConfig = osbuild.NewOscapRemediationStageOptions(oscapStageOptions)
 	}
 
 	osc.ShellInit = imageConfig.ShellInit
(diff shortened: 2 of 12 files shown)


test/configs/oscap-customizations.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "oscap-customizations",
```
kingsleyzissou, line 2: RHEL7 and some of the earlier RHEL8 versions don't support openscap customizations... any ideas on how I could run these tests conditionally?

internal/oscap/oscap.go
```diff
@@ -35,6 +38,9 @@ const (
 	defaultCentos9Datastream string = "/usr/share/xml/scap/ssg/content/ssg-cs9-ds.xml"
 	defaultRHEL8Datastream   string = "/usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml"
 	defaultRHEL9Datastream   string = "/usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml"
+
+	// autotailor directory path
+	autotailorDirPath string = "/usr/share/osbuild"
```
thozza, line 43: I'm no expert on SCAP, but wouldn't it make sense to use something like `/usr/share/xml/scap/osbuild` to make it obvious what is this directory for? :thinking:

pkg/manifest/os.go
```diff
@@ -93,35 +93,36 @@ type OSCustomizations struct {
...
+	Sysctld                  []*osbuild.SysctldStageOptions
+	DNFConfig                []*osbuild.DNFConfigStageOptions
+	DNFAutomaticConfig       *osbuild.DNFAutomaticConfigStageOptions
+	YUMConfig                *osbuild.YumConfigStageOptions
+	YUMRepos                 []*osbuild.YumReposStageOptions
+	SshdConfig               *osbuild.SshdConfigStageOptions
+	GCPGuestAgentConfig      *osbuild.GcpGuestAgentConfigOptions
+	AuthConfig               *osbuild.AuthconfigStageOptions
+	PwQuality                *osbuild.PwqualityConfStageOptions
+	OpenSCAPAutotailorConfig *osbuild.OscapAutotailorStageOptions
```
thozza, line 118: This is not needed, since the tailoring information is already part of `OpenSCAPConfig`

pkg/manifest/os.go
```diff
@@ -726,6 +727,17 @@ func (p *OS) serialize() osbuild.Pipeline {
 		}))
 	}
 
+	if p.OpenSCAPConfig != nil {
+		// nest this within the openscap config, since
+		// it doesn't make sense to have tailoring configs
+		// without openscap config
+		if p.OpenSCAPAutotailorConfig != nil {
```
thozza, line 734: You can use `p.OpenSCAPConfig.Tailoring` instead

thozza: Also, shouldn't you call `GetAutotailorFile()` here and append dirs and files to the list?

internal/oscap/oscap.go
thozza: I would check with the OSCAP team...

pkg/manifest/os.go
kingsleyzissou: The OpenSCAP autotailoring config object isn't in the OpenSCAPConfig though... the only thing that is saved there is the `filepath` to the tailoring file. So basically the two stages depend on each other here, but not completely.

pkg/manifest/os.go
thozza: Right, my bad. I didn't check everything thoroughly and switched the osbuild stage options for the BP customizations structure.... :facepalm:

pkg/manifest/os.go
thozza: fair point... It is fine as is...

internal/oscap/oscap.go
```diff
@@ -70,3 +76,15 @@ func IsProfileAllowed(profile string, allowlist []Profile) bool {
 
 	return false
 }
+
+func GetAutotailorFile(profile string) (string, string, *fsnode.Directory, error) {
+	newProfile := fmt.Sprintf("%s_osbuild_tailoring", profile)
+	filepath := fmt.Sprintf("%s/%s", autotailorDirPath, "tailoring.xml")
```
achilleas-k, line 82: Not very important, but `filepath.Join()` would be nicer here for avoiding potential double `//` if the `autotailorDirPath` happens to have a trailing `/`.

pkg/manifest/build.go
```diff
@@ -45,6 +45,7 @@ func (p *Build) addDependent(dep Pipeline) {
 func (p *Build) getPackageSetChain(distro Distro) []rpmmd.PackageSet {
 	// TODO: make the /usr/bin/cp dependency conditional
 	// TODO: make the /usr/bin/xz dependency conditional
+	// TODO: make the openscap-utils package conditional
```
achilleas-k, line 48: Is `openscap-utils` going to be included here?

pkg/manifest/os.go
```diff
@@ -726,6 +727,17 @@ func (p *OS) serialize() osbuild.Pipeline {
 		}))
 	}
 
+	if p.OpenSCAPConfig != nil {
+		// nest this within the openscap config, since
+		// it doesn't make sense to have tailoring configs
+		// without openscap config
```
achilleas-k, line 733: Should tailoring without openscap be an error?

pkg/distro/fedora/images.go
```diff
@@ -174,6 +158,49 @@ func osCustomizations(
...
+
+		oscapStageOptions := osbuild.OscapConfig{
+			Datastream: datastream,
+			ProfileID:  oscapConfig.ProfileID,
+		}
+
+		if oscapConfig.Tailoring != nil {
+			newProfile, tailoringFilepath, tailoringDir, err := oscap.GetAutotailorFile(oscapConfig.ProfileID)
+			if err != nil {
+				panic("unexpected error creating autotailoring file options")
```
achilleas-k, line 178: It'd be good to print the error (`err`) here in the panic message for easier troubleshooting.

pkg/distro/rhel9/images.go
```diff
@@ -136,14 +136,41 @@ func osCustomizations(
 		}
 		var datastream = oscapConfig.DataStream
 		if datastream == "" {
-			datastream = oscap.DefaultRHEL9Datastream(t.arch.distro.isRHEL())
+			datastream = oscap.DefaultFedoraDatastream()
```
achilleas-k, line 139: Is this right?
Copy-paste error from Fedora?

test/configs/oscap-customizations.json
achilleas-k: Clearly, I didn't think this through well enough.  I'll have to include a `distro` field in the config changes I'm working on.

achilleas-k (changes requested):
Thanks for this.  Some minor comments inline.

pkg/manifest/build.go
kingsleyzissou: I had it here first in testing and changed it... I must have missed this. Will fix

test/configs/oscap-customizations.json
kingsleyzissou: Okay cool

internal/oscap/oscap.go
kingsleyzissou: Waiting to hear back on this.

jan-cerny: I'm not sure what to suggest. The `autotailor` utility can store it's output to any file path and the `oscap` command accepts any file path in the `--tailoring-file` option, so in theory, you can put the file anywhere.

But, the path `/usr/share/xml/scap` is owned by the `scap-security-guide` RPM package, and I'm afraid that it isn't a good practice to add files to paths owned by some package, so I would rather avoid saving this there. From this point of view, the `/usr/share/osbuild` seems a better solution.

We had a similar situation in OSCAP Anaconda Addon, which is our plugin in the Anaconda installer. There, we put the artifacts to `/root/openscap-data` so that users can review them easily after the installation finishes.

kingsleyzissou:
Thanks for the feedback @jan-cerny... so the `customizations` here is internal api for how a user-provided bluepritn customizations are used when building the image - I can update the comments and commit messages to make this clearer.

The `autotailor` stage is a stage that wraps the `autotailor` binary so that's why I named it as such, but I don't feel strongly on this... @thozza do you have any opinions?

internal/oscap/oscap.go
kingsleyzissou: Thanks for the input @jan-cerny

thozza:
> The `autotailor` stage is a stage that wraps the `autotailor` binary so that's why I named it as such, but I don't feel strongly on this... @thozza do you have any opinions?

I do :sweat_smile:. See https://github.com/osbuild/images/pull/43#discussion_r1281695407

internal/oscap/oscap.go
kingsleyzissou: > Based on FHS (https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html) and the discussion, I suggest to put the tailoring file into a subdirectory under /usr/share/xml, e.g. /usr/share/xml/osbuild-openscap-data. This makes it explicit that the file was created by osbuild, but it is not related to osbuild as the tool in any way...

Awesome, I like this suggestion `/usr/share/xml/osbuild-openscap-data` and I like, as you mentioned, that it makes it explicit that it was created by osbuild.

achilleas-k (changes requested):
Nice work!  Thanks.

The config needs fixing though.
Also, it's possible now to filter configs based on distro, arch, or image type.
The all customizations config applies to all qcow2s except RHEL 7: https://github.com/osbuild/images/blob/efaf740181666aef5fc499cc9abc40bab24aed7c/test/config-map.json#L6-L11

But for this, maybe it's better to make a new config, `all-with-oscap` or something, and only enable it for the distro versions that support it, so that we don't disable the `all-customizations` test for the other distros too.

achilleas-k (approved):
Thanks!

PR #50, opened 2023-07-25
github: Remove the bogus condition from the auto-merge job

ondrejbudai:
I accidentally left the condition from the example in the code. We actually don't need it, we are happy to merge whatever dependabot proposes.

Since the Dependabot metadata job is no longer needed, I removed it as well.


achilleas-k (approved):
Oh, that explains why it didn't work.  Didn't get the chance to look into it.
Thank you!! :)

PR #51, opened 2023-07-25
Disable grub-boot-success in RHEL 9 edge commits to fix greenboot interference

say-paul:
grub_boot_success is stepping over greenboot as it is marking boot as successful in case of any ssh connection that persists over two min. The fix  is to update the unit stage to add dropin for user service and then adding the stage for edge-container and commits with the condition that `grub-boot-success.timer` not to trigger if greenboot exists.

This pull request includes:

- [x] adequate testing for the new functionality or fixed issue
- [x] adequate documentation informing people about the change such as
  - [x] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/


supakeen:
The change looks fine to disable the services if I understand the issue. I'd like acks from @nullr0ute and @runcom.

say-paul:
It will stay until we evaluate, if we want to drop the entire grub-tools.

say-paul:
Fixes https://github.com/fedora-iot/greenboot/issues/108

runcom:
> This change correctly disables the service(s) though I think this should be done through systemd presets; I'd like acks from @nullr0ute and @runcom. I'd assume this change to be temporary (?) and I'd like the bug report to be linked in the comments you've added to the `disabledServices` and `disabledIotServices`.
> 
> I'm also reasonably sure Fedora IoT already disables this in their presets as it's not running on my IoT instances.

@supakeen sure thing - for fedora-iot it definitely makes sense to disable this via presets, although I can't see it disabled in https://src.fedoraproject.org/rpms/fedora-release/blob/rawhide/f/80-iot.preset - does it make sense to understand why it's not running on your iot systems them too? then we can just disable it in the presets
For RHEL we don't have presets so _for now_ yes, this change makes sense

pkg/distro/rhel9/edge.go
```diff
@@ -197,6 +201,11 @@ var (
 		"NetworkManager.service", "firewalld.service", "sshd.service", "fdo-client-linuxapp.service",
 	}
 
+	disabledServices = []string{
+		//grub-boot-success interfering with greenboot
+		"grub-boot-success.timer",
```
runcom, line 206: @say-paul with this change in RHEL9 have you tested it correctly disables the user service?

say-paul:
>  I just figured it's here https://src.fedoraproject.org/rpms/fedora-release/blob/rawhide/f/80-iot-user.preset - so we probably don't need this change at all for IoT - @say-paul is there a reason you changed it for fedora too?

It is ignorance from my side as I didn't know where to look for the fedora presets, now I know, will revert  the changes accordingly.

pkg/distro/rhel9/edge.go
say-paul: Yes , tested it on rhel 9.

achilleas-k (dismissed):
LGTM.  Thanks!

runcom (dismissed):

achilleas-k:
Builds are failing with
```
Failed to disable unit, unit grub-boot-success.timer does not exist.
```

say-paul:
I guess we can leverage : org.osbuild.systemd.preset to disable as presets similar to fedora.

say-paul:
tried to disable it via  preset: using org.osbuild.systemd.preset in the new commit , but I am not seeing that its executed as part of the pipeline in osbuild-composer, what am I missing?

supakeen:
Stuffing things in `ImageConfig` won't make them magically appear on the pipeline object; you can either add the code that puts it in the `OSCustomizations` (`osCustomizations` in `images.go` reads from `ImageConfig` if set and uses that). Or you can directly set it on the pipeline; for example how `InstallWeakDeps` is set in `images.go:337`.

pkg/manifest/ostree_deployment.go
```diff
@@ -324,6 +326,15 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 		pipeline.AddStage(systemdStage)
 	}
 
+	if len(p.EnabledPresets) != 0 || len(p.DisabledPresets) != 0 {
```
say-paul, line 329: @supakeen so changes here is meaningless, I understood it as-  where it presets added to pipeline.

runcom:
> I guess we can leverage : org.osbuild.systemd.preset to disable as presets similar to fedora.

achilleas mentioned the unit fails to be located at all when disabling, why is that tho? how presets can change that?

achilleas-k:
Converted to draft to avoid merging accidentally on green.

achilleas-k:
@say-paul Current approach looks good, you just need to wire up the `DisabledPresets` from `ImageConfig` down to the `NewOSTreeDeployment` pipeline.

Here:
https://github.com/osbuild/images/blob/17eb47bb0876f269ed47e5793fa1c26b6c1bdbe7/pkg/image/ostree_raw.go#L80-L81

But since it's a new option, you should also look for other pipelines to wire it up for other image types, like in `OSCustomizations`.

runcom:
> @say-paul Current approach looks good, you just need to wire up the `EnabledPresets` and `DisabledPresets` from `ImageConfig` down to the `NewOSTreeDeployment` pipeline.

although, we want it in the commit right?

say-paul:
@runcom we need in both commit and deployment too i guess?

runcom:
> @runcom we need in both commit and deployment.

why would that be?

say-paul:
@runcom don't we want it anyway if the commit doesn't have it setup? Though I am thinking  presets will be written twice if we have add the stage  during both commit and deployment..

runcom:
> @runcom don't we want it anyway if the commit doesn't have it setup? Though I am thinking presets will be written twice if we have add the stage during both commit and deployment..

if composer builds the commit with presets and then the deployment there's no way the commit doesn't have it setup - that's the process, I can't see how would have a commit without and then we build a deployment from that and it does have it cause we "fixed" the deployment

say-paul:
Pending on: https://github.com/osbuild/osbuild/pull/1353

achilleas-k:
We don't have CI set up here to pull a non-released version of osbuild and unfortunately the osbuild PR just missed today's release.  I'll set up the testing later today and update the PR.

achilleas-k:
Rebasing on (or cherry-picking) #119 should work.

achilleas-k:
Thanks for rebasing.  Seems it didn't work.  Looking into it.

pkg/osbuild/systemd_unit_stage.go
```diff
@@ -18,10 +26,17 @@ func NewSystemdUnitStage(options *SystemdUnitStageOptions) *Stage {
...
 
 // 'Service' configuration section of a unit file
 type SystemdUnitServiceSection struct {
 	// Sets environment variables for executed process
 	Environment string `json:"Environment,omitempty"`
 }
+
+// 'Unit' configuration section of a unit file
+type SystemdUnitSection struct {
+	// Sets condition to to check if path exits
```
say-paul, line 40: need to fix comment

achilleas-k:
Found my mistake.  The generator scripts also use osbuild to generate the manifest ID by inspecting the manifests, so they need the pinned osbuild as well and I missed it.
Should be good now.

achilleas-k (approved):

pkg/manifest/os.go
```diff
@@ -732,6 +732,10 @@ func (p *OS) serialize() osbuild.Pipeline {
 		pipeline.AddStage(osbuild.NewOscapRemediationStage(p.OpenSCAPConfig))
 	}
 
+	for _, dropin := range p.SystemdUnit {
+		pipeline.AddStage(osbuild.NewSystemdUnitStage(dropin))
+	}
+
```
achilleas-k, line 738: I didn't catch this during the review.  I noticed that a lot of cloud images were being rebuilt when they shouldn't have been affected (https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/4890881075).
This is duplicating the systemd unit stage in the pipeline since it's already defined.  The stage isn't new.

achilleas-k (changes requested):

pkg/manifest/os.go
achilleas-k: Fixed and amended.

achilleas-k: @thozza Can you take a quick look at this so I don't self-approve?

thozza (changes requested):
Changes look good, except for putting unconditional configuration of the image in `distro/rhel9/images.go`. I would say that even the conditional configuration above should not be really here. It should be conditionally added to the image's `ImageConfig`.

 This is not the place for specific image configuration, this function is only the "glue" to pass configuration values to other parts of the code.

pkg/manifest/os.go
thozza: done, but I don't like some parts...

pkg/distro/rhel9/images.go
```diff
@@ -299,6 +299,18 @@ func edgeCommitImage(workload workload.Workload,
...
+			Unit:     "grub-boot-success.timer",
+			Dropin:   "10-disable-if-greenboot.conf",
+			UnitType: osbuild.Global,
+			Config: osbuild.SystemdServiceUnitDropin{
+				Unit: &osbuild.SystemdUnitSection{
+					FileExists: "!/usr/libexec/greenboot/greenboot",
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
say-paul: @thozza I get the ImageConfig, `where to the define the above dropin` which I think you referred as

> be set in the image definition itself

 so that I can refer that in the commitImage and  edgeContainerImage

pkg/distro/rhel9/edge.go
```diff
@@ -196,7 +198,19 @@ var (
...
+			Unit:     "grub-boot-success.timer",
+			Dropin:   "10-disable-if-greenboot.conf",
+			UnitType: osbuild.Global,
+			Config: osbuild.SystemdServiceUnitDropin{
+				Unit: &osbuild.SystemdUnitSection{
+					FileExists: "!/usr/libexec/greenboot/greenboot",
+				},
+			},
+		},
+	}
```
say-paul, line 213: @achilleas-k @thozza I dont see its getting added to the manifest, Not sure what I am missing.

achilleas-k: I see it added to the edge commit but not the edge container.  You did add it to the edge commit above (Line 27), so add it the same way to the edge container (Line 51) and it will generate the same manifests as before.

achilleas-k (approved):

say-paul:
> Changes look good, except for putting unconditional configuration of the image in `distro/rhel9/images.go`. I would say that even the conditional configuration above should not be really here. It should be conditionally added to the image's `ImageConfig`.
> 
> This is not the place for specific image configuration, this function is only the "glue" to pass configuration values to other parts of the code.

@thozza  Done

thozza (approved):
LGTM

PR #55, opened 2023-07-26
update to go 1.19

ondrejbudai:
UBI and the oldest supported Fedora (37) now all have go 1.19, so we are
 cleared to switch.


achilleas-k (approved):
THANK YOU!

PR #56, opened 2023-07-26
test: Use last stage ID as manifest ID

achilleas-k:
The biggest change for this PR is the change in the way we calculate the manifest ID for our test build cache.
Instead of hashing the manifest contents, use the last stage ID computed from osbuild --inspect.  This has the following advantages:
- It is not affected by formatting (indentation, newlines).
- It is not affected by content URLs when the content is the same.  For example, an ostree commit from a different URL will not affect the stage IDs if the content hash is the same.

This gets around an issue we were going to face when generating manifests from local ostree containers when the IP address or port of the local container changes.  It also lets us stop worrying about the exact format of the manifest file when comparing.

For now, the build config generator checks for both filenames (hash-based and stage-id-based) in the build cache from s3.  This is temporary until all or most files in the build cache are updated to the new filenames.  The script renames and syncs back with the new filename any file it finds that matches the old way of computing the manifest checksum.

Other improvements and changes:
- gen-manifests can now resolve ostree commits with --commits flag (mocks the resolution by default)
- gen-manifests can now mock the resolution of packages (depsolve) and containers with --packages=false and --containers=false respectively (does not mock the resolution by default)
- ostree container image types (iot-container and edge-container) are pushed to the gitlab registry for the project.  These will be used to test installer and raw image builds (followup PR).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -45,6 +45,7 @@ generate-build-config:
     INTERNAL_NETWORK: "true"
   script:
     - sudo dnf -y install go python3 gpgme-devel s3cmd
+      osbuild osbuild-luks2 osbuild-lvm2 osbuild-ostree osbuild-selinux
     - ./test/cases/generate-build-config build-config.yml
   artifacts:
     paths:
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -2,7 +2,6 @@
 package main
 
 import (
-	"crypto/sha256"
 	"encoding/json"
 	"flag"
 	"fmt"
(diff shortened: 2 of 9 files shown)


teg (dismissed):
At some point in the future we probably want to support manifests producing several artefacts, likely the combination of pipeline name + stage id for each.

For now that is not relevant and this looks perfect.

teg (approved):

achilleas-k:
> At some point in the future we probably want to support manifests producing several artefacts, likely the combination of pipeline name + stage id for each.
> 
> For now that is not relevant and this looks perfect.

I thought about this a bit too.  Our image definitions currently define their own exports so if we ever write an image type that has multiple, the exported pipeline names could be considered image type metadata that we can attach to the test manifest alongside the other metadata (sidenote: we should probably come up with a new name for "those manifests we generate for testing that include a bunch of metadata").  Then we could talk about "build IDs" (instead of manifest IDs), which would be the ID of the last stage of an exported pipeline.

PR #59, opened 2023-07-26
distro/rhel: update distro aliases (no minor)

achilleas-k:
RHEL distro names without minor version (rhel-8 and rhel-9) should point to the current GA.

RHEL 8 -> RHEL 8.8
RHEL 9 -> RHEL 9.2

(Better late than never?)


thozza (dismissed):
LGTM

achilleas-k:
Some repo gpg keys (gcp and eap/rhui) failed to validate :/
Will investigate tomorrow.

teg (approved):
This update is clearly correct in the current model.

However, I thought I'd use this as an opportunity to raise the fact that we may want to reconsider the logic (I feel responsible as I was at least partially responsible for introducing it).

We probably want to eventually get to the place where releasing new minor versions of RHEL will not require changes in our code. I think something like this might work:
- there is no hard-coded default minor version, but the cloud API take the minor version as an argument
 - on-prem, the welder API supports not specifying a minor version by detecting it at run-time from the host
 - the image builder API (hosted) supports not specifying a minor version by hard-coding what is the most recent release (moving this up several layers in the stack, maybe eventually this can be detected automatically)
 - our various conditionals always assume we will always be forward compatible between minor versions, so they will only ever enable new features based on the minor version being greater or equal than a certain value

This should mean that if I try to build `RHEL-8.12` either by passing it explicitly, running on-prem on a RHEL 8.12 host, or doing so through image builder once 8.12 is out (which of course it never will be, but that's another story), everything will "just work" even though we never added support for it.

thozza:
> We probably want to eventually get to the place where releasing new minor versions of RHEL will not require changes in our code. I think something like this might work:

thozza:
Sorry, I misclicked :innocent:

achilleas-k:
The gpg keys for the google repos, even when the gpg check is disabled, are failing to import when building on F38, breaking the set up of the build root:

```
error: Certificate FEEA9169307EA071:
  The certificate is expired: The primary key is not live
error: /tmp/gpgkey.zqjj6b_b: key 1 import failed.
error: Certificate 8B57C5C2836F4BEB:
  The certificate is expired: The primary key is not live
```

Updating the repository and adding the key from the current repository doesn't fix the issue.

I updated the Google repositories to the most recent snapshots and dropped the key to unbreak test builds until we can find the issue.

thozza (approved):

PR #62, opened 2023-07-27
image/live: renamed to `disk`

supakeen:
Renames the `image/live` to `image/disk`. Als renames all image types in all distros to be called `diskImage` instead of `liveImage`.

This better describes what the old `liveImage` did (build various disk images).


achilleas-k (approved):
This is the image function for all image types except things like ostree commits, containers, and ISOs.  This makes sense to me.

PR #64, opened 2023-07-27
resolver: Always return results sorted by Digest

bcl:
The output of the resolver can be inconsistent. This causes problems with the diff-manifest CI script because consecutive runs on the same code may have different outputs, causing a false manifest diff error from Schutzbot.

The tests was already sorting by Digest, so move that up into Finished() and remove the sort from the test. Now the results will always be in a consistent order.

This pull request includes:

- [X] adequate testing for the new functionality or fixed issue
- [X] adequate documentation informing people about the change such as
  - [ ] submit a PR for the [guides](https://github.com/osbuild/guides) repository if this PR changed any behavior described there: https://www.osbuild.org/guides/

### DIFF
--- a/pkg/container/resolver.go
+++ b/pkg/container/resolver.go
@@ -3,6 +3,7 @@ package container
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


achilleas-k (approved):
Good idea.  Thanks for this!

PR #81, opened 2023-08-04
Add URL to commit when printing build cache info

achilleas-k:
Adds a clickable URL in the job log to the commit when an image configuration was built.


achilleas-k:
Note that the commit IDs (and by extension the URLs) will always be for commits that aren't part of the main repository, because rebuilds almost always happen on PRs from forks.  We're relying on GitHub's retention to be able to see them.

There's an edge case where a rebuild can happen on the head of the merge queue if something changes in the rebase, which would mean the commit ID would be available in `main`, but I don't think that's going to happen often.

ochosi (approved):
Love it! Thanks!

PR #83, opened 2023-08-04
Strict decoding for test configs

achilleas-k:
With test configs, we want to make sure options aren't ignored if they're misspelled or in any other way malformed, so let's be strict about decoding them and DisallowUnknownFields.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -86,6 +86,9 @@ func loadConfig(path string) BuildConfig {
 	var conf BuildConfig
 
 	check(dec.Decode(&conf))
+	if dec.More() {
+		fail(fmt.Sprintf("multiple configuration objects or extra data found in %q", path))
+	}
 	return conf
 }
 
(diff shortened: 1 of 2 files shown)


achilleas-k:
https://github.com/osbuild/images/blob/03212162ff492db563b5805442a1f084a623a835/test/generators/generate-ostree-build-config#L234-L238

Who's petard is this and where did it come from? >_>

cmd/build/main.go
```diff
@@ -69,33 +69,31 @@ type crBlueprint struct {
...
-		fail(fmt.Sprintf("failed to unmarshal config %q: %s", filepath, err.Error()))
-	}
-	if config.Name == "" {
-		fail(fmt.Sprintf("config %q does not specify a name", filepath))
+
+	dec := json.NewDecoder(fp)
+	dec.DisallowUnknownFields()
+	var conf BuildConfig
+
+	if err := dec.Decode(&conf); err != nil {
```
ondrejbudai: `json.Decoder` supports decoding of multiple JSON values, so if you want to be extra sure that the file contains just one, you should also check for `!dec.More()`. Might be worth creating a helper for this (I suggest one with the same API as `json.Unmarshal`).

achilleas-k: Great idea.  Thanks!

ondrejbudai (approved):

PR #84, opened 2023-08-04
Store images in S3 after building

achilleas-k:
In addition to the build info, also store the manifest and the actual image by uploading the whole build directory to s3.
We should start using these images for further testing in other projects, like osbuild-composer and maybe to bring back image info in manifest-db without needing to rebuild every time.

This PR will cause a rebuild of all image types since the build cache directory structure is changing.

### DIFF
--- a/test/cases/build-image.sh
+++ b/test/cases/build-image.sh
@@ -9,13 +9,13 @@ arch=$(uname -m)
 imgtype="${2}"
 config="${3}"
 
-echo "Building image ${distro}/${imgtype} using config ${config}"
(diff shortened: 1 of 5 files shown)


ochosi (approved):
Beautiful! I love how simple this was :heart_eyes:

achilleas-k:
There was a permission issue on build artifacts.  They're owned by root and we were uploading them as non-root.  Added a line opening up the permissions.

ondrejbudai (dismissed):
Splendid!

ochosi (approved):
Love the emojis!

PR #91, opened 2023-08-07
Support checksums for ostree parents when building ostree commits

achilleas-k:
If a ref to be resolved is already a checksum, just copy it to the checksum field and don't resolve.  This is only valid for parent refs.  Image refs that look like checksums will be caught by a new Validate() function.

Supporting checksums for parent refs means that users can specify a parent ostree commit that is not the head of a branch.

Resolves COMPOSER-1976

See also https://github.com/osbuild/osbuild-composer/issues/3107

### DIFF
--- a/pkg/distro/rhel8/distro_test.go
+++ b/pkg/distro/rhel8/distro_test.go
@@ -489,11 +489,11 @@ func TestDistro_ManifestError(t *testing.T) {
 			if imgTypeName == "edge-commit" || imgTypeName == "edge-container" {
 				assert.EqualError(t, err, "kernel boot parameter customizations are not supported for ostree types")
 			} else if imgTypeName == "edge-raw-image" {
-				assert.EqualError(t, err, "edge raw images require specifying a URL from which to retrieve the OSTree commit")
+				assert.EqualError(t, err, fmt.Sprintf("%q images require specifying a URL from which to retrieve the OSTree commit", imgTypeName))
 			} else if imgTypeName == "edge-installer" || imgTypeName == "edge-simplified-installer" {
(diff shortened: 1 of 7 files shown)


pkg/ostree/ostree.go
```diff
@@ -83,6 +134,10 @@ func VerifyRef(ref string) bool {
 	return len(ref) > 0 && ostreeRefRE.MatchString(ref)
 }
 
+func VerifyChecksum(commit string) bool {
```
thozza, line 137: nitpick: This function and the one above could be private, no?

pkg/distro/rhel8/distro_test.go
```diff
@@ -489,7 +489,7 @@ func TestDistro_ManifestError(t *testing.T) {
 			if imgTypeName == "edge-commit" || imgTypeName == "edge-container" {
 				assert.EqualError(t, err, "kernel boot parameter customizations are not supported for ostree types")
 			} else if imgTypeName == "edge-raw-image" {
-				assert.EqualError(t, err, "edge raw images require specifying a URL from which to retrieve the OSTree commit")
+				assert.EqualError(t, err, fmt.Sprintf("\"%s\" images require specifying a URL from which to retrieve the OSTree commit", imgTypeName))
```
thozza, line 492: nitpick: I know that this is consistent with the message below, but we could use `%q` instead of `\"%s\"`.

thozza (dismissed):
Nice work. I added a few nitpicks, but nothing serious.

pkg/distro/rhel8/distro_test.go
achilleas-k: Yes, totally.  I copy pasted from RHEL 9 but I much prefer `%q` too.

pkg/ostree/ostree.go
achilleas-k: Yup.  We can make them public if we need to later.

PR #95, opened 2023-08-08
Use packaged Fedora IoT remote configuration

achilleas-k:
_Migration of https://github.com/osbuild/osbuild-composer/pull/3488._

The default Fedora IoT remote is packaged in the `fedora-iot-config` package.  This PR removes the ostree remote config we hard coded for the `iot-raw-image` (and associated image types `iot-qcow2-image` and `iot-simplified-installer`) and installs the package in the `iot-commit`, which means every Fedora IoT build (both commit and container) are configured using the package.


achilleas-k:
Marking as ready even though it includes #124

yih-redhat:
Hi @achilleas-k,
Does this pr fix any issue of Fedora Iot Remote? I didn't see anything special in our test script to test or configure Fedora Iot Remote, could you please explain more how this pr will impact our test script for iot-raw-image (and also the iot-simplified-installer and iot-qcow2-image) ? 
What test scenario will you suggest us to add into our test script? or just run current test script, and if it pass then this pr is ok to merge? Thanks.

achilleas-k:
The purpose of this PR is to configure the default remote for the image by installing an RPM that's managed by the Fedora IoT team rather than having it as part of the image definition.  This is similar to how RPM repositories are configured in RPM-based images: we don't configure repositories ourselves, but they are packaged and distributed by the distribution and we install them in our images.

I'm not sure we need to test that a particular remote configuration is present.  Perhaps we can have a check that there is a remote configured (any remote).  But more specifically, there was an issue the last time we tried to make this change that blocked us [1].  The tests we have in osbuild-composer were failing when setting up a **custom** remote in order to test the upgrade commit.  We still haven't figured out what's causing this issue and how the change affects it.

[1] https://github.com/osbuild/osbuild-composer/pull/3488

miabbott:
@paulwhalen you were looking at this from the Fedora side, right?

cgwalters:
In https://github.com/ostreedev/ostree/discussions/3079 I believe

pkg/manifest/ostree_deployment.go
```diff
@@ -164,7 +171,7 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 		&osbuild.OSTreeDeployStageOptions{
 			OsName: p.osName,
 			Ref:    commit.Ref,
-			Remote: p.Remote.Name,
+			Remote: remoteName,
```
achilleas-k, line 174: This is the cause of the problem with the upgrades (see https://github.com/achilleas-k/fedora-iot-remote).  When the remote configuration is coming through the package, we don't specify a remote name so the deployment doesn't have a configured remote.  And even adding one, with `ostree remote add ...`, it doesn't change the configuration of the deployment.

Compare
```
$ cat /ostree/deploy/fedora-iot/deploy/22543d2fc8d14087616f07ccf48d75eab535c26e81ecb31f4adf4d70f313f923.0.origin
[origin]
refspec=fedora/39/x86_64/iot
```
with
```
$ cat /ostree/deploy/fedora-iot/deploy/22543d2fc8d14087616f07ccf48d75eab535c26e81ecb31f4adf4d70f313f923.0.origin
[origin]
refspec=fedora-iot:fedora/39/x86_64/iot
```

achilleas-k: CC @henrywang

achilleas-k:
The remote issue should be fixed now!

pkg/image/ostree_disk.go
```diff
@@ -30,7 +30,7 @@ type OSTreeDiskImage struct {
 
 	SysrootReadOnly bool
 
-	Remote ostree.Remote
+	Remote *ostree.Remote
```
bcl, line 33: You can guess what I'm going to ask :) Why switch to a pointer instead of just looking for an empty `Name`?

achilleas-k: I was writing up a whole thing about how it lets us have a nil to signal "no remote at all" but we use parts of it in different places (name, url) so might as well have it concrete and rely on string values.

achilleas-k: Done.

supakeen (approved):
Awesome.

achilleas-k:
I'm going to merge this and move on to getting it into osbuild-composer.  My local tests worked so we'll have to catch any other issues in CI there.

PR #98, opened 2023-08-09
distro: add Fedora 40

ondrejbudai:
It's that time again and we are blocking Fedora IoT.


teg:
Would be nice if this could somehow be turned on its head so new versions are noop unless they introduce breaking changes ...

ondrejbudai:
I have a branch that attempts to do it. It brings interesting challenges - if you support an arbitrary version of Fedora (or RHEL), you cannot enumerate them all. However, we depend on enumerating the distributions on quite a lot of places. Without it, we e.g. cannot generate the test matrix: https://github.com/osbuild/images/blob/3554502aa561acb7eb21f21b3024ebb1e35192cf/pkg/distroregistry/distroregistry.go#L102

@achilleas-k and I think that a solution might to accept arbitrary distro versions in the API and then have a test matrix somewhere internally used just for the needs of testing.

achilleas-k:
> @achilleas-k and I think that a solution might to accept arbitrary distro versions in the API and then have a test matrix somewhere internally used just for the needs of testing.

Maybe the test matrix can just depend on what repos we have defined.  No repos => No distro support.  As soon as you add `test/data/repositories/fedora-99.json` with valid URLs, then any test enumerators will start testing F99...?

ondrejbudai:
For the record, we both got the same idea at basically the same moment.

image

This must mean something, right?

achilleas-k:
> Would be nice if this could somehow be turned on its head so new versions are noop unless they introduce breaking changes ...

Yes but!
Taking a moment to appreciate how 6b6bb6f3649d256c30786dbc1cd86ec15c76558e compares to how we introduced new distro versions 2 years ago.

test/generators/generate-ostree-build-config
```diff
@@ -281,7 +281,7 @@ def generate_configs(build_requests, pull_configs, container_configs, pipeline_f
         config_name = config["name"]
 
         # generate script line to pull and start container
-        container = container_configs[config_name]["container"]
+        container = container_configs[config_name]
```
achilleas-k, line 284: :flushed:

achilleas-k:
@ondrejbudai SNAPSHOTS ARE READY :tada:

ondrejbudai:
OK, the snapshot for F40 failed, because the mirror isn't apparently ready yet for F40. Let's go with F39 repositories for F40 short-term. We will fix this once F40 gets in a bit better shape. Note that the branching happens today, so this is expected.

achilleas-k (approved):
:tada: Thank you!

achilleas-k:
Package signature check failure
```
Signature check failed on sha256:ac2440f3fdaafc610f37ad4052a33b7d8889f86b3b0a520a09a71576f96b9a3c, lookup package name in manifest.
```

ondrejbudai:
I reverted the GPG key change back to F39. I need to look up how exactly package signing works in Fedora regarding the lifecycle.

PR #99, opened 2023-08-09
Enable test retries (1 retrry)

achilleas-k:
It seems the tests and the specifically the test case generators aren't as stable as I thought.  Some network flakiness can occur at times and they probably went unnoticed before due to the retries, making it look like it was all stable.

This reverts commit dd14277c09d257661441bf7c8297ba84bae56e29.


achilleas-k:
The top-level `generator-trigger` in the GitLab CI configuration doesn't need a retry.  It's a very simple script that just prints out yaml files and doesn't depend on any external resources.

ondrejbudai (approved):

PR #104, opened 2023-08-10
Refactor distros so we don't have to add a new one whenever it gets branched

ondrejbudai:
This is very WIP, I want to have a feedback from the testsuite firstly.


thozza:
@ondrejbudai this PR looks very promising. Do you plan to finish it any time soon? I'm asking, because the lack of support for 9.4 / 8.10 is currently blocking the latest `osbuild-composer` releases in RHEL gating. Thus I created https://github.com/osbuild/images/pull/195 as a quick fix. But if this PR would be finished earlier, I'd be happy to close mine.

ondrejbudai:
I hope to finish it! :) Will try to take a look at this in the upcoming days.

PR #106, opened 2023-08-10
devcontainer: update to work for images

teg:
There is no rpm for images, so we drop all rpm related bits. This also means we need to include the `gpgme-devel` dependency explicitly.

Update the container from f36 to f38.


ondrejbudai (approved):
Thanks!

PR #108, opened 2023-08-11
Add test for invalid hash algorithm for package checksum

atodorov:


pkg/osbuild/rpm_stage_test.go
```diff
@@ -180,3 +180,96 @@ func TestNewRpmStageSourceFilesInputs(t *testing.T) {
...
+
+	pkgSpecs := []rpmmd.PackageSpec{
+		{
+			Name:           "openssl-libs",
+			Epoch:          1,
+			Version:        "3.0.1",
+			Release:        "5.el9",
+			Arch:           "x86_64",
+			RemoteLocation: "https://example.com/repo/Packages/openssl-libs-3.0.1-5.el9.x86_64.rpm",
+			Checksum:       "invalid:fcf2515ec9115551c99d552da721803ecbca23b7ae5a974309975000e8bef666",
```
atodorov, line 197: @achilleas-k  I was expecting this to fail, however it's passing.

The `invalid:` prefix is the only difference between this test and `TestNewRpmStageSourceFilesInputs` above it.

atodorov (commented):
@achilleas-k I wasn't expecting the test to pass in its current form. Can you review ?

pkg/osbuild/rpm_stage_test.go
achilleas-k: We don't verify the checksum string in osbuild-composer or images.  This would have to be run against the stage in osbuild to fail.  If we want to have that verification early, we can add it as a `validate()` function that gets called as part of `New...()`.

We do have a `validate()` function for file references: https://github.com/osbuild/images/blob/41bfe61e14d32c7b176c86e79b4d37c53cd6fe23/pkg/osbuild/files_input.go#L309-L313
but it only checks if a checksum is included.

We can expand the validate to check for the algorithm strings that are supported by osbuild.  That function panics if it's invalid, so the test would have to be adapted to catch the panic instead.

atodorov: > We can expand the validate to check for the algorithm strings that are supported by osbuild. That function panics if it's invalid, so the test would have to be adapted to catch the panic instead.

That is the original intention, I didn't modify the test initially in order to see where exactly it will fail.

achilleas-k: Adding stricter checks to the checksum validator.

pkg/osbuild/curl_source.go
```diff
@@ -3,14 +3,47 @@ package osbuild
...
 
+func NewCurlSource() *CurlSource {
+	return &CurlSource{
+		Items: make(map[string]CurlSourceItem),
+	}
+}
+
+func NewCurlPackageItem(pkg rpmmd.PackageSpec) CurlSourceItem {
+	if !curlDigestPattern.MatchString(pkg.Checksum) {
+		panic("curl package source item has invalid digest")
```
thozza, line 27: I'm not convinced that using `panic()` is the right approach here. The digest hash comes from depsolving using DNF and it is AFAIK not hardcoded in the code. So it may as well happen that composer will get unsupported hash algorithm, but there is in fact no programming error in composer.

Returning an `error` and failing the image build with sensible error message seems like a more reasonable approach to me.

I think that this is a special case, compared to other stages, where we really want to panic.

pkg/osbuild/curl_source.go
```diff
@@ -3,14 +3,47 @@ package osbuild
...
+	return item
+}
+
+// AddPackage adds a pkg to the curl source to download. Will panic if any of
+// the supplied options are invalid or missing.
+func (source *CurlSource) AddPackage(pkg rpmmd.PackageSpec) {
+	item := NewCurlPackageItem(pkg)
+	source.Items[pkg.Checksum] = item
+}
+
```
thozza, line 46: minor nitpick: while reading the file, I was searching for `CurlSourceItem` definition while reading the `CurlSource` definition. I would suggest to move the whole block below `CurlSourceItem` definition, so that the code is easy to read from the top down...

thozza (commented):
I added a few inline comments. Otherwise, the PR looks good :+1:

atodorov (commented):
Test LGTM.

pkg/osbuild/curl_source.go
achilleas-k: You make a great point.  I'll change it.

achilleas-k: Done!

pkg/osbuild/curl_source.go
achilleas-k: Done.

thozza (approved):
Nice work! Thanks.

PR #109, opened 2023-08-11
test: ignore stage ordering for distro-arch jobs

achilleas-k:
Using the 'needs' keyword [1], a job will start as soon as its dependency is done, without waiting for the whole previous stage to finish.  This means that build jobs will start as soon as the corresponding generate jobs are done (i.e. for the same distro-arch), ostree-generators will run as soon as the corresponding build jobs are done, and ostree-build jobs will run as soon as the corresponding ostree-generators are done.

This might speed up execution in cases where one generator job takes longer than usual, for whatever reason, and ends up not generating any builds, whereas another generator job finishes early and does generate builds.

[1] https://docs.gitlab.com/ee/ci/yaml/index.html#needs

### DIFF
--- a/test/generators/configure-generators
+++ b/test/generators/configure-generators
@@ -47,6 +47,8 @@ image-build-trigger-{distro}-{arch}:
       - artifact: build-config.yml
         job: generate-build-config-{distro}-{arch}
     strategy: depend
+  needs:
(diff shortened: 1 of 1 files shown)


achilleas-k:
image

teg (approved):
Nice!

PR #119, opened 2023-08-16
Set up osbuild pre-release repos for testing

achilleas-k:
New script for setting a repository for installing an osbuild RPM built from a specific commit for testing against versions before they're released.
Based on osbuild-composer's `schutzbot/deploy.sh` script.


achilleas-k:
Included in #51

PR #121, opened 2023-08-17
Set up weekly releases

thozza:
Set up GH action to perform releases every Monday at 5 AM UTC. Use semantic versioning and bump only the minor version part of it.


thozza:
Marked as Draft, since I realized that our release action requires the Slack webhook URL.

achilleas-k (approved):
Alright, let's do this!!

PR #125, opened 2023-08-17
Delete duplicate GH release action

thozza:
I missed that there was already a release action in the repo, since I copied actions from the `ansible-osbuild-worker` repo, which named the action file in a different way... Let's remove one of the actions.


achilleas-k (approved):
Thanks!  :broom:

achilleas-k (approved):

PR #126, opened 2023-08-17
pathpolicy: allow /boot/efi customisation

croissanne:


internal/pathpolicy/policies.go
```diff
@@ -2,16 +2,17 @@ package pathpolicy
...
-	"/opt":  {},
-	"/srv":  {},
-	"/usr":  {},
-	"/app":  {},
-	"/data": {},
-	"/home": {},
-	"/tmp":  {},
+	"/":         {Exact: true},
+	"/boot":     {Exact: true},
+	"/boot/efi": {Exact: true},
```
achilleas-k, line 7: This will cause issues.  `/boot/efi` needs to be FAT so we'll need a special case because (I THINK) by default we make custom mountpoints xfs on RHEL.

achilleas-k (changes requested):

internal/pathpolicy/policies.go
achilleas-k: I might be wrong.  Checking.

achilleas-k: I WAS WRONG

achilleas-k: We generate the partition table in two passes: first pass resizes existing partitions and doesn't create new ones, to avoid changing filesystem types (I guess we thought of this).
Second pass adds new partitions/mountpoints.

achilleas-k (approved):

PR #130, opened 2023-08-22
Enable / disable services also using presets for OSTree deployment images if Ignition is used.

thozza:
Extend the `OSTreeDeployment` pipeline implementation to enable / disable services also using systemd presets in case Ignition is being used.

The pipeline already enables / disables services using the Systemd stage, which calls `systemctl`, but on some systems (since Fedora 37), Systemd will reset all enabled / disabled services based on system presets on first boot. This is not a problem for our images in general, since they contain fake machine-id, which makes systemd think that the system is not booting for the first time. This is however not the case when Ignition is being used, since it signals to systemd to treat the first boot as an actual first boot.

Enabling systemd services using `systemctl` is still kept in the pipeline, since the end result will be the same even if services are enabled / disabled multiple times via different means.

Related to:
- https://github.com/osbuild/osbuild-composer/pull/3634
- https://fedoraproject.org/wiki/Changes/Preset_All_Systemd_Units_on_First_Boot

### DIFF
--- a/pkg/manifest/ostree_deployment.go
+++ b/pkg/manifest/ostree_deployment.go
@@ -247,6 +247,18 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 				"systemd.condition-first-boot=true",
 			},
 		}))
+
+		// We enable / disable services below using the systemd stage, but its effect
+		// may be overridden by systemd which may reset enabled / disabled services on
+		// firstboot (which happend on F37+). This behavior, if available, is triggered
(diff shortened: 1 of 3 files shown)


thozza:
I still need to test this change in a PR in osbuild-composer repo to be sure it fixes the problem, but opening it for a review. I can also mark it as a Draft if people would feel better about it.

supakeen (approved):

ondrejbudai (approved):
Thanks!

achilleas-k:
Thanks for taking care of this :heart:

PR #132, opened 2023-08-23
distro/fedora: align the vhd image to the official version

ondrejbudai:
I took the kickstart from Fedora and reimplemented it here. This should fix a lot of issues like missing cloud-init. Since the kickstart for azure is inheriting the kickstart from the base cloud image, I did the same thing in our code to mimic that.

See
https://pagure.io/fedora-kickstarts/blob/58c856ae882a93362f025b4efae143fbd778edc9/f/fedora-cloud-base-azure.ks

### DIFF
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -247,35 +247,6 @@ var (
 		basePartitionTables: defaultBasePartitionTables,
 	}
 
-	vhdImgType = imageType{
-		name:     "vhd",
-		filename: "disk.vhd",
-		mimeType: "application/x-vhd",
-		packageSets: map[string]packageSetFunc{
-			osPkgsKey: vhdCommonPackageSet,
-		},
-		defaultImageConfig: &distro.ImageConfig{
-			Locale: common.ToPtr("en_US.UTF-8"),
-			EnabledServices: []string{
-				"sshd",
-			},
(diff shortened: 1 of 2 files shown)


pkg/distro/fedora/distro.go
```diff
@@ -572,6 +543,25 @@ func newDistro(version int) distro.Distro {
 	openstackImgType := qcow2ImgType
 	openstackImgType.name = "openstack"
 
+	vhdImgType := qcow2ImgType
```
thozza, line 546: What's the benefit of making a copy the `qcow2ImgType` if most of its properties are overridden anyway? :thinking: 

I think that this approach introduces unnecessary surprises when the `qcow2` image gets modified, with very little benefit...

ondrejbudai: Well, it's mimicking how these images are defined in the Fedora cloud kickstarts. The Azure image inherits from the qcow2 one. 

> I think that this approach introduces unnecessary surprises when the qcow2 image gets modified, with very little benefit...

...which matches our "upstream".

thozza: Fair enough, although personally I still don't think that doing so brings much value.

achilleas-k (approved):
I agree with @thozza regarding the inheritance (creating a copy of qcow2 and modifying), but I understand the logic, so let's keep it and see if makes life easier or harder down the line.

LGTM

PR #137, opened 2023-08-25
System Ready Minimal Images fixes

nullr0ute:
These are some fixes for the minimal image that are needed to fix up a few minor issues.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -17,6 +17,7 @@ type PartitionTable struct {
 
 	SectorSize   uint64 // Sector size in bytes
 	ExtraPadding uint64 // Extra space at the end of the partition table (sectors)
+	StartOffset  uint64 // Starting offset of the first partition in the table (Mb)
 }
 
 func NewPartitionTable(basePT *PartitionTable, mountpoints []blueprint.FilesystemCustomization, imageSize uint64, lvmify bool, requiredSizes map[string]uint64, rng *rand.Rand) (*PartitionTable, error) {
@@ -77,6 +78,7 @@ func (pt *PartitionTable) Clone() Entity {
 		Partitions:   make([]Partition, len(pt.Partitions)),
 		SectorSize:   pt.SectorSize,
 		ExtraPadding: pt.ExtraPadding,
+		StartOffset:  pt.StartOffset,
 	}
 
 	for idx, partition := range pt.Partitions {
@@ -364,6 +366,7 @@ func (pt *PartitionTable) relayout(size uint64) uint64 {
 	}
 
 	start := pt.AlignUp(header)
+	start += pt.StartOffset
 	size = pt.AlignUp(size)
 
 	var rootIdx = -1
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -224,6 +224,9 @@ func osCustomizations(
 	osc.PwQuality = imageConfig.PwQuality
 	osc.WSLConfig = imageConfig.WSLConfig
 
+	osc.Files = append(osc.Files, imageConfig.Files...)
+	osc.Directories = append(osc.Directories, imageConfig.Directories...)
+
 	return osc
 }
 
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/internal/fdo"
+	"github.com/osbuild/images/internal/fsnode"
 	"github.com/osbuild/images/internal/ignition"
 	"github.com/osbuild/images/internal/oscap"
 	"github.com/osbuild/images/internal/users"
@@ -686,4 +687,14 @@ func makeOSTreePayloadCommit(options *ostree.ImageOptions, defaultRef string) (o
 		Ref:  commitRef,
 		RHSM: options.RHSM,
 	}, nil
+}
+
+// initialSetupKickstart returns the File configuration for a kickstart file
+// that's required to enable initial-setup to run on first boot.
+func initialSetupKickstart() *fsnode.File {
+	file, err := fsnode.NewFile("/root/anaconda-ks.cfg", nil, "root", "root", []byte("# Run initial-setup on first boot\n# Created by osbuild\nfirstboot --reconfig\n"))
+	if err != nil {
+		panic(err)
+	}
+	return file
 }
--- a/pkg/platform/aarch64.go
+++ b/pkg/platform/aarch64.go
@@ -28,21 +28,21 @@ func (p *Aarch64) GetPackages() []string {
 	return packages
 }
 
-type Aarch64_IoT struct {
+type Aarch64_Fedora struct {
 	BasePlatform
 	UEFIVendor string
 	BootFiles  [][2]string
 }
 
-func (p *Aarch64_IoT) GetArch() Arch {
+func (p *Aarch64_Fedora) GetArch() Arch {
 	return ARCH_AARCH64
 }
 
-func (p *Aarch64_IoT) GetUEFIVendor() string {
+func (p *Aarch64_Fedora) GetUEFIVendor() string {
 	return p.UEFIVendor
 }
 
-func (p *Aarch64_IoT) GetPackages() []string {
+func (p *Aarch64_Fedora) GetPackages() []string {
 	packages := p.BasePlatform.FirmwarePackages
 
 	if p.UEFIVendor != "" {
@@ -57,6 +57,6 @@ func (p *Aarch64_IoT) GetPackages() []string {
 	return packages
 }
 
-func (p *Aarch64_IoT) GetBootFiles() [][2]string {
+func (p *Aarch64_Fedora) GetBootFiles() [][2]string {
 	return p.BootFiles
 }
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -357,7 +357,7 @@ var (
 		buildPipelines:      []string{"build"},
 		payloadPipelines:    []string{"os", "image", "xz"},
 		exports:             []string{"xz"},
-		basePartitionTables: defaultBasePartitionTables,
+		basePartitionTables: minimalrawPartitionTables,
 	}
 )
 
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -711,7 +711,9 @@ func newDistro(version int) distro.Distro {
 				FirmwarePackages: []string{
 					"arm-image-installer", // ??
 					"bcm283x-firmware",
-					"iwl7260-firmware",
+					"brcmfmac-firmware",
+					"iwlwifi-mvm-firmware",
+					"realtek-firmware",
 					"uboot-images-armv8", // ??
 				},
 			},
@@ -796,8 +798,10 @@ func newDistro(version int) distro.Distro {
 						"grub2-tools",
 						"grub2-tools-extra",
 						"grub2-tools-minimal",
+						"brcmfmac-firmware",
 						"iwlwifi-dvm-firmware",
 						"iwlwifi-mvm-firmware",
+						"realtek-firmware",
 						"microcode_ctl",
 						"syslinux",
 						"syslinux-nonlinux",
@@ -819,6 +823,10 @@ func newDistro(version int) distro.Distro {
 						"grub2-tools",
 						"grub2-tools-extra",
 						"grub2-tools-minimal",
+						"brcmfmac-firmware",
+						"iwlwifi-dvm-firmware",
+						"iwlwifi-mvm-firmware",
+						"realtek-firmware",
 						"uboot-images-armv8",
 					},
 				},
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -64,6 +64,13 @@ var (
 		"dbus-parsec",
 	}
 
+	minimalRawServices = []string{
+		"NetworkManager.service",
+		"firewalld.service",
+		"initial-setup.service",
+		"sshd.service",
+	}
+
 	// Image Definitions
 	imageInstallerImgType = imageType{
 		name:        "image-installer",
@@ -348,6 +355,9 @@ var (
 		mimeType:    "application/xz",
 		packageSets: map[string]packageSetFunc{
 			osPkgsKey: minimalrpmPackageSet,
+		},
+		defaultImageConfig: &distro.ImageConfig{
+			EnabledServices: minimalRawServices,
 		},
 		rpmOstree:           false,
 		kernelOptions:       defaultKernelOptions,
@@ -709,12 +719,12 @@ func newDistro(version int) distro.Distro {
 		&platform.Aarch64{
 			BasePlatform: platform.BasePlatform{
 				FirmwarePackages: []string{
-					"arm-image-installer", // ??
+					"arm-image-installer",
 					"bcm283x-firmware",
 					"brcmfmac-firmware",
 					"iwlwifi-mvm-firmware",
 					"realtek-firmware",
-					"uboot-images-armv8", // ??
+					"uboot-images-armv8",
 				},
 			},
 			UEFIVendor: "fedora",
@@ -726,7 +736,7 @@ func newDistro(version int) distro.Distro {
 		liveInstallerImgType,
 	)
 	aarch64.addImageTypes(
-		&platform.Aarch64_IoT{
+		&platform.Aarch64_Fedora{
 			BasePlatform: platform.BasePlatform{
 				ImageFormat: platform.FORMAT_RAW,
 			},
@@ -776,10 +786,18 @@ func newDistro(version int) distro.Distro {
 		minimalrawImgType,
 	)
 	aarch64.addImageTypes(
-		&platform.Aarch64{
+		&platform.Aarch64_Fedora{
 			UEFIVendor: "fedora",
 			BasePlatform: platform.BasePlatform{
 				ImageFormat: platform.FORMAT_RAW,
+				FirmwarePackages: []string{
+					"arm-image-installer",
+					"bcm283x-firmware",
+					"uboot-images-armv8",
+				},
+			},
+			BootFiles: [][2]string{
+				{"/usr/share/uboot/rpi_arm64/u-boot.bin", "/boot/efi/rpi-u-boot.bin"},
 			},
 		},
 		minimalrawImgType,
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -31,8 +31,11 @@ const (
 	// blueprint package set name
 	blueprintPkgsKey = "blueprint"
 
-	//Kernel options for ami, qcow2, openstack, vhd and vmdk types
-	defaultKernelOptions = "ro no_timer_check console=ttyS0,115200n8 biosdevname=0 net.ifnames=0"
+	//Default kernel command line
+	defaultKernelOptions = "ro"
+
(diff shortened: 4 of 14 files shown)


nullr0ute:
@sarmahaj FYI

runcom:
Lgtm

ondrejbudai (dismissed):

thozza (changes requested):
I think that the PR is missing changes to https://github.com/osbuild/images/blob/81ab54167675ff8f7f319be23c1a4e386c80a7f7/pkg/distro/rhel8/package_sets.go#L77-L83 and nothing in the commits suggests otherwise

Also the Fedora `minimal-raw` does not reflect the changes to kernel command line arguments done in RHEL -> https://github.com/osbuild/images/blob/81ab54167675ff8f7f319be23c1a4e386c80a7f7/pkg/distro/fedora/distro.go#L370-L388
https://github.com/osbuild/images/blob/81ab54167675ff8f7f319be23c1a4e386c80a7f7/pkg/distro/fedora/distro.go#L35

miabbott:
Do these changes prevent the `minimal-raw` artifacts from being used on devices or are they nice to have?

Given that we are past dev-freeze for RHEL 9.3, I don't think we should be making changes to the `minimal-raw` type unless they are absolutely necessary.

blomquisg:
I've asked around to figure out what is controversial about this change, and the main thing I've heard is the inclusion of `initial-setup`.

The workflow as I understand it is to use Image Builder to produce a bootable image from these minimal-raw images. 

It sounds like the compromise here is to remove `initial-setup` from this PR, then document two available paths for users:

1. Use the minimal-raw image and a blueprint with Image Builder to create an initial user account, or
2. Add `initial-setup` back using Image Builder.

If users are looking for an experience that's similar to building other images with Image Builder, they would choose option 1.  If users are looking to test out a simple bootable image and don't mind manually configuring the installation, they could choose option 2.

Thoughts? 

/cc @nullr0ute @thozza @runcom @miabbott @mrguitar

nullr0ute:
> I've asked around to figure out what is controversial about this change, and the main thing I've heard is the inclusion of `initial-setup`.
> 
> The workflow as I understand it is to use Image Builder to produce a bootable image from these minimal-raw images.

That's the first step of the whole process, the next step is to provision it onto the device.

> It sounds like the compromise here is to remove `initial-setup` from this PR, then document two available paths for users:

I did not take that conclusion away from yesterday's meeting.

>     1. Use the minimal-raw image and a blueprint with Image Builder to create an initial user account, or

This will cause potential security implications. We don't ship default credentials because of that.

>     2. Add `initial-setup` back using Image Builder.
> 
> 
> If users are looking for an experience that's similar to building other images with Image Builder, they would choose option 1. If users are looking to test out a simple bootable image and don't mind manually configuring the installation, they could choose option 2.

That depends on the usecase. In the case of 1) that may be useful for a single user, but it falls over if it's being provided to end users, partners or other cases.

nullr0ute:
I have found more fixes needed so I'm going to update this PR with more commits soon once I've worked it all out. New update soon.

nullr0ute:
> I think that the PR is missing changes to
> and nothing in the commits suggests otherwise

You are correct, but TBH this was never meant to land for el8 as these devices will only ever be supported on el9, but I will fix now it's there, I thought the package set was the same across both, I missed the 9 in the name.
 
> Also the Fedora `minimal-raw` does not reflect the changes to kernel command line arguments done in RHEL ->

And I missed that variable, I have other fixes coming in too, will update.

miabbott:
> I did not take that conclusion away from yesterday's meeting. The decision from what I understand it was "what ever we do we can't change it once it's shipped". Also they're not mutually exclusive.

Correct; we want to understand what (if any) changes are **required** in order for users/customers to be able to boot devices using this artifact type.

> That depends on the usecase. In the case of 1) that may be useful for a single user, but it falls over if it's being provided to end users, partners or other cases.

I'd like to understand what would be the more common use case for this artifact.

Even if it is the latter use case, including `initial-setup` by default would cause this artifact to be the unicorn among all the other artifacts produced by osbuild. Maybe that is the decision we come to, but it does feel anti-pattern compared to how most users will handle user provisioning/configuration when using osbuild.

The other consideration when adding packages to the default set is that end-users have no means of **removing** those default packages as part of the osbuild compose process, so we need to be very certain that any defaults are going to be applicable to the majority of users.

Regardless of the decisions made about this PR, we will need strong documentation explaining the outcome and how user provisioning/configuration will be done with this artifact.

nullr0ute:
Documentation for RHEL-8 here: https://access.redhat.com/articles/4323781

miabbott:
> For osbuild maybe, but from an install experience on devices (AKA phase two in this process) the user generally does have to deal with users and root passwords in the traditional installer (anaconda) use case, whether that be specifying it in the UX doing a "CD style installer (via traditional media or USB stick)" or as an option in the kickstart so for the "install on device" part it is not differing from the standard user experience. 

If the user composed an image using osbuild and the blueprint was missing any user customizations, then yes they would have to handle it during the install phase. But that can be avoided if the blueprint customization is used.

> If say Partner Management provided a image to a partner to put on a device the partner would never deal with osbuild so from their perspective it would not be different because with anaconda you have to specify those pieces. In fact if you do that install over a serial or text console it's that exact same UX that is shown to the user.

I think this use case is where I would dig into...is the most common use case for `minimal-raw` going to be something like this?  What happens if someone provides user customization in a blueprint for a `minimal-raw` image?  Will `initial-setup` still run and have to be dealt with?

nullr0ute:
> > For osbuild maybe, but from an install experience on devices (AKA phase two in this process) the user generally does have to deal with users and root passwords in the traditional installer (anaconda) use case, whether that be specifying it in the UX doing a "CD style installer (via traditional media or USB stick)" or as an option in the kickstart so for the "install on device" part it is not differing from the standard user experience.
> 
> If the user composed an image using osbuild and the blueprint was missing any user customizations, then yes they would have to handle it during the install phase. But that can be avoided if the blueprint customization is used.

The question then becomes how many people will take the defaults vs customising it? I suspect the answer to that is impossible to predict BUT we do need to provide *SANE* defaults, if a user does nothing in the blueprint as it stands now they will have an image that they have no way to login to, that is confusing and I am absolutely against providing a default username and password (as I'm sure product security will be as well), on the other side of that in Fedora we have had no issues with users understanding how to use initial-setup.
 
> > If say Partner Management provided a image to a partner to put on a device the partner would never deal with osbuild so from their perspective it would not be different because with anaconda you have to specify those pieces. In fact if you do that install over a serial or text console it's that exact same UX that is shown to the user.
> 
> I think this use case is where I would dig into...is the most common use case for `minimal-raw` going to be something like this? What happens if someone provides user customization in a blueprint for a `minimal-raw` image? Will `initial-setup` still run and have to be dealt with?

Well the intention of the minimal-raw was intended for basic development, board bring up etc, which would generally be a single user..... BUT like all these things predicting what is intended vs how customers end up using things is impossible to predict.

nullr0ute:
> > Documentation for RHEL-8 here: https://access.redhat.com/articles/4323781
> 
> That kbase describes how to answer the questions in the TUI. That part is fine and self explanatory. It's more about making the call to force this on users by default, when blueprints make it crazy trivial to add it when you need it. We don't have any product plans to distribute these images so I don't know why we care if customers embed users/credentials. I 

What happens if they don't specify a username/password in the blueprint? They can't login. In the case with initial-setup they should be able to just exit and login, or just ssh in straight up if they're doing it remotely. Also initial-setup may just exit if a user or root password is set (I need to verify this, because it's nuanced and I don't remember exact details, it may just not make them create it and they can just continue).

blomquisg:
I'm mostly stuck on various types of users for the `minimal-raw` image.  Looking back over the comments, it seems to be a mostly open question, honestly.  We have some ideas of some users, but I don't see anything concrete.

That said, if we have a definite way to bypass the `initial-setup` tui, and we can document that somewhere, then I guess it's less of a big deal.  But, that needs to be included in testing this before we release it.  If including `initial-setup` necessarily drops the user into an interactive tui, that feels like a miss.

Honestly, since we'll have to document the user interaction either way (whether we have to show them how to skip `initial-setup` with a blueprint, or how the user would add `initial-setup` back in via Image Builder), I would lean towards including fewer packages, and document how to build an image that includes what they need.  But, this goes back to the understanding who we think will use the `minimal-raw`, and how.

nullr0ute:
> I think that the PR is missing changes to

I think these are now all fixed, along with a bunch of other changes :)

thozza:
> That's probably overkill because there's a ConditionPathExists=/.unconfigured in the initial-setup-reconfigure unit file. I don't see that file as part of the rpm, so I'm not sure if a post script creates it. My guess is the firstboot --reconfigure (from memory so could be wrong) kickstart options tickles this. I know we can create files in the blueprint, but I don't know if we can remove or "untouch" a file like this. A drop-in feels more complicated and a worse customer experience to override. I think knowing the easiest/best way to bypass initial-setup would really help us weight the pros & cons here.

No, there is no option to delete files from the image.

From what I can tell, disabling the `initial-setup.service` unit looks like the most reasonable way to disable `initial-setup` from running in the image on the first boot.

> If I can dream a bit, an even better experience would be keeping the defaults @nullr0ute provided and empowering users to override/remove the base packages via a blueprint (or other method). That way we get away from hard coding things and backing ourselves (and customers) into corners. This causes so much pain right now particularly with packages like podman, when users don't need/want containers. We can still provide support because we can point to an image override if something doesn't work as expected.

I don't see this happening any time soon (or at all). We intentionally don't want users to be able to exclude packages from the image base package set. @supakeen is experimenting with an "empty" image in Fedora and an option to exclude packages in BP, but that would still be possible only with user-specified packages, not the base package set.

On a general note, all image types produced by image builder come without any user by default. Things are different for various image types, but in general there is always some form of mechanism for creating users on the first boot (`cloud-init`, `ignition`, cloud-specific guest tools, etc.).

So having an image type which does not have any default user (which is desired), does not have any way to create it on the first-boot and needs a user in order to use the OS (e.g. container images don't need one) is an anti-pattern from my PoV and something that would not make sense...

pkg/disk/partition_table.go
```diff
@@ -373,6 +373,7 @@ func (pt *PartitionTable) relayout(size uint64) uint64 {
 			rootIdx = idx
 			continue
 		}
+		start += partition.Start
```
supakeen, line 376: Let me comment on this that this change was suggested by me. Normally `PartitionTable` will find the start offset by itself based on the header and partition sizes. This adds the `Start` attribute if it already had a value (otherwise it will be `0`) to correctly offset the partitions (and keeps track of it since it happens in the loop).

I've built a manifest and image with these changes and the result is:
```
user@muja image € fdisk -l disk.img
Disk disk.img: 3.69 GiB, 3965714432 bytes, 7745536 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: D209C89E-EA5E-4FBD-B161-B461CCE297E0

Device       Start     End Sectors  Size Type
disk.img1    18432  428031  409600  200M EFI System
disk.img2   428032 1452031 1024000  500M Linux extended boot
disk.img3  1452032 7745502 6293471    3G Linux filesystem
```

However, this change *does* hit all partition table generation and as such please do take a good look at this change too :)

supakeen:
Looks OK in my book; I agree with @thozza on the users/initial-setup.

@nullr0ute Are these images to be used on Raspberry Pi's as well (at least for Fedora)? If so you might want to move the `minimalRawImgType` in `distro/fedora/distro.go` next to the `iotRawImgType` instead.

nullr0ute:
> @nullr0ute Are these images to be used on Raspberry Pi's as well (at least for Fedora)? If so you might want to do something similar as done for `Aarch64_IoT` in `distro/fedora/distro.go`.

For Fedora, yes, just pushed a commit for that, and also to fix up WiFi firmware in Fedora too.

pkg/disk/partition_table.go
nullr0ute: we may still have a bug here, I will verify Monday it's not my testing process.

mrguitar:
I really appreciate all the attention on this, and thanks for the discussion around initial-setup. I think we can easily doc up what's necessary here and proceed IMO. I feel much better about this now.

@nullr0ute, thanks so much for getting this done.

@thozza, I would love to show you some real world implications around the current position regarding base image inflexibility. I completely understand and think the status quo is justifiable, but I do not believe the implications are known which is why I'm extremely nervous about this stance moving forward. I'll set up some time to chat. 

Cheers!

pkg/disk/partition_table.go
nullr0ute: It was my testing process, it works for our offset needs. That concern is resolved for me, but I won't resolve the conversation due to the need of wider review.

nullr0ute:
Investigating a few other bits.

pkg/disk/partition_table.go
achilleas-k: I think this change makes things hard to follow and while it does have the intended effect when combined with the next commit, it's doing it in a way that could lead to bugs or confusing behaviour in the future.

The `partition.Start` is meant to be calculated from the partition table generator based on the order of partitions and their size.  With this change, any `Start` defined in the static partition table becomes essentially an offset or gap ahead of the partition.  Also, it's ignored for `/`.

The next commit in the PR adds a `Start` value to the first partition in the minimal raw partition tables which has the intended effect but consider that:
1. If, for whatever reason, the first partition in a table is `/`, any start/offset, wont be applied.
2. If, for whatever reason, we ever need to add gaps between partitions, it wont be added for `/`.  More importantly, the desire for this is questionable but it is something we're enabling with this change.
3. Even if we do ever want to add gaps between partitions, I'd rather add a new field rather than give the `Start` field the double meaning of "offset/gap" (when defined on the static partition table) and "absolute start position" (after processing).

Since the purpose of this change is to add a start offset for the entire PT, I propose we add a new field to the `PartitionTable` struct called `Offset` (or `StartOffset`), like we do for `ExtraPadding`: https://github.com/osbuild/images/blob/4f0f01389dec364f151b4b96f840dd2aa9dedeba/pkg/disk/partition_table.go#L19
and simply add that to the `header` after it's calculated at the top of the `relayout()` function.

achilleas-k (changes requested):
Generally this all looks good.  I'm requesting changes for the partition table offset handling (https://github.com/osbuild/images/pull/137/commits/c0cc14194bdcfff3f02368ad2a13ab3b24093d41#r1318749476).

The `initial-setup` discussion is also not resolved yet.  In general, I'd be in favour of installing and enabling `initial-setup` unconditionally, even if it wont be needed, rather than having a default image configuration that locks users out (when they don't define a user).

pkg/disk/partition_table.go
nullr0ute: This was based on what you had done in commit 48466ae44. From my experience at least there has been numerous occasions to define an offset for the start of all partitions. EG all vmware virt images with certain releases of VMWare used to need to start at a particular offset due to enterprise arrays and how they used to deal with blocks. I've never see a use/need of similarly specific spaces between concurrent partitions, only ever at the very beginning.

nullr0ute:
> just cleaning up some typos in the commits @nullr0ute

All fixed

pkg/disk/partition_table.go
nullr0ute: I've replaced this with a patch similar to the extrapadding.

pkg/disk/partition_table.go
```diff
@@ -364,6 +366,7 @@ func (pt *PartitionTable) relayout(size uint64) uint64 {
 	}
 
 	start := pt.AlignUp(header)
+	start += pt.StartOffset
```
nullr0ute, line 369: @achilleas-k it also needed this to actually apply the offset :)

nullr0ute:
OK I think this is ready for review and merge, just generating test images.

achilleas-k (commented):
LGTM but I wont approve since I wrote some of the changes myself.  I'd like to get an extra pair of eyes on this before merging.

nullr0ute:
> OK I think this is ready for review and merge, just generating test images.

I've tested both RHEL and Fedora images on an imx8m device and works as expected :partying_face:

pkg/platform/aarch64.go
```diff
@@ -28,21 +28,21 @@ func (p *Aarch64) GetPackages() []string {
 	return packages
 }
 
-type Aarch64_IoT struct {
+type Aarch64_Fedora struct {
```
thozza, line 31: Can you elaborate on the reason to rename this platform? I always thought that the platform is distro-agnostic until you instantiate it set custom values...

pkg/distro/fedora/images.go
```diff
@@ -687,3 +688,13 @@ func makeOSTreePayloadCommit(options *ostree.ImageOptions, defaultRef string) (o
 		RHSM: options.RHSM,
 	}, nil
 }
+
+// initialSetupKickstart returns the File configuration for a kickstart file
+// that's required to enable initial-setup to run on first boot.
+func initialSetupKickstart() *fsnode.File {
+	file, err := fsnode.NewFile("/root/anaconda-ks.cfg", nil, "root", "root", []byte("firstboot --reconfig\n"))
```
thozza, line 695: @achilleas-k could you please add the comment to this file that it was created by osbuild? Or something along those lines...

I won't comment on each distro file, but ideally please add it to each of them.

thozza (commented):
I added a few minor comments, but in general this looks OK.

@nullr0ute I don't see any partition table start offset being set for RHEL-8 image. Is this intentional?

pkg/distro/fedora/images.go
achilleas-k: Added the following comment to the first two lines of all `anaconda-ks.cfg` files:
```
# Run initial-setup on first boot
# Created by osbuild
```

pkg/platform/aarch64.go
paulwhalen: I think the intention was to remove reference to IoT as its applicable to all of Fedora. Perhaps `Aarch64_Sbc` or `Aarch64_Sr`  to be more generic?

thozza: Personally, I would even merge it all into the `Aarch64` platform and have just one. I don't like distro-specific code leaking outside of the `distro` directory. @achilleas-k do you have any opinions on this?

achilleas-k: I agree with both things you said but I'm okay with having it messy like this for now and cleaning up after.  It was already a bit weird that we had this `IoT` variant in the platform package.  Like you said, there's no reason we can't have the extra `BootFiles` in the base struct and leave it empty for the ones that don't need it.

But we can merge them in a follow-up and keep the rename.

thozza: OK, I'm fine with a follow-up. Nevertheless, there is still the question of the missing offset for RHEL-8 images...

nullr0ute:
> @nullr0ute I don't see any partition table start offset being set for RHEL-8 image. Is this intentional?

Yes, RHEL-8 only supports ServerReady and that doesn't need the offsets if people wish to use it on that release.

thozza (approved):

thozza:
> I had planned, had a brief discussion with @achilleas-k, on doing some more cleanups around the Fedora pieces on a some follow up clean ups (I have some other ideas for cleanups too), the el8 offset one has now been answered, what's left?

Nothing, I got all the answers I needed... Approved...

miabbott:
> on doing some more cleanups around the Fedora pieces on a some follow up clean ups (I have some other ideas for cleanups too),

@nullr0ute could you file issues for these changes?  It would enable us to get others working on the changes while you are on PTO (or even when you are not on PTO).

PR #159, opened 2023-09-13
Test distro and mock dnf json cleanup

achilleas-k:
The dnfjson_mock package isn't used by the images project and was left over from osbuild-composer.  The "test-distro-2" test distribution that was calling it was never used in the images tests.  The mock-dnf-json cmd is also duplicated in osbuild-composer and is used there by API tests.

The PR also includes a very small functional change: It adds the package name and digest to the error message when the digest/checksum fails to validate.

CI changes: I updated the version of golangci-lint that we use in the GitHub actions and disabled gosec G101 (hardcoded credentials) because it produces too many (and exclusively) false positives.  The updated linter revealed a lot of instances of G601 (implicit memory aliasing in for loop) when I ran it locally and I fixed all of them.


thozza (dismissed):
Nice work!

pkg/distro/test_distro/distro.go
```diff
@@ -392,47 +389,3 @@ func NewRegistry() *distroregistry.Registry {
 	registry.SetHostArchName(TestArchName)
 	return registry
 }
-
-// New2 returns new instance of TestDistro named "test-distro-2".
-func New2() *TestDistro {
```
bcl, line 397: This is needed by osbuild_composer tests in `client/unit_test.go` and `weldr/api_test.go`

bcl (dismissed):
Tried updating osbuild-composer with the changes and ran into at least 1 problem.

pkg/distro/test_distro/distro.go
achilleas-k: I was doing the same just now and I think it might be better to copy `test-distro-2` back into osbuild-composer.

achilleas-k:
Should we keep this PR open until https://github.com/osbuild/osbuild-composer/pull/3683 gets worked out in case we need to bring some changes here?

bcl:
This just needs one commit from this draft (I can never seem to make adding a commit to an existing PR work) - https://github.com/osbuild/images/pull/160/commits/5325bf58b559dee1f2978e3ef6a3eb0b7ae677b3 to switch newTestDistro to a public function.

I've updated my PR https://github.com/osbuild/osbuild-composer/pull/3683 to use this, and to fix the cloudapi test failures I was seeing. That PR will of course fail until a new images is vendored, I can do that once this PR is merged.

achilleas-k:
> This just needs one commit from this draft (I can never seem to make adding a commit to an existing PR work) - [5325bf5](https://github.com/osbuild/images/commit/5325bf58b559dee1f2978e3ef6a3eb0b7ae677b3) to switch newTestDistro to a public function.

Cherry picked and pushed.  Thanks!

bcl (approved):
Works for me :)

achilleas-k:
The GitLab CI test generator failed.  Adding back on the queue.

PR #163, opened 2023-09-14
[RHEL 9.3.0] System Ready Minimal Images fixes

achilleas-k:
#137 cherry picked on top of v0.3.0 for RHEL 9.3.0.

PR #166, opened 2023-09-18
distro: fix kernel options for Fedora Azure image

achilleas-k:
During rebase of 3d3649014465603f700cfe4191edcc39ddb64514, this change was missed due to 3d8388edf8c5e7f5b62ef0b5f7d1c7314e7eb2ed.

### DIFF
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -276,7 +276,7 @@ var (
 				"loadmodules.service",
 			},
 		},
-		kernelOptions:       defaultKernelOptions,
+		kernelOptions:       cloudKernelOptions,
 		bootable:            true,
 		defaultSize:         2 * common.GibiByte,
 		image:               diskImage,


kingsleyzissou (approved):
lgtm

ondrejbudai (approved):
I don't fully understand why we need a Fedora-related change in a RHEL branch, but it also shouldn't do any harm, and you have my full trust as always. :)

kingsleyzissou:
> I don't fully understand why we need a Fedora-related change in a RHEL branch, but it also shouldn't do any harm, and you have my full trust as always. :)

I believe it's because https://github.com/osbuild/images/pull/163 contained some of those changes already

achilleas-k:
> I don't fully understand why we need a Fedora-related change in a RHEL branch, but it also shouldn't do any harm, and you have my full trust as always. :)

I considered disabling the tests in the RHEL branch in osbuild-composer but decided fixing the issue would be better, even if a bit pointless.

PR #169, opened 2023-09-21
simplified-installer/fedora: extend rootlv size

paulwhalen:
This PR fixes [#3529](https://github.com/osbuild/osbuild-composer/issues/3529) , and extends the `iotSimplifiedInstallerPartitionTables` `rootlv` to 8GiB.

### DIFF
--- a/pkg/distro/fedora/partition_tables.go
+++ b/pkg/distro/fedora/partition_tables.go
@@ -350,7 +350,7 @@ var iotSimplifiedInstallerPartitionTables = distro.BasePartitionTableMap{
 						Description: "built with lvm2 and osbuild",
 						LogicalVolumes: []disk.LVMLogicalVolume{
 							{
-								Size: 2569 * common.MebiByte,
+								Size: 8 * common.GibiByte,
 								Name: "rootlv",
 								Payload: &disk.Filesystem{
 									Type:         "ext4",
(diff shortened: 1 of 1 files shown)


achilleas-k:
This shouldn't be necessary.  The coreos-installer should be growing the root filesystem after installation to fill the disk and the raw image file in the simplified installer is already set to 10 GiB [1].

I did a quick test locally and saw that the growfs stage of the installation is failing with `/dev/sda3 is in use by /dev/dm-0: found busy partitions`

image

We should figure out what's going on there instead.

[1] https://github.com/osbuild/images/blob/03a22660a9f889036332dee640b9d55b738bf9c7/pkg/distro/fedora/distro.go#L181

achilleas-k (changes requested):
Adding `Request changes` until we figure out what's wrong.

paulwhalen:
> This shouldn't be necessary. The coreos-installer should be growing the root filesystem after installation to fill the disk and the raw image file in the simplified installer is already set to 10 GiB:

This is to closer match the sizing in `edgeBasePartitionTables`. Currently `coreos-installer-growfs` grows the partition, but not the root filesystem. I believe this is to leave storage for other services, similar to Fedora Server.

achilleas-k:
> > This shouldn't be necessary. The coreos-installer should be growing the root filesystem after installation to fill the disk and the raw image file in the simplified installer is already set to 10 GiB:
> 
> This is to closer match the sizing in `edgeBasePartitionTables`. Currently `coreos-installer-growfs` grows the partition, but not the root filesystem. I believe this is to leave storage for other services, similar to Fedora Server.

We also keep it small because it's straightforward to grow after deployment if needed, whereas shrinking isn't.

All that said, if this is how we (well, you) want to define the image, then let's do it.  I'd still like to look into the growfs failure I saw above, but that can be done separately.

PR #172, opened 2023-09-22
rhel9: GCE: Add correct grub2 tools for UEFI platforms

nullr0ute:
The grub2-tools-efi package is actually contains tools just for booting Linux on x86 EFI Apple Macs, they're three tools that are uesd for the Mac bless process. The actually useful UEFI tools for normal platforms are contained in the minimal and main tools package so ensure we have the right ones in place for GCE images.


achilleas-k (approved):
Thanks.  LGTM.

PR #176, opened 2023-09-25
Edge remove hybrid boot

mmartinv:
Remove hybrid boot and all the legacy BIOS partitions from Edge images as UEFI is the only supported boot mechanism.

### DIFF
--- a/pkg/distro/rhel8/distro.go
+++ b/pkg/distro/rhel8/distro.go
@@ -251,21 +251,23 @@ func newDistro(name string, minor int) *distribution {
 		amiImgTypeX86_64(rd),
 	)
 
+	x86FirmwarePackages := []string{
+		"microcode_ctl", // ??
+		"iwl1000-firmware",
+		"iwl100-firmware",
+		"iwl105-firmware",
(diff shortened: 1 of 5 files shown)


pkg/distro/rhel8/distro.go
```diff
@@ -267,7 +267,7 @@ func newDistro(name string, minor int) *distribution {
 				"iwl6050-firmware",
 			},
 		},
-		BIOS:       true,
+		BIOS:       false,
```
achilleas-k, line 270: The `bareMetalX86Platform` isn't used only for edge.  It's also used for the image installer (see L279 below), which is an anaconda ISO that deploys a prebuilt image.  For this, we want to keep BIOS enabled, so we'll have to use a separate platform object for that.

achilleas-k (changes requested):
The change disables BIOS for the image installer as well, which we want to have enabled.  I made a note below on RHEL 8, but the same applies for RHEL 9.

achilleas-k (approved):

achilleas-k:
Thanks!

PR #182, opened 2023-09-29
Minimal image builds

achilleas-k:
This PR adds a new feature to the build configuration called `minimal` which affects payload packages in two ways:
- It ignores any statically defined packages for an image type (the packages defined on `ImageType.packageSets`).
- It disables weak dependencies for all payload packages.

The option therefore creates a **minimal** image where the packages installed are only the ones required to boot the image and provide basic functionality according to Platform (for firmware and bootloader configuration for example) and the Environment (`cloud-init` for cloud environment for example), removing any packages from an image type that are added purely by policy.

For the qcow2 (and its derivatives), the enabled services have been removed from the base image type configuration and moved to the respective environments: EC2 (ami), Azure (vhd), and KVM (new env for qcow2).

---

**THIS IS AN EXPERIMENTAL FEATURE**

I don't intend for this to be exposed in osbuild-composer just yet.  It's currently enabled only in Fedora.  I tested locally a minimal build of Fedora 38 x86_64 qcow2 and it works.  It shrinks the image from 500 MiB to 300 MiB, and reduces the package count from 453 to 215.

I'd like to get this in even if untested to have it as a proof of concept.
I intend for this to address a very common feature request which is the removal of base packages from image types.  Instead of allowing package exclusion, a minimal image with weak dependencies disabled should allow users to add just the packages they **want** while the platform and environment define the packages they **need**.  This will become even more powerful when we let users define the platform and environment directly.  If (when) we introduce a null environment, and make it user-configurable, a user would be able to build a minimal qcow2 (as enabled in this PR) and also remove `cloud-init`, which is defined in the new KVM environment.

### DIFF
--- a/internal/environment/azure.go
+++ b/internal/environment/azure.go
@@ -5,9 +5,18 @@ type Azure struct {
 }
 
 func (p *Azure) GetPackages() []string {
-	return []string{"WALinuxAgent"}
+	return []string{
+		"cloud-init",
+		"WALinuxAgent",
+	}
 }
 
 func (p *Azure) GetServices() []string {
-	return []string{"waagent"}
+	return []string{
+		"cloud-init.service",
+		"cloud-config.service",
+		"cloud-final.service",
+		"cloud-init-local.service",
+		"waagent",
+	}
 }
--- a/internal/environment/ec2.go
+++ b/internal/environment/ec2.go
@@ -9,5 +9,10 @@ func (p *EC2) GetPackages() []string {
 }
 
 func (p *EC2) GetServices() []string {
-	return []string{"cloud-init.service"}
+	return []string{
+		"cloud-init.service",
+		"cloud-config.service",
+		"cloud-final.service",
+		"cloud-init-local.service",
+	}
 }
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -233,20 +233,15 @@ var (
 	}
 
 	qcow2ImgType = imageType{
-		name:     "qcow2",
-		filename: "disk.qcow2",
-		mimeType: "application/x-qemu-disk",
+		name:        "qcow2",
+		filename:    "disk.qcow2",
+		mimeType:    "application/x-qemu-disk",
+		environment: &environment.KVM{},
 		packageSets: map[string]packageSetFunc{
 			osPkgsKey: qcow2CommonPackageSet,
 		},
 		defaultImageConfig: &distro.ImageConfig{
 			DefaultTarget: common.ToPtr("multi-user.target"),
-			EnabledServices: []string{
-				"cloud-init.service",
-				"cloud-config.service",
-				"cloud-final.service",
-				"cloud-init-local.service",
-			},
 		},
 		kernelOptions:       cloudKernelOptions,
 		bootable:            true,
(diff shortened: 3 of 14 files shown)


supakeen:
On first sight this is a great addition to the possibilities and will make any future Fedora editions-as-blueprints likely much easier as we can always start from a 'minimal' image type variant.

What I dislike is that this expands the combinations matrix we already have for our images with another dimension. I'm going to generate a few manifests and see what comes out of that because I feel that some image types with minimal turned on might produce the same manifest with perhaps only a different output format.

I'll go over it :)

achilleas-k:
> I'm going to generate a few manifests and see what comes out of that because I feel that some image types with minimal turned on might produce the same manifest with perhaps only a different output format.

I'd be curious to see how many of them boot.  We're pretty good with the definitions in Fedora I think, maybe not for IoT, but RHEL package sets are still not properly segmented so removing the static sets will likely make them unbootable.

internal/environment/ec2.go
```diff
@@ -9,5 +9,10 @@ func (p *EC2) GetPackages() []string {
 }
 
 func (p *EC2) GetServices() []string {
-	return []string{"cloud-init.service"}
+	return []string{
+		"cloud-init.service",
+		"cloud-config.service",
+		"cloud-final.service",
+		"cloud-init-local.service",
+	}
```
thozza, line 17: I'm just curious, if this was really needed (IOW, these services were previously not enabled and should have been) or is it to make this consistent with the `KVM` platform, which also installs and enables `cloud-init`?

pkg/distro/fedora/distro.go
```diff
@@ -233,20 +233,15 @@ var (
 	}
 
 	qcow2ImgType = imageType{
-		name:     "qcow2",
-		filename: "disk.qcow2",
-		mimeType: "application/x-qemu-disk",
+		name:        "qcow2",
+		filename:    "disk.qcow2",
+		mimeType:    "application/x-qemu-disk",
+		environment: &environment.KVM{},
```
thozza, line 239: Shouldn't you also remove `cloud-init` from the image base package set? :thinking:

thozza (commented):
Overall, nice work and interesting experimental feature. :rocket:  I added few comments, but nothing serious.

internal/environment/ec2.go
achilleas-k: On the qcow2, they were set on the image config which was inherited by the ami and vhd images.  I don't know if they're necessary, but more importantly I didn't want to change any base definitions.

pkg/distro/fedora/distro.go
achilleas-k: It was never explicitly part of the base package sets in Fedora.  For the image types we modified here, it was always pulled in by `@Fedora Cloud Server`.

internal/environment/azure.go
```diff
@@ -5,9 +5,18 @@ type Azure struct {
 }
 
 func (p *Azure) GetPackages() []string {
-	return []string{"WALinuxAgent"}
+	return []string{
+		"cloud-init",
```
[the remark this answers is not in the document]
achilleas-k: It was never explicitly part of the base package sets in Fedora.  For the image types we modified here, it was always pulled in by `@Fedora Cloud Server`.

internal/environment/ec2.go
thozza: ACK

thozza (approved):
LGTM

PR #188, opened 2023-10-04
Boot test AMIs

achilleas-k:
This PR rewrites the `osbuild-upload-aws` command to a new `boot-aws` that supports three subcommands: `setup`, `teardown`, and `run`.

- `setup` uploads an image to AWS and boots it, setting up user credentials and security policies for SSH.  It prints the IP address and saves the IDs of all newly created objects to a file (`resources.json` by default).
- `teardown` reads the file created by the `setup` command and cleans up all the resources/objects.
- `run` runs `setup`, uploads an executable to the remote host, runs it, then runs `teardown`.

This is set up to run every time a new AMI is built in CI.  It's a basic boot test that simply makes sure that the image boots and can run some basic commands to print information about the image (`rpm -qa`, `cat /etc/os-release`, `uname -a`).


achilleas-k:
Let's rebuild all the test images while we're at it :)

test/scripts/base-host-check.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -euxo pipefail
```
teg, line 2: I think at some point we used `systemctl is-system-running --wait` as a sanity check, might that be appropriate here? It would catch any services that failed to start.

achilleas-k: Added.

achilleas-k:
Things got a bit messy with some outdated local copies and minor mistakes in the command line arguments to the boot test script.  Should be okay now.

achilleas-k:
> Things got a bit messy with some outdated local copies and minor mistakes in the command line arguments to the boot test script. Should be okay now.

5 force-pushes later, I think it might actually be okay this time.

test/scripts/boot-image
```diff
@@ -0,0 +1,118 @@
...
+    return os.path.join(build_path, last_pipeline, files[0])
+
+
+def main():
+    desc = "Boot an image in the cloud environment it is built for and validate the configuration"
+    parser = argparse.ArgumentParser(description=desc)
+    parser.add_argument("distro", type=str, default=None, help="distro for the image to boot test")
+    parser.add_argument("arch", type=str, default=None, help="architecture of the image to boot test")
+    parser.add_argument("image_type", type=str, default=None, help="type of the image to boot test")
+    parser.add_argument("config", type=str, help="config used to build the image")
```
achilleas-k, line 85: We don't need this argument.  I'll remove it to simplify.

cmd/osbuild-upload-aws/main.go
```diff
@@ -59,10 +79,42 @@ func main() {
...
+
+	runResult, err := a.RunInstanceEC2(ami, securityGroup.GroupId, userData, "t3.micro")
+	instanceID := runResult.Instances[0].InstanceId
+
+	ip, err := a.GetInstanceAddress(instanceID)
+	check(err)
+
+	fmt.Printf("Instance %s is running and has IP address %s\n", *instanceID, ip)
+
+	defer func() {
```
ondrejbudai: This should be before `a.GetInstanceAddress`.

cmd/boot-aws/main.go
```diff
@@ -83,42 +92,157 @@ func main() {
...
+	}
+
+	return 0
+}
+
+func run(c string, args ...string) ([]byte, []byte, error) {
+	fmt.Printf("> %s %s\n", c, strings.Join(args, " "))
+	cmd := exec.Command(c, args...)
+
+	stdout, err := cmd.StdoutPipe()
```
ondrejbudai:
```golang
var cmdout, cmderr bytes.Buffer
cmd.Stdout = &cmdout
cmd.Stderr = &cmderr
err = cmd.Run()
```

would make this method much simpler I think.

cmd/boot-aws/main.go
```diff
@@ -12,10 +10,71 @@ import (
...
+func exitCheck(err error) {
+	if err != nil {
+		fmt.Fprintf(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+}
+
+// checkFlags checks that all the required flags are set and returns error if
+// one is not.
+func checkFlags(flags *pflag.FlagSet) error {
```
ondrejbudai, line 32: Use https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired instead of a custom function?

cmd/boot-aws/main.go
```diff
@@ -272,7 +273,8 @@ func doSetup(a *awscloud.AWS, filename string, flags *pflag.FlagSet, res *resour
 
 	fmt.Printf("AMI registered: %s\n", aws.StringValue(ami))
 
-	securityGroup, err := a.CreateSecurityGroupEC2("image-tests", "image-tests-security-group")
+	securityGroupName := uuid.New().String()
```
ondrejbudai, line 276: Please, prefix this.

thozza (commented):
Great stuff!

I didn't comment on anything, but agree with Ondrej's comments. Although some of them are resolved by subsequent commits.

There's one thing that I don't like too much, specifically the two forks of the internal AWS cloud library in this repo and osbuild-composer repo. This PR modifies it, which will only make things worse going forward. Did you think about actually making the library public in one of these repos and using it in the other? :thinking:

cmd/boot-aws/main.go
```diff
@@ -83,42 +92,157 @@ func main() {
...
-	fmt.Printf("Press RETURN to terminate instance")
-	bufio.NewReader(os.Stdin).ReadBytes('\n')
+	ip, err := a.GetInstanceAddress(instanceID)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "GetInstanceAddress(): %s\n", err.Error())
+		return 1
+	}
+	fmt.Printf("Instance %s is running and has IP address %s\n", *instanceID, ip)
+
+	hostsfile := "/tmp/test_hosts"
```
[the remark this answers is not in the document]
achilleas-k: In the final version, the `known_hosts` file gets created in a temporary directory created with the pattern `boot-test-*`.

achilleas-k:
> There's one thing that I don't like too much, specifically the two forks of the internal AWS cloud library in this repo and osbuild-composer repo. This PR modifies it, which will only make things worse going forward. Did you think about actually making the library public in one of these repos and using it in the other? 🤔

Yes, I think we should make this one public and import it in composer.  We've talked a few times (all of us I think) about whether it's weird to have cloud upload + boot code in images vs composer and I think we all agree (to varying degrees maybe) that it's okay to have it here.

None of the changes here are very big, so I think it wont be much work to adapt composer.

cmd/boot-aws/main.go
achilleas-k: Done!

cmd/boot-aws/main.go
```diff
@@ -0,0 +1,494 @@
...
+	if len(stderr) > 0 {
+		fmt.Fprintf(os.Stderr, string(stderr)+"\n")
+	}
+	return stdout, stderr, err
+}
+
+func sshRun(ip, user, key, hostsfile string, command ...string) error {
+	sshargs := []string{"-i", key, "-o", fmt.Sprintf("UserKnownHostsFile=%s", hostsfile), "-l", user, ip}
+	sshargs = append(sshargs, command...)
+	_, _, err := run("ssh", sshargs...)
```
thozza, line 84: Would it make sense to print the `stdout` when running the "check" script on the remote host? :thinking:

thozza (commented):
Overall, this looks very good, but I sanity checked some CI results and I'm confused... I may be reading them wrong, but e.g. https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/5283268188#L2927 looks as if the `base-host-check.sh` was never executed on the remote VM? I double-checked the scripts and all looks correct to me, so I'm not sure what's happening :thinking:

cmd/boot-aws/main.go
achilleas-k: It should be printed by the `run()` function (L70-72).  I'm not sure what's happening, but I think it might be some weirdness with stdout buffers.  The same log you linked prints
```
Testing image at ./build/rhel_9-x86_64-ami-empty/image/image.raw
✅ Marking boot successful
```
after the boot test is finished, but the `Testing image at` line should be printed before the boot test is triggered.

achilleas-k:
I think it might be that we don't set `PYTHONUNBUFFERED: 1` when running the build, boot, and upload scripts like we do for the generators.  Since it was a bash script, I guess I didn't add it before but we should now.  That should fix the log output.  I'll add it now.

achilleas-k:
The `PYTHONUNBUFFERED` fixed the ordering of the `Testing image at ./build/fedora_40-x86_64-ami-empty/image/image.raw` message, but I'm still not seeing the script output.  Looking into it.

achilleas-k:
Found the issue.  It was actually a bad combination of two things:
- The scp command was failing because it was running `scp path/to/file host:path/to/file` and failing because `path/to` doesn't exist on the host.
- The error from the scp command was being clobbered by the cleanup functionality returning without errors and so the `exitCheck()` was returning success.

cmd/boot-aws/main.go
```diff
@@ -0,0 +1,503 @@
...
+		return err
+	}
+
+	// copy the executable
+	if err := scpFile(ip, username, key, hostsfile, filename); err != nil {
+		return err
+	}
+
+	// run the executable
+	return sshRun(ip, username, key, hostsfile, fmt.Sprintf("./%s", filename))
```
thozza, line 396: You need to use `filepath.Base(filename)` here as well. It may be IMO cleaner to extract the base name here and pass the same value as `destination` to the `scpFile()` and `ssRun()`, instead of extracting the filename in `scpFile()`. My reasoning is that you should be able to expect that the provided destination passed to `scpFile()` will be used as is, not modified.

cmd/boot-aws/main.go
thozza: I saw that the check script output should be printed in the logs, but I can't see the output of the commands run in it (such as `rpm -qa`, etc...). The reason is most probably still the fact that the script was not yet run in the CI :confused:

cmd/boot-aws/main.go
achilleas-k: Yeah, good catch.  Late night hacking fail :/

achilleas-k: Done!

achilleas-k:
Fixed and updated.  Running a local test in the meantime as well.

achilleas-k:
I checked RHEL 9.0 x86 AMI and it's degraded because `kdump` fails to start with `kdump: No memory reserved for crash kernel`.

test/scripts/build-image
```diff
@@ -0,0 +1,67 @@
...
+    parser.add_argument("distro", type=str, default=None, help="distro for the image to boot test")
+    parser.add_argument("image_type", type=str, default=None, help="type of the image to boot test")
+    parser.add_argument("config", type=str, help="config used to build the image")
+
+    args = parser.parse_args()
+    distro = args.distro
+    image_type = args.image_type
+    config_path = args.config
+
+    print("👷 Building image {distro}/{imgtype} using config {config}")
```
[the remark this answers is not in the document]
achilleas-k: Woopsie.  Thanks!

test/scripts/base-host-check.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "❓ Checking system status"
+if ! sudo systemctl is-system-running --wait; then
```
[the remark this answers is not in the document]
achilleas-k: Oh good catch.  I was focused on the kdump failure that I didn't notice we were getting `starting` as the output.  I'll have to loop on `initializing` and `starting` it seems.  Not too difficult at least.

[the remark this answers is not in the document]
achilleas-k: Wrote a little bashy version of `--wait`

achilleas-k:
Did some testing on EC2 and locally and I think increasing the memory of the machines might solve the issue.  I switched them to `small` from `micro`.  It still doesn't explain why some were failing and others weren't, within the same major distro version.  Maybe 1 GiB is just borderline?

One thing I did when testing the RHEL 9.0 x86 AMI in EC2 was (sort of accidentally) upgrade the kernel to the one in 9.2 and on reboot, kdump was fine.  Not sure if this means that kdump will fail with specific kernel versions when the amount of memory isn't large enough.

Let's see if the change in instance type fixed it and take it from there.

thozza (commented):
The PR looks good and CI passes, but I'm still having hard time navigating the results. I can't find any job which actually boot-tested the image and all the image tests which I pick have `NullBuild` :shrug:

Is this due to the fact that these images have been tested by previous pipelines? If so, is there a way to get to the job which actually tested the image? :thinking:

supakeen (approved):
Very nice stuff; @thozza I think the boot tests only run if there's actually a changed image which this PR doesn't have (hence all the `NullBuild`).

achilleas-k:
Yes, that's a bit of an issue.  I've been thinking about this and I don't have the best solution yet.  The problem is when PRs like this get rebased a lot and different images get built at different times.  If you go into the generator for each distro version, you can see the commit SHA and a GitHub link to the commit when it was built last.  Also, in the top level coverage check, it now lists the images that were built in the CI run.

The last run here only built the RHEL 9.0 aarch64 AMI, because that was the last one that was failing to boot cleanly before I increased the instance memory for ARM.

I think it might be worth spending some time figuring out a good strategy to report image builds on a PR.  But it shouldn't just list every image built at every rebase.  I'm imagining an edge case where a PR changes an image accidentally and is then rebased to revert the change.  You wouldn't want it listed in the PR changes in the end.  So maybe what we need is a job that collects every image that differs from main and lists the images somewhere with a link to the commit.

achilleas-k:
This for example links to the last time the Fedora 38 x86 images were built for each one

https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/5329921859

thozza (approved):

achilleas-k:
> This is good, but it is still not obvious whether the commit when an image was built was/is part of the PR. Anyway, this is probably not something to figure out in this PR, but a general issue of the testing approach for images and trying to make it efficient.

Agree 100%.  I'll figure something out :)

PR #190, opened 2023-10-05
OSBuild result: improve readability of the text format

thozza:
I would find it nicer if the pipeline and stage names would be separated by a colon in the text format of osbuild output, instead of e.g.:

```
Pipeline build
Stage org.osbuild.rpm
Output:
[/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
...
```

In addition, ensure that there is a newline after each stage results are printed, regardless if there were any metadata to print. This change ensures that each stage results are visually separated by one newline.

The text format is now uploaded to Koji builds, therefore I'm trying to make it a bit easier to read.

### DIFF
--- a/pkg/osbuild/result.go
+++ b/pkg/osbuild/result.go
@@ -147,10 +147,10 @@ func (res *Result) Write(writer io.Writer) error {
 	sort.Strings(pipelineNames)
 
 	for _, pipelineName := range pipelineNames {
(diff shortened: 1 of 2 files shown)


achilleas-k:
Looks like we have some unit tests that check for this exact output.

ondrejbudai (approved):
Thanks!

PR #194, opened 2023-10-09
revert edge remove hybrid boot

mmartinv:
- Revert "distro: Disable BIOS boot for all the RHEL edge images"
- Revert "distro: remove unused BIOS partition from RHEL edge images"


ondrejbudai (approved):
Thanks!

PR #195, opened 2023-10-09
Add support for RHEL 8.10 / 9.4 distros

thozza:
This is currently blocking new upstream releases of `osbuild-composer` in RHEL Gating.

In addition, I had to fix the `common.VersionLessThan()` function, which did not work for version parts containing more than one digit (e.g. `8.10`).

Lastly, I had to fix the RHEL-8 AMI/EC2 default image configuration related to RHSM.


ondrejbudai (approved):

internal/common/distro_test.go
```diff
@@ -52,3 +52,89 @@ VARIANT_ID=workstation`,
...
+			Expected: false,
+		},
+		{
+			Name:     "8-stream < 9-stream",
+			VersionA: "8-stream",
+			VersionB: "9-stream",
+			Expected: true,
+		},
+		{
+			Name:     "9-stream < 9.1",
```
ondrejbudai, line 126: This one is weird, I think that semantically, these should be incomparable. Not a biggie, though, it's just a silly cornercase.

pkg/distro/rhel9/edge.go
```diff
@@ -586,7 +586,7 @@ func edgeCommitPackageSet(t *imageType) rpmmd.PackageSet {
 		ps = ps.Append(aarch64EdgeCommitPackageSet(t))
 	}
 
-	if !common.VersionLessThan(t.arch.distro.osVersion, "9.2") || !common.VersionLessThan(t.arch.distro.osVersion, "9-stream") {
+	if !common.VersionLessThan(t.arch.distro.osVersion, "9.2") || t.arch.distro.osVersion == "9-stream" {
```
ondrejbudai, line 589: I think I prefer `!t.arch.isRHEL()` for determining if a distro is Centos Stream, but it's also a bit weird. Maybe we should introduce a `isCentosStream` helper at some point.

ondrejbudai (dismissed):
Wonderful! :)

internal/common/distro_test.go
thozza: It is, but if you think about it more, the suffix is treated as a specifier or the version (e.g. rc or beta). Then `9` is just `9.0.0`, thus `9.1` is of higher version. But I agree that in our case and in case of CS vs RHEL versions, this does not make sense and is logically incorrect :man_shrugging:

pkg/distro/rhel9/edge.go
thozza: I fixed all comparisons of the `osVersion` to `9-stream`.

internal/common/distro_test.go
achilleas-k: Would it make more sense to have a special case in the function and make it fail?

achilleas-k (approved):
Thanks for all this.  LGTM.  I'm ok with merging even without figuring out the exact behaviour for comparisons with `stream`.

internal/common/distro_test.go
thozza: FWIW, I'm OK with the current behavior as is. This is a situation where we technically compare versions of two different distros with different versioning scheme.

achilleas-k: Yeah let's merge it and if it gets weird we can change it.

PR #198, opened 2023-10-11
Fedora: add base support for ppc64le and s390x architectures

thozza:
Add base support for ppc64le and s390x architectures, based on RHEL-9. Support qcow2 and container image types for now.

I manually tested these on ppc64le and s390x hosts.
[ppc64le-qcow2-boot.log](https://github.com/osbuild/images/files/12872361/ppc64le-qcow2-boot.log)
[s390x-qcow2-boot.log](https://github.com/osbuild/images/files/12872363/s390x-qcow2-boot.log)

Marking as draft until I replace Fedora repos with rpmrepo snapshots (https://github.com/osbuild/rpmrepo/pull/86)


supakeen (dismissed):
<3

pkg/distro/fedora/distro.go
```diff
@@ -541,6 +541,16 @@ func newDistro(version int) distro.Distro {
...
 
+	ppc64le := architecture{
+		distro: &rd,
+		name:   platform.ARCH_PPC64LE.String(),
+	}
+
+	s390x := architecture{
+		distro: &rd,
+		name:   platform.ARCH_S390X.String(),
+	}
```
praiskup, line 552: \o/

test/data/repositories/fedora-39.json
```diff
@@ -14,5 +14,21 @@
...
...AJU\nSFShBlctWKnlhQ7ehockqtgXtWckkqPZZjGiMXwHde9b9Yyi+VqtUQWxSWny+9g9\n6tcoa3AdnmpqSTHQxYajD0EGXJ0z0NXfqxkI0lo8UxzypEBy4sARZ4XhTU73Zwk0\nLGhEUHlfyxXgRs6RRvM2UIoo+gou2M9rn/RWkhuHJNSfgrM0BmIBCjhjwGiS33Qh\nysLDWJMdch8lsu1fTmLEFQrOB93oieOJQ0Ysi5gQY8TOT+oZvVi9pSMJuwARAQAB\ntDFGZWRvcmEgKDM5KSA8ZmVkb3JhLTM5LXByaW1hcnlAZmVkb3JhcHJvamVjdC5v\ncmc+iQJOBBMBCAA4FiEE6PI5lvIyGGQMtEy+dc9axBi450wFAmLykg8CGw8FCwkI\nBwIGFQoJCAsCBBYCAwECHgECF4AACgkQdc9axBi450yd4w//ZtghbZX5KFstOdBS\nrcbBfCK9zmRvzeejzGl6lPKfqwx7OOHYxFlRa9MYLl8QG7Aq6yRRWzzEHiSb0wJw\nWXz5tbkAmV/fpS4wnb3FDArD44u317UAnaU+UlhgK1g62lwI2dGpvTSvohMBMeBY\nB5aBd+sLi3UtiSRM2XhxvxaWwr/oFLjKDukgrPQzeV3F/XdxGhSz/GZUVFVprcrB\nh/dIo4k0Za7YVRhlVM0coOIcKbcjxAK9CCZ8+jtdIh3/BN5zJ0RFMgqSsrWYWeft\nBI3KWLbyMfRwEtp7xSi17WXbRfsSoqwIVgP+RCSaAdVuiYs/GCRsT3ydYcDvutuJ\nYZoE53yczemM/1HZZFI04zI7KBsKm9NFH0o4K2nBWuowBm59iFvWHFpX6em54cq4\n45NwY01FkSQUqntfqCWFSowwFHAZM4gblOikq2B5zHoIntCiJlPGuaJiVSw9ZpEc\n+IEQfmXJjKGSkMbU9tmNfLR9skVQJizMTtoUQ12DWC+14anxnnR2hxnhUDAabV6y\nJ5dGeb/ArmxQj3IMrajdNwjuk9GMeMSSS2EMY8ryOuYwRbFhBOLhGAnmM5OOSUxv\nA4ipWraXDW0bK/wXI7yHMkc6WYrdV3SIXEqJBTp7npimv3JC+exWEbTLcgvV70FP\nX55M9nDtzUSayJuEcfFP2c9KQCE=\n=J4qZ\n-----END PGP PUBLIC KEY BLOCK-----",
```
praiskup, line 22: You likely know this, but @xsuchy maintains up-to-date keys in https://github.com/xsuchy/distribution-gpg-keys which you can easily depend on.

thozza: Thanks. I don't know the exact reasons why we tend to use literal key values instead of pointing to URL. I think we originally lacked a mechanism to fetch GPG keys from URLs and embedding them into osbuild manifest or fetching them also during image build. Thinking about it now, I think it would be doable, we just currently don't have support for it AFAIK.

thozza: Yeah. However I think that depending on the package would introduce additional friction to our image definitions for the following reasons:

- using it would mean importing a different GPG key file on each Fedora version before verifying signatures during image build.
- this package is not available on RHEL, so we can't use it there.
- some users (or specific image types) care a lot about the image content, thus always including this package in the image could be undesirable and would force us to keep using the current approach anyway.

xsuchy: It is used in tests only. So the package needs to be presented only on CI systems. On host only. Right?

thozza: When we build image, we import the key on that system, install rpms and then verify their signature. So the package would need to be installed in the image and we would need to import a specific GPG key file shipped by it before verifying package signatures. Moreover, we use the same repository configuration format for both testing and for what is then shipped within our packages in Fedora / CS / RHEL. Therefore using the `distribution-gpg-keys ` package only for testing would not make our life easier :confused:

xsuchy: Or you can copy only that one file from distribution-gpg-keys to the image. 

Just trying to ease your life as maintaining the set of gpg separately is a pain for sure.

thozza: Don't take me wrong, I appreciate all your suggestion :slightly_smiling_face:

It is just that these GPG keys in our repo definitions haven't been a source of pain so far. We create these definitions only when a new Fedora version gets branched and then we just copy&paste them as needed. I'll take all the suggestions into consideration if we end up reworking this in the future, but I don't plan to touch it as part of this PR.

achilleas-k (dismissed):

achilleas-k:
Oops, I merged the snapshot update and it's conflicting on the test repo definitions.

test/data/repositories/fedora-39.json
praiskup: On @xsuchy's point
> Or you can copy only that one file from distribution-gpg-keys to the image*

This is what I meant actually (not that the package should be installed into every single image).  This is actually what we [do in Mock](https://github.com/rpm-software-management/mock/blob/3f351f5c269880c561e8ee1a6be592d682da7220/mock/py/mockbuild/package_manager.py#L388) (We copy all because there's no reason to "minimize" the target, that's small difference from osbuild). Sorry for confusion.

> Don't take me wrong, I appreciate all your suggestion 🙂

Yeah, no problem, that's my obsession to try to avoid any manual copy-pasting :-). It is surprising this is not a painful thing for you; gathering and maintaining all the GPG keys is one of the most painful things during Mock maintenance. Perhaps we support just too many chroots now.

> this package is not available on RHEL, so we can't use it there.

So far there was no motivation to ship, but this could be changed. Alternatively, you can bundle the released tarball.

supakeen (approved):

achilleas-k:
🎉🎉🎉

thozza:
I'll submit a PR in osbuild-composer shortly...

PR #201, opened 2023-10-12
packages: propagate excludes for installers

supakeen:
Installers use `installerPkgsKey` to propagate packages necessary for the installer environment. This didn't send through excluded packages leading to depsolve failures on Fedora 40 where both `sdubby` and `grubby` are selected; but `sdubby` conflicts and should be excluded.


ondrejbudai (dismissed):
Splendid! Thanks for the quick fix!

achilleas-k (dismissed):

thozza (dismissed):
Nice catch!

supakeen:
Disabled the auto-merge; this change breaks dracut as we unconditionally enable modules which are now not available due to excluded packages. I'll update the PR :)

ondrejbudai (approved):

thozza (approved):

PR #212, opened 2023-10-17
github: Merge dependabot PRs as a proper bot account

ondrejbudai:
Currently, we merge these PRs as the default GitHub Actions bot account.
However, I'm fairly sure that this bot account cannot spawn new actions
as a protection against recursive actions. Thus, the merge train will
fail because no actions will be spawned for it since the train was
created by the default bot.

This commit uses the Schutzbot's token in order to create the merge
train. This should allow GitHub to create the neccessary workflows when
Schutzbot auto-merges the PR.

Since secrets are not available on pull_request workflows, I changed
the trigger to pull_request_target. This shouldn't have any other
effects on the workflow.


achilleas-k (approved):
Very interesting.  Let's see if this works!

PR #216, opened 2023-10-20
Support arbitrary custom mountpoints

thozza:
- Add unit test to test specific mountpoint paths against the mountpoint policy. Replace distro-specific unit tests which were testing basically the same thing, but in which distro. This makes little sense, since all distros use the same mountpoint policy.
- Explicitly disallow mountpoints which can't / shouldn't be on a separate partition.
- Relax the mountpoint policy to allow arbitrary custom mountpoints and adjust the unit test accordingly.

Related to https://issues.redhat.com/browse/COMPOSER-2030

### DIFF
--- a/internal/pathpolicy/policies.go
+++ b/internal/pathpolicy/policies.go
@@ -7,7 +7,9 @@ var MountpointPolicies = NewPathPolicies(map[string]PathPolicy{
 	"/var":  {},
 	"/opt":  {},
 	"/srv":  {},
-	"/usr":  {},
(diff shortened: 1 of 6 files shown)


achilleas-k (commented):
Nice work and I really like the merged testing.  Is there more work to be done here (draft)?

thozza:
> Nice work and I really like the merged testing. Is there more work to be done here (draft)?

Not that I know of. I was just not sure if I didn't forget to fix some tests. :innocent:  But in general, this should be sufficient for this repo.

achilleas-k (approved):

PR #217, opened 2023-10-20
AWS boot testing for all supported image types

achilleas-k:
Enables boot testing on AWS for all supported image types.

The following changes were made to image definitions and tests:
- The `edge-ami` image uses the kernel options that are applied to all other EC2/AMI images.
- The command line flags for the `boot-aws` command were modified.
- The `ec2-sap` image for RHEL 8.4 is built with the `lm_sensors.service` disabled (in CI) because of a known issue.


supakeen (approved):

supakeen:
Awesome :)

achilleas-k:
EC2 images are created compressed (xz) and should be decompressed before pushing to boot.

thozza (commented):
The changes look good, but the failing EDGE-ami boot test is IMHO caused by the fact that EDGE images don't have `cloud-init` installed by default... :thinking: Thus the ssh key is never configured as authorized in the VM...

achilleas-k:
> The changes look good, but the failing EDGE-ami boot test is IMHO caused by the fact that EDGE images don't have `cloud-init` installed by default... 🤔 Thus the ssh key is never configured as authorized in the VM...

I've been thinking about how to test these (while working on other stuff) and I guess the only way is to add a user to the image BP.  I really wanted to have an empty baseline test for everything, but I guess there's no way to automate initial-setup (at least not the way we configure it?).  We'll need to do the same with the installers as well anyway.

achilleas-k:
It worked!  Kinda https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/5366641949

test/configs/edge-ostree-pull-user.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "edge-ostree-pull-user",
+  "blueprint": {
+    "customizations": {
+      "user": [
+        {
+          "name": "osbuild",
+          "key": "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPDyeUyVWvuSSOdrikiDWgNoCz0oYP4Ir68jJecn+XYc github.com/osbuild/images"
+        }
+      ]
```
thozza, line 10: Looking at the CI results, the user does not seem to be in any of the sudoers group and I'm not sure if password-less sudo is enabled at all in the image. This is what `cloud-init` does on first boot, so we need to do all of this on our own :disappointed:

achilleas-k: Yeah, that's what I meant by "kinda" :laughing: 

We don't have a clean way to create passwordless sudo.  I think there are a couple of options here:
1. Log in as root to run the script.
2. Add the user to the appropriate groups so it can run `systemctl` commands without root.

I'm going to try 2. first and if we ever need root to run things for the test (which I imagine we eventually will) we can figure it out then.

thozza: we could use the files customization:

```toml
[[customizations.files]]  
path = "/etc/sudoers.d/wheel-passwordless-sudo"  
mode = "0400"  
data = """  
%wheel ALL=(ALL) NOPASSWD: ALL  
"""
```

achilleas-k: Of course, file customizations aren't enabled for `edge-ami`

achilleas-k: Did a quick manual test and I think simply adding the user to the wheel group might be enough.  Let's find out.

achilleas-k: Silly me.  I was running `sudo systemctl is-system-running` and as far as I can tell, that doesn't need `sudo`.

achilleas-k:
Alright, we are booting and... failing on ARM for some reason:
```
● systemd-firstboot.service                                                                 loaded failed failed    First Boot Wizard
```

cgwalters:
FWIW in coreos and derivatives https://github.com/coreos/fedora-coreos-config/blob/6b1b56baab105ffac5cb290d5083f75e23437d2e/manifests/ignition-and-ostree.yaml#L25
(systemd-firstboot also makes little sense in a system that uses cloud-init)

achilleas-k:
> FWIW in coreos and derivatives https://github.com/coreos/fedora-coreos-config/blob/6b1b56baab105ffac5cb290d5083f75e23437d2e/manifests/ignition-and-ostree.yaml#L25 (systemd-firstboot also makes little sense in a system that uses cloud-init)

We don't add `cloud-init` to the `edge-ami`.  The image was added by the Edge team so I'm not sure if that was an oversight or deliberate.  Maybe @7flying can shed some light?

It's also strange that it only fails on aarch64 (and consistently).
I haven't gotten around to checking the log for the failure yet.

7flying:
> > FWIW in coreos and derivatives https://github.com/coreos/fedora-coreos-config/blob/6b1b56baab105ffac5cb290d5083f75e23437d2e/manifests/ignition-and-ostree.yaml#L25 (systemd-firstboot also makes little sense in a system that uses cloud-init)
> 
> We don't add `cloud-init` to the `edge-ami`. The image was added by the Edge team so I'm not sure if that was an oversight or deliberate. Maybe @7flying can shed some light?
> 
> It's also strange that it only fails on aarch64 (and consistently). I haven't gotten around to checking the log for the failure yet.

We deliberately did not add it since that would mean adding the RPM to every image using the same `edge-commit` and that did not make sense. Our docs recommend to add `cloud-init` in the blueprint instead.

achilleas-k:
> Our docs recommend to add `cloud-init` in the blueprint of the `edge-ami` instead.

But we can't add packages just to the ami.  The build just deploys the commit and creates the image and doesn't support layering packages on top.  Is that something we should consider?

There's also the (IMO better) alternative that I want to do eventually but might take some time: Build the image directly without pulling in an ostree commit and export both the image and the commit tarball.  That way we could build an `edge-ami` and get out both an `edge-commit.tar` that's specific to the AMI and the raw image for EC2.

7flying:
> > Our docs recommend to add `cloud-init` in the blueprint of the `edge-ami` instead.
> 
> But we can't add packages just to the ami. The build just deploys the commit and creates the image and doesn't support layering packages on top. Is that something we should consider?
> 

Good point, I meant to say, "to the commit that is just going to be used for the AMI". But I'll double check with docs to make that point clear.

Yes, ideally we would want a single edge commit to have all the packages that all the possible edge images would need, and then, depending on the image that you want to build some of them would be "disabled/removed", or we could use layers, but regardless, there shouldn't be more configuration burden for the user, if possible. 

> There's also the (IMO better) alternative that I want to do eventually but might take some time: Build the image directly without pulling in an ostree commit and export both the image and the commit tarball. That way we could build an `edge-ami` and get out both an `edge-commit.tar` that's specific to the AMI and the raw image for EC2.

Yes, having a single command-line spell that builds it all in one go would be the best solution.

7flying:
> Also of course, adding it would require it being in the commit, I don't know why I thought it would be included in the `ami`. But for our test, we can build an ostree commit that has it to test against the `edge-ami` specifically.
> 
> @7flying do you know if adding `cloud-init` to an `edge/iot-commit` conflict will conflict with `initial-setup`?

@achilleas-k no sorry, I did not test that

achilleas-k:
> > Also of course, adding it would require it being in the commit, I don't know why I thought it would be included in the `ami`. But for our test, we can build an ostree commit that has it to test against the `edge-ami` specifically.
> > @7flying do you know if adding `cloud-init` to an `edge/iot-commit` conflict will conflict with `initial-setup`?
> 
> @achilleas-k no sorry, I did not test that

I think I'll add this as part of our test configs: Build an edge commit with cloud-init installed and use it as the base image for an edge-ami and then boot test it with cloud-init.

But I'll leave that for a follow-up PR.
For now, here, I'll see if I can figure out what's making the firstboot service fail to start (could be nothing, but it could be a side effect of some bug) and take it from there.

Thanks for the input!

achilleas-k:
Looked into the failure and the first boot service is failing with
```
Failed to set up standard input: No such device
Failed at step STDIN spawning systemd-firstboot: No such device
```

I also noticed the following in the journal:
```
kernel: Warning: unable to open an initial console.
```

Adding `console=tty0` to the kernel args fixes the issue.

achilleas-k:
I modified the edge-ami image type to include all the kernel options that we add to other EC2/AMI image types.

achilleas-k:
One failure left: https://gitlab.com/redhat/services/products/image-builder/ci/images/-/pipelines/1055416885 (ec2-sap on RHEL 8.4).

achilleas-k:
> One failure left: https://gitlab.com/redhat/services/products/image-builder/ci/images/-/pipelines/1055416885 (ec2-sap on RHEL 8.4).

Most likely it's an issue with just being slower to start:
```
$ sudo systemd-analyze
Startup finished in 2.601s (kernel) + 9.615s (initrd) + 1min 33.448s (userspace) = 1min 45.665s
multi-user.target reached after 38.479s in userspace
```

achilleas-k:
> > One failure left: https://gitlab.com/redhat/services/products/image-builder/ci/images/-/pipelines/1055416885 (ec2-sap on RHEL 8.4).
> 
> Most likely it's an issue with just being slower to start:
> 
> ```
> $ sudo systemd-analyze
> Startup finished in 2.601s (kernel) + 9.615s (initrd) + 1min 33.448s (userspace) = 1min 45.665s
> multi-user.target reached after 38.479s in userspace
> ```
> 
> The keyscan, which we use to wait for the system, quits after 10 tries with 10 second intervals.
> 
> EDIT: Though it also failed to start `lm_sensors.service`, so the system is in any case `degraded`.

So the timeout increase worked and the test script ran but like I saw in my local test, the `lm_sensors` service didn't start:
https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/5420205170

This is only an issue on RHEL 8.4 SAP, so I'm guessing the problem is:
-  lm_sensors is only installed on SAP images
- 8.4 kernel is too old..?

thozza (approved):
LGTM

supakeen:
Absolutely love it.

PR #219, opened 2023-10-20
test: Add PR number to build info and share test caches

achilleas-k:
This PR makes two small changes to the test pipelines:
1. Adds the PR number to the build info when a new image is built.  This number is printed when filtering builds to make it easier to determine why image builds are skipped and it's also used to list images build in all pipelines of a given PR in the `check-build-coverage` jobs.  Eventually, we can use this information to add messages to the PR so it's easier to track what is being built by each pipeline run.
2. Puts the rpmmd and build info cache in the working directory and configures build pipeline generators to share that cache.  This should speed up test generation.  It'll be especially good for speeding up depsolving for manifest generation.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -42,6 +42,10 @@ check-build-coverage:
   script:
     - apk add -U go python3 gpgme-dev s3cmd
     - ./test/scripts/check-build-coverage ./s3configs/
+  cache:
+    key: testcache
+    paths:
+      - .cache/osbuild-images
 
 finish:
   stage: finish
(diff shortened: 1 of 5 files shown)


ochosi (approved):
Can't judge how the cache bits work, but the PR info looks good! Thank you so much for this - CI speedups ftw ❤️

achilleas-k:
Looking at the numbers of the jobs, it doesn't seem to have had much of an effect.  Maybe there's a smarter way to do it, but I think it doesn't hurt to keep it.

achilleas-k:
I want to add a starter job that pulls in all the `info.json` files from the build cache and then shares it with all the following pipelines through the cache.  That should make things a bit faster.  I'll do that in a followup because I'd like to get the PR number part in for jobs to start adding it.

PR #227, opened 2023-10-26
Fix implicit memory aliasing in for loops

bcl:
Fix all instances of gosec G601: Implicit memory aliasing in for loop.

(cherry picked from commit 9a445e6af94e209a786695b315cc9adf8f2e383c)

### DIFF
--- a/cmd/osbuild-pipeline/main.go
+++ b/cmd/osbuild-pipeline/main.go
@@ -140,14 +140,15 @@ func main() {
 	}
 
 	repos := make([]rpmmd.RepoConfig, len(composeRequest.Repositories))
-	for i, repo := range composeRequest.Repositories {
+	for idx := range composeRequest.Repositories {
+		repo := composeRequest.Repositories[idx]
 		repoName := repo.Name
 		if repoName == "" {
-			repoName = fmt.Sprintf("repo-%d", i)
+			repoName = fmt.Sprintf("repo-%d", idx)
 		}
 		repoId := repo.Id
 		if repoId == "" {
-			repoId = fmt.Sprintf("repo-%d", i)
+			repoId = fmt.Sprintf("repo-%d", idx)
 		}
 		var urls []string
 		if repo.BaseURL != "" {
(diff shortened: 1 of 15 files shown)


bcl:
I am fairly confident this fixes the issue in https://issues.redhat.com/browse/COMPOSER-2055 -- I setup a VM, set check_gpg on an extra source and all of the manifest entries (as shown by `composer-cli compose info --json UUID` were missing `check_gpg`. I added this patch to the osbuild-composer branch and the next build had correct `check_gpg` settings. I turned it off for my extra source, and just that one package showed it as false.

NOTE: osbuild-composer rhel-9.3.0 branch will need to pull in a new images and be rebuilt once this is merged.

cmd/osbuild-pipeline/main.go
```diff
@@ -140,14 +140,15 @@ func main() {
 	}
 
 	repos := make([]rpmmd.RepoConfig, len(composeRequest.Repositories))
-	for i, repo := range composeRequest.Repositories {
+	for idx := range composeRequest.Repositories {
+		repo := composeRequest.Repositories[idx]
```
cgwalters, line 144: I think your fix is correct, but the more idiomatic Go solution is "rebinding" in the loop instead of re-indexing the array.  Basically

```
for i, repo in composeRequest.Repositories {
    repo := repo
```

There's a lot more on this in https://go.dev/blog/loopvar-preview

achilleas-k: I personally find using the index more readable than making a copy / rebinding and I've used it all over the place, even when not strictly necessary.  It's also often useful when needing to index two aligned arrays (for copying or comparison), which admittedly doesn't apply here.

Great to see there's a change coming though.  I wasn't aware.  Thanks for the heads-up!

achilleas-k (approved):

cmd/osbuild-pipeline/main.go
achilleas-k: Either way, since this is a backport of an existing change, there's no need to add extra work modifying the solution.

PR #237, opened 2023-11-01
osbuild: rework device mount names handling

thozza:
The ability to create arbitrary mountpoints under the filesystem root revealed an issue in the code generating device names for mount options of osbuild stages. Specifically, it would generate the same device name for `/` and `/root` mountpoints. Moreover, other places were making assumptions which should not be relied on, e.g.:

- `GenCopyFSTreeOptions()` could happily replace existing device mount with a different device, but the same name.
- `GenCopyFSTreeOptions()` expected that the mount name for entity with `/` mountpoint will be always called `root`.
- `RawOSTreeImage.serialize()` expected that the mount name for entity with `/` mountpoint will be always called `root`.

Remove these assumptions from the code and explicitly check for device name conflicts.

Rework the function escaping mountpoint paths to use a similar algorithm as `systemd-escape`. This will assure that it will produce different escaped path for different paths,  but still the same escaped path if the path did not change.

As a minor change, change the osbuild code handling LVM to not append `vg` to LVM VG names, but use it as is. This is consistent with how LVM LV names are handled. All partition table definitions and code converting PT to LVM always use `rootvg` as the name got the root LVM VG. Which previously resulted in the VG having name `rootvgvg`.

### DIFF
--- a/pkg/osbuild/device.go
+++ b/pkg/osbuild/device.go
@@ -153,7 +153,7 @@ func deviceName(p disk.Entity) string {
 
 	switch payload := p.(type) {
 	case disk.Mountable:
-		return pathdot(payload.GetMountpoint())
+		return pathEscape(payload.GetMountpoint())
 	case *disk.LUKSContainer:
 		return "luks-" + payload.UUID[:4]
 	case *disk.LVMVolumeGroup:
@@ -206,12 +206,21 @@ func getDevices(path []disk.Entity, filename string, lockLoopback bool) (map[str
 	return do, parent
 }
 
-func pathdot(path string) string {
-	if path == "/" {
-		return "root"
+// pathEscape implements similar path escaping as used by systemd-escape
+// https://github.com/systemd/systemd/blob/c57ff6230e4e199d40f35a356e834ba99f3f8420/src/basic/unit-name.c#L389
+func pathEscape(path string) string {
+	if len(path) == 0 || path == "/" {
+		return "-"
 	}
 
-	path = strings.TrimLeft(path, "/")
+	path = strings.Trim(path, "/")
 
-	return strings.ReplaceAll(path, "/", ".")
+	escapeChars := func(s, char string) string {
+		return strings.ReplaceAll(s, char, fmt.Sprintf("\\x%x", char[0]))
+	}
+
+	path = escapeChars(path, "\\")
+	path = escapeChars(path, "-")
+
+	return strings.ReplaceAll(path, "/", "-")
 }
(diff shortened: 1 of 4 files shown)


pkg/osbuild/device.go
```diff
@@ -206,12 +206,21 @@ func getDevices(path []disk.Entity, filename string, lockLoopback bool) (map[str
 	return do, parent
 }
 
-func pathdot(path string) string {
-	if path == "/" {
-		return "root"
+// pathEscape implements similar path escaping as used by systemd-escape
+// https://github.com/systemd/systemd/blob/main/src/basic/unit-name.c#L389
```
achilleas-k, line 210: NITPICK: Let's get the permalink for this so it doesn't link to a different line when the file changes in `main`.

achilleas-k (dismissed):
Nice work.  I agree with the decision to use something that exists already for escaping paths.  It might not be the most readable (when there are `-` or other non-standard characters), but in most cases it's perfectly readable.

Just a tiny nitpick for the link in the comment.

ondrejbudai (approved):
Love this!

pkg/osbuild/device.go
thozza: fixed

thozza:
CI revealed one more place which assumed that the FS root mount name is always `root`. Specifically the `RawOSTreeImage.serialize()` method. I fixed it in the same way as the `GenCopyFSTreeOptions()` function.

Besides that, I replaced the link to systemd escaping code to be a permalink and rebased the PR on top of `main`.

thozza:
> Nice work. I agree with the decision to use something that exists already for escaping paths. It might not be the most readable (when there are `-` or other non-standard characters), but in most cases it's perfectly readable.

I agree with the single dash (`-`) not being very obvious or self-explanatory when used for the FS root mountpoint. However, the `org.osbuild.copy` stage options would still make it reasonably possible to determine what `-` represents:

```json
{
  "type": "org.osbuild.copy",
  "inputs": {
    "root-tree": {
      "type": "org.osbuild.tree",
      "origin": "org.osbuild.pipeline",
      "references": [
        "name:os"
      ]
    }
  },
  "options": {
    "paths": [
      {
        "from": "input://root-tree/",
        "to": "mount://-/"
      }
    ]
  },
  "devices": {
    "-": {
      "type": "org.osbuild.lvm2.lv",
      "parent": "rootvg",
      "options": {
        "volume": "rootlv"
      }
    },
    ...
  },
  "mounts": [
    {
      "name": "-",
      "type": "org.osbuild.ext4",
      "source": "-",
      "target": "/"
    },
    {
      "name": "boot",
      "type": "org.osbuild.ext4",
      "source": "boot",
      "target": "/boot"
    },
    ...
  ]
}
```

I'm open to suggestions how to improve it if anyone has some...

achilleas-k (approved):
Wonderful!

PR #239, opened 2023-11-02
gen-manifest: add a shortcut option for providing single image config

thozza:
The gen-manifest accepts an image config map via the `-config` option,
which is good for CI and testing purposes.

However, imagine a situation when:
- you are making changes to code in the repository which affects
  generated manifests,
- you want to use a custom image config to test these changes and
  generate manifests to build them using osbuild

In this case, one can filter out distro / arch / image type using the
`gen-manifest` CLI options, instead of creating the image config map.
Passing an ad-hoc image config to `gen-manifests` on the CLI to use for
all image types and filtering what gets generated using the CLI options
is IMO much nicer (or easier to use) for development, than the current
approach (which involves basically creating two JSON files and editing
one of them each time one wants to use a different image config).

Change the behavior of the original `-config` option to accept a
single image config which will be used for all image types / distros /
arches. Note that this will almost certainly fail for some images, but
the intention of this option is to be used for development and together
with filtering options such as `-distros`, `-arches` and `-image-types`.

The original option is now available under a new name `-config-map`,
which is better aligned with what it does.

Example use:
```
$ go run cmd/gen-manifests/* -distros fedora-38 -images minimal-raw -arches x86_64 -config test-fs-customizations.json
```

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -215,6 +215,16 @@ func loadConfigMap(configPath string) BuildConfigs {
 	return cm
 }
 
+// loadImgConfig loads a single image config from a file and returns a
+// a BuildConfigs map with the config mapped to all distros, arches, and
+// image types.
+func loadImgConfig(configPath string) BuildConfigs {
+	cm := make(BuildConfigs)
+	config := loadConfig(configPath)
+	cm.Insert("*", "*", "*", config)
+	return cm
+}
+
 type manifestJob func(chan string) error
 
 func makeManifestJob(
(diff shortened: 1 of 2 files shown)


cmd/gen-manifests/main.go
```diff
@@ -602,14 +612,15 @@ func u(s string) string {
...
-	var outputDir, cacheRoot, configPath string
+	var outputDir, cacheRoot, configPath, imgConfigPath string
 	var nWorkers int
 	var metadata, skipNoconfig, skipNorepos bool
 	flag.StringVar(&outputDir, "output", "test/data/manifests/", "manifest store directory")
 	flag.IntVar(&nWorkers, "workers", 16, "number of workers to run concurrently")
 	flag.StringVar(&cacheRoot, "cache", "/tmp/rpmmd", "rpm metadata cache directory")
 	flag.BoolVar(&metadata, "metadata", true, "store metadata in the file")
 	flag.StringVar(&configPath, "config", "test/config-map.json", "configuration file mapping image types to configs")
+	flag.StringVar(&imgConfigPath, "image-config", "", "image config file to use for all images (overrides -config)")
```
achilleas-k, line 623: Let's rename `-config` to `-config-map` and use `-config` for the new one.  I think that would make the flag names clearer.  And it's a development tool so we don't really care about backwards compatibility.

achilleas-k (changes requested):

achilleas-k:
Not a bad idea.  It does require filtering the image types though because you can never have one config that works for everything (simplified-installer requires `installation_device` while adding it  breaks all the other types), but it's a dev tool so it should be convenient and doesn't need to make sense for all cases.

thozza:
> Not a bad idea. It does require filtering the image types though because you can never have one config that works for everything (simplified-installer requires `installation_device` while adding it breaks all the other types), but it's a dev tool so it should be convenient and doesn't need to make sense for all cases.

I know and you are of course 100% right.

I started with a workflow such as:

1. Make changes to `images` repo.
2. Replace the `images` package with the one I modified in `osbuild-composer`.
3. Build and install composer binaries or RPMs.
4. Trigger a compose with a BP which contains desired customizations.

And once I ran into an issue, I had to repeat all of these steps again. This takes ages and it very inefficient for development.

So I moved to just generating manifests using `gen-manifest` and building it with `osbuild` directly. This worked well, but it required creating the config map and the config itself. So in this case, the need to filter our specific distro / arch / image type is assumed when using just one image config. However, that was my whole goal - have a workflow to iterate quickly on image definitions, while being able to actually build images and boot them.

I can also expand the help message a bit more if you would find it helpful.

cmd/gen-manifests/main.go
thozza: I didn't want to break backward compatibility, but I agree with your suggestion :wink:

thozza: Done (hopefully I didn't miss any place where to change the option name :innocent: ).

achilleas-k (approved):
Looks good.  Thank you!!

thozza:
> Don't get me wrong, I wasn't against the change, just pointing out an edge case. As a side note though, based on your workflow, there's also the `./cmd/build` command that will generate the manifest (save it in the output directory) and then build it immediately, and that takes a single config as input.

No worries, I didn't think that you were against the change.

Thanks for pointing out the `./cmd/build` command. It does look helpful even for development purposes :thinking:

PR #242, opened 2023-11-06
FIPS mode support

mmartinv:
Enables system FIPS mode at build time by defining a new 'fips' blueprint 
customization. Depends on osbuild/osbuild/pull/1423

- blueprint(fips): add FIPS blueprint
- distro(rhel): enable FIPS customization for edge images
- distro(rhel): add required kernel cmdline for FIPS customization
- image(anaconda): allow additional kernel cmdline options
- osbuild(stages): Add the new FIPS stage
- manifest(fips): enable FIPS stage
- osbuild(fips): add 'boot' kernel cmdline option in FIPS mode
- test(fips): add tests for FIPS enabled edge images

### DIFF
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -742,6 +742,13 @@ func (p *OS) serialize() osbuild.Pipeline {
 		}))
 	}
 
+	if osbuild.ContainsFIPSKernelOption(p.KernelOptionsAppend) {
+		fipsStage := osbuild.NewFIPSStage(&osbuild.FIPSStageOptions{
+			BootCfg: false,
+		})
+		pipeline.AddStage(fipsStage)
+	}
+
 	if p.SElinux != "" {
 		pipeline.AddStage(osbuild.NewSELinuxStage(&osbuild.SELinuxStageOptions{
 			FileContexts:     fmt.Sprintf("etc/selinux/%s/contexts/files/file_contexts", p.SElinux),
--- a/pkg/osbuild/fips_stage.go
+++ b/pkg/osbuild/fips_stage.go
@@ -1,0 +1,28 @@
+package osbuild
+
+import (
+	"strings"
+)
+
+type FIPSStageOptions struct {
+	BootCfg bool `json:"boot_cfg"`
+}
+
+func (FIPSStageOptions) isStageOptions() {}
+
+// NewFIPSStage creates FIPSStage
+func NewFIPSStage(options *FIPSStageOptions) *Stage {
+	return &Stage{
+		Type:    "org.osbuild.fips",
+		Options: options,
+	}
+}
+
+func ContainsFIPSKernelOption(kernelOpts []string) bool {
+	for _, kernelOption := range kernelOpts {
+		if strings.Contains(kernelOption, "fips=1") {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/blueprint/customizations.go
+++ b/pkg/blueprint/customizations.go
@@ -24,6 +24,7 @@ type Customizations struct {
 	Directories        []DirectoryCustomization  `json:"directories,omitempty" toml:"directories,omitempty"`
 	Files              []FileCustomization       `json:"files,omitempty" toml:"files,omitempty"`
 	Repositories       []RepositoryCustomization `json:"repositories,omitempty" toml:"repositories,omitempty"`
+	FIPS               *bool                     `json:"fips,omitempty" toml:"fips,omitempty"`
 }
 
 type IgnitionCustomization struct {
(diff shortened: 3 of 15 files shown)


bcl:
Is all of this really needed? Why not just document setting it in the kernel customization append string? If it were doing more than just passing it through to osbuild I could see a need for it, but it seems like quite a bit of code to support adding a simple string.

mmartinv:
> Is all of this really needed? Why not just document setting it in the kernel customization append string? If it were doing more than just passing it through to osbuild I could see a need for it, but it seems like quite a bit of code to support adding a simple string.

When I started the implementation I also thought that the approach you are mentioning would be simpler and more elegant but it turns out that it's not and it also implies some additional concerns: It requires all the image types to support the `Kernel` blueprint customization. Enabling this support for all of them needs more code and It would allow the user to add additional kernel arguments other than `fips=1` to images that currently don't support any kind of `Kernel` customization. This is something I am not sure we want to allow.

runcom:
> Is all of this really needed? Why not just document setting it in the kernel customization append string? If it were doing more than just passing it through to osbuild I could see a need for it, but it seems like quite a bit of code to support adding a simple string.

I think it's important that we provide an easy/opinionated way of doing that to fully make sure the images we produce are just doing their job without user interference/bugs/typos - a knob like `fips=true` in the blueprint makes sure of that imo (similarly to what the MCO does in OpenShift too, you don't have to worry about kargs and such)

achilleas-k:
I agree with having a higher level, abstracted option that translates to a group of changes that need to be done together.  But even beyond that, the option enables a stage that runs `fips-mode-setup` so either way, we would need an explicit way to enable it.

pkg/manifest/os.go
```diff
@@ -742,6 +742,13 @@ func (p *OS) serialize() osbuild.Pipeline {
 		}))
 	}
 
+	if osbuild.ContainsFIPSKernelOption(p.KernelOptionsAppend) {
+		fipsStage := osbuild.NewFIPSStage(&osbuild.FIPSStageOptions{
+			BootCfg: false,
+		})
+		pipeline.AddStage(fipsStage)
+	}
```
achilleas-k, line 750: Adding the stage based on the existence of the kernel argument is a bit unusual for the way we do things in the pipeline generators.  I can kind of see the logic behind this: the stage will be added even if the user sets `fips=false` in the blueprint (or omits it) but instead adds `fips=1` to the kernel options.  But since we're exposing it as a high level option, I don't think we need this "automatic" way of detecting its inclusion and we should instead use the user-provided to option for all conditions.

Let's instead add a `Fips bool` to the `OS struct` (and `OSTreeDeployment`) and make all changes based on that.

pkg/osbuild/fips_stage.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func ContainsFIPSKernelOption(kernelOpts []string) bool {
+	for _, kernelOption := range kernelOpts {
+		if strings.Contains(kernelOption, "fips=1") {
+			return true
+		}
+	}
+	return false
+}
```
achilleas-k, line 28: Given the other comment, this function isn't needed.

achilleas-k (changes requested):

bcl:
> It requires all the image types to support the `Kernel` blueprint customization.

Ah! Right, I forgot that customization support is inconsistent (and difficult for users to figure out). Thanks!

pkg/manifest/os.go
mmartinv: I have been thinking a lot about this and I changed my mind and I would prefer to use the `fips=1` kernel argument as the user API for enabling FIPS mode to avoid this kind of confusion. This way we will be coherent and will maintain the current user API for all the supported FIPS installations.

thozza: I understand your reasoning and motivation for this decision, but I'm not in favor of this approach for the following reasons:

- Blueprints are higher-level API and the customizations tend to abstract lower level configuration done to the image.
- Specifying `fips=1` on the kernel cmdline is not really sufficient to build image which uses FIPS mode. Based on this and the other PRs, one needs to potentially append the `boot=` option to kernel cmdline, set the crypto policy and modify dracut config. Gating "FIPS enabled image" customization behind just adding `fips=1` to the kernel cmdline IMO creates false expectations that this is really enough to have FIPS enables system.
- Users with older osbuild-composer, which won't support this feature, may still think that appending `fips=1` to kernel cmdline will do the trick. If there was an explicit option, they would get an error instead.

mmartinv:
This PR is based on the `fips-mode-setup` stage which is not going to be merged so closing this in favor of osbuild/images/pull/253 which uses `update-crypto-policies` stage.

pkg/manifest/os.go
cgwalters: > Specifying fips=1 on the kernel cmdline is not really sufficient to build image which uses FIPS mode. Based on this and the other PRs, one needs to potentially append the boot= option to kernel cmdline, set the crypto policy and modify dracut config

The idea here is that those things should *automatically follow* the `fips=1` karg; that's what the code in both Anaconda and RHEL CoreOS are doing.  I think it makes sense to do here too.

PR #243, opened 2023-11-06
Add support for building images from ostree containers and the Fedora CoreOS qcow2 image type

achilleas-k:
This PR adds support for the org.osbuild.ostree.deploy.container stage and a new image type that uses it called `coreos-qcow2` for Fedora.

I tested the new image type locally with the container we added to the gitlab registry for testing the sample manifest in osbuild.
Fedora CoreOS booted

Some notes and open questions:
- I use the same pipeline as the regular ostree deployment with a condition on the type of source.  I think it wasn't worth making a whole separate pipeline for this, but we can revisit this decision if the two start diverging.
- The new image type is added to Fedora which means that if this is merged and updated in osbuild-composer it will be available to build on-prem.  If we want this, we'll have to figure out the command line options for specifying a container source and potentially a target-imgref.
- When specifying a container source to deploy, you must specify the ref (otherwise the default will be assumed).  This requirement also exists for ostree commits from a repo, but I wonder: Is there a way to determine the container's ostree ref without downloading it?

You can generate the manifest for this image with:
```
go run ./cmd/gen-manifests --distros fedora-38 --arches x86_64 --images coreos-qcow2 --output .
```
or build it directly with
```
go build -o bin/build ./cmd/build
sudo ./bin/build --distro fedora-38 --image coreos-qcow2 --output builds --config ./test/configs/ostree-container-pull.json
```

You can also add a user to the config file which will create a user after the deployment before finalising the image:
```json
{
  "name": "ostree-container-pull",
  "ostree": {
    "ref": "ostree/1/1/0",
    "container": "registry.gitlab.com/redhat/services/products/image-builder/ci/images/fedora-coreos:stable",
    "tls-verify": true
  },
  "blueprint": {
    "customizations": {
      "user": [
        {
          "name": "<login>",
          "password": "<password>",
          "key": "<pubkey>"
        }
      ]
    }
  }
}
```


pkg/image/ostree_disk.go
```diff
@@ -26,21 +27,22 @@ type OSTreeDiskImage struct {
...
 
 	KernelOptionsAppend []string
 	Keyboard            string
 	Locale              string
 
 	Filename string
 
-	Ignition         bool
-	IgnitionPlatform string
+	IgnitionPlatform *string
```
bcl, line 45: Why make this a pointer instead of using empty string to indicate it isn't set?

achilleas-k: I don't really have a good answer.  I think I picked up a habit of making optionals `nil`-able but that's not a good reason.  You're right, an plain string with empty as the "off" value would work better and require less checks.

cgwalters: mumble mumble `Option<String>`

pkg/distro/fedora/images.go
```diff
@@ -641,6 +639,82 @@ func iotSimplifiedInstallerImage(workload workload.Workload,
...
+	if err != nil {
+		return nil, err
+	}
+
+	img.KernelOptionsAppend = []string{
+		"rw",
+		"console=tty0",
+		"console=ttyS0",
+	}
+	img.Keyboard = "us"
```
ondrejbudai, line 676: Does `Keyboard` and `Locale` options have any effects on the manifest?

achilleas-k: Adds the appropriate stages to the deployment pipeline: https://github.com/achilleas-k/images/blob/b19437c5d35c5c9028a37b14048352360288e79a/pkg/manifest/ostree_deployment.go#L361-L377

achilleas-k: Those should actually be part of some image specific configuration (imageconfig or default workload or something) rather than hardcoded in the image function, but that's a more general problem I'm cleaning up separately.

pkg/image/ostree_disk.go
achilleas-k: Done!

pkg/distro/fedora/images.go
ondrejbudai: Makes me wonder: Should we meddle with the default keyboard/locale while deploying the container? I think I would rather leave the defaults as defined in the container.

achilleas-k: Dropped.  But we should allow customizing these through the blueprint.  For all ostree images.  And the Anaconda ISO through kickstart (now that we added the commands in osbuild).

I think it's time to do the rewrite of the allowed options for image types.

ondrejbudai: Agreed, for every image type, we should support as many customizations possible. :)

ondrejbudai (commented):
I'm very happy with this PR and I haven't found any issues. However, it's a big one, so I would love a third pair of eyes to take a look. @achilleas-k should we ask someone explicitly? Any ideas whom? :)

pkg/distro/fedora/distro.go
```diff
@@ -689,6 +707,7 @@ func newDistro(version int) distro.Distro {
 			UEFIVendor: "fedora",
 		},
 		iotQcow2ImgType,
+		coreosQcow2ImgType,
```
teg, line 710: What about aarch64?

teg (dismissed):
I think I noticed an oversight, but would rather fix that in a follow-up than rebasing it tbh, this is plenty useful as it is.

pkg/distro/fedora/distro.go
achilleas-k: I added it at first but then realised I needed to get the base container in our CI registry to test it and decided to get this open so people can review and I can add it in a follow-up.

pkg/manifest/ostree_deployment.go
```diff
@@ -197,6 +193,11 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 		)
 	}
 
+	pipeline.AddStage(osbuild.NewOSTreePullStage(
+		&osbuild.OSTreePullStageOptions{Repo: repoPath, Remote: p.Remote.Name},
+		osbuild.NewOstreePullStageInputs("org.osbuild.source", commit.Checksum, commit.Ref),
+	))
+
```
dustymabe: We don't really need a pull stage at all anymore since https://github.com/osbuild/osbuild/pull/1393 - just specify the commit as an input to the deploy stage.

pkg/manifest/ostree_deployment.go
```diff
@@ -193,51 +188,58 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
...
+						Name:        p.Remote.Name,
+						URL:         remoteURL,
+						ContentURL:  p.Remote.ContentURL,
+						GPGKeyPaths: p.Remote.GPGKeyPaths,
+					},
 				},
 			},
-		},
-	))
+		))
```
dustymabe, line 244: Is this stage always added? IIUC it's not always needed. For example, in Fedora we deliver remote configuration via the `fedora-repos-ostree` RPM:

```
[core@rpi4 ~]$ rpm -ql fedora-repos-ostree
/etc/ostree/remotes.d
/etc/ostree/remotes.d/fedora-compose.conf
/etc/ostree/remotes.d/fedora.conf
```

- [source](https://src.fedoraproject.org/rpms/fedora-repos/blob/rawhide/f/fedora.conf)

pkg/manifest/ostree_deployment.go
```diff
@@ -193,51 +188,58 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
...
+		pipeline.AddStage(osbuild.NewOSTreeFillvarStage(
+			&osbuild.OSTreeFillvarStageOptions{
+				Deployment: osbuild.OSTreeDeployment{
+					OSName: p.osName,
+					Ref:    ref,
+				},
 			},
-		},
-	))
+		))
```
dustymabe, line 253: same for this one. we might not always want it.

pkg/image/ostree_disk.go
```diff
@@ -50,12 +53,27 @@ type OSTreeDiskImage struct {
 func NewOSTreeDiskImage(commit ostree.SourceSpec) *OSTreeDiskImage {
 	return &OSTreeDiskImage{
 		Base:         NewBase("ostree-raw-image"),
-		CommitSource: commit,
+		CommitSource: &commit,
+	}
+}
+
+func NewOSTreeContainerDiskImage(container container.SourceSpec, ref string) *OSTreeDiskImage {
```
dustymabe, line 59: to tamp down on confusion we could name this one `NewFromContainerOSTreeDiskImage()` and rename the other one to be `NewFromOSTreeCommitOSTreeDiskImage()`. The words "Container" and "Image" are both getting overloaded a lot and when I first read this I could have though we were creating a Container Image.

pkg/manifest/ostree_deployment.go
```diff
@@ -204,13 +194,10 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 	kernelOpts := osbuild.GenImageKernelOptions(p.PartitionTable)
 	kernelOpts = append(kernelOpts, p.KernelOptionsAppend...)
 
-	if p.ignition {
-		if p.ignitionPlatform == "" {
-			panic("ignition is enabled but ignition platform ID is not set")
-		}
+	if p.IgnitionPlatform != "" {
 		kernelOpts = append(kernelOpts,
 			"coreos.no_persist_ip", // users cannot add connections as we don't have a live iso, this prevents connections to bleed into the system from the ign initrd
```
dustymabe, line 199: this looks very specific to a particular caller (Automotive? Edge?)

I think I would pull out `coreos.no_persist_ip` into something else. There is a new `kernel-cmdline.bls-append` stage in https://github.com/osbuild/osbuild/pull/1429

pkg/manifest/ostree_deployment.go
```diff
@@ -318,7 +305,7 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 		pipeline.AddStage(grpStage)
 	}
 
-	if p.ignition {
+	if p.IgnitionPlatform != "" {
```
dustymabe, line 308: I suspect a lot of this is specific to IoT/Edge/Automotive

One thing that is interesting is that the `NewIgnitionStage` `Network:` stuff where `systemd.firstboot=off` and `systemd.condition-first-boot=true` get set is kind of an abuse of something that isn't really public API. It's meant to be an internal implementation detail.

pkg/ostree/ostree.go
```diff
@@ -72,6 +77,9 @@ type ImageOptions struct {
 	// Indicate if the 'org.osbuild.rhsm.consumer' secret should be added when pulling from the
 	// remote.
 	RHSM bool `json:"rhsm"`
+
+	// TLS verification for container sources.
+	TLSVerify *bool `json:"tls-verify"`
```
dustymabe, line 82: since there are multiple things here that are "remote" artifacts it might be a little confusing to have this just apply to one of them and still name it `TLSVerify`.

One thing we could do here is break out sets of options that go together into subordinate objects.

i.e.

```
type ContainerImageOptions struct {
    Container string `json:"container"`
    TLSVerify *bool `json:"tls-verify"`
}
...
```

pkg/distro/fedora/images.go
```diff
@@ -639,6 +639,80 @@ func iotSimplifiedInstallerImage(workload workload.Workload,
...
+	if err != nil {
+		return nil, err
+	}
+	img.PartitionTable = pt
+
+	img.Filename = t.Filename()
+	img.Compression = t.compression
+
+	return img, nil
+}
```
dustymabe, line 714: There is a lot going on here. Maybe @achilleas-k we can get together and step through some of it.

pkg/distro/fedora/distro.go
```diff
@@ -368,6 +368,24 @@ var (
...
+			Locale: common.ToPtr("en_US.UTF-8"),
+		},
+		defaultSize:         10 * common.GibiByte,
+		rpmOstree:           true,
+		bootable:            true,
+		image:               coreosImage,
+		buildPipelines:      []string{"build"},
+		payloadPipelines:    []string{"ostree-deployment", "image", "qcow2"},
+		exports:             []string{"qcow2"},
+		basePartitionTables: iotBasePartitionTables,
```
dustymabe, line 387: I probably need to dig into `iotBasePartitionTables` a little bit more to understand why they are being used here.

pkg/manifest/ostree_deployment.go
```diff
@@ -340,7 +340,8 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 		}
 	}
 
-	if !hasRoot {
+	if !hasRoot && p.ostreeSpec != nil {
+		// NOTE: fails on container-based deployments
```
dustymabe, line 344: not sure either.. from my side this feels like something that would be done when composing the ostree commit or oscontainer rather than when composing an image. IOW I feel like this shouldn't be part of Image Builder at all?

pkg/distro/distro_test.go
```diff
@@ -46,6 +46,10 @@ func TestImageType_PackageSetsChains(t *testing.T) {
 							URL: "https://example.com", // required by some image types
 						},
 					}
+					if imageType.Name() == "coreos-qcow2" {
+						options.OSTree.Container = "https://registry.example.com/image"
```
dustymabe, line 50: is this supposed to be a container pullspec? those don't often include `https://` do they?

pkg/osbuild/ostree_deploy_container_stage.go
```diff
@@ -0,0 +1,68 @@
...
+func (OSTreeDeployContainerInputs) isStageInputs() {}
+
+func (inputs OSTreeDeployContainerInputs) validate() error {
+	if ncontainers := inputs.Images.References.Len(); ncontainers != 1 {
+		return fmt.Errorf("stage requires exactly 1 input container (got %d)", ncontainers)
+	}
+	return nil
+}
+
+// A new org.osbuild.ostree.init stage to create an OSTree repository
```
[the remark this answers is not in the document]
achilleas-k: Very much so.

pkg/manifest/ostree_deployment.go
achilleas-k: I want to keep the existing images unchanged while adding support for the new stuff to avoid any side effects or confusion.  I intend to drop the pull stage in a follow-up, but for now I'd like the existing images to change as little as possible and so I'm limiting their changes to just a stage reordering.

pkg/manifest/ostree_deployment.go
achilleas-k: We used to configure this before the package was created.  I have [a separate PR](https://github.com/osbuild/images/pull/95) that's been open for a while but configuring the remote through the package has been causing issues with our tests in osbuild-composer (https://github.com/osbuild/osbuild-composer/pull/3635).

I've been looking into it on and off and have no idea what the issue is, but I'd love to get your input on it.

pkg/manifest/ostree_deployment.go
achilleas-k: In what situation would we not need this?  We've always had it for deployments so I always considered it necessary.  The osbuild stage referenceshttps://github.com/coreos/fedora-coreos-config/blob/0b50084c9605d40c01c083e8b7b8efd9695d13a5/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-populate-var.sh.

pkg/image/ostree_disk.go
achilleas-k: That's a fair point.  The name of the pipeline in the manifest is `ostree-raw-image` for both and the type they return is `OSTreeDiskImage` so maybe that could be a prefix.
I think I'll go with
`NewOSTreeDiskImageFromCommit` and `NewOSTreeDiskImageFromContainer`.

pkg/manifest/ostree_deployment.go
achilleas-k: This was added by the edge/iot team as part of adding ignition support https://github.com/osbuild/osbuild-composer/pull/3130
@runcom: is this necessary for all ignition-enabled images?

pkg/manifest/ostree_deployment.go
achilleas-k: Again, I'll defer to @runcom.

pkg/ostree/ostree.go
achilleas-k: There's two things going on here that I don't like and I'm unsure of:
- The `ostree.ImageOptions` and more generally the [`distro.ImageOptions`](https://github.com/osbuild/images/blob/86c3f08097f0d409019f70cc0770fd4bcc564868/pkg/distro/distro.go#L132-L138) that exist separately from things like the blueprint customizations.  These I want to merge into a build config.
- The way that users will be specifying an ostree container for a build when this gets into osbuild-composer.

The way I added the options now sort of reflects things that might happen in composer, but that's just leftover internal structure from the split.

All that said, none of this is considered "stable API" (yet) in images, so I did the quickest thing that felt natural at the time.  You're right though, a completely separate object is nicer.  Getting the "build config" done soon would be even better.

pkg/distro/fedora/images.go
achilleas-k: Sure, we can do that.  For a high level description though: Almost all of this function is essentially creating the `img`
```go
img := image.NewOSTreeContainerDiskImage(container, ref)
```
followed by setting options on the image based on image type settings and user customizations.

pkg/distro/fedora/distro.go
achilleas-k: I based the image off the iot-raw-image and reused that partition table.  I would say, instead of digging, point me to the default partition table you use in coreos images and I'll replicate it here.

pkg/manifest/ostree_deployment.go
dustymabe: ok

pkg/manifest/ostree_deployment.go
achilleas-k: Locking the root password during deployment is something that's done for the official Fedora IoT image: https://pagure.io/fedora-kickstarts/blob/main/f/fedora-iot.ks#_9

We were replicating that here.

pkg/manifest/ostree_deployment.go
dustymabe: looks like https://github.com/ostreedev/ostree/discussions/3079#discussioncomment-7339766 seems to have figured it all out?

pkg/distro/distro_test.go
achilleas-k: It could be any string since the test doesn't really do anything with it, but you're right.  Might as well make it look like a real spec.

pkg/manifest/ostree_deployment.go
dustymabe: well, for Fedora/RH CoreOS for instance, we have the file you linked that runs on boot and does those things, so we don't need them in the disk image.

pkg/image/ostree_disk.go
dustymabe: I like those even better.

pkg/manifest/ostree_deployment.go
dustymabe: I can say at least for CoreOS, we don't want it :)

pkg/ostree/ostree.go
dustymabe: I'm OK with whatever.

pkg/manifest/ostree_deployment.go
achilleas-k: It seems to be something else.  I should write it up somewhere else but here's how the problem manifests:
- Build an image with the ostree remote stage (like we do now).
- Boot it
- Remove the remote
- Add a new remote with the same name, but with a different URL
- Add an upgrade commit to the location of the new remote
- Run `rpm-ostree upgrade`

This works.  The update gets pulled and all is well.

If I build an image without the ostree remote stage but with the packaged remote configuration and perform the same steps, then `rpm-ostree upgrade` doesn't do anything.  I can `ostree remote log` and `ostree remote summary` etc, I can see the upgrade commit in the repo.  I can even `rpm-ostree rebase` onto it.  But `rpm-ostree upgrade` refuses to pick it up.

pkg/distro/fedora/distro.go
dustymabe: [this](https://github.com/osbuild/osbuild/blob/5c345fb3fa5d9fa5804ba74e7dc4445ae8ff1010/test/data/manifests/fedora-coreos-container.mpp.yaml#L2-L27) is what I've been using for now.

There is one theme here, though, that I think we keep bumping up into, which is where definitions like this should live. I think it was also touched on in https://github.com/osbuild/osbuild-deploy-container/issues/4 
      I

achilleas-k: For Image Builder (osbuild-composer and the service), the defaults are defined here.  For all the RHEL images we build for example as part of the composes in Brew and for all the images in c.rh.c/ib, this is the source of truth right now.

pkg/manifest/ostree_deployment.go
dustymabe: dumb question.. there is a new commit in the history for that ref to upgrade to?

achilleas-k: Yes, I create the repo, serve it over a web server, and pull an upgrade commit into it in both scenarios.

I'd love to show you the whole thing.  I'll tidy up my repro scripts and put them somewhere tomorrow as well, show them around to anyone who might have some idea about what's going on.

achilleas-k: Solved this issue after 5 mins of showing Dusty the issue.  I wasn't defining a remote name when deploying.  Fixed separately.  Thanks @dustymabe!!

pkg/manifest/ostree_deployment.go
achilleas-k: This stage isn't run when building the CoreOS image.  Resolving.

achilleas-k:
Partially addressed comments (and rebased on `main`).  Will resume addressing comments later.

pkg/osbuild/ostree_deploy_container_stage.go
```diff
@@ -0,0 +1,68 @@
+package osbuild
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const ostreeContainerTargetImgrefRegex = "^(ostree-remote-registry|ostree-image-signed|ostree-unverified-registry):.*$"
```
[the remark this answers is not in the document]
achilleas-k: This is a fair point, but I also like that the regex string is a `const`.  I don't expect anything will modify the regex at runtime, but it's nice to have that guarantee.

pkg/manifest/ostree_deployment.go
achilleas-k: Fixed in #272.

pkg/image/ostree_disk.go
```diff
@@ -50,12 +52,27 @@ type OSTreeDiskImage struct {
...
+		Base:            NewBase("ostree-raw-image"),
+		ContainerSource: &container,
+		Ref:             ref,
 	}
 }
 
 func baseRawOstreeImage(img *OSTreeDiskImage, m *manifest.Manifest, buildPipeline *manifest.Build) *manifest.RawOSTreeImage {
-	osPipeline := manifest.NewOSTreeDeployment(buildPipeline, m, img.CommitSource, img.OSName, img.Ignition, img.IgnitionPlatform, img.Platform)
+	var osPipeline *manifest.OSTreeDeployment
+	if img.CommitSource != nil {
```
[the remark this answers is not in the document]
achilleas-k: Fixed in #272.

achilleas-k:
Moved most of the commits to #272.  I'll rebase this PR so that it only adds the coreos image once that's merged.

pkg/manifest/ostree_deployment.go
runcom: So, I think we can’t really say that’s not public api as it has been used in the past by coreos too (where we’ve learned about it) and even if undocumented it acts as an api to me 🤷🏻‍♂️ anyhow, I agree it’s a huge ack and it’s meant as a workaround that we need to solve (I’ll prioritize investigation and fix soon)

supakeen:
@achilleas-k should be able to rebase now as #272 has landed \o/

achilleas-k:
Will rebase and fix conflicts then start addressing the remaining comments.

PR #246, opened 2023-11-08
internal/dnfjson: fix size calculation bug with multidir caches

diaasami:
COMPOSER-2068


internal/dnfjson/cache.go
```diff
@@ -186,9 +190,12 @@ func (r *rpmCache) updateCacheDirInfo(path string) {
 	// sort IDs by mtime (oldest first)
 	sort.Slice(repoIDs, sortFunc)
```
ondrejbudai, line 191: I think that need to sort the repos in `updateInfo`, otherwise `r.repoRecency` will be only "partially" sorted.

diaasami: right, I guess there are no tests for this as well, I will add one

achilleas-k: The way it works now is that it keeps a per-distro cache and the recency and size are only relevant within each distro.  I wonder if that's the source of the issue.  It means that the max size becomes `maxSize * n distros`.  But we had a lot more than a handful of GiB right?

achilleas-k:
Something to note here: I'm trying to test the difference between main and this PR with some extreme cases and while I know there's a difference, I'm not seeing an issue with the existing things in `main`.  Try this, for example:
- Create a directory for the rpmmd cache `/var/cache/rpmmd`
- Watch the directory sizes (with sudo): `sudo watch -n1 du -hd2 /var/cache/rpmmd`
- Chance the max cache size to 10 MB (change [this](https://github.com/osbuild/images/blob/63a1eead26a7c802dbcebe863439f591be6dc6e5/internal/dnfjson/dnfjson.go#L51C1-L51C1) to `1024*1024*10`).
- Run some depsolves
```
go run ./cmd/gen-manifests --workers=10 --cache=/var/cache/rpmmd --distros=rhel-92,rhel-93 --arches=x86_64,aarch64 --output=/media/scratch/osbuild/manifests/ --images=qcow2
```

You'll see that some distro repos are deleted completely (because the limit is so low) and others (like RHEL 9.3 x86) stay at 4.6 MiB.

It also gets a bit more interesting if you set the limit to 17 MiB.  You'll see some distro caches reach 20-30 MiB and then get deleted to 14-15 MiB after the depsolve is done.  Which to me means the cleanup is actually working (but with the per-distro limit), so I wonder what went wrong originally.

achilleas-k:
Actually, `gen-manifests` is a bad example because it uses a separate solver and subdirectory for each arch+distro.  We should maybe change that.

ondrejbudai:
The issue is that the service workers set an empty distribution: https://github.com/osbuild/osbuild-composer/blob/a1e428fc539271333d71902672dff2407947ae62/cmd/osbuild-worker/jobimpl-depsolve.go#L23

So these caches are not stored in a distribution subdirectory, but in a cache root, which breaks the whole algorithm. Maybe the fix could be to disallow passing an empty distribution (or use an `unknown` subdir). Nevertheless, I think that we should one global cache size, because the number of distributions isn't bound.

achilleas-k:
If we use `unknown` in place of an empty distro name, then all repos will share the same cache root, which is probably fine.  I think in that case, depsolving will only block on individual repo locks from dnf itself.  When depsolving, we only take a non-exclusive read lock on the whole cache to prevent a cleanup from running while calling dnf-json.  Write locking of individual repos for updating metadata is deferred to libdnf.

Given all this, it makes me wonder if the per-repo cache roots are even necessary in the end.  Maybe it's fine to use a single cache for everything?

A solution that fixes the current issue is to drop the `distro` arg from `NewWithConfig()` and just use `modulePlatformID + releaseVer + arch`, which would have the same effect.

internal/dnfjson/cache_test.go
```diff
@@ -103,24 +103,24 @@ var testCfgs = map[string]testCache{
 		"df2665154150abf76f4d86156228a75c39f3f31a79d4a861d76b1edd89814b62-98177081b9162766/repodata/repomd.xml":                                                                                      fileInfo{12817, 300},
 	},
 	"fake-real": { // fake but resembling real data
-		"1111111111111111111111111111111111111111111111111111111111111111.solv":           fileInfo{100, 0},
```
diaasami, line 106: Unfortunately golang's os.ReadDir() returns entries in alphabetical order and since for our test data, entries are sorted both alphabetically and by mtime, tests pass even if no sorting is done at all.
So this change breaks the alphabetical order of data.

internal/dnfjson/cache.go
```diff
@@ -114,81 +114,94 @@ func newRPMCache(path string, maxSize uint64) *rpmCache {
...
+		//
 
-		// if for some reason the mtimes of the various entries of a single
-		// repository are out of sync, use the most recent one
-		if repo.mtime.Before(mtime) {
-			repo.mtime = mtime
+		r.size += totalSize
+		// TODO: replace loop with maps.Copy() introduced in Go 1.21
+		for k, v := range repos {
+			r.repoElements[k] = v
```
achilleas-k, line 191: Could we avoid the extra loop by doing `r.repoElements[repoID] = repo` at the end of the previous loop right after doing `repos[repoID] = repo`?  Would that interfere with anything?  It doesn't seem like it, but I could be missing something.

internal/dnfjson/cache.go
```diff
@@ -114,81 +114,94 @@ func newRPMCache(path string, maxSize uint64) *rpmCache {
...
 	}
 
-	// sort IDs by mtime (oldest first)
 	sort.Slice(repoIDs, sortFunc)
 
-	r.size = totalSize
-	r.repoElements = repos
-	r.repoRecency = repoIDs
+	// sort IDs by mtime (oldest first)
+	r.repoRecency = append(r.repoRecency, repoIDs...)
```
achilleas-k, line 204: `repoRecency` is nilled at the start of this function and is only referenced again here at the end.  Is there a reason we can't do `r.repoRecency = repoIDs`?

internal/dnfjson/cache_test.go
achilleas-k: Good thinking.  Though I think it doesn't matter here because maps are unordered.  Did this have any effect on the result of the test?
I wouldn't revert it, keep the change, I'd be curious to know though.

achilleas-k (dismissed):
LGTM.  A few non-functional nitpicks and questions :)

One more thing though:
I think we should remove the `distro` here: https://github.com/osbuild/images/blob/54a72ee04cec5d10d33e85765257246f2ff1b852/internal/dnfjson/dnfjson.go#L72
and instead use the other components to make a cache subdirectory here: https://github.com/osbuild/images/blob/54a72ee04cec5d10d33e85765257246f2ff1b852/internal/dnfjson/dnfjson.go#L123-L130

The other components (platform ID, os version, architecture) are mandatory for depsolving, so they can't be empty strings or the depsolver will fail.  That way, we avoid running into any issues (like we already did) with the solver being initialised with an empty distro string, essentially pushing the distro caches one level up, and we still get distro-specific caches.

I  don't know if there might be a conflict between RHEL 9 and CS9 (if osversion of rhel 9 is `9`, then they both get `el9` platform ID and os version `9`).

Another alternative is to make the `distro` mandatory and fail if it's an empty string.

achilleas-k:
I have some comments and a recommendation, but approving since they're not critical and can be done in a follow-up if we need this PR to be merged quickly.

internal/dnfjson/cache_test.go
diaasami: This is only about `rpmCache.reporecency` since it's an ordered array, but not any other fields.
Since the the test data files are written to disk then read using `os.ReadDir(r.root)`, they come out in the right order automatically.

PR #250, opened 2023-11-10
Add single-purpose command for building bootable images from an ostree native container

achilleas-k:
_This PR is based on #243._

This PR adds a new command for building a qcow2 image from an ostree container.  The binary is very specific to Fedora ELN and embeds the repository configurations so it can be run from anywhere.

Build and run with:
```
go build -o bin/osbuild-deploy-container ./cmd/osbuild-deploy-container && sudo ./bin/osbuild-deploy-container --output <outputdir> --imageref <container ref>
```

Tested working with `quay.io/centos-boot/fedora-tier-1:eln`.
Other options:
- `--store`: osbuild cache/store location.  This is where rpms and containers used to build the image are cached and can get quite large (defaults to `.osbuild` in the working directory).
- `--rpmmd`: rpm metadata cache (defaults to `/var/cache/osbuild/rpmmd`).
- `--config`: image configuration/customization options (see below).

### Configuration

The most useful customization is creating a user to log in to the system.  The format of the configuration file with a user is:
```json
{
  "name": "image-with-user",
  "blueprint": {
    "customizations": {
      "user": [
        {
          "name": "username",
          "password": "password",
          "key": "public-ssh-key",
          "groups": [
            "wheel"
          ]
        }
      ]
    }
  }
}
```

Kernel command line options can also be specified that will be appended to the pre-defined ones:
```json
{
  "name": "image-with-kargs",
  "blueprint": {
    "customizations": {
      "kernel": {
        "append": "debug"
      }
    }
  }
}
```


cmd/osbuild-deploy-container/main.go
```diff
@@ -0,0 +1,181 @@
...
+
+	arch := common.CurrentArch()
+	repos := loadRepos(arch)
+
+	var outputDir, osbuildStore, rpmCacheRoot, configFile, imageref string
+	flag.StringVar(&outputDir, "output", ".", "artifact output directory")
+	flag.StringVar(&osbuildStore, "store", ".osbuild", "osbuild store for intermediate pipeline trees")
+	flag.StringVar(&rpmCacheRoot, "rpmmd", "/var/cache/osbuild/rpmmd", "rpm metadata cache directory")
+	flag.StringVar(&configFile, "config", "", "build config file")
+	flag.StringVar(&imageref, "imageref", "", "container image to deploy")
```
achilleas-k, line 141: Might be better to rename this to `--imgref` to match the argument in `rpm-ostree container image deploy ...`.

cmd/osbuild-deploy-container/main.go
```diff
@@ -0,0 +1,181 @@
...
+
+	flag.Parse()
+
+	if imageref == "" {
+		fail("imageref is required")
+	}
+
+	if err := os.MkdirAll(outputDir, 0777); err != nil {
```
bcl, line 149: Is there a reason this is `0777`? I'm worried it will get forgotten, this tool will become popular, and we'll have a security issue down the road :)

achilleas-k: Those bits are applied to the umask so it ends up just creating it with the default permissions of the running environment.

ondrejbudai:
This became https://github.com/osbuild/bootc-image-builder/!

PR #253, opened 2023-11-13
Enable system FIPS mode for Edge Images

mmartinv:
Defines a new `fips` blueprint customization to enable the system FIPS mode on the generated images.

Enables `fips` blueprint customization support for the following images: 
- Edge images based on OSTree deployments:
   - `edge-raw-image`
   - `edge-simplified-installer`
   - `edge-vsphere`
   - `edge-ami`
- Edge images based on the Anaconda OSTree installer:
  -  `edge-installer`

Depends on osbuild/osbuild/pull/1443

### DIFF
--- a/pkg/osbuild/fips.go
+++ b/pkg/osbuild/fips.go
@@ -1,0 +1,43 @@
+package osbuild
+
+import (
+	"os"
+
+	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/internal/fsnode"
+	"github.com/osbuild/images/pkg/disk"
+)
+
+func GenFIPSKernelOptions(pt *disk.PartitionTable) []string {
+	cmdline := make([]string, 0)
+	cmdline = append(cmdline, "fips=1")
+	if bootMnt := pt.FindMountable("/boot"); bootMnt != nil {
+		boot := bootMnt.GetFSSpec()
+		if label := boot.Label; label != "" {
+			karg := "boot=LABEL=" + label
+			cmdline = append(cmdline, karg)
+		} else if uuid := boot.UUID; uuid != "" {
+			karg := "boot=UUID=" + uuid
+			cmdline = append(cmdline, karg)
+		}
+	}
+	return cmdline
+}
+
+func GenFIPSFiles() (files []*fsnode.File) {
+	file, _ := fsnode.NewFile("/etc/system-fips", common.ToPtr(os.FileMode(0644)),
+		"root", "root", []byte("# FIPS module installation complete\n"))
+	files = append(files, file)
+	return
+}
+
+func GenFIPSStages() (stages []*Stage) {
+	stages = []*Stage{
+		NewUpdateCryptoPoliciesStage(
+			&UpdateCryptoPoliciesStageOptions{
+				Policy: "FIPS",
+			}),
+	}
+	stages = append(stages, GenFileNodesStages(GenFIPSFiles())...)
+	return
+}
--- a/Schutzfile
+++ b/Schutzfile
@@ -2,7 +2,7 @@
   "fedora-38": {
     "dependencies": {
       "osbuild": {
-        "commit": "61da6124edee4f9735d50a412d5ecc16c8eec674"
+        "commit": "f3d740aaf8531e55b99632f579f2fae13f1511b7"
       }
     },
     "repos": [
--- a/pkg/blueprint/customizations.go
+++ b/pkg/blueprint/customizations.go
@@ -24,6 +24,7 @@ type Customizations struct {
 	Directories        []DirectoryCustomization  `json:"directories,omitempty" toml:"directories,omitempty"`
 	Files              []FileCustomization       `json:"files,omitempty" toml:"files,omitempty"`
 	Repositories       []RepositoryCustomization `json:"repositories,omitempty" toml:"repositories,omitempty"`
+	FIPS               *bool                     `json:"fips,omitempty" toml:"fips,omitempty"`
 }
 
 type IgnitionCustomization struct {
@@ -355,4 +356,11 @@ func (c *Customizations) GetRepositories() ([]RepositoryCustomization, error) {
 	}
 
 	return c.Repositories, nil
+}
+
+func (c *Customizations) GetFIPS() bool {
+	if c == nil || c.FIPS == nil {
+		return false
+	}
+	return *c.FIPS
 }
--- a/pkg/distro/rhel9/images.go
+++ b/pkg/distro/rhel9/images.go
@@ -404,6 +404,7 @@ func edgeRawImage(workload workload.Workload,
 
 	img.Users = users.UsersFromBP(customizations.GetUsers())
 	img.Groups = users.GroupsFromBP(customizations.GetGroups())
+	img.FIPS = customizations.GetFIPS()
 
 	// The kernel options defined on the image type are usually handled in
 	// osCustomiztions() but ostree images don't use OSCustomizations, so we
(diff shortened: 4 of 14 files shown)


pkg/osbuild/fips.go
```diff
@@ -0,0 +1,56 @@
...
+				} else if uuid := ent.UUID; uuid != "" {
+					karg := "boot=UID=" + uuid
+					cmdline = append(cmdline, karg)
+				}
+			}
+		}
+		return nil
+	}
+
+	_ = pt.ForEachEntity(genOptions)
```
achilleas-k, line 30: Since we only need to do something for `/boot`, you can use:
```go
	if bootMnt := pt.FindMountable("/boot"); bootMnt != nil {
		boot := bootMnt.GetFSSpec()
		if label := boot.Label; label != "" {
			karg := "boot=LABEL=" + label
			cmdline = append(cmdline, karg)
		} else if uuid := boot.UUID; uuid != "" {
			karg := "boot=UID=" + uuid
			cmdline = append(cmdline, karg)
		}
	}

```
which I find more readable.

pkg/osbuild/fips.go
```diff
@@ -0,0 +1,56 @@
...
+
+	genOptions := func(e disk.Entity, path []disk.Entity) error {
+		switch ent := e.(type) {
+		case *disk.Filesystem:
+			if ent.Mountpoint == "/boot" {
+				if label := ent.Label; label != "" {
+					karg := "boot=LABEL=" + label
+					cmdline = append(cmdline, karg)
+				} else if uuid := ent.UUID; uuid != "" {
+					karg := "boot=UID=" + uuid
```
achilleas-k, line 22: Should this be `boot=UUID`?

pkg/osbuild/fips.go
```diff
@@ -0,0 +1,56 @@
...
+	_ = pt.ForEachEntity(genOptions)
+	return cmdline
+}
+
+func GenFIPSFiles() (files []*fsnode.File) {
+	file, _ := fsnode.NewFile("/etc/system-fips", common.ToPtr(os.FileMode(0644)),
+		"root", "root", []byte("# FIPS module installation complete\n"))
+	files = append(files, file)
+	return
+}
```
achilleas-k, line 39: What's the purpose of this file?  I looked around a bit and it seems to be more of a signal and not really functional.  Also it seems it's sometimes in the initrd?
Either way: If this is part of the spec, or required somehow, and if it's always done, should it be part of the stage in osbuild?

test/configs/edge-ostree-pull-empty-fips.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "edge-ostree-pull-empty-fips",
```
achilleas-k, line 2: Tiny nitpick, but since we have the `fips` customisation here, it's not `empty`, so a more appropriate name would be `edge-ostree-pull-fips`.  Not an issue though, it's just a test name.

achilleas-k (changes requested):
Thanks for this.  Some comments in line.  Mostly nitpicks, though I'm curious about the file in `/etc` and whether it should be part of the osbuild stage.

pkg/osbuild/fips.go
mmartinv: This file is required for `fips-mode-setup --check` to return
```
 FIPS mode is enabled
```
instead of:
```
Installation of FIPS modules is not completed.
FIPS mode is enabled.
Inconsistent state detected.
```
It's not really functional indeed, but we need it at least for RHEL Edge (I think fedora does not check its existence anymore)

pkg/osbuild/fips.go
mmartinv: Yes, that's a typo

pkg/osbuild/fips.go
mmartinv: Done

pkg/osbuild/fips.go
mmartinv: I moved this part to the `GenFIPSStages()` function. Please take a look.

test/configs/edge-ostree-pull-empty-fips.json
mmartinv: Done

pkg/osbuild/fips.go
achilleas-k: So my original thought was adding this to the stage in osbuild but I realise now that that was wrong, because the stage in osbuild isn't FIPS-specific, it just runs `update-crypto-policies`.

Now that said, the change here actually breaks the feature and makes the manifest ubuildable, because it adds the stage but not the inline source which defines the content.  This part should be reverted to the earlier version which added the file to the pipeline using `p.Files = append(p.Files, osbuild.GenFIPSFiles()...)`.

pkg/osbuild/fips.go
```diff
@@ -0,0 +1,53 @@
...
+			}),
+		NewDracutConfStage(
+			&DracutConfStageOptions{
+				Filename: "40-fips.conf",
+				Config: DracutConfigFile{
+					AddModules: []string{"fips"},
+				},
+			}),
+	}
+	stages = append(stages, GenFileNodesStages(files)...)
```
thozza, line 50: This won't work, since the `GenFileNodesStages()` generates only stages, but no inline sources, for the file. These need to be added to the final manifest, so that the generate stages can copy them to the image FS tree.

I suggest to modify the function to return `([]*Stage, []*fsnode.File)` and then add those stages to the pipeline and the files to the custom files in the pipeline implementation:
https://github.com/osbuild/images/blob/63a1eead26a7c802dbcebe863439f591be6dc6e5/pkg/manifest/ostree_deployment.go#L52
https://github.com/osbuild/images/blob/63a1eead26a7c802dbcebe863439f591be6dc6e5/pkg/manifest/os.go#L133

The existing pipeline implementation will then handle the generation of necessary stages to copy the file into the image and set its attributes (this also ensures that instead of multiple instances of those stages, there will be only one handling all custom files). The following code then handles the creation of inline sources for custom files:
https://github.com/osbuild/images/blob/63a1eead26a7c802dbcebe863439f591be6dc6e5/pkg/manifest/os.go#L817-L820
https://github.com/osbuild/images/blob/63a1eead26a7c802dbcebe863439f591be6dc6e5/pkg/manifest/ostree_deployment.go#L362-L365

mmartinv: Yeah, you are right, but doesn't this complicate things a lot in comparison with the previous implementation? I personally prefer the previous approach.

mmartinv: Actually, If I do that the file stages will be appended twice to the pipepline unless I remove the file stages from the `GenFIPSStages()` function. That doesn't make much sense to me honestly. Having the stages and files generation in two different functions feels more natural to me.

mmartinv: Reverted to previous implementation

pkg/osbuild/fips.go
mmartinv: Reverted to previous implementation

pkg/osbuild/fips.go
thozza: > Actually, If I do that the file stages will be appended twice to the pipepline unless I remove the file stages from the `GenFIPSStages()` function. That doesn't make much sense to me honestly. Having the stages and files generation in two different functions feels more natural to me.

You are right, I didn't realize that...

pkg/osbuild/fips.go
achilleas-k: @thozza let's get it merged and we can rethink this in a followup if necessary.

achilleas-k (approved):
LGTM.  Good to merge as soon as we fix the build config filtering in CI.

achilleas-k (approved):

PR #257, opened 2023-11-14
docs/developer: manifest generation

achilleas-k:
This PR adds developer documentation under docs/developer/.

It moves the content of the HACKING.md doc to docs/developer/ and links to it from the original location of the doc.

The developer docs are separated into topics now.  So far only two:
- "Useful cmds" which was originally part of HACKING.md.
- "Manifest generation code" which describes how manifest generation works in code.

I wanted to emphasise the **code** part in the second topic to differentiate from generating manifest using `cmd/gen-manifests`.  The purpose of this document is to describe the flow of function calls required to generate a manifest.  I consider this a starting point for understanding how customizations become pipeline options and how those in turn affect pipelines, stages, and stage options.  This isn't emphasised in the document currently, but I wanted to submit it as is and work on adding more content later to avoid having it be forgotten on some branch on my fork for months.

Note that when describing types or functions, I chose to link to pkg.go.dev instead (and selected `main` as the version) instead of code files or lines to make the document resistant to changes in line numbers.  The same could be achieve by linking to specific lines in specific commits, but that would risk linking to outdated information if code changes too much.  Linking to pkg.go.dev makes sure that as long as the type or method still exists in the same package, the link will work, and anyone browsing can quickly get from there to the code in this repository.


bcl (approved):
Looks good!

ochosi (approved):
Looks amazing 😻

PR #260, opened 2023-11-16
fix: filter non relevant configurations with dependencies

mmartinv:
Currently the configurations with dependencies are filtered
based on the "depends" field presence only. This makes
configurations from other distros to be processed even
when they are not relevant.

### DIFF
--- a/test/scripts/generate-ostree-build-config
+++ b/test/scripts/generate-ostree-build-config
@@ -33,7 +33,7 @@ def read_config_map():
         return json.load(config_map_file)
 
 
-def configs_with_deps(configs):
+def configs_with_deps(configs, distro=None):
     """
     Return a config map with only the config files that have dependencies.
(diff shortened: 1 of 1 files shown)


achilleas-k:
I think the bug is in `setupdependences()`, here:https://github.com/osbuild/images/blob/8a852bfc07e155c2c7b20d9638f91e5ea2ce36b8/test/scripts/generate-ostree-build-config#L211-L214

I think what that should be doing is checking if `distros` is empty and if it's not checking that the current distro is in the list.  If the list it not empty and the distro is not in the list, then this config should be skipped.

achilleas-k:
Looking at it again, your way is correct.  My way would create dependency manifests unnecessarily.

achilleas-k (approved):
Thanks for the fix (and discovering the bug)!

PR #262, opened 2023-11-17
test: filter on architecture as well when generating config map of dependencies

achilleas-k:
Following up from #260, this PR also filters the config map on the architecture.


bcl (approved):
Looks good!

PR #265, opened 2023-11-20
Support globs in distro and image filtering

mmartinv:
Support globs in distro and image filtering.
Currently distro filters like `rhel-8*` or arch
filters like `*64` don't work as expected.

### DIFF
--- a/test/scripts/generate-ostree-build-config
+++ b/test/scripts/generate-ostree-build-config
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import fnmatch
 import json
 import os
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


test/scripts/generate-ostree-build-config
```diff
@@ -33,6 +34,15 @@ def read_config_map():
         return json.load(config_map_file)
 
 
+def matches(a: str, items: list[str]) -> bool:
+    matched = False
+    for item in items:
+        if fnmatch.fnmatch(a, item):
+            matched = True
```
bcl, line 41: Just return True here, and False at the end. Then you can drop `matched` variable.
Or you could simplify it to:
`return any(fnmatch.fnmatch(a, i) for i in items)`

bcl (approved):
Looks good to me!

PR #268, opened 2023-11-21
OpenSCAP results compression

kingsleyzissou:
The OpenSCAP remediation stage has been reworked to operate in three
stages:
1. Scan the filesystem tree and save results 
2. Generate the remediation script 
3. Run the offline remediation.

Steps one and two both require a data directory for the output that is then used by the next step. This PR creates the necessary config options and enables compression for the results file, since the file could be potentially quite large.

### DIFF
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -165,6 +165,16 @@ func osCustomizations(
 		if t.rpmOstree {
 			panic("unexpected oscap options for ostree image type")
 		}
+
+		// although the osbuild stage will create this directory,
+		// it's probably better to ensure that it is created here
+		dataDirNode, err := fsnode.NewDirectory(oscapDataDir, nil, nil, nil, true)
+		if err != nil {
+			panic("unexpected error creating OpenSCAP data directory")
+		}
+
+		osc.Directories = append(osc.Directories, dataDirNode)
+
 		var datastream = oscapConfig.DataStream
 		if datastream == "" {
 			datastream = oscap.DefaultFedoraDatastream()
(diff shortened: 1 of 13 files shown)


pkg/distro/fedora/images.go
```diff
@@ -211,6 +211,9 @@ func osCustomizations(
 			osc.Directories = append(osc.Directories, tailoringDir)
 		}
 
+		// add this after tailoring
+		oscapStageOptions.Compression = true
```
kingsleyzissou, line 215: For the most part, `oscapStageOptions` and `oscapTailoringStageOptions` are the same, but this field breaks things.
I didn't split the definitions, but I was very tempted to here, let me know if this would be preferred, see the below comment for further clarification.

pkg/osbuild/oscap_remediation_stage.go
```diff
@@ -28,6 +28,7 @@ type OscapConfig struct {
 	HtmlReport   string              `json:"html_report,omitempty" toml:"html_report,omitempty"`
 	VerboseLog   string              `json:"verbose_log,omitempty" toml:"verbose_log,omitempty"`
 	VerboseLevel OscapVerbosityLevel `json:"verbose_level,omitempty" toml:"verbose_level,omitempty"`
+	Compression  bool                `json:"compress_results,omitempty" toml:"compress_results,omitempty"`
```
kingsleyzissou, line 31: To clarify the above comment, these stage options are passed to the tailoring stage too... which is probably not great since some of these fields will break the tailoring stage if they're added somehow. I'm inclined to separate these.

pkg/distro/fedora/images.go
achilleas-k: Are they that much the same?  I see a few common fields in the stages in osbuild, but there's a bunch that are unique to each, no?
I  think they might have diverged enough that we can write different configs for each stage and avoid confusion.

pkg/osbuild/oscap_remediation_stage.go
achilleas-k: Agree with your inclination :)
Let's separate them.

achilleas-k (changes requested):

achilleas-k (approved):
Wonderful!  Thank you
LGTM

PR #272, opened 2023-11-24
Add support for building images from ostree containers

achilleas-k:
Cherry-picked commits from #243 without the CoreOS image.

This PR adds support for the org.osbuild.ostree.deploy.container stage and addresses the relevant comments from #243.  The other PR will be rebased to only add the CoreOS image after this is merged.

This PR doesn't define a new image type, we will instead instantiate the manifest directly as a separate project, in the new repository https://github.com/osbuild/osbuild-deploy-container/.

Part of HMS-3121.


pkg/manifest/ostree_deployment.go
```diff
@@ -93,42 +101,67 @@ func (p *OSTreeDeployment) getBuildPackages(Distro) []string {
...
+		return []container.SourceSpec{}
+	}
+	return []container.SourceSpec{
+		*p.containerSource,
 	}
 }
 
 func (p *OSTreeDeployment) serializeStart(packages []rpmmd.PackageSpec, containers []container.Spec, commits []ostree.CommitSpec) {
-	if p.ostreeSpec != nil {
+	if p.ostreeSpec != nil && p.containerSpec != nil {
```
supakeen, line 152: Should this be `||`, it's invaldi to call `serializeStart` if either is set right?

pkg/manifest/ostree_deployment.go
```diff
@@ -93,42 +101,67 @@ func (p *OSTreeDeployment) getBuildPackages(Distro) []string {
...
-	p.ostreeSpec = &commits[0]
 }
 
 func (p *OSTreeDeployment) serializeEnd() {
-	if p.ostreeSpec == nil {
+	if p.ostreeSpec == nil && p.containerSpec == nil {
 		panic("serializeEnd() call when serialization not in progress")
 	}
 
 	p.ostreeSpec = nil
```
supakeen, line 170: Should `p.containerSpec` be nilled here too?

pkg/manifest/ostree_deployment.go
```diff
@@ -344,7 +344,8 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 		}
 	}
 
-	if !hasRoot {
+	if !hasRoot && p.ostreeSpec != nil {
+		// NOTE: fails on container-based deployments
```
supakeen, line 350: Do we want to error out here instead if it fails anyhow?

pkg/distro/fedora/images.go
```diff
@@ -524,6 +524,7 @@ func iotImage(workload workload.Workload,
 	img.OSName = "fedora-iot"
 
 	if !common.VersionLessThan(distro.Releasever(), "38") {
+		img.KernelOptionsAppend = append(img.KernelOptionsAppend, "coreos.no_persist_ip")
```
supakeen, line 527: Does this have no effect because the kernel option is ignored? If so I'd still like it to be conditional.

pkg/osbuild/ostree_deploy_container_stage.go
```diff
@@ -0,0 +1,67 @@
+package osbuild
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const ostreeContainerTargetImgrefRegex = "^(ostree-remote-registry|ostree-image-signed|ostree-unverified-registry):.*$"
```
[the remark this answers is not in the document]
achilleas-k: Yeah we do this often for validating stage options before they reach the stage to catch things early, but it requires keeping them in sync manually, which isn't the best solution.

pkg/manifest/ostree_deployment.go
achilleas-k: Moved it to the image function so it works like the other change we made with the kernel options.

achilleas-k:
Reverted the upgrade of the CI runners to F39.  Looked like there might be some misconfiguration with the x86 image.  We can investigate separately.

pkg/osbuild/ostree_deploy_container_stage.go
```diff
@@ -0,0 +1,67 @@
...
+	Rootfs *Rootfs `json:"rootfs,omitempty"`
+
+	// Mount points of the final file system
+	Mounts []string `json:"mounts,omitempty"`
+}
+
+func (OSTreeDeployContainerStageOptions) isStageOptions() {}
+
+func (options OSTreeDeployContainerStageOptions) validate() error {
+	exp := regexp.MustCompile(ostreeContainerTargetImgrefRegex)
```
mvo5, line 32: (nitpick) Maybe it's worth moving this out of the function and into the global context. The advantage is that any regex error becomes an error at `go test` time (and it's compiled onlyonce) Something like:
```go
diff --git a/pkg/osbuild/ostree_deploy_container_stage.go b/pkg/osbuild/ostree_deploy_container_stage.go
index 78e01f6af..f5a36f455 100644
--- a/pkg/osbuild/ostree_deploy_container_stage.go
+++ b/pkg/osbuild/ostree_deploy_container_stage.go
@@ -5,7 +5,7 @@ import (
        "regexp"
 )
 
-const ostreeContainerTargetImgrefRegex = "^(ostree-remote-registry|ostree-image-signed|ostree-unverified-registry):.*$"
+var ostreeContainerTargetImgrefRegex = regexp.MustCompile("^(ostree-remote-registry|ostree-image-signed|ostree-unverified-registry):.*$")
 
 // Options for the org.osbuild.ostree.deploy.container stage.
 type OSTreeDeployContainerStageOptions struct {
@@ -29,7 +29,7 @@ type OSTreeDeployContainerStageOptions struct {
 func (OSTreeDeployContainerStageOptions) isStageOptions() {}
 
 func (options OSTreeDeployContainerStageOptions) validate() error {
-       exp := regexp.MustCompile(ostreeContainerTargetImgrefRegex)
+       exp := ostreeContainerTargetImgrefRegex
        if !exp.MatchString(options.TargetImgref) {
                return fmt.Errorf("'target_imgref' %q doesn't conform to schema (%s)", options.TargetImgref, exp.String())
        }
```
when the regex is invalid this yields:
```
$ go test ./pkg/osbuild/
panic: regexp: Compile(`(abc* ^(ostree-remote-registry|ostree-image-signed|ostree-unverified-registry):.*$`): error parsing regexp: missing closing ): `(abc* ^(ostree-remote-registry|ostree-image-signed|ostree-unverified-registry):.*$`

goroutine 1 [running]:
regexp.MustCompile({0xfdd886, 0x52})
	/usr/lib/go-1.21/src/regexp/regexp.go:319 +0xb4
github.com/osbuild/images/pkg/osbuild.init()
	/home/egon/devel/osbuild/images/pkg/osbuild/ostree_deploy_container_stage.go:8 +0x59
FAIL	github.com/osbuild/images/pkg/osbuild	0.020s
FAIL
```
the downside of course is an unnecessary compile if this function is never used so tradeoffs :)

pkg/osbuild/ostree_deploy_container_stage.go
```diff
@@ -0,0 +1,67 @@
...
+type OSTreeDeployContainerStageOptions struct {
+
+	// Name of the stateroot to be used in the deployment
+	OsName string `json:"osname"`
+
+	// Additional kernel command line options
+	KernelOpts []string `json:"kernel_opts,omitempty"`
+
+	// Image ref used as the source of truth for updates
+	TargetImgref string `json:"target_imgref"`
```
mvo5, line 20: (super-nitpick) I personalyl like `TargetImageRef` but that is probably the mother of all bikesheding (sorry!)

pkg/osbuild/ostree_deploy_container_stage.go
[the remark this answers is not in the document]
mvo5: maybe worthwhile to add a comment like `// keep in sync with the osbuild stage`  ? or does that go without saying :) ?

mvo5 (approved):
Thanks for this PR - I added some comments and nitpicks inline but a lot of it it personal preference and/or comemnts due to my ignorance. I hope it's still useful.

pkg/osbuild/ostree_deploy_container_stage.go
achilleas-k: I commented on this in the previous PR here: https://github.com/osbuild/images/pull/243#discussion_r1404372500

I still think I prefer having the safety of a const.  And like you said there's the tiny addition of it only being compiled when needed.  I don't know if that outweighs catching it at test/compile time though.

pkg/manifest/ostree_deployment.go
```diff
@@ -167,43 +213,74 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
...
-	pipeline.AddStage(osbuild.NewOSTreeDeployStage(
-		&osbuild.OSTreeDeployStageOptions{
-			OsName: p.osName,
-			Ref:    commit.Ref,
-			Remote: p.Remote.Name,
-			Mounts: []string{"/boot", "/boot/efi"},
-			Rootfs: osbuild.Rootfs{
-				Label: "root",
+	var ref string
+	if p.ostreeSpec != nil {
```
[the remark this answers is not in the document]
achilleas-k: I agree.  Let's pull them out.

pkg/osbuild/ostree_deploy_container_stage.go
mvo5: Aha, sure - this defined outside? that makes sense and then yes, it should be as close as possible.

pkg/osbuild/ostree_deploy_container_stage.go
mvo5: Thanks and sorry, I must have missed the reply last time. Fine of course

pkg/osbuild/ostree_deploy_container_stage.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+type OSTreeDeployContainerInputs struct {
+	Images ContainersInput `json:"images"`
+}
+
+func (OSTreeDeployContainerInputs) isStageInputs() {}
+
+func (inputs OSTreeDeployContainerInputs) validate() error {
+	if ncontainers := inputs.Images.References.Len(); ncontainers != 1 {
```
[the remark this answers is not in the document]
achilleas-k: `Images` is a struct so it's fine.  `References` is an interface so I guess it depends on the implementation.  Better if we check it.

pkg/osbuild/ostree_deploy_container_stage.go
achilleas-k: > this defined outside?

Yup, it gets serialised into the manifest for the stage options.

pkg/manifest/ostree_deployment.go
```diff
@@ -88,53 +121,70 @@ func (p *OSTreeDeployment) getBuildPackages(Distro) []string {
...
 	}
 
-	p.ostreeSpecs = nil
+	p.ostreeSpec = nil
+	p.containerSpec = nil
 }
 
 func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
-	if len(p.ostreeSpecs) == 0 {
+	if p.ostreeSpec == nil && p.containerSpec == nil {
```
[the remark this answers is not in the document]
achilleas-k: Done!

achilleas-k:
@mvo5 I went all in on the `switch`es for you

supakeen (approved):
Awesome.

mvo5 (approved):
Thank you for this PR! It looks good to merge, I made a bunch of inline comments with ideas/suggestions/questions but all totally optional.

cmd/build/main.go
```diff
@@ -34,47 +34,11 @@ func check(err error) {
 	}
 }
 
-type repository struct {
```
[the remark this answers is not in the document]
achilleas-k: You're not wrong, but let's get this merged to unblock other work that depends on it.  Resolving.

PR #273, opened 2023-11-24
Add org.osbuild.ostree.encapsulate stage

achilleas-k:
Add the ostree encapsulate stage and a pipeline for creating ostree native containers.


supakeen (approved):
The only question really is, `BootContainer` or `BootableContainer` (the `.tar` filename is the latter, the struct has the former) but that's non-important :)

achilleas-k:
I'm completely open on the naming of the option and the filename.  I gave it almost no thought in the moment.

PR #274, opened 2023-11-24
Make dnfjson module public

achilleas-k:
Make the dnfjson module public so we can use it in osbulid-composer instead of duplicating the code.
This will also be needed for https://github.com/osbuild/osbuild-deploy-container/ as part of HMS-3121.


croissanne (approved):

achilleas-k:
Just realised that #246 is still open.  Let's merge that one first and avoid conflicts.

achilleas-k:
I wonder what will happen if I put this on the queue behind the other PR... let's find out!

achilleas-k:
image
Cool

croissanne (approved):

PR #275, opened 2023-11-24
pkg/arch: split arch from pkg/platform

croissanne:
This allows importing the architecture separately, and will help getting rid of inline string literals in composer.

---

Alternatively we import platform everywhere.

### DIFF
--- a/internal/boot/context-managers.go
+++ b/internal/boot/context-managers.go
@@ -13,7 +13,7 @@ import (
 
 	"github.com/osbuild/images/cmd/osbuild-image-tests/constants"
 	"github.com/osbuild/images/internal/common"
-	"github.com/osbuild/images/pkg/platform"
+	"github.com/osbuild/images/pkg/arch"
 )
 
 // WithNetworkNamespace provides the function f with a new network namespace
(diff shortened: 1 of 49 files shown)


achilleas-k:
Not a bad idea.  Can you also move the `CurrentArch()` function from `common/helpers.go` into the new arch package?  It's nothing big, but can be useful to have public and I think it makes sense in the new package.

internal/boot/context-managers.go
```diff
@@ -111,7 +111,7 @@ func WithBootedQemuImage(image string, ns NetNS, f func() error) error {
 		}
 
 		var qemuCmd *exec.Cmd
-		if common.CurrentArch() == arch.ARCH_X86_64.String() {
+		if arch.CurrentArch() == arch.ARCH_X86_64 {
```
croissanne, line 114: Could also add something like `arch.IsX86()` (and `arch.IsAarch64` etc...)

achilleas-k: yeah that'd be nice.  Also should we rename it to `arch.Current()` (or `arch.GetCurrent()`) to avoid repetition of the word `arch`?

croissanne: done! reads much better

achilleas-k:
Merge conflict :(

achilleas-k (commented):
THANK YOU!!

achilleas-k (approved):

PR #280, opened 2023-11-27
Strip PR prefix from PR number in build cache

achilleas-k:
We've been saving the PR number to the build cache metadata by reading the branch name in GitLab CI, but the branch name is prefixed with `PR-`.  Save it as a plain number so we can use it for links etc.


ochosi (approved):
LGTM!

PR #282, opened 2023-11-28
FIPS on Fedora and RHEL 8/9 images

mmartinv:
- feat: enable FIPS customization in OS pipeline
- feat: enable FIPS customization in Anaconda tar installer
- feat: Enable FIPS customization for RHEL 9
- feat: Enable FIPS customization for RHEL 8
- feat: Enable FIPS customization for Fedora

### DIFF
--- a/pkg/distro/rhel8/edge.go
+++ b/pkg/distro/rhel8/edge.go
@@ -7,6 +7,7 @@ import (
 	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/customizations/fsnode"
 	"github.com/osbuild/images/pkg/distro"
+	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/rpmmd"
 )
 
@@ -21,6 +22,7 @@ func edgeCommitImgType(rd distribution) imageType {
 		},
 		defaultImageConfig: &distro.ImageConfig{
 			EnabledServices: edgeServices(rd),
+			DracutConf:      []*osbuild.DracutConfStageOptions{osbuild.FIPSDracutConfStageOptions},
 		},
 		rpmOstree:        true,
 		image:            edgeCommitImage,
@@ -47,6 +49,7 @@ func edgeOCIImgType(rd distribution) imageType {
 		},
 		defaultImageConfig: &distro.ImageConfig{
 			EnabledServices: edgeServices(rd),
+			DracutConf:      []*osbuild.DracutConfStageOptions{osbuild.FIPSDracutConfStageOptions},
 		},
 		rpmOstree:        true,
 		bootISO:          false,
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -122,6 +122,7 @@ var (
 		},
 		defaultImageConfig: &distro.ImageConfig{
 			EnabledServices: iotServices,
+			DracutConf:      []*osbuild.DracutConfStageOptions{osbuild.FIPSDracutConfStageOptions},
 		},
 		rpmOstree:        true,
 		image:            iotCommitImage,
@@ -143,6 +144,7 @@ var (
 		},
 		defaultImageConfig: &distro.ImageConfig{
 			EnabledServices: iotServices,
+			DracutConf:      []*osbuild.DracutConfStageOptions{osbuild.FIPSDracutConfStageOptions},
 		},
 		rpmOstree:        true,
 		bootISO:          false,
--- a/pkg/distro/fedora/distro_test.go
+++ b/pkg/distro/fedora/distro_test.go
@@ -492,11 +492,11 @@ func TestDistro_ManifestError(t *testing.T) {
 				} else if imgTypeName == "iot-installer" || imgTypeName == "iot-simplified-installer" {
 					assert.EqualError(t, err, fmt.Sprintf("boot ISO image type \"%s\" requires specifying a URL from which to retrieve the OSTree commit", imgTypeName))
 				} else if imgTypeName == "image-installer" {
-					assert.EqualError(t, err, fmt.Sprintf("unsupported blueprint customizations found for boot ISO image type \"%s\": (allowed: User, Group)", imgTypeName))
+					assert.EqualError(t, err, fmt.Sprintf("unsupported blueprint customizations found for boot ISO image type \"%s\": (allowed: User, Group, FIPS)", imgTypeName))
 				} else if imgTypeName == "live-installer" {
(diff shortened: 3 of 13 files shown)


evgenyz:
Just to be clear here. Current ssg (0.1.69) won't try to enable FIPS during the oscap remediation phase since it caused troubles with bootloader. Idea is to keep it that way and use Blueprint instead as we do with partitions and packages.

What would be the Blueprint snippet to enable FIPS, BTW?

test/configs/all-with-oscap-fips.json
```diff
@@ -0,0 +1,155 @@
...
+      ],
+      "openscap": {
+        "profile_id": "xccdf_org.ssgproject.content_profile_pci-dss",
+        "tailoring": {
+          "selected": [
+            "bind_crypto_policy"
+          ],
+          "unselected": [
+            "rpm_verify_hashes",
+            "enable_fips_mode"
```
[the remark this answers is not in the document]
kingsleyzissou: I think this might be a copy and paste from the initial tailoring customisations

kingsleyzissou:
The docs PR for the syntax is here:

https://github.com/osbuild/guides/pull/154

Should just be:
```
fips = true
```

kingsleyzissou:
FTR I'm testing this locally. Just ran into one or two issues, will report back

mmartinv:
> FTR I'm testing this locally. Just ran into one or two issues, will report back

It looks like the oscap remediation stage was setting the system's crypto policy to `DEFAULT` after it was set to `FIPS` by the `update-crypto-policies` stage so I moved the latter to be run just after the former.

test/configs/all-with-oscap-fips.json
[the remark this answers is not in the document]
mmartinv: Yeas, It's a copy of of `all-with-oscap.json` but adding the fips customization to it

kingsleyzissou:
> It looks like the oscap remediation stage was setting the system's crypto policy to `DEFAULT` after it was set to `FIPS` by the `update-crypto-policies` stage so I moved the latter to be run just after the former.

Yeah this was the issue I was running into, I haven't had time for getting back to it. Thanks for that

evgenyz:
> > FTR I'm testing this locally. Just ran into one or two issues, will report back
> 
> It looks like the oscap remediation stage was setting the system's crypto policy to `DEFAULT` after it was set to `FIPS` by the `update-crypto-policies` stage so I moved the latter to be run just after the former.

That's not good. Ideally oscap should not switch policy from FIPS if it is set in the profile and use the FIPS mode it to adjust its behaviour. 

Which profile does that?

evgenyz:
OTOH, if the profile requires a different policy it should be OK that the FIPS mode is reset as the result of remediation.

TLDR: FIPS mode should be set before oscap stage. If there is a discrepancy in the behaviour it should be solved on the oscap/CaC side.

evgenyz:
I'd even say that you should isolate enablement of Blueprint's `fips=1` itself from the `oscap` stages. We'll deal with oscap later.

mmartinv:
> That's not good. Ideally oscap should not switch policy from FIPS if it is set in the profile and use the FIPS mode it to adjust its behaviour.
> 
> Which profile does that?

The blueprint configuration regarding oscap is:
```
      "openscap": {
        "profile_id": "xccdf_org.ssgproject.content_profile_pci-dss",
        "tailoring": {
          "selected": [
            "bind_crypto_policy"
          ],
          "unselected": [
            "rpm_verify_hashes",
            "enable_fips_mode"
          ]
        }
      }
```

I also tested by removing the `enable_fips_mode` from the unselected rules but the result is the same: the crypto policy is reset to the `DEFAULT` policy.

mmartinv:
Instead of using a copy of the `all-with-oscap.json` configuration I used `all-customizations.json` and added the `fips-=true`. Everything should work now.

evgenyz:
> > That's not good. Ideally oscap should not switch policy from FIPS if it is set in the profile and use the FIPS mode it to adjust its behaviour.
> > Which profile does that?
 
```
>         "profile_id": "xccdf_org.ssgproject.content_profile_pci-dss",
>         "tailoring": {
>           "selected": [
>             "bind_crypto_policy"
```

The PCI-DSS profile does not require FIPS mode. It also ensures that DEFAULT crypto policy is in effect. No surprises here.

```
>           "unselected": [
>             "rpm_verify_hashes",
>             "enable_fips_mode"
```

> I also tested by removing the `enable_fips_mode` from the unselected rules but the result is the same: the crypto policy is reset to the `DEFAULT` policy. I guess the problem is the `bind_crypto_policy` maybe?

The PCI-DSS profile actually does not have `enable_fips_mode` rule enabled. But even if it would have, the won't be executed with the current version of `ssg` content as it is disabled globally in osbuild env.

Anyhow, as I said, just get rid of oscap section. Test only BP customization itself. The oscap stage will be bootstrapped with initial BP customization and won't interfere unless misconfigured (but we'll deal with that ourselves).

pkg/distro/rhel8/edge.go
```diff
@@ -7,9 +7,21 @@ import (
...
 
+var (
+	edgeDracutConf = []*osbuild.DracutConfStageOptions{
+		{
+			Filename: "40-fips.conf",
+			Config: osbuild.DracutConfigFile{
+				AddModules: []string{"fips"},
+			},
+		},
+	}
```
achilleas-k, line 22: Why is this defined here instead of using the stages defined in `obsuild/fips.go` where it was added in the first commit of this PR?

achilleas-k (commented):
Looks good overall.  Just one small question about the dracut stage and where it should be defined and reused.

pkg/distro/rhel8/edge.go
mmartinv: We need to include the dracut `fips` module within the initramfs in `edge-commit` and `edge-container` images by default because I didn't find a way to regenerate the initramfs when building an `edge-raw-image` (or any other image based on an ostree deployment) from an `edge-container`:
- Adding the dracut conf works as expected (the conf file is there in the resulting image), but the initramfs is not regenerated.
- With the current `org.ostree.dracut` implementation we can't regenerate the initramfs without knowing the actual kernel version within the remote ostree.
It's also added to `edge-commit` images to make sure it's always there and we don't break anything when updating (not sure if it's really needed though)

achilleas-k: Right, that's all good, I'm just wondering if it would be easier to define the stage options for the fips dracut module in one place and reuse it from there where needed instead of defining the same config in multiple places.

evgenyz (approved):

kingsleyzissou (approved):
Looks good to me :)

PR #287, opened 2023-11-29
image: add new `ContainerBuildable` flag to OSTreeDiskImage (HMS-3141)

mvo5:
One objective for bifrost images is that it should be possible to run osbuild inside a container. This can interfere with the selinux policies of the buildroot.

Inside the container everything is labeled
`system_u:object_r:container_files_t`. Labeling /usr/bin/osbuild as `osbuild_exec_t` is not possible in the general case because the host may not have `osbuild-selinux` installed that contains this type. The workaround is that the container labels osbuild itself as `install_exec_t`. This works fine however there is a selinux denial warning when the `{,u}mount` binary is called because the transition from `install_t`->`mount_t` is not allowed. The warning is "harmless" because `install_t` has enough privs to allow the `{,u}mount` binaries to work.

To silence this warning we can label `{,u}mount` in the buildroot as `install_exec_t` directly.

This commit allows to control this now via the `ContainerBuildable` flag that can be set on `manifest.Build` to enable this behavior.

---

Once this has landed https://github.com/osbuild/osbuild-deploy-container/pull/11/files#diff-895717f8731ac528a1e229abd5c4b2d22eb2c471412ea30c1df61b88dd90e4f5R49 will need an update to do `img.ContainerBuildlable = true`

### DIFF
--- a/pkg/image/ostree_disk_test.go
+++ b/pkg/image/ostree_disk_test.go
@@ -27,14 +27,16 @@ func TestOSTreeDiskImageManifestSetsContainerBuildable(t *testing.T) {
 		Name:   "name",
 	}
 
-	var buildPipeline *manifest.Build
-	restore := image.MockManifestNewBuild(func(m *manifest.Manifest, r runner.Runner, repos []rpmmd.RepoConfig) *manifest.Build {
-		buildPipeline = manifest.NewBuild(m, r, repos)
-		return buildPipeline
+	var buildOpts []*manifest.BuildOptions
+	restore := image.MockManifestNewBuild(func(m *manifest.Manifest, r runner.Runner, repos []rpmmd.RepoConfig, opts *manifest.BuildOptions) *manifest.Build {
+		buildOpts = append(buildOpts, opts)
+		return manifest.NewBuild(m, r, repos, opts)
 	})
 	defer restore()
 
 	for _, containerBuildable := range []bool{true, false} {
+		buildOpts = nil
+
 		mf := manifest.New()
 		img := image.NewOSTreeDiskImageFromContainer(containerSource, ref)
 		require.NotNil(t, img)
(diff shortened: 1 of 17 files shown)


pkg/manifest/build.go
```diff
@@ -22,6 +22,11 @@ type Build struct {
 	dependents   []Pipeline
 	repos        []rpmmd.RepoConfig
 	packageSpecs []rpmmd.PackageSpec
+
+	// TODO: make private?
```
mvo5, line 26: Would love to get ideas/feedback on this - it seems a pattern we use in `manifest.Build` but not too sure if it's appropriate here?

pkg/image/ostree_disk_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/osbuild/images/pkg/runner"
+)
+
+func TestOSTreeDiskImageManifestSetsContainerBuildable(t *testing.T) {
+	rng := rand.New(rand.NewSource(0))
+
+	repos := []rpmmd.RepoConfig{}
+	r := &runner.Fedora{Version: 39}
+
+	ref := "ostree/1/1/0"
```
cgwalters, line 24: Would still like to drop this from pipeline flows, it's hardcoding an ostree implementation detail

ondrejbudai: Yup, but that's a discussion for https://github.com/osbuild/osbuild-deploy-container/issues/18 :)

pkg/manifest/build.go
achilleas-k: The pattern/rule for these things is that private members should be options that are mandatory and shouldn't be changed outside the constructor.  So we make them private and force them to be set by making them an argument to the `NewX()` function.  Anything optional, options that aren't absolutely required for a pipeline or image, are public and are set by any caller that needs them.

mvo5: Thanks for your input on this! This sounds like my approach was okay(?). I did push a small commit for review that uses a slightly different route by introducing a new optional "BuildOptions" that can be omited. But happy to remove it again, I realize it probably does not fit that well. Just let me know what you prefer :) (or if there is a third option that I might have overlooked).

achilleas-k: It could be a good solution for extending the options in the future.  Might be worth considering the same for other pipelines at some point.

PR #288, opened 2023-11-29
osbuild: remove unneccessary debug print during the test

mvo5:
Instead of printing the metadata do some basic precondition checks and stop printing. The (no longer) printed data was:
```
Result: &osbuild.PipelineMetadata{"org.osbuild.rpm":(*osbuild.RPMStageMetadata)(0xc0005b71d0)}
```


bcl (approved):
Ack

PR #289, opened 2023-11-29
osbuild: include context in skopeo error messages

mvo5:
When trying to create enough data for a full pipeline run for https://github.com/osbuild/images/pull/287 I struggled a bit because it was not clear which items where faulty.

This commit adds more context to the errors. Sadly I had to implement `assertPanicsWithErrorRegexp` because of https://github.com/stretchr/testify/issues/1304

I can contribute it upstream but for that it needs some tests around it first :)

### DIFF
--- a/pkg/osbuild/skopeo_source_test.go
+++ b/pkg/osbuild/skopeo_source_test.go
@@ -1,10 +1,12 @@
 package osbuild
 
 import (
+	"regexp"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
+	"github.com/osbuild/images/internal/assertx"
 	"github.com/osbuild/images/internal/common"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/osbuild/skopeo_source_test.go
```diff
@@ -33,27 +52,30 @@ func TestNewSkopeoSource(t *testing.T) {
 	assert.Nil(t, item.Image.TLSVerify)
 
 	// empty name
-	assert.Panics(t, func() {
+	expectedErr := regexp.MustCompile(`source item osbuild.SkopeoSourceItem.* has empty name`)
```
bcl, line 55: Since this is needed every time I'd just pass the regex as an argument and let the function compile it.

achilleas-k:
Fixed a typo in the PR name.  Hope you don't mind :)

bcl (approved):
Looks good!

PR #290, opened 2023-11-29
many: use `common.{Ki,Me,Gi}Bi` instead of manually doing them

mvo5:
This commit uses `common.{Ki,Me,Gi}Bi` in places that used to write `10 * 1024 * 1024` etc. It's not really important but feels quicker to scan when having the constants.

Also I wonder if we should also have common.{KiB,MiB,GiB,TiB} as a shorthand?

Also just close the PR  if you feel this is not worth it :)


cmd/osbuild-playground/playground.go
```diff
@@ -21,7 +22,7 @@ func RunPlayground(img image.ImageKind, d distro.Distro, arch distro.Arch, repos
 	solver.SetDNFJSONPath(findDnfJsonBin())
 
 	// Set cache size to 3 GiB
```
ondrejbudai, line 24: hehe, this comment is funny.

ondrejbudai (approved):
Much better, thanks! :)

bcl:
> Also I wonder if we should also have common.{KiB,MiB,GiB,TiB} as a shorthand?

YES :) I don't know anyone who likes to use the long form...

achilleas-k:
> Also I wonder if we should also have common.{KiB,MiB,GiB,TiB} as a shorthand?

Yes please!

thozza (dismissed):
Thanks a lot!

achilleas-k:
Needs a rebase and merge conflict fixing

cmd/osbuild-playground/playground.go
mvo5: Indeed it is, I was inclined to remove it entirely but wanted to avoid mixing too many concerns into a single PR. I at least fixed it - thanks for pointing this out :)

mvo5:
> Needs a rebase and merge conflict fixing

Thank you! This is rebased and updated to add/use the shorthand form - fwiw, I like the shorthand a lot better as well :)

thozza (approved):
LGTM

PR #292, opened 2023-11-30
upload/azure: create v2 storage account

ondrejbudai:
We got feedback that v1 storage accounts are not allowed by certain IT departments. According to the documentation, there are no downsides to just switching to v2. V2 incorporates all features of v1.

This commit thus switches the default version of NEW storage accounts to v2.

See
https://learn.microsoft.com/en-us/azure/storage/common/storage-account-upgrade

While doing this, I noticed that we have two copies of `upload` in composer and images, so I made the one public to deduplicate them.


croissanne (approved):

achilleas-k:
> While doing this, I noticed that we have two copies of `upload` in composer and images, so I made the one public to deduplicate them.

I was thinking about this a lot when doing the pulp integration in composer.  I started trying to do it "properly" by making all uploaders and targets public (or maybe just uploaders) then gave up and did it all in composer.
I think uploaders in images makes sense, so happy to see this change now.

achilleas-k (approved):

PR #296, opened 2023-12-01
Make customization types public under pkg/customizations

achilleas-k:
ImageKind and Pipeline structs allow defining customizations using types that we define in the image definitions.  These are intermediate representations of customizations and options between the front-facing Blueprint customizations and the osbuild options.  Sometimes, they are identical to one of the two, but this is not a rule and are meant to give us the freedom to change them freely without affecting the UI (Blueprints) or the Manifest (osbuild options).  For example, a single Blueprint customization can be combined with the target distribution to set a value on the Pipeline using an intermediate customization which in turn can be applied to multiple stage options that need to work together to apply the customization.

Making these public is necessary to support users of the images library to interact with the Pipeline generators.  For example, it's currently not possible for an external user of the library to add Users and Groups to any Pipeline that embeds OSCustomizations, since the types it requires are internal.

We need this to support user creation on the `OSTreeDiskImage` in https://github.com/osbuild/osbuild-deploy-container.


mvo5 (commented):
Fwiw, this looks fine to me but a second review by someone more knowledgeable/experienced would probably be a good idea.

thozza (approved):
LGTM, this makes a lot of sense!

PR #297, opened 2023-12-01
distro/el9: refactor package sets to again use `@core` pkg group (COMPOSER-2074)

thozza:
Previously, we used a fixed package list from RHEL-9.0 times, instead of
the `@core` pkg group, for all images which used it on el8. The
motivation was that some packages got removed from the group in 9.0
(specifically tuned), which went almost unnoticed. The reasoning was that
a fixed list would give us more control over the package set and prevent
unexpected package removals.

However, with COMPOSER-2074 [1], it turned out that this also meant that
any new packages added to the `@core` group won't be automatically added
to any image which previously used the `@core` group. People still have
(anyhow incorrect) expectation that `@core` pkg group is always
installed on RHEL images and thus that adding a package to the group
will also add it to vanilla RHEL images.

We discussed this shortly within the team and reverting back to using
`@core` package group, instead of a fixed package list, seems like the
lesser evil.

Modify package set definitions, which used `coreOsCommonPackageSet()` to
use `@core` package group instead. Modify the exclude lists accordingly
to achieve almost the same image content as before. There are a few
exceptions:

 - `rhc` package is now installed, which is desired.
 - `insights-client` is now installed on some images where it was
   previously not. This is OK, since RHEL is aiming for a "connected
   experience" by default. This also results in a few additional
   packages to be pulled in as dependencies, specifically:
   - `python3-file-magic`
   - `policycoreutils-python-utils`
   - `tar`
   - `pciutils`
 - `initscripts-rename-device` is now installed. This sub-package has
   been split out of the `initscripts` package on el9, but it was not
   included in the `@core` group definition, when we created the fixed
   pkg list. However, it is now the default member of the group. The
   functionality was previously installed by default on el8 images via
   the `initscripts` package. After talking to the maintainer, I kept it
   in the images as a new package.

Affected images by this change:
 - ami
 - gce
 - gce_rhui
 - image_installer
 - oci
 - openstack
 - ova
 - qcow2
 - vmdk

Some of the listed images already contained `rhc` or `insights-client`,
but now, these are consistently installed on all of them. This is the
case since 9.2, when `rhc` was added to the `@core` group.

`initscripts-rename-device` has been added to all of these image types,
inlcuding on CentOS Stream 9 images.

[1] https://issues.redhat.com/browse/COMPOSER-2074

This resolves COMPOSER-2074.


achilleas-k (approved):
This is a nice cleanup.  Flip-flopping on things like this can get annoying but I'm glad we tried it at least and Now We Know!

Thanks!!

PR #299, opened 2023-12-04
repository: Normalize handling of false for module_hotfixes

ezr-ondrej:
This removes nil boolean values from Hash calculation.


ezr-ondrej:
This is just a clean-up  after introducing new function in hash just for one boolean flag.
I do not know the consequences of changing Hash calculation, so happy to lear if someone knows what else needs to happen for this not to break stuff :)

ezr-ondrej:
Shall I self-merge? :t-rex:

PR #300, opened 2023-12-04
OpenSCAP tailoring: add key/value rule overrides

kingsleyzissou:
Expose a field in the blueprints to be able to override rule values
using the OpenSCAP autotailor feature. The customization exposes an
array with two values, the name of the var to overwrite and the value
to write to the var.

This depends on https://github.com/osbuild/osbuild/pull/1407
Jira: https://issues.redhat.com/browse/COMPOSER-1994

### DIFF
--- a/pkg/blueprint/customizations.go
+++ b/pkg/blueprint/customizations.go
@@ -109,7 +109,7 @@ type ServicesCustomization struct {
 }
 
 type OpenSCAPCustomization struct {
-	DataStream string                           `json:"datastream,omitempty" toml:"datastream,omitempty"`
+	Datastream string                           `json:"datastream,omitempty" toml:"datastream,omitempty"`
 	ProfileID  string                           `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
 	Tailoring  *OpenSCAPTailoringCustomizations `json:"tailoring,omitempty" toml:"tailoring,omitempty"`
 }
--- a/pkg/blueprint/customizations.go
+++ b/pkg/blueprint/customizations.go
@@ -108,17 +108,6 @@ type ServicesCustomization struct {
 	Disabled []string `json:"disabled,omitempty" toml:"disabled,omitempty"`
 }
 
-type OpenSCAPCustomization struct {
-	Datastream string                           `json:"datastream,omitempty" toml:"datastream,omitempty"`
-	ProfileID  string                           `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
-	Tailoring  *OpenSCAPTailoringCustomizations `json:"tailoring,omitempty" toml:"tailoring,omitempty"`
-}
-
-type OpenSCAPTailoringCustomizations struct {
-	Selected   []string `json:"selected,omitempty" toml:"selected,omitempty"`
-	Unselected []string `json:"unselected,omitempty" toml:"unselected,omitempty"`
-}
-
 type CustomizationError struct {
 	Message string
 }
@@ -314,13 +303,6 @@ func (c *Customizations) GetFDO() *FDOCustomization {
 		return nil
 	}
 	return c.FDO
-}
-
-func (c *Customizations) GetOpenSCAP() *OpenSCAPCustomization {
-	if c == nil {
-		return nil
-	}
-	return c.OpenSCAP
 }
 
 func (c *Customizations) GetIgnition() *IgnitionCustomization {
--- a/pkg/blueprint/openscap_customizations.go
+++ b/pkg/blueprint/openscap_customizations.go
@@ -1,0 +1,78 @@
+package blueprint
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type OpenSCAPCustomization struct {
+	Datastream string                           `json:"datastream,omitempty" toml:"datastream,omitempty"`
+	ProfileID  string                           `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
+	Tailoring  *OpenSCAPTailoringCustomizations `json:"tailoring,omitempty" toml:"tailoring,omitempty"`
+}
+
+type OpenSCAPTailoringCustomizations struct {
+	Selected   []string                    `json:"selected,omitempty" toml:"selected,omitempty"`
+	Unselected []string                    `json:"unselected,omitempty" toml:"unselected,omitempty"`
+	Overrides  []OpenSCAPTailoringOverride `json:"overrides,omitempty" toml:"overrides,omitempty"`
+}
+
+type OpenSCAPTailoringOverride struct {
+	Var   string      `json:"var,omitempty" toml:"var,omitempty"`
+	Value interface{} `json:"value,omitempty" toml:"value,omitempty"`
+}
+
+func (c *Customizations) GetOpenSCAP() *OpenSCAPCustomization {
+	if c == nil {
+		return nil
+	}
+	return c.OpenSCAP
+}
+
+func (ot *OpenSCAPTailoringOverride) UnmarshalTOML(data interface{}) error {
+	d, _ := data.(map[string]interface{})
+
+	switch d["var"].(type) {
+	case string:
+		ot.Var = d["var"].(string)
+	default:
+		return fmt.Errorf("TOML unmarshal: override var must be string, got %v of type %T", d["var"], d["var"])
+	}
+
+	switch d["value"].(type) {
+	case int64:
+		ot.Value = uint64(d["value"].(int64))
+	case string:
+		ot.Value = d["value"].(string)
+	default:
+		return fmt.Errorf("TOML unmarshal: override value must be integer or string, got %v of type %T", d["value"], d["value"])
+	}
+
+	return nil
+}
+
+func (ot *OpenSCAPTailoringOverride) UnmarshalJSON(data []byte) error {
+	var v interface{}
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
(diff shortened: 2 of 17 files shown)


pkg/osbuild/oscap_autotailor_stage.go
```diff
@@ -8,9 +8,15 @@ type OscapAutotailorStageOptions struct {
...
-	Unselected []string `json:"unselected,omitempty"`
+	NewProfile string                    `json:"new_profile"`
+	Selected   []string                  `json:"selected,omitempty"`
+	Unselected []string                  `json:"unselected,omitempty"`
+	Overrides  []OscapAutotailorOverride `json:"overrides,omitempty"`
+}
+
+type OscapAutotailorOverride struct {
+	Var   string      `json:"var"`
+	Value interface{} `json:"value"`
```
thozza, line 19: I think that it would be good to ensure that the `Value` can be only `int` or `string`. This way, one could set anything as the value (`bool`, `nil`, etc..).

Probably an equivalent of https://github.com/osbuild/images/blob/82de94ad6cf7979156e274e1d5e9c94d718e6f62/pkg/osbuild/pam_limits_conf_stage.go#L79-L102

pkg/distro/rhel9/images.go
```diff
@@ -199,9 +199,18 @@ func osCustomizations(
 				panic(fmt.Sprintf("unexpected error creating tailoring file options: %v", err))
 			}
 
+			var overrides []osbuild.OscapAutotailorOverride
+			for _, override := range oscapConfig.Tailoring.Overrides {
+				overrides = append(overrides, osbuild.OscapAutotailorOverride{
+					Var:   override.Var,
+					Value: override.Value,
+				})
+			}
```
thozza, line 208: Seeing this multiple times in each distro, it feels like a common conversion function would make sense, instead of duplicating the logic. Basically something to convert `oscapConfig.Tailoring` to `osbuild.OscapAutotailorConfig`. What do you think? :thinking:

thozza (commented):
Nice work. I added a few comments / ideas inline.

pkg/distro/rhel9/images.go
kingsleyzissou: Yeah, that's a good idea. I will do that :smile:

kingsleyzissou: I did quite a big refactor here, but in the long run I think it's worth it. Let me know what you think

pkg/blueprint/customizations.go
```diff
@@ -114,8 +114,14 @@ type OpenSCAPCustomization struct {
...
-	Selected   []string `json:"selected,omitempty" toml:"selected,omitempty"`
-	Unselected []string `json:"unselected,omitempty" toml:"unselected,omitempty"`
+	Selected   []string                    `json:"selected,omitempty" toml:"selected,omitempty"`
+	Unselected []string                    `json:"unselected,omitempty" toml:"unselected,omitempty"`
+	Overrides  []OpenSCAPTailoringOverride `json:"overrides,omitempty" toml:"overrides,omitempty"`
+}
+
+type OpenSCAPTailoringOverride struct {
+	Var   string      `json:"var,omitempty" toml:"var,omitempty"`
+	Value interface{} `json:"value,omitempty" toml:"value,omitempty"`
```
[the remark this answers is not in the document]
kingsleyzissou: Ah I added it anyway, I'll do the same in composer too

pkg/osbuild/oscap_autotailor_stage.go
kingsleyzissou: I did a simple implementation, but I think the above is more elegant. I'll come back to this.

pkg/customizations/oscap/stage_options.go
```diff
@@ -29,3 +33,34 @@ func CreateRequiredDirectories(createTailoring bool) ([]*fsnode.Directory, error
 
 	return directories, nil
 }
+
+func getTailoringProfileID(profileID string) string {
+	return fmt.Sprintf("%s_osbuild_tailoring", profileID)
+}
+
+func CreateTailoringStageOptions(oscapConfig *blueprint.OpenSCAPCustomization, d distro.Distro) *osbuild.OscapAutotailorStageOptions {
```
thozza, line 41: I think that it would make more sense to move this to the `osbuild` package, instead of the `oscap`. We usually generate osbuild stage options from our internal abstractions in the `osbuild` package.

Lastly, a constructor function for stage options would IMHO feel more natural in the `osbuild` package.

pkg/distro/fedora/images.go
```diff
@@ -174,25 +174,14 @@ func osCustomizations(
...
-
-			// overwrite the profile id with the new tailoring id
-			oscapStageOptions.ProfileID = newProfile
-			oscapStageOptions.Tailoring = tailoringFilepath
+		osc.OpenSCAPTailorConfig = oscap.CreateTailoringStageOptions(
+			oscapConfig,
+			t.arch.distro,
+		)
+
+		if tailorConfig := osc.OpenSCAPTailorConfig; tailorConfig != nil {
```
thozza: Is this only to make the following two lines shorter? I must admit that I had to stop for a second to carefully read the code and understand that you are assigning a variable to another variable just to check if it is `nil`...

pkg/customizations/oscap/stage_options.go
```diff
@@ -64,3 +64,38 @@ func CreateTailoringStageOptions(oscapConfig *blueprint.OpenSCAPCustomization, d
 		},
 	)
 }
+
+func CreateRemediationStageOptions(
```
thozza, line 77: Again, this should IMO go into the `osbuild` package.

pkg/customizations/oscap/stage_options.go
```diff
@@ -64,3 +64,38 @@ func CreateTailoringStageOptions(oscapConfig *blueprint.OpenSCAPCustomization, d
...
+	oscapConfig *blueprint.OpenSCAPCustomization,
```
thozza, line 78: It would be better if this function would accept some internal abstraction of the OSCAP config, instead of BP customization.

pkg/customizations/oscap/stage_options.go
```diff
@@ -64,3 +64,38 @@ func CreateTailoringStageOptions(oscapConfig *blueprint.OpenSCAPCustomization, d
...
+	isOSTree bool,
+	d distro.Distro,
+) (*osbuild.OscapRemediationStageOptions, []*fsnode.Directory, error) {
+	if oscapConfig == nil {
+		return nil, nil, nil
+	}
+
+	if isOSTree {
+		return nil, nil, fmt.Errorf("unexpected oscap options for ostree image type")
+	}
```
thozza, line 88: This is IMO not the place to check if the image type is ostree or not. This is not the job if a function which generates an osbuild stage options. Instead, this should be checked on a higher-lervel in the distro implementation and we should return an error there in case one wants to use BP customizations with an image type that does not support them.

pkg/distro/fedora/images.go
```diff
@@ -161,39 +161,22 @@ func osCustomizations(
...
-
-		directories, err := oscap.CreateRequiredDirectories(oscapConfig.Tailoring != nil)
-		if err != nil {
-			panic(err)
-		}
-
-		if len(directories) > 0 {
-			osc.Directories = append(osc.Directories, directories...)
-		}
+	var directories []*fsnode.Directory
```
thozza, line 164: nitpick: consider renaming this to something like `oscapDirectories` to make it obvious what the variable is used for.

pkg/blueprint/openscap_customizations.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (ot *OpenSCAPTailoringOverride) UnmarshalTOML(data interface{}) error {
+	d, _ := data.(map[string]interface{})
+
+	switch d["var"].(type) {
+	case string:
+		ot.Var = d["var"].(string)
+	default:
+		return fmt.Errorf("TOML unmarshal: override var must be string, got %v of type %T", d["var"], d["var"])
```
thozza, line 39: nitpick: you could use this instead of specifying the same value twice:
```go
return fmt.Errorf("TOML unmarshal: override var must be string, got %[1]v of type %[1]T", d["var"])
```
and basically in all similar cases in this file.

pkg/osbuild/oscap_autotailor_stage.go
```diff
@@ -27,6 +33,17 @@ func (c OscapAutotailorConfig) validate() error {
...
+	for _, override := range c.Overrides {
+		if _, ok := override.Value.(uint64); ok {
+			continue
+		}
+
+		if _, ok := override.Value.(string); ok {
+			continue
+		}
+
+		return fmt.Errorf("override 'value' must be an integere or a string")
```
thozza, line 45: would it make sense to also include the name of the override value in the error?

pkg/osbuild/oscap_autotailor_stage.go
thozza: The validation check is better than nothing :slightly_smiling_face: Thanks!

thozza (commented):
Very nice rework (plus of course the added support for override values), I like it!

I added a few comments, mostly of a cosmetic nature. On top of that, I think that moving all functions that generate osbuild stage options should be moved to the `osbuild` package.

pkg/distro/rhel9/images.go
thozza: I had a few comments, but in general I like it a lot. Nice work!

PR #306, opened 2023-12-07
HMS-3235 import containers in local containers-storage

kingsleyzissou:
This PR adds the ability to import containers from local containers-storage. It additionally, adds the ability to specify a custom storage path that a user can configure. This will then update the configurations for the `/etc/containers/storage.conf` file and add the storage path to the additional image stores.

Jira: https://issues.redhat.com/browse/HMS-3235


pkg/blueprint/blueprint.go
```diff
@@ -29,10 +29,12 @@ type Group struct {
...
 type Container struct {
-	Source string `json:"source" toml:"source"`
-	Name   string `json:"name,omitempty" toml:"name,omitempty"`
+	Source string  `json:"source,omitempty" toml:"source"`
+	Name   string  `json:"name,omitempty" toml:"name,omitempty"`
+	Digest *string `json:"digest,omitempty" toml:"digest,omitempty"`
 
-	TLSVerify *bool `json:"tls-verify,omitempty" toml:"tls-verify,omitempty"`
+	TLSVerify         *bool `json:"tls-verify,omitempty" toml:"tls-verify,omitempty"`
+	ContainersStorage *bool `json:"containers-storage,omitempty" toml:"containers-storage,omitempty"`
```
kingsleyzissou, line 37: Just went with this as a key rather than something like `local` or `local-container`... but one of those might be clearer.

kingsleyzissou:
I've made it possible to configure the image store to which the containers get saved, given the following blueprint:
```json
{
  "name": "embed-containers",
  "blueprint": {
    "packages": [
      {
        "name": "podman",
        "version": "*"
      }
    ],
    "containers": [
      {
        "source": "registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/fedora-minimal"
      },
      {
        "source": "localhost/hello",
        "containers-transport": "containers-storage"
      }
    ],
    "customizations": {
      "containers-storage": {
        "storage-path": "/ostree/container-storage"
      }
    }
  }
}

```

We get the following output from the `image-info` tool:

```json
[
  {
    "Created": "2023-12-12T12:54:40.001637346Z",
    "Digest": "sha256:d8bac64df06e506dd2e285a9e52ea86ee276637d1ad1cf1c89bb1f3b80b0d5ad",
    "Id": "5dc7f15873a23033de66f8734cb61969a9c749ded0cfb56300e785bfaf5ef528",
    "Names": [
      "localhost/hello:latest"
    ]
  },
  {
    "Created": "2022-07-11T13:16:13Z",
    "Digest": "sha256:4d76a7480ce1861c95975945633dc9d03807ffb45c64b664ef22e673798d414b",
    "Id": "d4ee87dab8193afad523b1042b9d3f5ec887555a704e5aaec2876798ebb585a6",
    "Names": [
      "registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/fedora-minimal:latest"
    ]
  }
]

```

achilleas-k:
Does the ContainersStorage customization in the blueprint control both where to get containers from and where to install them?

I think we need to separate two things here:
1. The path to the local container storage on the host where a container source will be pulled from.
2. The path where a container will be stored in the image when it's embedded during a build.

The customization in the blueprint is, I think, good for 2., configuring the container storage path for the image itself.
But this path should be the same for both.  For example, it should be possible to have containers on the host at `/var/lib/containers`, pull them from there, and then install them into the image at `/usr/share/containers`, if you're building an ostree-based image from a non-ostree host.
But for 1. I think the option should be separate for each container.  I'm imagining a scenario where we're building an image (of any type) and we want to pull several containers, some from a registry and others from the local storage, and then we also want to control where they're stored in the image.  In that case, I'm imagining the blueprint looking like this:
```json
{
  "name": "embed-containers",
  "blueprint": {
    "packages": [
      {
        "name": "podman",
        "version": "*"
      }
    ],
    "containers": [
      {
        "source": "registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/fedora-minimal"
      },
      {
        "source": "localhost/hello",
        "containers-transport": "containers-storage",
        "storage-location": "/var/lib/containers/storage"
      }
    ],
    "customizations": {
      "containers-storage": {
        "storage-path": "/data/containers/storage"
      }
    }
  }
}
```

kingsleyzissou:
> Does the ContainersStorage customization in the blueprint control both where to get containers from and where to install them?
> 
> 
> 
> I think we need to separate two things here:
> 
> 1. The path to the local container storage on the host where a container source will be pulled from.
> 
> 2. The path where a container will be stored in the image when it's embedded during a build.
> 
> 
> 
> The customization in the blueprint is, I think, good for 2., configuring the container storage path for the image itself.
> 
> But this path should be the same for both.  For example, it should be possible to have containers on the host at `/var/lib/containers`, pull them from there, and then install them into the image at `/usr/share/containers`, if you're building an ostree-based image from a non-ostree host.
> 
> But for 1. I think the option should be separate for each container.  I'm imagining a scenario where we're building an image (of any type) and we want to pull several containers, some from a registry and others from the local storage, and then we also want to control where they're stored in the image.  In that case, I'm imagining the blueprint looking like this:
> 
> ```json
> 
> {
> 
>   "name": "embed-containers",
> 
>   "blueprint": {
> 
>     "packages": [
> 
>       {
> 
>         "name": "podman",
> 
>         "version": "*"
> 
>       }
> 
>     ],
> 
>     "containers": [
> 
>       {
> 
>         "source": "registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/fedora-minimal"
> 
>       },
> 
>       {
> 
>         "source": "localhost/hello",
> 
>         "containers-transport": "containers-storage",
> 
>         "storage-location": "/var/lib/containers/storage"
> 
>       }
> 
>     ],
> 
>     "customizations": {
> 
>       "containers-storage": {
> 
>         "storage-path": "/data/containers/storage"
> 
>       }
> 
>     }
> 
>   }
> 
> }
> 
> ```

Yeah spot on. The customisation is for the destination so that answers #2.

And your suggestion for number #1 (the storage path) will work, it was only checking `/var/lib/containers/storage`. Maybe we could leave that as a sane default?

achilleas-k:
> Maybe we could leave that as a sane default?

Definitely yeah.  Unless it's more correct to read the storage.conf file.

kingsleyzissou:
> > Maybe we could leave that as a sane default?
> 
> Definitely yeah. Unless it's more correct to read the storage.conf file.

Ah you're right, skopeo reads the the `storage.conf` file. So makes more sense to use that.

kingsleyzissou:
This is needed: https://github.com/osbuild/osbuild/pull/1504

bcl:
```
    {
        "source": "localhost/hello",
        "containers-transport": "containers-storage",
        "storage-location": "/var/lib/containers/storage"
      }
```

This is the source storage location, right? It seems mildly confusing to call it storage-location instead of something like source-location.

kingsleyzissou:
> ```
>     {
>         "source": "localhost/hello",
>         "containers-transport": "containers-storage",
>         "storage-location": "/var/lib/containers/storage"
>       }
> ```
> 
> This is the source storage location, right? It seems mildly confusing to call it storage-location instead of something like source-location.

Yeah this is a great point. Maybe we could flip this on it's head a bit? Rename the customization to `destination-path` or `dest-storage`? And then rename the source to `storage-path` (or even be more explicit and go with `source-path`?). Location seems a bit too vague.

```json
{
    "source": "localhost/hello",
    "containers-transport": "containers-storage",
    "source-path": "/var/lib/containers/storage"
}
  ```
  
  and
  
  ```json
  {
      "customizations": {
          "containers-storage": {
              "destination-path": "/data/containers/storage"
          }
      }
  }
  ```

**Comment by bcl** [comment]
I like that! It's consistent between the two related parts with `*-path` and I don't have to read it twice to understand what it's supposed to be doing.

**Review by achilleas-k — APPROVED** [review]
LGTM! :rocket:

**Review by achilleas-k — CHANGES_REQUESTED** [review]
Removing approval.  Option rename pending.

**Review by achilleas-k — APPROVED** [review]
All sorted now.  LGTM!

**Comment by achilleas-k** [comment]
Oh, the CI runners need btrfs-tools installed now too.  It's a build dependency of the project.  Can you update the test configs to install it?
- .gitlab-ci.yml
- test/scripts/configure-generators
- test/scripts/generate-build-config
- test/scripts/generate-ostree-build-config

Also update the contribution guides (docs/developer/README.md and CONTRIBUTING.md.

Now that I'm listing it all I realise this should all probably be defined in one location.  A script or some equivalent to a Python package's `requirements.txt` that we can point to and read from every other location.

**Comment by kingsleyzissou** [comment]
> Now that I'm listing it all I realise this should all probably be defined in one location. A script or some equivalent to a Python package's requirements.txt that we can point to and read from every other location.

Sounds like a good idea. Since we're referencing it in multiple places. I will update the PR

File: pkg/osbuild/skopeo_source_test.go [hunk 2]
```diff
@@ -37,28 +37,28 @@ func TestNewSkopeoSource(t *testing.T) {
...
 	expectedErr = regexp.MustCompile(`item osbuild.SkopeoSourceItem.* has invalid digest`)
 	assertx.PanicsWithErrorRegexp(t, expectedErr, func() {
-		source.AddItem("name", "foo", imageID, nil)
+		source.AddItem("name", "foo", imageID, nil, nil, nil)
 	})
 
 	// invalid image id
 	assert.PanicsWithError(t, `item "sha256:foo" has invalid image id`, func() {
-		source.AddItem("name", testDigest, "sha256:foo", nil)
+		source.AddItem("name", testDigest, "sha256:foo", nil, nil, nil)
```
mvo5, line 62: Would it make sense to test the new validations here as well? Like:
```go
--- a/pkg/osbuild/skopeo_source_test.go
+++ b/pkg/osbuild/skopeo_source_test.go
@@ -61,4 +61,9 @@ func TestNewSkopeoSource(t *testing.T) {
        assert.PanicsWithError(t, `item "sha256:foo" has invalid image id`, func() {
                source.AddItem("name", testDigest, "sha256:foo", nil, nil, nil)
        })
+
+       // invalid image transport
+       assert.PanicsWithError(t, `invalid container transport`, func() {
+               source.AddItem("name", testDigest, testDigest, nil, common.ToPtr("invalid-transport"), nil)
+       })
 }
```
(maybe more, I have not looked too deeply, sorry for that)

pkg/osbuild/skopeo_source.go
```diff
@@ -47,6 +54,12 @@ func (item SkopeoSourceItem) validate() error {
 		return fmt.Errorf("source item %#v has invalid digest", item)
 	}
 
+	if transport := item.Image.ContainersTransport; transport != nil {
+		if *transport != DockerTransport && *transport != ContainersStorageTransport {
```
mvo5, line 58: I wonder if a common helper like `validateTransport` could be extracted here? I think the same code is used `pkg/osbuild/skopeo_index_source.go` ?

mvo5 (commented):
Thank you for this PR! I gave it a read mostly of curiosity to learn a bit more about this area and put in some suggestions/ideas/questions. Looks nice and nothing is a blocker but maybe some of the comments are helpful :)

pkg/container/client.go
```diff
@@ -340,9 +344,35 @@ func (m RawManifest) Digest() (digest.Digest, error) {
 	return manifest.Digest(m.Data)
 }
 
+func getImageRef(target reference.Named, transport string, storagePath string) (types.ImageReference, error) {
+	switch transport {
+	case "":
+		// use docker as default transport
+		fallthrough
+	case dockerTransport:
```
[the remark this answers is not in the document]
kingsleyzissou: TIL 😄

pkg/distro/rhel7/images.go
```diff
@@ -159,6 +159,10 @@ func osCustomizations(
 		panic(fmt.Sprintf("failed to convert file customizations to fs node files: %v", err))
 	}
 
+	// if containerStorage := c.GetContainerStorage(); containerStorage != nil {
```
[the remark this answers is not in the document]
kingsleyzissou: Thanks, I missed this

pkg/distro/fedora/images.go
```diff
@@ -138,6 +138,19 @@ func osCustomizations(
...
+	// OSTree commits do not include data in `/var` since that is tied to the
+	// deployment, rather than the commit. Therefore the containers need to be
+	// stored in a different location, like `/usr/share`, and the container
+	// storage engine configured accordingly.
+	if t.rpmOstree && len(containers) > 0 {
+		storagePath := "/usr/share/containers/storage"
+		osc.ContainersStorage = &storagePath
+	}
+
+	if containerStorage := c.GetContainerStorage(); containerStorage != nil {
```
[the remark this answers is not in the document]
kingsleyzissou: No, good catch 😃

test/scripts/configure-generators
```diff
@@ -31,8 +31,7 @@ generate-build-config-{distro}-{arch}:
     INTERNAL_NETWORK: "true"
   script:
     - sudo ./test/scripts/setup-osbuild-repo
-    - sudo dnf -y install go python3 gpgme-devel s3cmd
-      osbuild osbuild-luks2 osbuild-lvm2 osbuild-ostree osbuild-selinux osbuild-depsolve-dnf
+    - sudo ./test/scripts/install-dependencies
```
kingsleyzissou, line 34: Not sure if this is a bit of overkill here because it is going to install all the other dependencies too. Or should I split this out a bit further and have a flag to install all and a flag for just this subset?

achilleas-k: It's a bit of extra work every time we run CI but it's not terrible.  It's fine.

PR #313, opened 2023-12-12
Rm/dnf json

achilleas-k:
Remove the dnf-json script from the repository.  Instead, the library will depend on osbuild-depsolve-dnf, an osbuild subpackage, for depsolving.

Add `findDepsolveDnf()` to the `dnfjson` package and call it when initialising the solver.  This used to be a helper function in some of the cmds and would look for dnf-json in the current directory for running it from sources then fall back to the installed locations.  Now that it's no longer part of the repository, it only searches for the installed locations, `/usr/libexec` for the official Fedora, CentOS, and RHEL paths, and `/usr/lib/osbuild` for other distros that might not user `/usr/libexec`.


ochosi (approved):
LGTM

supakeen (approved):

PR #318, opened 2023-12-13
test: compare osbuild version for image builds and don't specify boot mode

achilleas-k:
This PR was originally written to test builds for osbuild/osbuild#1493.
Changes:
- Update osbuild commit ID to current main (includes https://github.com/osbuild/osbuild/pull/1493).
- Use the osbuild version and commit ID (in addition to the manifest ID) when deciding if an image needs to be rebuilt.
- Don't specify a boot mode when boot testing in AWS.  We might specify it in the future, knowing which mode each image supports, but for now let's leave it off.  See also #323.


achilleas-k:
Embarrassing typo fixed 😳

dustymabe:
There were three test failures from Shutzbot, but I'm not sure I understand  the failures. Looks like they built the disk images fine and uploaded them to AWS and launched instances, but something after instance launch failed?

Screenshot 2023-12-13 at 23-11-58 Pipeline · Red Hat _ Red Hat Services _ Products _ Image Builder _ ci _ images CI · GitLab

achilleas-k:
image

achilleas-k:
I think the thing the failures have in common is that they're all BIOS only.  I think the issue is that when we boot them we don't specify so maybe they're defaulting to UEFI in EC2.  But that doesn't explain why they succeeded in the past, unless something changed with our uploader or the AWS library defaults.

achilleas-k:
Oops, there it is: 15e9d28e1596c8d1e398b9e27dc1d984d1a9e5fc

achilleas-k:
I'm going to clean this up (remove the osbuild PR pin) and mark it ready since I found a few issues while setting it up.

test/scripts/imgtestlib.py
```diff
@@ -254,3 +291,32 @@ def clargs():
...
+def read_osrelease():
+    """Read Operating System Information from `os-release`
+
+    This creates a dictionary with information describing the running operating system. It reads the information from
+    the path array provided as `paths`.  The first available file takes precedence. It must be formatted according to
+    the rules in `os-release(5)`.
+    """
+    osrelease = {}
+
+    with open(OS_RELEASE_FILE, encoding="utf8") as orf:
```
bcl, line 305: Another way to do this (possibly overkill?) https://github.com/weldr/lorax/blob/master/src/pylorax/sysutils.py#L115

bcl (approved):
Looks good to me.

test/scripts/imgtestlib.py
achilleas-k: Maybe a little overkill yeah.
Or we can just copy the example from [the manpage](https://www.freedesktop.org/software/systemd/man/latest/os-release.html) :smile:

thozza (approved):
LGTM, thanks!

PR #321, opened 2023-12-14
dnfjson: skip TestDepsolver() if no osbuild-depsolve-dnf is installed

mvo5:
Do not fail if `osbuild-depsolve-dnf` is missing.

It also fixes `dnfjson.go` so that the comment (and I think intention) matches the code :)

### DIFF
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -66,7 +66,6 @@ func findDepsolveDnf() string {
 func NewBaseSolver(cacheDir string) *BaseSolver {
 	return &BaseSolver{
 		cache:       newRPMCache(cacheDir, 1024*1024*1024), // 1 GiB
-		dnfJsonCmd:  []string{findDepsolveDnf()},
 		resultCache: NewDNFCache(60 * time.Second),
 	}
 }
@@ -636,6 +635,9 @@ func ParseError(data []byte) Error {
 }
 
 func run(dnfJsonCmd []string, req *Request) ([]byte, error) {
+	if len(dnfJsonCmd) == 0 {
+		dnfJsonCmd = []string{findDepsolveDnf()}
+	}
 	if len(dnfJsonCmd) == 0 {
 		return nil, fmt.Errorf("osbuild-depsolve-dnf command undefined")
 	}
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -57,7 +57,7 @@ func findDepsolveDnf() string {
 
 	// if it's not found, return empty string; the run() function will fail if
 	// it's used before setting.
-	return locations[0]
+	return ""
 }
 
 // Create a new unconfigured BaseSolver (without platform information). It can
--- a/pkg/dnfjson/dnfjson_test.go
+++ b/pkg/dnfjson/dnfjson_test.go
@@ -4,8 +4,6 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	"os"
-	"os/exec"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


bcl (approved):
Looks good to me. The tests run with `-force-dnf` so we won't accidentally pass if it's missing :)

achilleas-k:
Maybe we should remove the `python3 -c "import dnf"` check now.  The `osbuild-depsolve-dnf` package should depend on libdnf so all we need to do is check if that exists.  The shelling out to python thing is a bit silly.

> It also fixes `dnfjson.go` so that the comment (and I think intention) matches the code :)

aaah yeah the comment matches my initial idea and then I thought it might be better to set the the default so that, when (if) it fails, you get an error that shows the primary default location.  I'm not sure which is better.
We can't fail in the initialiser though because for one, we should be able to initialise a solver without using it and also it should be possible to create a solver and then change the path externally.

mvo5:
> Maybe we should remove the `python3 -c "import dnf"` check now. The `osbuild-depsolve-dnf` package should depend on libdnf so all we need to do is check if that exists. The shelling out to python thing is a bit silly.

Thanks, I think that makes sense, I updated the PR.

> > It also fixes `dnfjson.go` so that the comment (and I think intention) matches the code :)
> 
> aaah yeah the comment matches my initial idea and then I thought it might be better to set the the default so that, when (if) it fails, you get an error that shows the primary default location. I'm not sure which is better.

I have no strong opinion either way, I pushed a small commit that makes `New{,Base}Solver()` return an error when no `osbuild-depsolve-dnf` is found with a descriptive error message. But it has drawbacks in the test (as more are skipped now) so I can just drop this commit again if desired (it feels like the cleanest option though).

achilleas-k:
> I pushed a small commit that makes `New{,Base}Solver()` return an error when no `osbuild-depsolve-dnf` is found with a descriptive error message.

We don't want this.  It makes it impossible (or harder) to set up the solver externally with a depsolver in a non-standard location.  For example, this should be possible:
```go
solver := NewBaseSolver()
solver.SetDNFJSONPath("/some/odd/path/to/solver")
```
without having `osbuild-depsolve-dnf` installed, either for testing or for non-standard setups.

achilleas-k (changes requested):
See https://github.com/osbuild/images/pull/321#issuecomment-1878617616

mvo5:
> > I pushed a small commit that makes `New{,Base}Solver()` return an error when no `osbuild-depsolve-dnf` is found with a descriptive error message.
> 
> We don't want this. It makes it impossible (or harder) to set up the solver externally with a depsolver in a non-standard location. For example, this should be possible:
> 
> ```go
> solver := NewBaseSolver(cacheDir)
> solver.SetDNFJSONPath("/some/odd/path/to/solver")
> ```
[..]

Thank you, that is an interesting use-case and I have no considered it. I tweaked the PR slightly to run the findDepsolverDnf later. This seems to be the most minimal change and I pushed it. *But* below another idea if you prefer to detect the missing dnfjson early :)

If we don't want this and make any detection of missing dnfjson part of the constructor I wonder if it would  make sense to change the signature of `NewBaseSolver()` to make sure that the solver path is passed in when constructing?  Something like:
```go
diff --git a/pkg/dnfjson/dnfjson.go b/pkg/dnfjson/dnfjson.go
index 9505e09b3..d11ce164a 100644
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -61,13 +61,24 @@ func findDepsolveDnf() (string, error) {
        return "", fmt.Errorf("cannot find required osbuild-depsolve-dnf in any of %q", depSolveDnfPaths)
 }
 
+type SolverOptions struct {
+       DNFJsonPath string
+}
+
 // Create a new unconfigured BaseSolver (without platform information). It can
 // be used to create configured Solver instances with the NewWithConfig()
 // method.
-func NewBaseSolver(cacheDir string) (*BaseSolver, error) {
-       depsolveDnf, err := findDepsolveDnf()
-       if err != nil {
-               return nil, err
+func NewBaseSolver(cacheDir string, opts *SolverOptions) (*BaseSolver, error) {
+       if opts == nil {
+               opts = &SolverOptions{}
+       }
+       depsolveDnf := opts.DNFJsonPath
+       if depsolveDnf == "" {
+               foundSolver, err := findDepsolveDnf()
+               if err != nil {
+                       return nil, err
+               }
+               depsolveDnf = foundSolver
        }
        return &BaseSolver{
                cache:       newRPMCache(cacheDir, 1024*1024*1024), // 1 GiB
@@ -135,7 +146,7 @@ type Solver struct {
 
 // Create a new Solver with the given configuration. Initialising a Solver also loads system subscription information.
 func NewSolver(modulePlatformID, releaseVer, arch, distro, cacheDir string) (*Solver, error) {
-       s, err := NewBaseSolver(cacheDir)
+       s, err := NewBaseSolver(cacheDir, nil)
        if err != nil {
                return nil, err
        }
```
or is an API break like this too annoying?

achilleas-k (approved):
Current state LGTM

PR #315, opened 2023-12-13
stage(skopeo): add oci

supakeen:

PR #322, opened 2023-12-14
anaconda: container installer

supakeen:
Based on: #315

This introduces a new image that installs a bootc bootable container with Anaconda.

### DIFF
--- a/pkg/manifest/anaconda_installer.go
+++ b/pkg/manifest/anaconda_installer.go
@@ -72,6 +72,9 @@ type AnacondaInstaller struct {
 	AdditionalDrivers       []string
 
 	Files []*fsnode.File
+
+	// Temporary
+	UseRHELLoraxTemplates bool
 }
 
 func NewAnacondaInstaller(m *Manifest,
@@ -135,8 +138,18 @@ func (p *AnacondaInstaller) getBuildPackages(Distro) []string {
 	packages := p.anacondaBootPackageSet()
 	packages = append(packages,
 		"rpm",
-		"lorax-templates-generic",
 	)
+
+	if p.UseRHELLoraxTemplates {
+		packages = append(packages,
+			"lorax-templates-rhel",
+		)
+	} else {
+		packages = append(packages,
+			"lorax-templates-generic",
+		)
+	}
+
 	return packages
 }
 
@@ -269,9 +282,17 @@ func (p *AnacondaInstaller) serialize() osbuild.Pipeline {
 	}
 
 	if p.Type == AnacondaInstallerTypePayload {
+		var LoraxPath string
+
+		if p.UseRHELLoraxTemplates {
+			LoraxPath = "80-rhel/runtime-postinstall.tmpl"
+		} else {
+			LoraxPath = "99-generic/runtime-postinstall.tmpl"
+		}
+
 		pipeline.AddStage(osbuild.NewAnacondaStage(osbuild.NewAnacondaStageOptions(p.AdditionalAnacondaModules)))
 		pipeline.AddStage(osbuild.NewLoraxScriptStage(&osbuild.LoraxScriptStageOptions{
-			Path:     "99-generic/runtime-postinstall.tmpl",
+			Path:     LoraxPath,
 			BaseArch: p.platform.GetArch().String(),
 		}))
 	}
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -363,18 +363,39 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 			images,
 			nil))
 
+		// do what we can in our kickstart stage
+		kickstartOptions, err := osbuild.NewKickstartStageOptionsWithOSTreeContainer(
+			"/osbuild-base.ks",
+			p.Users,
+			p.Groups,
+			path.Join("/run/install/repo", p.PayloadPath),
+			"oci",
+			"",
+			"")
+
+		if err != nil {
+			panic("failed to create kickstartstage options")
+		}
+
+		pipeline.AddStage(osbuild.NewKickstartStage(kickstartOptions))
+
+		// and what we can't do in a separate kickstart that we include
+
 		kickstartFile, err := fsnode.NewFile(p.KSPath, nil, nil, nil, []byte(`
-ostreecontainer --url=/run/install/repo/container --transport=oci --no-signature-verification
+%include /run/install/repo/osbuild-base.ks
+
 rootpw --lock
-user --name fedora --groups wheel
+
 lang en_US.UTF-8
 keyboard us
 timezone UTC
+
 clearpart --all
 part /boot/efi --fstype=efi --size=512 --fsoptions="umask=0077"
 part /boot --fstype=ext2 --size=1024 --label=boot
 part swap --fstype=swap --size=1024
 part / --fstype=ext4 --grow
+
 reboot --eject
 `))
 
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -5,6 +5,7 @@ import (
 	"path"
 
 	"github.com/osbuild/images/pkg/container"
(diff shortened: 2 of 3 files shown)


pkg/manifest/anaconda_installer.go
```diff
@@ -135,7 +135,7 @@ func (p *AnacondaInstaller) getBuildPackages(Distro) []string {
 	packages := p.anacondaBootPackageSet()
 	packages = append(packages,
 		"rpm",
-		"lorax-templates-generic",
+		"lorax-templates-rhel", // XXX maybe always use -rhel?
```
supakeen, line 138: So, `-rhel` exists in Fedora; it seems to work. What do we do? At this point in the code we can't really switch on outside inputs.

bcl: It really needs to know if it's building for RHEL or not and switch which one it uses.

pkg/manifest/anaconda_installer.go
```diff
@@ -135,8 +138,20 @@ func (p *AnacondaInstaller) getBuildPackages(Distro) []string {
...
+	// - Explicit
+	if p.UseRHELLoraxTemplates {
+		packages = append(packages,
+			"lorax-templates-rhel",
+		)
+	} else {
+		packages = append(packages,
+			"lorax-templates-generic",
+		)
+	}
```
achilleas-k, line 153: I think t his is fine as a pipeline option set from outside.  The distro sets the option on the image kind, and the image function passes it down to the appropriate pipeline(s).  It's how we do all our distro-level switches.

supakeen: Right, but in this case we have no distro/image function (yet).

pkg/manifest/anaconda_installer.go
supakeen: This is now added as a property on the image.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -37,17 +37,20 @@ type AnacondaInstallerISOTree struct {
...
 	isoLabel string
 
 	SquashfsCompression string
 
-	OSPipeline         *OS
-	OSTreeCommitSource *ostree.SourceSpec
+	OSPipeline *OS
 
-	ostreeCommitSpec *ostree.CommitSpec
+	CommitSource *ostree.SourceSpec
```
achilleas-k, line 50: Why the rename?  Not really that important, but I think the ostree prefix is useful here.  A "commit" doesn't really mean much on its own and might be misunderstood out of context.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -74,69 +77,61 @@ func NewAnacondaInstallerISOTree(buildPipeline *Build, anacondaPipeline *Anacond
...
 }
 
-func (p *AnacondaInstallerISOTree) serializeStart(_ []rpmmd.PackageSpec, _ []container.Spec, commits []ostree.CommitSpec) {
-	if len(commits) == 0 {
-		// nothing to do
-		return
-	}
-
-	if len(commits) > 1 {
-		panic("pipeline supports at most one ostree commit")
```
achilleas-k, line 117: Where did this check go?   We should also be making the same check for containers, though that might change in the future if we also support bundling application containers into the install process.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -74,69 +77,61 @@ func NewAnacondaInstallerISOTree(buildPipeline *Build, anacondaPipeline *Anacond
...
+func (p *AnacondaInstallerISOTree) serializeStart(_ []rpmmd.PackageSpec, containers []container.Spec, commits []ostree.CommitSpec) {
+	if p.commitSpec != nil || p.containerSpec != nil {
+		panic("double call to serializeStart()")
 	}
-
-	p.ostreeCommitSpec = &commits[0]
```
achilleas-k, line 120: Why was this removed and why isn't the container source set as well (when it's defined)?  Does the pipeline work without setting the specs?

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -74,69 +77,61 @@ func NewAnacondaInstallerISOTree(buildPipeline *Build, anacondaPipeline *Anacond
...
-	if p.anacondaPipeline.Type == AnacondaInstallerTypePayload {
-		if p.ostreeCommitSpec == nil && p.OSPipeline == nil {
-			panic("missing ostree or ospipeline parameters in ISO tree pipeline")
-		}
-
-		// But not both payloads
-		if p.ostreeCommitSpec != nil && p.OSPipeline != nil {
-			panic("got both ostree and ospipeline parameters in ISO tree pipeline")
-		}
-	}
```
achilleas-k, line 138: Don't we still need these checks but with the container payload included?

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -287,6 +282,40 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 		pipeline.AddStage(osbuild.NewKickstartStage(kickstartOptions))
 	}
 
+	if p.containerSpec != nil {
+		images := osbuild.NewContainersInputForSources([]container.Spec{*p.containerSpec})
+		manifests := osbuild.NewFilesInputForManifestLists([]container.Spec{*p.containerSpec})
```
achilleas-k, line 309: We probably don't need the manifest list in this case.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -300,20 +322,28 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
...
+rootpw --lock
+user --name fedora --groups wheel
+lang en_US.UTF-8
+keyboard us
+timezone UTC
+clearpart --all
+part /boot/efi --fstype=efi --size=512 --fsoptions="umask=0077"
+part /boot --fstype=ext2 --size=1024 --label=boot
+part swap --fstype=swap --size=1024
+part / --fstype=ext4 --grow
```
achilleas-k, line 356: The PR mentioned in the commit message has been merged now.  Do we have time to do this properly or should we keep this for a bit?

achilleas-k (changes requested):

pkg/manifest/anaconda_installer_iso_tree.go
achilleas-k: I quickly generated the manifests and yeah the ostree commit pipelines for edge installers were dropped.
You can also see that the ISOs were all rebuilt in this PR [1], which means the manifests changed.  None of the existing images should be changing here.

[1] https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/5794994872

achilleas-k: You can very quickly inspect the manifest changes without resolving content (it generates fake content IDs) to compare:
```
go run ./cmd/gen-manifests -output manifests -packages=false -commits=false -containers=false
```

pkg/manifest/anaconda_installer_iso_tree.go
supakeen: Agreed, I think I was making it consistent with `container*`.

pkg/manifest/anaconda_installer_iso_tree.go
supakeen: Check is back, it seems I eradicated *all* checks at some point and it stayed in :(

pkg/manifest/anaconda_installer_iso_tree.go
supakeen: Check is back, it seems I eradicated all checks at some point and it stayed in :(

pkg/manifest/anaconda_installer_iso_tree.go
supakeen: We decided to do that bit in a followup.

achilleas-k (commented):
Looks good.  Tiny comment about the commit history though.

achilleas-k (approved):
Looks great.  Thanks!

PR #325, opened 2023-12-15
pkg: make internal/cloud public

ondrejbudai:
So we can share the code with osbuild/bootc-image-builder.

In order to clean up https://github.com/osbuild/bootc-image-builder/pull/59


ochosi (approved):
LGTM!

PR #332, opened 2023-12-20
Use dot to separate distro major and minor version and replace distro registry with factory

thozza:
This is a follow-up to https://github.com/osbuild/images/pull/104
Related to: https://issues.redhat.com/browse/COMPOSER-1962

Main changes:

- Replace `DistroRegistry` with a `DistroFactory`. The registry had a fixed list of tested and supported distros, while the factory can generate a distro for any (even untested) version of a supported distro. **This will allow building images for new RHEL and Fedora versions, without the need to make any code change.**
- Distros are constructed from a "distro ID string" which corresponds to that we used as distro names (e.g. `rhel-93` or `fedora-34`). Each distro is now responsible for parsing the distro ID string which corresponds to it. This allows one to define compatibility layers on per-distro basis. a new `distroidparser.Parser` is added as the single entry point for parsing any distro ID string with all existing distro-specific parsers.
- Distributions which use minor release versions (RHEL), now use dot `.` to separate the major and minor release version. This includes their `Name` as well as distro ID strings that can be used to construct a distro instance. A backward compatibility for RHEL-8 and RHEL-9 distro ID strings without a dot is part of the implementation.
- There are no longer a distro instances for RHEL distributions without a minor release version, so no `rhel-7`, `rhel-8` or `rhel-9`. A RHEL distro must always use a minor release version. These minor-version-less distros will be defined as an alias on higher-level (osbuild-composer).
- `Reporegistry` now also supports a dot as the separator for release major and minor version in the repo config filename, when loading repository definitions. It uses the `distroidparser.Parser` to parse repo config filename and get a standardized representation of the distro ID string. Meaning that `rhel-88.json` and `rhel-8.8.json` both represent repo definitions for RHEL-8.8.
- Extend the distrofactory to support specifying distro name aliases. This will allow any users of the package (osbuild-composer) to support defining any distro name aliases for supported distros.


thozza:
I want to integrate this first into the `osbuild-composer` repo, but otherwise this is ready for initial review...

achilleas-k:
Commit 6f9ce5af313961d5e403f548b0e6f6118179d923 has a `fixup` line in the message.

pkg/distroidparser/idparser.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type ParserFunc func(idStr string) (*distro.ID, error)
+
+// Parser is a list of distro-specific idStr parsers.
+type Parser struct {
+	parsers []ParserFunc
+}
+
+func NewParser(parsers ...ParserFunc) *Parser {
```
achilleas-k, line 20: Would it make sense to rename this to plain `New()`?
```
parser := distroidparser.New(...)
```

pkg/distro/fedora/distro.go
```diff
@@ -902,17 +902,26 @@ func newDistro(version int) distro.Distro {
 	return &rd
 }
 
-func DistroFactory(idStr string) distro.Distro {
+func DistroIDParser(idStr string) (*distro.ID, error) {
```
achilleas-k, line 891: I was thinking we could rename this to `IDParser()` so it gets called as:
```
id, err := distro.IDParser(idstr)
```
but since it's a function that parses, let's make it a verb phrase:
```
id, err := distro.ParseID(idstr)
```

achilleas-k (commented):
This is great.  Amazing work!!

I had to nitpick a couple of names of course.

achilleas-k:
You can blame @ondrejbudai for the merge conflicts there :sweat_smile:

ondrejbudai:
:see_no_evil: Should be trivial to rebase, though.

achilleas-k:
I was looking into why so many images were rebuilt [1] and it looks like the gpg keys aren't in the rpm options any more.

Here's a single diff:
```diff
diff --git a/fedora_39-x86_64-qcow2-empty.json b/fedora_39-x86_64-qcow2-empty.json
index 190b5fe2..8576da1f 100644
--- a/fedora_39-x86_64-qcow2-empty.json
+++ b/fedora_39-x86_64-qcow2-empty.json
@@ -9,9 +9,6 @@
         "baseurls": [
           "https://rpmrepo.osbuild.org/v2/mirror/public/f39/f39-x86_64-branched-20231101"
         ],
-        "gpgkeys": [
-          "-----BEGIN PGP PUBLIC KEY BLOCK-----\n\nmQINBGLykg8BEADURjKtgQpQNoluifXia+U3FuqGCTQ1w7iTqx1UvNhLX6tb9Qjy\nl/vjl1iXxucrd2JBnrT/21BdtaABhu2hPy7bpcGEkG8MDinAMZBzcyzHcS/JiGHZ\nd/YmMWQUgbDlApbxFSGWiXMgT0Js5QdcywHI5oiCmV0lkZ+khZ4PkVWmk6uZgYWf\nJOG5wp5TDPnoYXlA4CLb6hu2691aDm9b99XYqEjhbeIzS9bFQrdrQzRMKyzLr8NW\ns8Pq2tgyzu8txlWdBXJyAMKldTPstqtygLL9UUdo7CIQQzWqeDbAnv+WdOmiI/hR\netbbwNV+thkLJz0WD90C2L3JEeUJX5Qa4oPvfNLDeCKmJFEFUTCEdm0AYoQDjLJQ\n3d3q9M09thXO/jYM0cSnJDclssLNsNWfjJAerLadLwNnYRuralw7f74QSLYdJAJU\nSFShBlctWKnlhQ7ehockqtgXtWckkqPZZjGiMXwHde9b9Yyi+VqtUQWxSWny+9g9\n6tcoa3AdnmpqSTHQxYajD0EGXJ0z0NXfqxkI0lo8UxzypEBy4sARZ4XhTU73Zwk0\nLGhEUHlfyxXgRs6RRvM2UIoo+gou2M9rn/RWkhuHJNSfgrM0BmIBCjhjwGiS33Qh\nysLDWJMdch8lsu1fTmLEFQrOB93oieOJQ0Ysi5gQY8TOT+oZvVi9pSMJuwARAQAB\ntDFGZWRvcmEgKDM5KSA8ZmVkb3JhLTM5LXByaW1hcnlAZmVkb3JhcHJvamVjdC5v\ncmc+iQJOBBMBCAA4FiEE6PI5lvIyGGQMtEy+dc9axBi450wFAmLykg8CGw8FCwkI\nBwIGFQoJCAsCBBYCAwECHgECF4AACgkQdc9axBi450yd4w//ZtghbZX5KFstOdBS\nrcbBfCK9zmRvzeejzGl6lPKfqwx7OOHYxFlRa9MYLl8QG7Aq6yRRWzzEHiSb0wJw\nWXz5tbkAmV/fpS4wnb3FDArD44u317UAnaU+UlhgK1g62lwI2dGpvTSvohMBMeBY\nB5aBd+sLi3UtiSRM2XhxvxaWwr/oFLjKDukgrPQzeV3F/XdxGhSz/GZUVFVprcrB\nh/dIo4k0Za7YVRhlVM0coOIcKbcjxAK9CCZ8+jtdIh3/BN5zJ0RFMgqSsrWYWeft\nBI3KWLbyMfRwEtp7xSi17WXbRfsSoqwIVgP+RCSaAdVuiYs/GCRsT3ydYcDvutuJ\nYZoE53yczemM/1HZZFI04zI7KBsKm9NFH0o4K2nBWuowBm59iFvWHFpX6em54cq4\n45NwY01FkSQUqntfqCWFSowwFHAZM4gblOikq2B5zHoIntCiJlPGuaJiVSw9ZpEc\n+IEQfmXJjKGSkMbU9tmNfLR9skVQJizMTtoUQ12DWC+14anxnnR2hxnhUDAabV6y\nJ5dGeb/ArmxQj3IMrajdNwjuk9GMeMSSS2EMY8ryOuYwRbFhBOLhGAnmM5OOSUxv\nA4ipWraXDW0bK/wXI7yHMkc6WYrdV3SIXEqJBTp7npimv3JC+exWEbTLcgvV70FP\nX55M9nDtzUSayJuEcfFP2c9KQCE=\n=J4qZ\n-----END PGP PUBLIC KEY BLOCK-----"
-        ],
         "check_gpg": true,
         "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/public/f39/f39-x86_64-branched-20231101"
       }
@@ -1530,11 +1527,7 @@
                 ]
               }
             },
-            "options": {
-              "gpgkeys": [
-                "-----BEGIN PGP PUBLIC KEY BLOCK-----\n\nmQINBGLykg8BEADURjKtgQpQNoluifXia+U3FuqGCTQ1w7iTqx1UvNhLX6tb9Qjy\nl/vjl1iXxucrd2JBnrT/21BdtaABhu2hPy7bpcGEkG8MDinAMZBzcyzHcS/JiGHZ\nd/YmMWQUgbDlApbxFSGWiXMgT0Js5QdcywHI5oiCmV0lkZ+khZ4PkVWmk6uZgYWf\nJOG5wp5TDPnoYXlA4CLb6hu2691aDm9b99XYqEjhbeIzS9bFQrdrQzRMKyzLr8NW\ns8Pq2tgyzu8txlWdBXJyAMKldTPstqtygLL9UUdo7CIQQzWqeDbAnv+WdOmiI/hR\netbbwNV+thkLJz0WD90C2L3JEeUJX5Qa4oPvfNLDeCKmJFEFUTCEdm0AYoQDjLJQ\n3d3q9M09thXO/jYM0cSnJDclssLNsNWfjJAerLadLwNnYRuralw7f74QSLYdJAJU\nSFShBlctWKnlhQ7ehockqtgXtWckkqPZZjGiMXwHde9b9Yyi+VqtUQWxSWny+9g9\n6tcoa3AdnmpqSTHQxYajD0EGXJ0z0NXfqxkI0lo8UxzypEBy4sARZ4XhTU73Zwk0\nLGhEUHlfyxXgRs6RRvM2UIoo+gou2M9rn/RWkhuHJNSfgrM0BmIBCjhjwGiS33Qh\nysLDWJMdch8lsu1fTmLEFQrOB93oieOJQ0Ysi5gQY8TOT+oZvVi9pSMJuwARAQAB\ntDFGZWRvcmEgKDM5KSA8ZmVkb3JhLTM5LXByaW1hcnlAZmVkb3JhcHJvamVjdC5v\ncmc+iQJOBBMBCAA4FiEE6PI5lvIyGGQMtEy+dc9axBi450wFAmLykg8CGw8FCwkI\nBwIGFQoJCAsCBBYCAwECHgECF4AACgkQdc9axBi450yd4w//ZtghbZX5KFstOdBS\nrcbBfCK9zmRvzeejzGl6lPKfqwx7OOHYxFlRa9MYLl8QG7Aq6yRRWzzEHiSb0wJw\nWXz5tbkAmV/fpS4wnb3FDArD44u317UAnaU+UlhgK1g62lwI2dGpvTSvohMBMeBY\nB5aBd+sLi3UtiSRM2XhxvxaWwr/oFLjKDukgrPQzeV3F/XdxGhSz/GZUVFVprcrB\nh/dIo4k0Za7YVRhlVM0coOIcKbcjxAK9CCZ8+jtdIh3/BN5zJ0RFMgqSsrWYWeft\nBI3KWLbyMfRwEtp7xSi17WXbRfsSoqwIVgP+RCSaAdVuiYs/GCRsT3ydYcDvutuJ\nYZoE53yczemM/1HZZFI04zI7KBsKm9NFH0o4K2nBWuowBm59iFvWHFpX6em54cq4\n45NwY01FkSQUqntfqCWFSowwFHAZM4gblOikq2B5zHoIntCiJlPGuaJiVSw9ZpEc\n+IEQfmXJjKGSkMbU9tmNfLR9skVQJizMTtoUQ12DWC+14anxnnR2hxnhUDAabV6y\nJ5dGeb/ArmxQj3IMrajdNwjuk9GMeMSSS2EMY8ryOuYwRbFhBOLhGAnmM5OOSUxv\nA4ipWraXDW0bK/wXI7yHMkc6WYrdV3SIXEqJBTp7npimv3JC+exWEbTLcgvV70FP\nX55M9nDtzUSayJuEcfFP2c9KQCE=\n=J4qZ\n-----END PGP PUBLIC KEY BLOCK-----"
-              ]
-            }
+            "options": {}
           },
           {
             "type": "org.osbuild.selinux",
@@ -5240,11 +5233,7 @@
                 ]
               }
             },
-            "options": {
-              "gpgkeys": [
-                "-----BEGIN PGP PUBLIC KEY BLOCK-----\n\nmQINBGLykg8BEADURjKtgQpQNoluifXia+U3FuqGCTQ1w7iTqx1UvNhLX6tb9Qjy\nl/vjl1iXxucrd2JBnrT/21BdtaABhu2hPy7bpcGEkG8MDinAMZBzcyzHcS/JiGHZ\nd/YmMWQUgbDlApbxFSGWiXMgT0Js5QdcywHI5oiCmV0lkZ+khZ4PkVWmk6uZgYWf\nJOG5wp5TDPnoYXlA4CLb6hu2691aDm9b99XYqEjhbeIzS9bFQrdrQzRMKyzLr8NW\ns8Pq2tgyzu8txlWdBXJyAMKldTPstqtygLL9UUdo7CIQQzWqeDbAnv+WdOmiI/hR\netbbwNV+thkLJz0WD90C2L3JEeUJX5Qa4oPvfNLDeCKmJFEFUTCEdm0AYoQDjLJQ\n3d3q9M09thXO/jYM0cSnJDclssLNsNWfjJAerLadLwNnYRuralw7f74QSLYdJAJU\nSFShBlctWKnlhQ7ehockqtgXtWckkqPZZjGiMXwHde9b9Yyi+VqtUQWxSWny+9g9\n6tcoa3AdnmpqSTHQxYajD0EGXJ0z0NXfqxkI0lo8UxzypEBy4sARZ4XhTU73Zwk0\nLGhEUHlfyxXgRs6RRvM2UIoo+gou2M9rn/RWkhuHJNSfgrM0BmIBCjhjwGiS33Qh\nysLDWJMdch8lsu1fTmLEFQrOB93oieOJQ0Ysi5gQY8TOT+oZvVi9pSMJuwARAQAB\ntDFGZWRvcmEgKDM5KSA8ZmVkb3JhLTM5LXByaW1hcnlAZmVkb3JhcHJvamVjdC5v\ncmc+iQJOBBMBCAA4FiEE6PI5lvIyGGQMtEy+dc9axBi450wFAmLykg8CGw8FCwkI\nBwIGFQoJCAsCBBYCAwECHgECF4AACgkQdc9axBi450yd4w//ZtghbZX5KFstOdBS\nrcbBfCK9zmRvzeejzGl6lPKfqwx7OOHYxFlRa9MYLl8QG7Aq6yRRWzzEHiSb0wJw\nWXz5tbkAmV/fpS4wnb3FDArD44u317UAnaU+UlhgK1g62lwI2dGpvTSvohMBMeBY\nB5aBd+sLi3UtiSRM2XhxvxaWwr/oFLjKDukgrPQzeV3F/XdxGhSz/GZUVFVprcrB\nh/dIo4k0Za7YVRhlVM0coOIcKbcjxAK9CCZ8+jtdIh3/BN5zJ0RFMgqSsrWYWeft\nBI3KWLbyMfRwEtp7xSi17WXbRfsSoqwIVgP+RCSaAdVuiYs/GCRsT3ydYcDvutuJ\nYZoE53yczemM/1HZZFI04zI7KBsKm9NFH0o4K2nBWuowBm59iFvWHFpX6em54cq4\n45NwY01FkSQUqntfqCWFSowwFHAZM4gblOikq2B5zHoIntCiJlPGuaJiVSw9ZpEc\n+IEQfmXJjKGSkMbU9tmNfLR9skVQJizMTtoUQ12DWC+14anxnnR2hxnhUDAabV6y\nJ5dGeb/ArmxQj3IMrajdNwjuk9GMeMSSS2EMY8ryOuYwRbFhBOLhGAnmM5OOSUxv\nA4ipWraXDW0bK/wXI7yHMkc6WYrdV3SIXEqJBTp7npimv3JC+exWEbTLcgvV70FP\nX55M9nDtzUSayJuEcfFP2c9KQCE=\n=J4qZ\n-----END PGP PUBLIC KEY BLOCK-----"
-              ]
-            }
+            "options": {}
           },
           {
             "type": "org.osbuild.fix-bls",
```

It's also strange that the builds succeeded when the packages have `check_gpg = true` but without any keys listed in the stage options :thinking: 

[1] https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/5804488140

achilleas-k:
Also weird: I see the keys in the build log: https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/5804589259

achilleas-k:
OH WAIT!  Of course everything will get rebuilt.  All the distro names have essentially changed so it can't match against old configs (this is okay, we like mass rebuilds from time to time :) )

achilleas-k:
Found the issue:
The private `rpmmd.repository` struct that's used to load the test repos only has `GPGKey`: https://github.com/thozza/osbuild-images/blob/0ca994ddcfd660829033bf2caa67692a01b5d687/pkg/rpmmd/repository.go#L20
And I recently switched the test repos to use a slice: https://github.com/osbuild/images/commit/81cc5e2d287d3c4ae1d073c3c672ece72a0d5cca

I think it'd be nice to support both (of course keep the old one for backwards compatibility).

```diff
diff --git a/pkg/rpmmd/repository.go b/pkg/rpmmd/repository.go
index 13158b40b..64591a2e5 100644
--- a/pkg/rpmmd/repository.go
+++ b/pkg/rpmmd/repository.go
@@ -18,6 +18,7 @@ type repository struct {
 	Metalink       string   `json:"metalink,omitempty"`
 	MirrorList     string   `json:"mirrorlist,omitempty"`
 	GPGKey         string   `json:"gpgkey,omitempty"`
+	GPGKeys        []string `json:"gpgkeys,omitempty"`
 	CheckGPG       bool     `json:"check_gpg,omitempty"`
 	IgnoreSSL      bool     `json:"ignore_ssl,omitempty"`
 	RHSM           bool     `json:"rhsm,omitempty"`
@@ -240,6 +241,9 @@ func LoadRepositoriesFromFile(filename string) (map[string][]RepoConfig, error)
 			if repo.GPGKey != "" {
 				keys = []string{repo.GPGKey}
 			}
+			if len(repo.GPGKeys) > 0 {
+				keys = append(keys, repo.GPGKeys...)
+			}
 			config := RepoConfig{
 				Name:           repo.Name,
 				BaseURLs:       urls,
```

The reason the build log shows keys is because `cmd/build/` still uses [its own loader into `rpmmd.RepoConfig`](https://github.com/thozza/osbuild-images/blob/0ca994ddcfd660829033bf2caa67692a01b5d687/cmd/build/main.go#L116-L138).  We should switch that like we did for gen-manifests too.

pkg/distro/fedora/distro.go
thozza: I like your suggestions. I was a bit struggling with the naming for the parser package and functions, so a fresh opinion is very welcomed :slightly_smiling_face:

thozza:
I implemented the following changes:

- Renamed distro ID string parser functions and the `disroidparser.Parser` constructor, as suggested.
- I extended `rpmmd.repository` to support `gpgkeys` array in the JSON representation.
- Modified `distrofactory.New()` to return pointer instead of a value, since it makes more sense and it turned out to be desired when integrating the change into `osbuild-composer`.
- Modified `cmd/build` to also use `reporegistry`, instead of a custom code to load repo configs.

pkg/distro/fedora/distro.go
thozza: Done.

pkg/distroidparser/idparser.go
thozza: Done

achilleas-k (dismissed):
Everything looks great.  Anything left for this to be marked as ready?

thozza:
> Everything looks great. Anything left for this to be marked as ready?

Well, I think that it would be for the best to have a PR in osbuild-composer ready, which would integrate the changes there, before we merge this. Because otherwise, we may end up having to make a small change in an image definition, but it would be impossible to make it quickly and re-deploy the service. Unfortunately that means more for for me to keep this PR up-to-date until the PR in composer is ready... :shrug:

achilleas-k:
Right.  Good thinking.  Let's see how quickly we can get the osbuild-composer side done then.

thozza:
Notes on what changed as a result of porting the work to osbuild-composer:

- `distro.ParseID()` has been modified to allow dashed in the distro name. Originally, this was not possible, but osbuild-composer used `centos-stream-X` for repo files, which filename we need to parse.
- `reporegistry.LoadAllRepositories()` has been modified to not fail in case the repo config filename can't be parsed by distro ID parser. The filename is used as is as the distro name if it can't be parsed. The goal is to prevent osbuild-composer from failing when loading repositories, since this was previously the case.

pkg/distrofactory/distrofactory.go
```diff
@@ -46,14 +65,48 @@ func (f *Factory) FromHost() distro.Distro {
 }
 
 // New returns a Factory of distro.Distro factories for the given distros.
-func New(factories ...FactoryFunc) *Factory {
-	return &Factory{factories: factories}
+// The provided aliases map has the following constraints:
+// - An alias must not mask an existing distro.
+// - An alias target must map to an existing distro.
+func New(aliases map[string]string, factories ...FactoryFunc) (*Factory, error) {
```
achilleas-k, line 71: TINIEST NITPICK: I think I would prefer if this functionality (registering aliases) was separate.  Having it as the first arg in the constructor makes it stand out too much and while we will definitely use it all the time in composer, I see it as a "secondary" functionality of the distrofactory.

What I'm thinking is basically the following usage pattern:
```go
df := distrofactory.New(fedora.DistroFactory, rhel7.DistroFactory, rhel8.DistroFactory)
err := df.RegisterAliases(aliases)
```

or also from the default level:
```go
df := distrofactory.NewDefault()
err := df.RegisterAliases(aliases)
```

I'm fine to keep it as is.  Change it only if you like it :)

pkg/reporegistry/reporegistry.go
```diff
@@ -38,26 +36,12 @@ func NewFromDistrosRepoConfigs(distrosRepoConfigs rpmmd.DistrosRepoConfigs) *Rep
...
-}
-
-// reposByImageTypeName returns a slice of rpmmd.RepoConfig instances, which should be used for building the specific
+// ReposByImageTypeName returns a slice of rpmmd.RepoConfig instances, which should be used for building the specific
 // image type name (of a given distribution and architecture). The method does not verify
 // if the given image type name is actually part of the architecture definition of the provided name.
 // Therefore in general, all common distro-arch-specific repositories are returned for any image type name,
 // even for non-existing ones.
-func (r *RepoRegistry) reposByImageTypeName(distro, arch, imageType string) ([]rpmmd.RepoConfig, error) {
+func (r *RepoRegistry) ReposByImageTypeName(distro, arch, imageType string) ([]rpmmd.RepoConfig, error) {
```
achilleas-k, line 44: Commit message issues

Subject reads:
```
reporegistry: export ReposByImageTypeName instead of ReposByImageTypeName
```
and it's 4 chars too long (which funnily enough, is the word `Name` which should be removed).

Also:
```
Also keep the ReposByImageType() is removed in favor of just ReposByImageTypeName().
```
I guess `keep` should be removed there.

pkg/reporegistry/repository.go
```diff
@@ -40,7 +40,11 @@ func LoadAllRepositories(confPaths []string) (rpmmd.DistrosRepoConfigs, error) {
 				// without a dot to separate major and minor release versions
 				distro, err := distroidparser.DefaultParser.Standardize(distroIDStr)
 				if err != nil {
-					return nil, fmt.Errorf("failed to parse distro ID string: %v", err)
+					log.Printf("failed to parse distro ID string, using it as is: %v", err)
+					// NB: Before the introduction of distro ID standardization, the filename
+					//     was used as the distro ID. This is kept for backward compatibility
+					//     if the filename can't be parsed.
+					distro = distroIDStr
```
achilleas-k, line 47: This is more of a question than a comment:
If I understand this change correctly, it means I can have a distro called `bob`, and a repo file called `bob.json`, and it would still work even if I don't register a parser for it?

achilleas-k (changes requested):
Wonderful work.  The osbuild-composer PR is also almost green, which is awesome.

I have:
- 1 nitpick.  Only address it if you think it's a good idea.
- 1 commit message that needs fixing (the actual changes I'm requesting).
- 1 question for clarify.

pkg/reporegistry/repository.go
thozza: Yes, exactly.

pkg/distrofactory/distrofactory.go
thozza: So, I'm not opposed to the suggestion. I'll have a closer look and consider making the change.

thozza: I made the change. I think that it is a bit nicer in all the places that do not set the alias and that one does not need to check the error...

pkg/reporegistry/reporegistry.go
thozza: Done.

achilleas-k (approved):
Amazing.  Thank you!

PR #334, opened 2023-12-21
distro/rhel9: add support for azure-sap-rhui

ondrejbudai:
We don't have a snapshot yet, so the test repository is just download.eng.bos.redhat.com for now. Let's fix it when we have a snapshot.

Regarding firewalld and libcanberra-gtk2:

The 1st party image has firewalld, but doesn't have libcanberra-gtk2. This is actually backwards that what is in out EC2 SAP images, so I had to do some shuffling to keep the behaviour in EC2, but have it different for Azure.


achilleas-k (dismissed):

ondrejbudai:
Forgot to update rhel-9 repos.

pkg/distro/rhel9/distro.go
```diff
@@ -460,6 +460,8 @@ func newDistro(name string, minor int) *distribution {
 		x86_64.addImageTypes(azureX64Platform, azureRhuiImgType, azureByosImgType)
 		aarch64.addImageTypes(azureAarch64Platform, azureRhuiImgType, azureByosImgType)
 
+		x86_64.addImageTypes(azureX64Platform, azureSapRhuiImgType(rd))
```
thozza, line 463: nitpick: you could have just extended the line 460

achilleas-k (approved):

thozza (approved):
LGTM

pkg/distro/rhel9/distro.go
thozza: :stuck_out_tongue_winking_eye:

PR #346, opened 2024-01-04
Simplify Pipeline constructors

achilleas-k:
This PR simplifies the Pipeline constructors (all the `manifest.New...` functions such as `NewOS()`, `NewXZ()` etc.) by removing the manifest argument.  Reasoning:

- All Pipeline implementations, excluding Build, require a build pipeline because they run in the build root.  This is not an osbuild restriction, but something we "enforce" or strongly incentivise in our image definitions.
- Any pipeline that uses a specific build pipeline must belong to the same manifest as the build pipeline.

BEFORE:
- Pipeline constructors would take a build pipeline argument and a manifest.  They would check that the build pipeline's manifest reference and the manifest argument were the same.
NOW:
- Pipeline constructors take a build pipeline argument and use the manifest reference from that argument to add themselves to the manifest

Sanity checks for manifest equivalence between the manifest's build pipeline and the manifest it is being added to are done in the `Manifest.addPipeline()` method.

See individual commit messages for more details.

---

This is part of a bigger plan to simplify the pipeline generation and configuration in general.  I still haven't figured everything out, but some of these smaller things are, I think, obvious improvements.  I'd like to get some of them merged as smaller PRs so there might be a few more like this coming.
For larger changes, like things that change the general manifest generation flow, I will discuss in advance before implementing.

### DIFF
--- a/cmd/osbuild-playground/my-container.go
+++ b/cmd/osbuild-playground/my-container.go
@@ -50,7 +50,7 @@ func (img *MyContainer) InstantiateManifest(m *manifest.Manifest,
 	build.Checkpoint()
 
 	// create a minimal non-bootable OS tree
-	os := manifest.NewOS(m, build, &platform.X86{}, repos)
+	os := manifest.NewOS(build, &platform.X86{}, repos)
 	os.ExtraBasePackages = []string{"@core"}
 	os.OSCustomizations.Language = "en_US.UTF-8"
 	os.OSCustomizations.Hostname = "my-host"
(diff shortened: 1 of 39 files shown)


supakeen (dismissed):
I like this.

thozza (approved):
LGTM, nice work!

achilleas-k:
Had to rebase and change the call to `manifest.NewAnacondaInstaller()` for the new `AnacondaContainerInstaller`.

thozza (approved):

PR #348, opened 2024-01-05
osbuild: add new ContainerDeployStage for org.osbuild.container-deploy

mvo5:
This commit adds support for the new `org.osbuild.container-deploy` stage (see https://github.com/osbuild/osbuild/pull/1509).

PR #349, opened 2024-01-05
osbuild: simplify containers_input.go

mvo5:
[only the second commit is new in this PR]

While writing tests to ensure the that the `ContainerDeployInputs`
generates the correct json I noticed that this was more involved
than I had hoped and it seems there are some indirections in the
code that may not be necessary. I removed them and IMHHO the code
is now a bit more direct and easier to read.

Feedback very welcome!

Build on top of https://github.com/osbuild/images/pull/348 (as I wanted the unit tests from it).

### DIFF
--- a/pkg/osbuild/container_deploy_stage.go
+++ b/pkg/osbuild/container_deploy_stage.go
@@ -12,7 +12,7 @@ func (inputs ContainerDeployInputs) validate() error {
 	if inputs.Images.References == nil {
 		return fmt.Errorf("stage requires exactly 1 input container (got nil References)")
 	}
-	if ncontainers := inputs.Images.References.Len(); ncontainers != 1 {
+	if ncontainers := len(inputs.Images.References); ncontainers != 1 {
 		return fmt.Errorf("stage requires exactly 1 input container (got %d)", ncontainers)
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


supakeen:
It's an interesting change. As far as I know it is idiomatic Go to custom type things like these and have an interface for if you ever want to put other stuff in there. However, that only needs to be done *when* we have other things.

achilleas-k:
There's a lot of stupid stuff in the osbuild package in this repo and most of it is from (sometimes, if not often, incorrect) mirroring of the osbuild schema and its flexibility.  In this case, the reference for containers input is an interface because inputs in osbuild *can* (but in the case of containers *don't*) have multiple schemas.  It's also possible for stages to define extra properties for their inputs, though I think we don't do this (though all inputs in osbuild have `additionalProperties` set to `True`.  For example, iirc, at one point in the development stage of the manifest list feature, the option to include the manifest list was an input option on the container input but only defined in the skopeo stage.

Anyway.  I agree with this change.  No need to complicate things unnecessarily.  Let's merge this.
We need more simplifications like this.  Maybe we can make a plan and do it all in one go instead of chipping away at it bit by bit.

achilleas-k (dismissed):
LGTM.  Let's add it to the queue after #348 is added.

achilleas-k (approved):

supakeen (approved):

PR #351, opened 2024-01-08
Add FIPS enabled image warning

mmartinv:
Show a warning message when building a FIPS enabled image in a non enabled FIPS host

### DIFF
--- a/pkg/distro/fedora/imagetype.go
+++ b/pkg/distro/fedora/imagetype.go
@@ -392,5 +392,10 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
 		return nil, err
 	}
 
+	if customizations.GetFIPS() && !common.IsBuildHostFIPSEnabled() {
+		w := fmt.Sprintln(common.FIPSEnabledImageWarning)
+		return []string{w}, nil
+	}
+
 	return nil, nil
 }
--- a/pkg/distro/fedora/imagetype.go
+++ b/pkg/distro/fedora/imagetype.go
@@ -280,7 +280,7 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
 	if t.name == "iot-raw-image" || t.name == "iot-qcow2-image" {
 		allowed := []string{"User", "Group", "Directories", "Files", "Services", "FIPS"}
 		if err := customizations.CheckAllowed(allowed...); err != nil {
-			return nil, fmt.Errorf("unsupported blueprint customizations found for image type %q: (allowed: %s)", t.name, strings.Join(allowed, ", "))
+			return nil, fmt.Errorf(distro.UnsupportedCustomizationError, t.name, strings.Join(allowed, ", "))
 		}
 		// TODO: consider additional checks, such as those in "edge-simplified-installer" in RHEL distros
 	}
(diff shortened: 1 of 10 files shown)


bcl:
I don't understand the reason for this -- if the resulting image is FIPS compliant, why print the warning? Also, long text strings like this may need to be translated, something which we don't support yet.

7flying:
> I don't understand the reason for this -- if the resulting image is FIPS compliant, why print the warning? Also, long text strings like this may need to be translated, something which we don't support yet.

This disclaimer is a recommendation that we got from the Crypto team so that the users know that if they're using our infra they shouldn't be generating keys (since composer doesn't run in FIPS enabled runners).

miabbott:
> This disclaimer is a recommendation that we got from the Crypto team so that the users know that if they're using our infra they shouldn't be generating keys (since composer doesn't run in FIPS enabled runners).

I believe the context for this warning comes from this comment in the associated Jira story - https://issues.redhat.com/browse/THEEDGE-3735?focusedId=23577973&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23577973

cc:  @mrguitar @neverpanic @evgenyz

evgenyz:
Yep. The UI/UX implementation could be different of course, as long as users understand the implications about generated cryptographic artifacts.

neverpanic:
Could we not make this warning dynamic, i.e., depend on whether the host is actually running in FIPS mode? That would reduce the noise in situations where somebody is actually running image builder on a FIPS-enabled host.

We can check `/proc/sys/crypto/fips_enabled` for this; if it doesn't contain `1`, the host isn't in FIPS mode.

neverpanic (approved):
lg2m now!

achilleas-k (dismissed):
LGTM.  I'll hold off on merging until @bcl approves since he commented.

7flying:
@mmartinv LGTM too, no more tests needed

pkg/distro/rhel9/distro_test.go
```diff
@@ -498,6 +501,43 @@ func TestDistro_ManifestError(t *testing.T) {
 	}
 }
 
+// Check that Manifest() function returns an warning when FIPS
+// customization is enabled and the host is not FIPS
+func TestDistro_ManifestFIPSWarning(t *testing.T) {
+	r8distro := rhel9.New()
```
bcl, line 507: copy pasta typo :)

Unrelated, but my brain sees these tests with almost exactly the same code and wishes it could be a function. But not really asking for that.

pkg/distro/fedora/imagetype.go
```diff
@@ -392,5 +392,10 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
 		return nil, err
 	}
 
+	if customizations.GetFIPS() && !common.IsBuildHostFIPSEnabled() {
+		w := fmt.Sprintln(common.FIPSEnabledImageWarning)
+		return []string{w}, nil
```
bcl, line 397: I think this also needs a log() like the RHEL versions. Unless there is a reason not to log on Fedora?

pkg/distro/rhel9/distro_test.go
mmartinv: @bcl  I moved the tests to `pkg/distro/distro_test.go` and simplified some things. I hope this is ok with you.

pkg/distro/fedora/imagetype.go
mmartinv: There's no log in the whole function so I kept it the same way.

bcl: I looked a bit deeper and I think no logging is ok for a warning. It's really meant to make it back to the user, who isn't going to read the logs. In most cases the warnings are currently ignored, but that can be fixed later. FWIW it does look like it makes it back up to the weldr API call and is returned as part of the compose start response.

bcl (approved):
Looks good, thanks!

PR #353, opened 2024-01-08
Fedora: Enable remote in iot-installer kickstart

paulwhalen:
Currently the Fedora IoT installer ISO's are not configured to use the upstream repository and instead use the OStree commit included on the installation ISO.  To use the intended upstream remote, users are currently required to run a few `ostree` commands to make the change.

This PR enables the `--remote` option in the iot-installer kickstart to use the repository listed in the `fedora-iot-config` package. 

Signed-off-by: Paul Whalen <pwhalen@fedoraproject.org>

### DIFF
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -23,6 +23,7 @@ type AnacondaInstallerISOTree struct {
 	// TODO: review optional and mandatory fields and their meaning
 	OSName  string
 	Release string
+	Remote  string
 	Users   []users.User
 	Groups  []users.Group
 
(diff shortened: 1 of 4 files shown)


pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -370,6 +372,7 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 			path.Join("/run/install/repo", p.PayloadPath),
 			"oci",
 			"",
+			"",
```
7flying, line 375: I think that the linter is right and this parameter should be deleted

7flying (approved):
LGTM!

achilleas-k (approved):
Nice and simple.  Thank you!

PR #354, opened 2024-01-08
manifest,images: add new manifest.NewBuildFromContainerSpec and use in new `BootcDiskImage` (HMS-3318)

mvo5:
This branch prepares for allowing to have build roots from container sources. This is a bit of a shortcut PR that is mainly tailored to `bootc-image-builder`. 

Alternatively we could change `InstantiateManifest()` to not take `repos []rpmmd.RepoConfig` but instead something like: 
```go
type RepoContainerStuff struct {
    Repos []rpmmd.RepoConfig
    Containers []container.SourceSpec
}

type ImageKind interface {
	Name() string
	InstantiateManifest(m *manifest.Manifest, repoContainer *RepoContainerStuff, runner runner.Runner, rng *rand.Rand) (*artifact.Artifact, error)
}
```
of course with a more sensible name. Happy to do this if that is the preferred way. I will move this out of draft (or close it) once I get feedback on this (or maybe there is a third option).

Once this is available, bib will use:
```go
img := image.NewBootcDiskImage(containerSource)
...
_, err = img.InstantiateManifestFromContainers(&mf, containerSources, runner, rng)
```

Note that this is not sufficient for a full container based buildroot, the following is missing:
1. get EFI binaries from an alternative location (e.g. via https://github.com/osbuild/osbuild/pull/1529 or bootupd)
2. create a second "bib-build" pipeline that uses the "bib" container as the buildroot and run "qemu-img" from there

### DIFF
--- a/pkg/image/bootc_disk.go
+++ b/pkg/image/bootc_disk.go
@@ -1,0 +1,72 @@
+package image
+
+import (
+	"fmt"
+	"math/rand"
+
+	"github.com/osbuild/images/pkg/artifact"
+	"github.com/osbuild/images/pkg/container"
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/images/pkg/platform"
+	"github.com/osbuild/images/pkg/runner"
+)
+
+type BootcDiskImage struct {
+	*OSTreeDiskImage
+}
+
+func NewBootcDiskImage(container container.SourceSpec) *BootcDiskImage {
+	// XXX: hardcoded for now
+	ref := "ostree/1/1/0"
+
+	return &BootcDiskImage{
+		&OSTreeDiskImage{
+			Base:            NewBase("bootc-raw-image"),
+			ContainerSource: &container,
+			Ref:             ref,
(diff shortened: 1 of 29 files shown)


mvo5:
> I think we discussed this elsewhere but I would prefer if instead we created a `Build` pipeline interface with two implementations: an RPM-based one, which is the current `Build` pipeline, and a new container-based one.

Thank you! I forgot, fixed this now. Please let me know if that matches what you had in mind :)

achilleas-k (approved):
Nice!  I like this :)

Any reason it's still draft?  Should we add an image build of some sort for testing?

mvo5:
> Nice! I like this :)
> 
> Any reason it's still draft? Should we add an image build of some sort for testing?

Mostly draft as I was not 100% certain if I got the idea right :) Un-drafted now. I am happy to add more testing, I will look into how to do this best, it's slightly incomplete still (but should be enough to build raw images already with `bib`).

achilleas-k:
Let's merge it then and see what happens :D

PR #358, opened 2024-01-12
osbuild: remove unused `NewGrub2StageOptions()`

mvo5:
This function was replaced with NewGrub2StageOptionsUnified() in early 2022 and nothing in this code, osbuild-composer or bootc-image-builder is using it. I would like to delete it therefore.

[draft for now as I am not sure what our policy for API stability is and what else might import the images library]

### DIFF
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -641,7 +641,7 @@ func (p *OS) serialize() osbuild.Pipeline {
 					),
 				)
 			} else {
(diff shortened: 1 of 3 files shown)


achilleas-k:
Let's also rename the current function to the old simpler name then?  If it's important to know that the unified option is fixed, we can add that to the docstring.

thozza (approved):
LGTM.

AFAIK our API stability promise is that there's none (the `0` as the major version means that it is not stable) and we have to deal with any breaking changes in our projects...

PR #360, opened 2024-01-12
anaconda: bios boot for container-installer

supakeen:


achilleas-k (approved):

supakeen:
This PR enables BIOS installation without "kickstart insufficient".

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -390,6 +390,8 @@ keyboard us
 timezone UTC
 
 clearpart --all
+
+part biosboot --size=1 --fstype=biosboot
```
cgwalters, line 394: I believe we should be able to use `reqpart` too now.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -390,6 +390,8 @@ keyboard us
...
 part /boot/efi --fstype=efi --size=512 --fsoptions="umask=0077"
 part /boot --fstype=ext2 --size=1024 --label=boot
```
cgwalters, line 396: We *really really* should not use ext2 here...there's no filesystem journaling.  Is this a hack-around for https://github.com/osbuild/bootc-image-builder/issues/20 ?

If so, then...we need a reminder to remove it ASAP.  If it's not, please fix to use ext4.

Of course this larger topic goes really into the rampant duplication we have of the filesystem layout in kickstarts, many times in Go code in this repo, anaconda, and bootc, and fedora-coreos etc.

What I think we should aim for is moving the default partitioning *into the container* - not hardcoded in this repo or in bootc code.   Doing that with anaconda would require https://github.com/rhinstaller/anaconda/discussions/5197

cgwalters (changes requested):

pkg/manifest/anaconda_installer_iso_tree.go
supakeen: Tracking here: #362.

pkg/manifest/anaconda_installer_iso_tree.go
supakeen: Agreed it feels icky, it was taken directly from a pre-existing kickstart which only solidifies the problem. Tracking here: #362.

PR #361, opened 2024-01-12
New image type: Fedora iot-bootable-container

achilleas-k:
Add new image type and its image function called "bootc-container".  The image function is very similar to the iot commit but enables the BootContainer flag so it exports a bootable base container instead of a tarball of the ostree repo.

The package set for the container mostly replicates the package selection from centos-bootc [1].  Comments are also retained from the original source files to assist in future decisions for modifications.

The two differences from the centos-bootc package set are:
- Kernel not specified: it is selected automatically along with user package selection.
- The packages specified in the user-experience [2] file aren't added.  We want to keep the image definition minimal, and the existing set is already bigger than minimal.

[1] https://github.com/CentOS/centos-bootc/tree/4d5e1d86fbe1028f47e08d5533b4825e3b42dc68
[2] https://github.com/CentOS/centos-bootc/blob/4d5e1d86fbe1028f47e08d5533b4825e3b42dc68/tier-1/user-experience.yaml

## DRAFT

Keeping draft until I include a boot test for this.

### DIFF
--- a/pkg/distro/fedora/package_sets.go
+++ b/pkg/distro/fedora/package_sets.go
@@ -188,6 +188,104 @@ func iotCommitPackageSet(t *imageType) rpmmd.PackageSet {
 	return ps
 }
 
+func bootableContainerPackageSet(t *imageType) rpmmd.PackageSet {
+	// Replicating package selection from centos-bootc:
+	// https://github.com/CentOS/centos-bootc/
+	ps := rpmmd.PackageSet{
+		Include: []string{
+			"acl",
+			"attr", // used by admins interactively
+			"bootc",
+			"bootupd",
+			"chrony", // NTP support
+			"container-selinux",
+			"container-selinux",
+			"crun",
+			"cryptsetup",
+			"dnf",
+			"e2fsprogs",
+			"fwupd",                 // if you're using linux-firmware, you probably also want fwupd
+			"iproute", "iproute-tc", // route manipulation and QoS
+			"iptables", "nftables", // firewall manipulation
+			"iptables-services", // additional firewall support
+			"kbd",               // i18n
+			"keyutils",          // Manipulating the kernel keyring; used by bootc
+			"libsss_sudo",       // allow communication between sudo and SSSD for caching sudo rules by SSSD
+			"linux-firmware",    // linux-firmware now a recommends so let's explicitly include it
+			"logrotate",         // There are things that write outside of the journal still (such as the classic wtmp, etc.). auditd also writes outside the journal but it has its own log rotation.  Anything package layered will also tend to expect files dropped in /etc/logrotate.d to work. Really, this is a legacy thing, but if we don't have it then people's disks will slowly fill up with logs.
+			"lsof",
+			"lvm2",                       // Storage configuration/management
+			"nano",                       // default editor
+			"ncurses",                    // provides terminal tools like clear, reset, tput, and tset
+			"NetworkManager-cloud-setup", // support for cloud quirks and dynamic config in real rootfs: https://github.com/coreos/fedora-coreos-tracker/issues/320
+			"NetworkManager", "hostname", // standard tools for configuring network/hostname
+			"NetworkManager-team", "teamd", // teaming https://github.com/coreos/fedora-coreos-config/pull/289 and http://bugzilla.redhat.com/1758162
+			"NetworkManager-tui",               // interactive Networking configuration during coreos-install
+			"nfs-utils-coreos", "iptables-nft", // minimal NFS client
+			"nss-altfiles",
+			"openssh-clients",
+			"openssh-server",
+			"openssl",
+			"ostree",
+			"passwd", "shadow-utils", // User configuration
+			"podman",
+			"rpm-ostree",
+			"selinux-policy-targeted",
+			"sg3_utils",
+			"skopeo",
+			"socat", "net-tools", "bind-utils", // interactive network tools for admins
+			"sssd-client", "sssd-ad", "sssd-ipa", "sssd-krb5", "sssd-ldap", // SSSD backends
+			"stalld",               // Boost starving threads https://github.com/coreos/fedora-coreos-tracker/issues/753
+			"subscription-manager", // To ensure we can enable client certs to access RHEL content
+			"sudo",
+			"systemd",
+			"systemd-resolved",  // resolved was broken out to its own package in rawhide/f35
+			"tpm2-tools",        // needed for tpm2 bound luks
+			"WALinuxAgent-udev", // udev rules for Azure (rhbz#1748432)
+			"xfsprogs",
+			"zram-generator", // zram-generator (but not zram-generator-defaults) for F33 change
+		},
+		Exclude: []string{
+			"cowsay", // just in case
+			"grubby",
+			"initscripts",                         // make sure initscripts doesn't get pulled back in https://github.com/coreos/fedora-coreos-tracker/issues/220#issuecomment-611566254
+			"NetworkManager-initscripts-ifcfg-rh", // do not use legacy ifcfg config format in NetworkManager See https://github.com/coreos/fedora-coreos-config/pull/1991
+			"nodejs",
+			"perl",
+			"perl-interpreter",
+			"plymouth",         // for (datacenter/cloud oriented) servers, we want to see the details by default.  https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HSMISZ3ETWQ4ETVLWZQJ55ARZT27AAV3/
+			"systemd-networkd", // we use NetworkManager
+		},
+	}
+
+	switch t.Arch().Name() {
+	case arch.ARCH_AARCH64.String():
+		ps.Append(rpmmd.PackageSet{
+			Include: []string{
+				"irqbalance",
+				"ostree-grub2",
+			},
+		})
+	case arch.ARCH_PPC64LE.String():
+		ps.Append(rpmmd.PackageSet{
+			Include: []string{
+				"irqbalance",
+				"librtas",
+				"powerpc-utils-core",
+				"ppc64-diag-rtas",
+			},
+		})
+	case arch.ARCH_X86_64.String():
+		ps.Append(rpmmd.PackageSet{
+			Include: []string{
+				"irqbalance",
+			},
+		})
+	}
+
+	return ps
+}
+
 // INSTALLER PACKAGE SET
 
 func installerPackageSet(t *imageType) rpmmd.PackageSet {
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -131,6 +131,20 @@ var (
 		exports:          []string{"commit-archive"},
 	}
 
+	iotBootableContainer = imageType{
+		name:     "iot-bootable-container",
+		filename: "iot-bootable-container.tar",
+		mimeType: "application/x-tar",
+		packageSets: map[string]packageSetFunc{
+			osPkgsKey: bootableContainerPackageSet,
+		},
+		rpmOstree:        true,
+		image:            bootableContainerImage,
+		buildPipelines:   []string{"build"},
+		payloadPipelines: []string{"os", "ostree-commit", "ostree-encapsulate"},
+		exports:          []string{"ostree-encapsulate"},
+	}
+
 	iotOCIImgType = imageType{
 		name:        "iot-container",
 		nameAliases: []string{"fedora-iot-container"},
@@ -843,6 +857,7 @@ func newDistro(version int) distro.Distro {
 			},
 			iotSimplifiedInstallerImgType,
 		)
+
 		aarch64.addImageTypes(
 			&platform.Aarch64{
 				BasePlatform: platform.BasePlatform{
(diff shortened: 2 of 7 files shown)


pkg/distro/fedora/package_sets.go
```diff
@@ -188,6 +188,101 @@ func iotCommitPackageSet(t *imageType) rpmmd.PackageSet {
 	return ps
 }
 
+func bootcContainerPackageSet(t *imageType) rpmmd.PackageSet {
+	// Replicating package selection from centos-bootc:
+	// https://github.com/CentOS/centos-bootc/
```
cgwalters, line 193: I'd really like to not fork this...how about instead auto-generating this Go code from that as Source Of Truth?

Or vendoring it as a git submodule...

cgwalters:
There's a bigger picture issue here too in that I think it'd be most useful to have a base image build process which works as part of a `Containerfile`...executed as part of `podman build` at least.  That process would just much more naturally fit into existing pipelines.

pkg/distro/fedora/package_sets.go
achilleas-k: Could read in the yaml files and autogenerate, keeping the commit ID pinned so it updates manually (a sort of manual submodule).

cgwalters:
Previous build-base-image-as-container discussion in https://github.com/coreos/rpm-ostree/issues/4688

achilleas-k:
> There's a bigger picture issue here too in that I think it'd be most useful to have a base image build process which works as part of a `Containerfile`...executed as part of `podman build` at least. That process would just much more naturally fit into existing pipelines.

The plan here is to build Fedora bootc containers using image builder in Koji and we need it soon.  There's certainly room for alternative, easier ways, but right now we have all the bits right here and can get the builds building (and iterating) with not a lot of effort in the existing pipelines.

cgwalters:
> The plan here is to build Fedora bootc containers using image builder in Koji and we need it soon. 

Is that the plan?  There was extensive prior work on
https://pagure.io/pungi/pull-request/1699
which also builds in Koji.

Don't get me wrong, I understand the thought behind this but...at a baseline *forking the image definition* is introducing a quite serious problem at this point in time.

I'd be much more interested in seeing an approach for sustainably exposing the ability to build base images for "external" users.  The rpm-ostree YAML is...okay, but not great.  We have issues like e.g. https://github.com/coreos/rpm-ostree/issues/4701 which this effectively re-introduces - and again more generally the lack of (obvious) integration with container-native pipeline tooling (e.g. stitching together a manifest listed image).

achilleas-k:
> _forking the image definition_ is introducing a quite serious problem at this point in time

Is it?  We can keep it in sync.  Question is: Who owns the definition, who will want to request changes for it, and where will that happen?  Having the definition be a specification with potentially multiple implementations isn't unsustainable as long as everyone agrees on where the spec lives and who "owns" it.

> I'd be much more interested in seeing an approach for sustainably exposing the ability to build base images for "external" users.

Sure.  But that's a different conversation.  The sustainability of a given solution is something we can theorise about.  We can talk about all sorts of ways to enable external users to build base image.  At the same time we can build base images with the tooling we have that can already do that.  I don't see why it needs to be one or the other.  

> We have issues like e.g. [coreos/rpm-ostree#4701](https://github.com/coreos/rpm-ostree/issues/4701) which this effectively re-introduces

Not sure I understand the linked issue fully.  Is it that it doesn't implement the full container config spec?  Then yeah we don't.  I wouldn't say we're _introducing_ a problem here, we're enabling an image type that we can build which exposes some customisation but is ultimately a specific configuration of a base image that we want to build.  A template of sorts.  If and when we need more, we can iterate, but we don't have to expose the full range of every kind of bootable container that can be built, just like when we build "Fedora 39 qcow2" we don't expose every single possible configuration of a Fedora 39 VM image.  We _do_ err on the side of limiting flexibility in favour of maintainability and predictability (and limiting the possibility of _broken_ or unusable images), maybe a bit too far, but that's something we're considering more globally in the context of image builder and isn't specific to containers (bootable or otherwise).

> more generally the lack of (obvious) integration with container-native pipeline tooling

IB isn't "container-native".  You might even say it's not even "container-friendly" (bootc-image-builder made this quite clear).  We all know this and we've discussed the reasons multiple times.  I still don't understand why building a container image in our own infra with tooling that isn't container friendly is a problem though.

> This hardcoded-package-lists-in-Go is not something I can see us taking down a path for giving to "external" folks for a nice way to build base images.

There are a lot of ideas for how to externalise this configuration.  We never landed on one we all felt comfortable maintaining or distributing, but there are some good candidates.  Other priorities pushed this down the list.  It might be something we're going to change for a whole distro definition at some point (maybe Fedora and not RHEL, for example).  Again though, it's a bigger story about how images are defined in IB and how much flexibility we want to enable from the outside.  I don't think this is something we're going to change because of the introduction of a single new image definition, but it might be something we change as a result of a bigger integration effort, which might start from here.

cgwalters:
> Is it? We can keep it in sync. Question is: Who owns the definition, who will want to request changes for it, and where will that happen? 

I would be the very first to say with a baseline statement that having content shipped by Fedora come from where the definition is currently (in https://github.com/CentOS/centos-bootc ) is weird.  I know others have expressed this.

There is just a fundamental tension here in that the OS/distribution needs to be capable of owning the things it ships...but there also is today a somewhat dedicated team that is intending to manage this base image - and crucially that team has a somewhat strong interest in keeping things in sync across Fedora and C9S (and C10S in the near future).  And that team operates cross-distribution.

This issue of course runs very much into the bigger picture infrastructural divergence between Fedora and CentOS Stream.  Having things in *this* github organization doesn't from my PoV improve that (but doesn't make it worse either).

There's been a lot of discussions about aligning things with https://gitlab.com/redhat/centos-stream/ - which this effectively replaces.  There's not small implications to that.

> Having the definition be a specification with potentially multiple implementations isn't unsustainable as long as everyone agrees on where the spec lives and who "owns" it.

Given that we know we want to also make it very convenient for users/customers to build custom base images, ultimately I am less concerned with trying to define a "spec" for the single base image versus just ensuring that we aren't splitting up development efforts and CI and build systems too much in the short term.

None of this is a new issue...we of course have divergence between the way things like how the cloud images are built across the different distributions today.

Anyways I guess bottom line from me: Given that https://github.com/CentOS/centos-bootc/ is still actively maintained right at this moment, let's please continue to maintain it as Source Of Truth in the short term (i.e. sync the Go code from the YAML) as we iterate towards rethinking the larger long term architecture.

pkg/distro/fedora/package_sets.go
```diff
@@ -188,6 +188,101 @@ func iotCommitPackageSet(t *imageType) rpmmd.PackageSet {
...
+			"sg3_utils",
+			"skopeo",
+			"socat", "net-tools", "bind-utils", // interactive network tools for admins
+			"sssd-client", "sssd-ad", "sssd-ipa", "sssd-krb5", "sssd-ldap", // SSSD backends
+			"stalld",               // Boost starving threads https://github.com/coreos/fedora-coreos-tracker/issues/753
+			"subscription-manager", // To ensure we can enable client certs to access RHEL content
+			"systemd",
+			"systemd-resolved",  // resolved was broken out to its own package in rawhide/f35
+			"tpm2-tools",        // needed for tpm2 bound luks
+			"WALinuxAgent-udev", // ???
```
cgwalters, line 241: Answering `???` goes like this:

```
git log -G WALinux                                                                                                                                                                                      01/15/24 13:23:50 PM
commit 60d8e77ee405b59ff96ebf7a3a4fd1f9a784272b
Author:     Colin Walters <walters@verbum.org>
AuthorDate: Fri Dec 9 15:30:03 2022 -0500
Commit:     Colin Walters <walters@verbum.org>
CommitDate: Fri Dec 9 15:32:25 2022 -0500

    Initial split from Fedora CoreOS
```

Then running the same process there turns up
```
commit b30c72bc80135c5ff2524272bf5e3d37dd57286c
Author:     Micah Abbott <miabbott@redhat.com>
AuthorDate: Wed Dec 16 13:51:32 2020 -0500
Commit:     Jonathan Lebon <jonathan@jlebon.com>
CommitDate: Wed Jan 6 09:12:08 2021 -0500

    use WALinuxAgent-udev package for Azure udev rules
    
    BZ#1748432 was filed while working on #159 and the udev rules were
    split out into a separate package as part of F32.  Let's stop carrying
    the rules in the overlay and use the provided package now.
```

Now, I was already not totally happy with effectively losing the git history when creating the initial centos-bootc repository, but...at the same time, the rationale is that we're building also isn't the same thing as Fedora CoreOS.

But anyways, I think a notable point here is that we should have a goal to align and keep these things in sync more.

pkg/distro/fedora/package_sets.go
```diff
@@ -188,6 +188,101 @@ func iotCommitPackageSet(t *imageType) rpmmd.PackageSet {
...
+			"NetworkManager-initscripts-ifcfg-rh", // do not use legacy ifcfg config format in NetworkManager See https://github.com/coreos/fedora-coreos-config/pull/1991
+			"nodejs",
+			"perl",
+			"perl-interpreter",
+			"plymouth",         // for (datacenter/cloud oriented) servers, we want to see the details by default.  https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HSMISZ3ETWQ4ETVLWZQJ55ARZT27AAV3/
+			"systemd-networkd", // we use NetworkManager
+		},
+	}
+
+	switch t.Arch().Name() {
```
cgwalters, line 258: Just skimming, I notice this drops s390x on the floor, which I am not happy with.

achilleas-k: s390x doesn't need extra packages included here (ref: https://github.com/CentOS/centos-bootc/blob/4d5e1d86fbe1028f47e08d5533b4825e3b42dc68/tier-1/manifest.yaml#L45-L54).

Though more to your point, the image type was only enabled for x86_64 and aarch64.  Added to ppc64le and s390x.

pkg/distro/fedora/package_sets.go
achilleas-k: Thanks! I updated the comment with a reference to the rhbz.

> I think a notable point here is that we should have a goal to align and keep these things in sync more.

I agree.

cgwalters:
Another big thing this just drops is all the content we inject that isn't RPM.  Now again...I feel the tension here because 98% of the content is RPM.  But it's rather annoying to do configuration-management-via-RPM.

We have some [important things](https://github.com/search?q=repo%3ACentOS%2Fcentos-bootc%20postprocess&type=code) in there today (some of which is already duplicated in osbuild, but definitely not all of it).

I am not happy myself with generating static files via scripts, and the direction I'd like to take some of these config files is a more container-native flow that looks like `COPY`.

achilleas-k:
I expect the tests will pass now so marking this as ready.

cgwalters:
I guess as a framing thing here...ultimately the current centos-bootc configs were forked from FCOS, which actually in turn distantly inherits from RHEL Atomic Host.  I've been maintaining variants of this thing for many years.

And we're pretty committed to continuing to maintain Fedora CoreOS for quite some time, alongside RHEL Atomic Host.

I had an outstanding PR in https://github.com/coreos/fedora-coreos-config/pull/2610 which would add some inheritance.

Anyways, from my PoV the expected way the bootc ecosystem involves is container-native buildsystems, and ultimately at least some baseline system that works as part of a Containerfile.

ondrejbudai:
@cgwalters I think these are good questions for @rhatdan and @nullr0ute. We got asked to build a minimal image for Fedora IoT that would include kernel, systemd and bootc. Thus, we are leveraging our experience of building Fedora IoT and RHEL for Edge to deliver this as soon as possible, because the deadlines are tight.

My suggestion is to keep things as they are in this PR, so we can start integrating this in Fedora asap. When the dust settles, we can certainly discuss how we can improve the setup. :)

Regarding container-native buildsystems: It would be quite cool to have this in `bootc-image-builder`, we should explore this option later.

cgwalters:
Today though, the technology involved here doesn't actually use bootc.  In fact, I'm chasing multiple bugs derived from that.  There's a core tension here in that what we're building isn't RHEL for Edge (though it does absolutely make sense to not rewrite the world too).  I'm not fundamentally opposed to experimentation in this direction, but let's be very clear: this is basically a fork.  And that has quite not-small implications.

I'd kindly ask given the state of things that the work here not be branded bootc.

cgwalters:
One very specific thing this change currently drops is https://github.com/CentOS/centos-bootc/pull/98 - and that's not a small thing!  It's actually a notable part of what we're trying to change here with bootc - as a distinct thing from ostree.

Hmm, now that I look...I'm quite surprised to see no one involved here as a watcher on https://github.com/CentOS/centos-bootc/watchers ...if you're not at least skimming and understanding changes in the source repository that you're forking code from, that's a recipe for problems exactly like this.

Somewhat less important, but notable given the above...there was some extensive debate in https://github.com/ostreedev/ostree/pull/3114 (and also in https://github.com/ostreedev/ostree/pull/3120 ) that it'd certainly be useful to have a smaller set of people watching there.

To rephrase this: what's key to the team maintaining the "bootc" experience is not rooted in making disk images from a filesystem tree.  What's important is *day 2 update consistency* - hence all the debate in these things.

Going to other repos, as of right now we have one other participant as a watcher in [bootc](https://github.com/containers/bootc/watchers) - I expect more than that.

achilleas-k:
Troubleshooting that installer build failure for F39 aarch64.

achilleas-k:
> Troubleshooting that installer build failure for F39 aarch64.

Fixed in #388.

ondrejbudai (approved):
We agreed on continuing with this effort until we figure out something better. Ideally, it should be possible to build the base container via a multi-stage Containerfile, but such a file doesn't exist yet, thus pursuing the goal of Fedora IoT bootable container images with Image Builder sounds like the only immediately actionable way.

PR #363, opened 2024-01-13
anaconda: reqpart for container-installer

### LINKED ISSUE #362: container-installer: use reqpart
In the container installer kickstart we should use `reqpart` instead of manually creating both BIOS and UEFI partitions: https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#reqpart

Ensure that created partitions are `ext4` instead of the current `ext2` for `/boot`.

supakeen:
Using `reqpart` in the kickstart automatically handles the hardware specific partitions that are necessary in UEFI or BIOS environments.

Closes #362.

Note that `reqpart` creates `/boot` as XFS.

@cgwalters we'd need to split off the discussion of the duplicate partition definitions in Anaconda, Go-code, and such; where should we take that?

### DIFF
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -391,9 +391,8 @@ timezone UTC
 
 clearpart --all
 
-part biosboot --size=1 --fstype=biosboot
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
Thanks.  LGTM!

PR #366, opened 2024-01-15
Print more informational branch info for previous builds

achilleas-k:
When a manifest has already been built, we would assume it was on a PR and print a link to the PR in the info message.  But new manifests can also appear on the merge queue (from the rebase) so let's handle that case too.  We can't link to it directly because the merge queue branch is deleted after it's merged, but the commit ID is enough info to recreate the build conditions if needed and the name of the merge queue branch includes the PR number if more context is needed.

If the PR field is empty, print an appropriate message too.  This shouldn't happen now, the condition was there while old info existed that is all gone now, but let's keep it in case something unexpected happens with the writing of the build info in the future.


bcl (approved):
Nice improvement!

PR #369, opened 2024-01-15
Report GitLab job failures globally

### LINKED ISSUE #368: GitLab CI: Report failures after retries have also failed
Flakiness + retries don't mix well with the merge-queue.

Whenever a test fails, before it's retried, it updates the test status on GitHub to failed.  If this happens on the merge-queue, the PR is removed from the queue and if the retry succeeds after that, you get into the interesting situation where:
- The CI run in GitLab is green (after retries) but,
- GitHub says it removed the PR from the queue because of a failure.

I think we can solve this by removing the `update_github_status` run from each pipeline and defining a new job with `when: on_failure` to update the status only when everything has been retried and is still failing.

_Originally posted by @achilleas-k in https://github.com/osbuild/images/issues/359#issuecomment-1892595780_

achilleas-k:
Never update the GitHub status at the end of a job.  Instead, let the
job retry and if another failure occurs, when everything is done, the
`fail` job will run in the `finish` stage and update the status
accordingly.

Fixes #368

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,3 +54,11 @@ finish:
     - shell
   script:
     - schutzbot/update_github_status.sh finish
+
+fail:
+  stage: finish
+  tags:
+    - shell
+  script:
+    - schutzbot/update_github_status.sh fail
+  when: on_failure
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -61,4 +61,5 @@ fail:
     - shell
   script:
     - schutzbot/update_github_status.sh fail
+    - exit 1  # make the pipeline fail so it doesn't look like success
   when: on_failure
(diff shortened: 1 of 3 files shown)


achilleas-k:
Forcing a failure in the generator pipeline to make sure failure reporting works as expected.

bcl (dismissed):
Looks good to me. Although the fact that the fail job has a green checkmark is ironic :)

jrusz (approved):
I thought it would run both `fail` and `finish` when the pipeline fails because it should have `when: always` by default, but apparently not, so LGTM :)

jrusz:
Also I suppose that if you want the `fail` job to really fail, you can add another line to the `script` section with `exit 1` that will make it actually fail and stop there

jrusz (changes requested):
I actually think that you should add the `exit 1` after running the `update_githu_status.sh` script. The green fail can be confusing.

achilleas-k:
> I actually think that you should add the `exit 1` after running the `update_githu_status.sh` script. The green fail can be confusing.

I had the same thought when Brian commented so good, on it!

jrusz (approved):

PR #372, opened 2024-01-15
Sync only necessary test info when filtering for generating build configs

achilleas-k:
When filtering builds for a CI pipeline, we download all the info.json files from the S3 bucket where we keep our test build cache.  Even though this isn't a lot of data, the number of files can grow fast which makes the operation take longer.  Since we only need the info for a specific distribution and architecture combination, there's no need to download all of the files every time.

Let's download only the `{distro}/{arch}/` subdirectory that contains the info.json files that we need.


test/scripts/imgtestlib.py
```diff
@@ -116,11 +116,17 @@ def s3_auth_args():
     return []
 
 
-def dl_s3_configs(destination):
+def dl_s3_configs(destination, distro=None, arch=None):
     """
     Downloads all the configs from the s3 bucket.
     """
     s3url = f"{S3_BUCKET}/{S3_PREFIX}/"
+    s3url = os.path.join(S3_BUCKET, S3_PREFIX)
```
supakeen, line 124: You're overwriting `s3url` here but also this is not the correct join operation for URIs which are different from filesystem paths, the previous one is correct (same goes for line 127).

supakeen (changes requested):

test/scripts/imgtestlib.py
achilleas-k: oops

supakeen (approved):

PR #373, opened 2024-01-16
osbuild: simplify code by removing `osbuild.Mounts type`

mvo5:
A quick drive-by commit for your consideration. I noticed while writing a test for the bootupd stage that the osbuild.Mounts type is not really helping me and hence wonder if we can do away with it. I did a grep and it seems the type is also not used outside images. Feel free to close this is you disagree or are attached to the type but to me the code feels easier to work with without it (and if you agree with this change I would like to see if `osbuild.Device` might also benefit from this).

--- 

The `osbuild.Mounts` type is just an alias to `[]osbuild.Mount` and it feels like it does not add clarity. From the type it's not clear if it's a slice or a map (like `osbuild.Devices`). While it is nice to abstract things away one needs to know it anyway when constructing the type so this commit removes the type and just exposes/uses `[]osbuild.Mount` directly.

Another benefit of this is that the pointers used are now clearer (IMHO) - i.e. `[]Mount` is a slice so no need to indirect it further via `*Mounts` in the function signatures.

### DIFF
--- a/pkg/manifest/iso_rootfs.go
+++ b/pkg/manifest/iso_rootfs.go
@@ -63,7 +63,7 @@ func (p *ISORootfsImg) serialize() osbuild.Pipeline {
 		},
 	}
 	copyStageInputs := osbuild.NewPipelineTreeInputs(inputName, p.installerPipeline.Name())
-	copyStageMounts := &osbuild.Mounts{*osbuild.NewExt4Mount(devName, devName, "/")}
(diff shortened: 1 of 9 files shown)


achilleas-k:
The rebase on the merge queue seems to be causing issues.  Doing a UI rebase to see what's up.

achilleas-k (changes requested):
Needs post-rebase fixes.

achilleas-k (approved):

PR #375, opened 2024-01-16
manifest: add missing `BuildrootFromContainer.getContainerSpecs()` (HMS-3318)

mvo5:
This was an oversight in PR#354. Without the container specs for the source are not always generated.


achilleas-k (approved):
Thanks!

PR #376, opened 2024-01-16
image,manifest: allow using the qcow2 pipeline without a build pipeline (HMS-3318)

mvo5:
This commit allows to use the qcow2 pipeline outside of a build pipeline. The use-case is the `bootc-image-builder` that will use arbitrary containers as buildroots and we cannot assume that they have `qemu-util` installed. And because `bootc-image-builder` is already a container we can install `qemu-util` there and be confident that it will work.

Note that there is one small drive-by one-ine improvement in one of the panics()s. I had a bit of trouble writing the test and this helped me to pinpoint what is missing. Happy to remove it again (as it's not needed for the PR itself).


achilleas-k (approved):
Neat!

PR #377, opened 2024-01-16
osbuild,manifest: simplify `osbuild.Devices`

mvo5:
The same idea as in https://github.com/osbuild/images/pull/373 :-) 

--- 

The osbuild.Devices type is just an alias to map[string]osbuild.Device and it feels like it might be easier to expose the map[string]Device directly. From the type it's not clear if it's a map and when using the type it needs to be used like map it seems easiest to expose it directly.

Another benefit of this is that the pointers used are now clearer (IMHO) i.e. no indirections required for maps.

### DIFF
--- a/pkg/manifest/iso_rootfs.go
+++ b/pkg/manifest/iso_rootfs.go
@@ -49,7 +49,7 @@ func (p *ISORootfsImg) serialize() osbuild.Pipeline {
 	)
 
 	devName := "device"
-	devices := osbuild.Devices{devName: *lodevice}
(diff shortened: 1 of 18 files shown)


achilleas-k (dismissed):
[Same answer](https://github.com/osbuild/images/pull/373#pullrequestreview-1823193363)

achilleas-k (dismissed):

achilleas-k (approved):

achilleas-k (approved):

PR #378, opened 2024-01-16
Fix ostree deployment target

cgwalters:
ostree: Fix target imgref

If we haven't been passed an ostree remote, don't use it for
verification; just use `ostree-unverified-image` which matches
https://github.com/containers/bootc/pull/230/commits/20d2d6cdc1bb2ddf7f8a6884ad83efaa708b4c65

---

ostree: Use local name when creating deployment

This will address https://github.com/osbuild/bootc-image-builder/issues/69

(But not marking closes because we need to update the vendor there)

---


runcom (commented):
lgtm - I'll leave it to @ondrejbudai to verify given the bug report

achilleas-k:
Thanks!  Doing a quick local test to verify before approving and merging.

achilleas-k (approved):

PR #381, opened 2024-01-17
many: use bootupd for image.BootcDiskImage

mvo5:
This commit enables bootupd for the `image.BootcDiskImage` image type.

### DIFF
--- a/pkg/image/bootc_disk.go
+++ b/pkg/image/bootc_disk.go
@@ -47,7 +47,8 @@ func (img *BootcDiskImage) InstantiateManifestFromContainers(m *manifest.Manifes
 		panic(fmt.Sprintf("no compression is allowed with %q format for %q", imgFormat, img.name))
 	}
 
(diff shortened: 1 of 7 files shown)


ondrejbudai:
Don't worry, that's still on our plan, this is just an intermediate step: We want to start firstly using the container itself as a buildroot, and then switch fully to bootc.

achilleas-k:
> This commit enables bootupd for the `iamge.BootcDiskImage` image type.

Typo in the commit message :)

achilleas-k:
Would you mind breaking this up into multiple commits?  Preferably 3 where:
1. Changes in `pkg/osbuild/`
2. Changes in `pkg/manifest/`
3. Changes in `pkg/image/`

There will of course be some changes that need to happen in `image` when things change in `manifest`, but that general separation will help with reading the PR (and the history in the future).

achilleas-k (approved):

PR #385, opened 2024-01-18
osbuild: extract `genMountsDevicesFromPt` helper

mvo5:
The `org.osbuild.copy` stage requires mounts/devices options when
run. Those will also be required by the coming `bootupd` stage.
To prepare for this the helper is extracted and tests added.

There is a second commit that extracts a common `testdisk.MakeFakePartitionTable("/", "/boot")` helper which is not strictly needed but feels nicer.

This is the first split out of https://github.com/osbuild/images/pull/381

### DIFF
--- a/internal/testdisk/partition.go
+++ b/internal/testdisk/partition.go
@@ -1,0 +1,34 @@
+package testdisk
+
+import (
+	"github.com/osbuild/images/pkg/disk"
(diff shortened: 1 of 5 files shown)


achilleas-k (approved):
LGTM.  Thanks!!

PR #382, opened 2024-01-17
Replace s3cmd with awscli2 in test scripts

achilleas-k:
Quick tests showed that the sync operation is much faster with awscli2 than with s3cmd.

PR #386, opened 2024-01-18
Test: return and print stdout from aws s3 sync only when filtering

achilleas-k:
Followup to #382.  This partially reverts 366d846fd3f749f448594f1f50fdb78d749cd84b.

Return the output from the aws s3 sync command and print it only when running the filter_builds() function, which syncs a limited set of build info.

When running from check-build-coverage, don't print the output since that will download all files and generates too much output for the job log (and isn't very useful).  Also, fail the job if the sync failed since it doesn't make sense to run anything after it and a failure in that job could be a sign that something's wrong.

In the filter_builds() function, don't fail if the sync failed, just let it run and generate a build config for all the image configurations.

### DIFF
--- a/test/scripts/imgtestlib.py
+++ b/test/scripts/imgtestlib.py
@@ -119,15 +119,17 @@ def dl_s3_configs(destination, distro=None, arch=None):
     print(f"⬇️ Downloading configs from {s3url}")
     # only download info.json (exclude everything, then include) files, otherwise we get manifests and whole images
     job = sp.run(["aws", "s3", "sync",
+                  "--no-progress",  # wont show progress but will print file list
                   "--exclude=*",
                   "--include=*/info.json",
                   s3url, destination],
-                 capture_output=False,
+                 capture_output=True,
                  check=False)
     ok = job.returncode == 0
     if not ok:
         print(f"⚠️ Failed to sync contents of {s3url}:")
-    return ok
+        print(job.stderr.decode())
+    return job.stdout.decode(), ok
 
 
 def get_manifest_id(manifest_data):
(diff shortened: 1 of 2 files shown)


test/scripts/imgtestlib.py
```diff
@@ -119,15 +119,17 @@ def dl_s3_configs(destination, distro=None, arch=None):
...
                   "--include=*/info.json",
                   s3url, destination],
-                 capture_output=False,
+                 capture_output=True,
                  check=False)
     ok = job.returncode == 0
     if not ok:
         print(f"⚠️ Failed to sync contents of {s3url}:")
-    return ok
+        print(job.stderr.decode())
```
mvo5, line 131: Iirc `capture_output` redirects both stdout/stderr to `job.stdout`, so this one maybe `None`

achilleas-k: https://docs.python.org/3/library/subprocess.html?highlight=capture_output#subprocess.run implies they're captured separately, so I decided to check to be sure:
```python
>>> import subprocess as sp
>>> c = sp.run(["ls", "doesnotexist"], capture_output=True)
>>> c.stdout
b''
>>> c.stderr
b"ls: cannot access 'doesnotexist': No such file or directory\n"
>>>
```

mvo5: Oh, sorry - TIL :) if we just use stderr here, some context may get lost, should be also include stdout?

mvo5 (approved):
Let's do it, it looks fine. I was wondering a bit if in the error case we lose context because stdout is not printed (afaict) but a) I may misread the code b) it may not be needed c) we can add stdout if it's really missing, this is just a testhelper after all :)

mvo5 (approved):

PR #388, opened 2024-01-19
test: update F39 repositories

achilleas-k:
The repository snapshots were updated as part of #361 (not yet merged) which caused the Fedora 39 aarch64 image installer build to start failing.      The packages we're installing for Anaconda are larger than 4 GiB (the size of the LiveOS image) on Fedora 39 aarch64.  The size increase isn't huge, checking the size of the packages, it goes from 3880533484 to 3974938171 bytes, but it's enough to tip it over the threshold of a 4 GiB partition.

Opening this PR as a standalone repo snapshot update to modify the LiveOS size separate from the #361 PR.
Increasing the size for all LiveOS images on the image installers to 5 GiB to avoid further issues.  After compression, the increase should have a negligible (if any) effect on the size of the ISO.


bcl (approved):
Looks like the perfect solution :)

PR #391, opened 2024-01-22
dnfjson: fix pointer aliasing issue

diaasami:


achilleas-k (approved):
LGTM.

Pinging @ezr-ondrej on this since he wrote the original code to ask if this might disrupt anything.  Afaict, the default value for the option should be `false`, so functionally this change should have no effect.

ezr-ondrej:
I don't understand why there are no pointers in this struct in contrast to other structs around the repo, but I did like to have a pointer in this case, because I like the extra option "flag not passed" which might be different then setting to false explicitly. For depsolving only, it is only filosophical problem tho, as false is a default and when it's one use and doesn't stick around anywhere, it's the same thing.

If it would be passed to yum repo file on the image itself at some point I'd say it would make a slight difference tho.

For the record, I've originally proposed this as a nonpointer, to keep the struct non-pointer, but we've come to the current design in this conversation: https://github.com/osbuild/images/pull/284#discussion_r1413828785

I hope the context helps, I don't know enough to decide, I'll leave decisions to smarter people :stuck_out_tongue:

achilleas-k:
> For the record, I've originally proposed this as a nonpointer, to keep the struct non-pointer, but we've come to the current design in this conversation: [#284 (comment)](https://github.com/osbuild/images/pull/284#discussion_r1413828785)
> 
> I hope the context helps, I don't know enough to decide

Thanks.  That's basically the context I was looking for.
Given the conversation you linked and your description, I think we should keep it as a pointer and make sure it's copied explicitly without aliasing.

ezr-ondrej (approved):
works for me :+1:

achilleas-k (approved):
Thanks!

PR #395, opened 2024-01-23
Move the actual path policies into a public package 'policies'

bcl:
This will allow them to be used along with the
blueprints.CheckMountpointsPolicy() function outside of images itself.

### DIFF
--- a/pkg/distro/fedora/imagetype.go
+++ b/pkg/distro/fedora/imagetype.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/internal/environment"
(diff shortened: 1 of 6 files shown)


achilleas-k:
For a followup: I think we'll also have to make things in `internal/pathpolicy/` public.  Right now, if you want to call `blueprint.CheckMountpointsPolicy()`, the second argument needs to be a `*pathpolicy.PathPolicies`, which is a private type, so you can only use the policies that are already defined:
```go
	blueprint.CheckMountpointsPolicy(bp.Customizations.Filesystem, policies.MountpointPolicies)
```

I can imagine it'd be useful for callers to also be able to define their own policies.  We might want to have different ones in BIB.

achilleas-k (approved):
LGTM.  Thanks!

PR #399, opened 2024-01-24
rhel9: systemd service to create mountpoints for filesystem customizations for edge images

say-paul:
Filesystem customization  mountpoints does not persists any upgrade due to the property of ostree which only persists data on /var and /etc. This feature will ingest the filesystem customization and create a unit file which can create the required mountpoints if it doesnot exist.

### DIFF
--- a/pkg/distro/rhel9/images.go
+++ b/pkg/distro/rhel9/images.go
@@ -490,6 +490,10 @@ func edgeRawImage(workload workload.Workload,
 	img.Filename = t.Filename()
 	img.Compression = t.compression
 
+	for _, fs := range customizations.GetFilesystems() {
+		img.CustomFilesystems = append(img.CustomFilesystems, fs.Mountpoint)
+	}
+
 	return img, nil
 }
 
@@ -545,6 +549,10 @@ func edgeSimplifiedInstallerImage(workload workload.Workload,
 	rawImg.PartitionTable = pt
 
 	rawImg.Filename = t.Filename()
+
+	for _, fs := range customizations.GetFilesystems() {
+		rawImg.CustomFilesystems = append(rawImg.CustomFilesystems, fs.Mountpoint)
+	}
 
 	// 92+ only
 	if kopts := customizations.GetKernel(); kopts != nil && kopts.Append != "" {
--- a/pkg/manifest/ostree_deployment.go
+++ b/pkg/manifest/ostree_deployment.go
@@ -74,6 +74,8 @@ type OSTreeDeployment struct {
 
 	// Use bootupd instead of grub2 as the bootloader
 	UseBootupd bool
+
+	CustomFileSystems []string
 }
 
 // NewOSTreeCommitDeployment creates a pipeline for an ostree deployment from a
@@ -353,6 +355,19 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
 			},
 		}))
 
+		// This will create a custom systemd unit that create
+		// mountpoints if its not present.This will safeguard
+		// any ostree deployment  which has custom filesystem
+		// during ostree upgrade.
+		// issue # https://github.com/osbuild/images/issues/352
+		if len(p.CustomFileSystems) != 0 {
+			serviceName := "osbuild-ostree-mountpoints.service"
+			stageOption := osbuild.NewSystemdUnitCreateStageOptions(createMountpointService(serviceName, p.CustomFileSystems))
+			stageOption.MountOSTree(p.osName, ref, 0)
+			pipeline.AddStage(stageOption)
+			p.EnabledServices = append(p.EnabledServices, serviceName)
+		}
+
 		// We enable / disable services below using the systemd stage, but its effect
 		// may be overridden by systemd which may reset enabled / disabled services on
 		// firstboot (which happend on F37+). This behavior, if available, is triggered
@@ -480,4 +495,39 @@ func (p *OSTreeDeployment) getInline() []string {
 	}
 
 	return inlineData
+}
+
+// Creates systemd unit stage by ingesting the servicename and mount-points
+func createMountpointService(serviceName string, mountpoints []string) *osbuild.SystemdUnitCreateStageOptions {
+	var conditionPathIsDirectory []string
+	for _, mountpoint := range mountpoints {
+		conditionPathIsDirectory = append(conditionPathIsDirectory, "|!"+mountpoint)
+	}
+	unit := osbuild.Unit{
+		Description:              "Ensure custom filesystem mountpoints exist",
+		DefaultDependencies:      false,
+		ConditionPathIsDirectory: conditionPathIsDirectory,
+	}
+	service := osbuild.Service{
+		Type:            osbuild.Oneshot,
+		RemainAfterExit: true,
+		//compatibility with composefs, will require transient rootfs to be enabled too.
+		ExecStartPre: []string{"/bin/sh -c \"if [ -z \"$(grep -Uq composefs /run/ostree-booted)\" ]; then chattr -i /; fi\""},
+		ExecStopPost: []string{"/bin/sh -c \"if [ -z \"$(grep -Uq composefs /run/ostree-booted)\" ]; then chattr +i /; fi\""},
+		ExecStart:    []string{"mkdir -p " + strings.Join(mountpoints[:], " ")},
+	}
+	install := osbuild.Install{
+		WantedBy: []string{"local-fs.target"},
+	}
+	options := osbuild.SystemdUnitCreateStageOptions{
+		Filename: serviceName,
+		UnitPath: osbuild.Etc,
+		UnitType: osbuild.System,
+		Config: osbuild.SystemdServiceUnit{
+			Unit:    &unit,
+			Service: &service,
+			Install: &install,
+		},
+	}
+	return &options
 }
--- a/pkg/osbuild/systemd_unit_create_stage_test.go
+++ b/pkg/osbuild/systemd_unit_create_stage_test.go
@@ -1,0 +1,70 @@
+package osbuild
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func createSystemdUnit() SystemdServiceUnit {
+
+	var unit = Unit{
+		Description:              "Create directory and files",
+		ConditionPathExists:      []string{"!/etc/myfile"},
+		ConditionPathIsDirectory: []string{"!/etc/mydir"},
+		Requires:                 []string{"dbus.service", "libvirtd.service"},
+		Wants:                    []string{"local-fs.target"},
+	}
+	var service = Service{
+		Type:            Oneshot,
+		RemainAfterExit: true,
+		ExecStartPre:    []string{"echo creating_files"},
+		ExecStopPost:    []string{"echo done_creating_files"},
+		ExecStart:       []string{"mkdir -p /etc/mydir", "touch /etc/myfiles"},
+	}
+
+	var install = Install{
+		RequiredBy: []string{"multi-user.target", "boot-complete.target"},
+		WantedBy:   []string{"sshd.service"},
+	}
+
+	var systemdUnit = SystemdServiceUnit{
+		Unit:    &unit,
+		Service: &service,
+		Install: &install,
+	}
+
+	return systemdUnit
+}
+
+func TestNewSystemdUnitCreateStage(t *testing.T) {
+	systemdServiceConfig := createSystemdUnit()
+	var options = SystemdUnitCreateStageOptions{
+		Filename: "create-dir-files",
+		Config:   systemdServiceConfig,
+	}
+	expectedStage := &Stage{
+		Type:    "org.osbuild.systemd.unit.create",
+		Options: &options,
+	}
+
+	actualStage := NewSystemdUnitCreateStageOptions(&options)
+	assert.Equal(t, expectedStage, actualStage)
+}
+
+func TestNewSystemdUnitCreateStageInEtc(t *testing.T) {
+	systemdServiceConfig := createSystemdUnit()
+	var options = SystemdUnitCreateStageOptions{
+		Filename: "create-dir-files",
+		Config:   systemdServiceConfig,
+		UnitPath: Etc,
+		UnitType: Global,
+	}
+	expectedStage := &Stage{
+		Type:    "org.osbuild.systemd.unit.create",
+		Options: &options,
+	}
+
+	actualStage := NewSystemdUnitCreateStageOptions(&options)
+	assert.Equal(t, expectedStage, actualStage)
+}
--- a/test/config-map.json
+++ b/test/config-map.json
@@ -234,5 +234,27 @@
     "image-types": [
       "image-installer"
     ]
+  },
+  "./configs/ostree-filesystem-customizations-installer.json": {
+    "image-types": [
+      "edge-simplified-installer"
+    ],
+    "distros": [
+      "rhel-9.2",
+      "rhel-9.3",
+      "rhel-9.4"
+    ]
+  },
+  "./configs/ostree-filesystem-customizations.json": {
+    "image-types": [
+      "edge-raw-image",
+      "edge-ami",
+      "edge-vsphere"
+    ],
+    "distros": [
+      "rhel-9.2",
+      "rhel-9.3",
+      "rhel-9.4"
+    ]
   }
 }
--- a/Schutzfile
+++ b/Schutzfile
@@ -8,7 +8,7 @@
   "fedora-39": {
     "dependencies": {
       "osbuild": {
-        "commit": "b42e1afddc37dae3199e039f2a3d3cb10f01485c"
+        "commit": "6b4bb850a77feade79d839d9ea075e29ca210ae9"
       }
     },
     "repos": [
--- a/pkg/distro/rhel9/distro_test.go
+++ b/pkg/distro/rhel9/distro_test.go
@@ -672,7 +672,7 @@ func TestDistro_CustomFileSystemManifestError(t *testing.T) {
 			imgType, _ := arch.GetImageType(imgTypeName)
 			_, _, err := imgType.Manifest(&bp, distro.ImageOptions{}, nil, 0)
 			if imgTypeName == "edge-commit" || imgTypeName == "edge-container" {
-				assert.EqualError(t, err, "Custom mountpoints are not supported for ostree types")
+				assert.EqualError(t, err, "Custom mountpoints are not supported for edge-container and edge-commit")
 			} else if imgTypeName == "edge-installer" || imgTypeName == "edge-simplified-installer" || imgTypeName == "edge-raw-image" || imgTypeName == "edge-ami" || imgTypeName == "edge-vsphere" {
 				continue
 			} else {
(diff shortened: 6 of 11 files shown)


pkg/distro/rhel9/images.go
```diff
@@ -698,3 +708,24 @@ func initialSetupKickstart() *fsnode.File {
 	}
 	return file
 }
+
+// Creates a unit file that will create the the mountpoints if it doesnot exits.
+// This ensures that the filesystem created using image builder continues to work after any upgrade.
+func createFilesystemMounpointService(customizations *blueprint.Customizations) *fsnode.File {
```
runcom, line 714: I think somewhere has been mentioned to create an osbuild stage just for this no?

runcom:
this is a silly cross-check from me, have we made sure that data is properly persisted within these mountpoints?

say-paul:
> this is a silly cross-check from me, have we made sure that data is properly persisted within these mountpoints?

I tested this out, data persists.

pkg/distro/rhel9/images.go
say-paul: We can get this out for as fix and having a osbuild stage is more more like a permanent fix to a temporary problem, 
I would prefer more of a template-able service that picks ups the mountpoints and create them on the go.
cc @achilleas-k

pkg/distro/rhel9/images.go
```diff
@@ -476,6 +476,11 @@ func edgeRawImage(workload workload.Workload,
 	}
 	img.PartitionTable = pt
 
+	customServiceFile := createFilesystemMounpointService(customizations)
+	if customServiceFile != nil {
+		img.Files = []*fsnode.File{customServiceFile}
```
7flying, line 481: since `img.Files` is an array, and it is already initialized, it does make more sense to me to append to it rather than create a new one

achilleas-k: You're right.  Even if it's known to be empty at this point (not sure it is), better append to it so a future modification that might end up being added before this line doesn't become a noop.

pkg/distro/rhel9/images.go
```diff
@@ -533,6 +538,11 @@ func edgeSimplifiedInstallerImage(workload workload.Workload,
 	}
 	rawImg.PartitionTable = pt
 
+	customServiceFile := createFilesystemMounpointService(customizations)
+	if customServiceFile != nil {
+		rawImg.Files = []*fsnode.File{customServiceFile}
```
achilleas-k, line 543: Same here (append).

pkg/distro/rhel9/images.go
achilleas-k: I think it's okay to inject this as a file for now and make a dedicated stage out of it later.

achilleas-k (changes requested):

achilleas-k:
> > this is a silly cross-check from me, have we made sure that data is properly persisted within these mountpoints?
> 
> I tested this out, data persists.

@henrywang When this is added to composer, can we get a test for this added?  We will need to add custom mountpoints to an edge image and then upgrade it, so I think it can be included in the existing test scripts.

pkg/distro/rhel9/images.go
say-paul: changed to append

pkg/distro/rhel9/images.go
say-paul: Appended

7flying (approved):
The changes that I requested were added, LGTM

say-paul:
The CI failure is expected as the, the unit condition fails on first boot, it runs when there is an upgrade when the mount-points gets deleted. Will be doing some manual testing on "edge-ami" to confirm the functionality.
 @achilleas-k @thozza  do you suggest removing the "edge-ami" as it will fail every time for first boot.

pkg/distro/rhel9/images.go
```diff
@@ -698,3 +708,25 @@ func initialSetupKickstart() *fsnode.File {
...
 }
+
+// fixes: https://github.com/osbuild/images/issues/352
+// Creates a unit file that create mountpoints if it doesnot exits.
+// This ensures that the filesystem created using image builder continues to work after upgrade.
+func createFilesystemMounpointService(customizations *blueprint.Customizations) *fsnode.File {
+	mntpnts := customizations.GetFilesystems()
+	if mntpnts != nil {
+		customServiceUnit := "[Unit]\nDescription=Create mountpoints if does not exists\nDefaultDependencies=no\n"
+		customService := "[Service]\nType=oneshot\nRemainAfterExit=yes\nExecStartPre=chattr -i /\nExecStopPost=chattr +i /\nExecStart=mkdir -p"
```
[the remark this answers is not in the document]
say-paul: > One thing to note is this hack will break when switching to composefs. Ultimately then a use case that wants this would need to enable root.transient instead.

I have been pondering on this from: https://gitlab.com/CentOS/automotive/sample-images/-/merge_requests/451

cc @runcom @7flying 

Adding a warning seems to be feasible considering ITM's dead line.
and  following up with a spike on root.transient.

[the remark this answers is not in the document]
cgwalters: > I have been pondering on this from: https://gitlab.com/CentOS/automotive/sample-images/-/merge_requests/451

It's important to understand this: The "signed OS image" model simply *cannot* have out of band arbitrary unsigned code injected - and that's exactly what this is.  It's state injected into a *disk image* that is unsigned.

Hence, this is why ultimately the model we need to get away from configuring disk images - and ensure that user desired mountpoints are created in the container (ostree commit).

pkg/distro/rhel9/images.go
```diff
@@ -698,3 +708,25 @@ func initialSetupKickstart() *fsnode.File {
 	}
 	return file
 }
+
+// fixes: https://github.com/osbuild/images/issues/352
+// Creates a unit file that create mountpoints if it doesnot exits.
```
cgwalters, line 713:
```suggestion
// Creates a unit file that ensures empty directories exist to be used for mountpoints.
```
maybe?

pkg/distro/rhel9/images.go
```diff
@@ -698,3 +708,25 @@ func initialSetupKickstart() *fsnode.File {
...
+// This ensures that the filesystem created using image builder continues to work after upgrade.
+func createFilesystemMounpointService(customizations *blueprint.Customizations) *fsnode.File {
+	mntpnts := customizations.GetFilesystems()
+	if mntpnts != nil {
+		customServiceUnit := "[Unit]\nDescription=Create mountpoints if does not exists\nDefaultDependencies=no\n"
```
cgwalters, line 718: It's also a best practice to include information about the *source* of systemd units.  I'd inject a comment into the generated unit at the top, like:
```suggestion
	customServiceUnit := "# Generated by osbuild/images\n[Unit]\nDescription=Create mountpoints if does not exists\nDefaultDependencies=no\n"
```

Further consider using e.g. `Documentation=https://github.com/osbuild/images` maybe?

cgwalters (approved):
I'm OK with this as is, we can't do better as far as I know without reworking how blueprints work for builds.

pkg/distro/rhel9/images.go
```diff
@@ -698,3 +708,25 @@ func initialSetupKickstart() *fsnode.File {
 	}
 	return file
 }
+
+// fixes: https://github.com/osbuild/images/issues/352
+// Creates a unit file that create mountpoints if it doesnot exits.
+// This ensures that the filesystem created using image builder continues to work after upgrade.
+func createFilesystemMounpointService(customizations *blueprint.Customizations) *fsnode.File {
```
thozza, line 715: In general, using custom files for things which are not one-off or a special case is highly discouraged in favor of explicit osbuild stage.

There is already https://github.com/osbuild/osbuild/blob/main/stages/org.osbuild.systemd.unit.

While it was developed mainly to create service unit dropins, The name of the stage is a bit unfortunate, since ideally it should be `org.osbuild.systemd.unit.dropin` or something similar. That said, I think that the `dropin` stage option could be made optional with the behavior of the stage changing to create the actual unit file, instead of the unit dropin. The stage will require additional options to be able to cover your use case, but I think that these will prove useful in the future for other scenarios as well.

Lastly, the function should then end up in the `osbuild` package and not in a distro definition.

thozza (changes requested):
Thank you for the fix.

Unfortunately, I'm not in favor of using `*fsnode.File` for this purpose. I propose to extend https://github.com/osbuild/osbuild/blob/main/stages/org.osbuild.systemd.unit and use it instead.

The rest looks OK to me.

In general, `*fsnode.File` is not intended as a workaround for not creating / enhancing osbuild stages, unless the problem is a special / corner case.

cgwalters:
See also https://github.com/ostreedev/ostree/pull/2681

pkg/distro/rhel9/images.go
say-paul: @thozza I think renaming the stage  `org.osbuild.systemd.unit` ->  `org.osbuild.systemd.unit.dropin` will make sense as it will correctly depict of whats the stage doing,  and modifying 'org.osbuild.systemd.unit' to create systemd units.

thozza: We can't rename existing stages and replace them when incompatible implementation. It is not an option, sorry.

achilleas-k:

say-paul:
waiting for https://github.com/osbuild/osbuild/pull/1578 to merge

achilleas-k:
Converted to draft since the PR was green and I felt it should be clearer that it's blocked on the osbuild PR.

pkg/distro/rhel9/images.go
say-paul: done.

pkg/distro/rhel9/images.go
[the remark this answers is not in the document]
say-paul: discussed the approach in meeting and we will keep the .service file in deployment stage with addition of the `create-mountpoint.service` for rhel9.4.

pkg/distro/rhel9/images.go
say-paul: Added them

pkg/manifest/ostree_deployment.go
```diff
@@ -481,3 +494,34 @@ func (p *OSTreeDeployment) getInline() []string {
...
 }
+
+// Creates systemd unit stage by ingesting the servicename and mount-points
+func createMountpointService(serviceName string, mountpoints []string) *osbuild.SystemdUnitCreateStageOptions {
+	var conditionPathExists []string
+	for _, mountpoint := range mountpoints {
+		conditionPathExists = append(conditionPathExists, "|!"+mountpoint)
+	}
+	unit := osbuild.Unit{
+		Description:         "Generated by osbuild/images - Create mountpoints if does not exist",
```
achilleas-k, line 505: Small syntax fix and changes to `osbuild` since `osbuild/images` doesn't really mean much in the end (it's osbuild that's doing the building and writing of files)
```suggestion
		Description:         "Generated by osbuild - Create mountpoints if they do not exist",
```

pkg/manifest/ostree_deployment.go
```diff
@@ -481,3 +494,34 @@ func (p *OSTreeDeployment) getInline() []string {
...
+		ConditionPathExists: conditionPathExists,
+	}
+	service := osbuild.Service{
+		Type:            osbuild.Oneshot,
+		RemainAfterExit: true,
+		ExecStartPre:    []string{"chattr -i"},
+		ExecStopPost:    []string{"chattr +i"},
```
achilleas-k, line 512: Shouldn't these specify a path to `chattr`?

test/config-map.json
```diff
@@ -206,9 +206,7 @@
     ],
     "image-types": [
       "edge-raw-image",
-      "edge-ami",
```
achilleas-k, line 209: Why was edge-ami removed from this config?

achilleas-k (changes requested):
Looks good overall.  A couple of comments and suggestions.

Also can you please rewrite the commit messages?  There should be an empty line between the subject and body.  For example currently the first commit message is:
```
    stage/systemd-unit-create:add systemd-unit-create stage
    Create systemd unit using the system-unit-create stage

    This new stage adds the ability to create a systemd service
    that is stored in `/usr/lib/systemd/system/`, the service name
...
```

but it should be:
```
```
    stage/systemd-unit-create: add systemd-unit-create stage

    Create systemd unit using the system-unit-create stage.
    This new stage adds the ability to create a systemd service
    that is stored in `/usr/lib/systemd/system/`, the service name
...
```

File: test/config-map.json (hunk 10 above)
Comment (say-paul) [reply test/config-map.json:209]: removed `edge-ami` as the the `create-mountpoint.service` will fail on first boot resulting in `test/scripts/base-host-check.sh` to fail too, which now I think can modify to make it state agnostic for this service.

File: pkg/manifest/ostree_deployment.go (hunk 9 above)
Comment (achilleas-k) [reply pkg/manifest/ostree_deployment.go:512]: Right, the workarounds there are running `chattr -i /` (on root).
`chattr -i` without a path isn't a valid command.

Comment (say-paul) [reply pkg/manifest/ostree_deployment.go]: my bad.. overlooked `/`

**Comment by say-paul** [comment]
CI is failing as , it does not have the osbuild with the systemd-unit-create stage yet.

**Comment by achilleas-k** [comment]
> CI is failing as , it does not have the latest osbuild with the systemd-unit-create stage yet.

You can change the version of osbuild used in the tests and point it to the commit ID for (current) main `90193d007f8a4f55e07290e4a3896227730c5505` in the Schutzfile: https://github.com/osbuild/images/blob/56b58ecef28e7303088dfdf3dc4ec6aef3b69478/Schutzfile#L6C20-L6C60

That will install osbuild from the RPMs built in its CI at that commit.

File: pkg/manifest/ostree_deployment.go [hunk 11]
```diff
@@ -353,6 +355,17 @@ func (p *OSTreeDeployment) serialize() osbuild.Pipeline {
...
 		}))
 
+		// This will create a custom systemd unit that create
+		// mountpoints if its not present.This will safeguard
+		// any ostree deployment  which has custom filesystem
+		// during ostree upgrade.
+		// issue # https://github.com/osbuild/images/issues/352
+		if len(p.CustomFileSystems) != 0 {
+			serviceName := "osbuild-ostree-mountpoints.service"
+			pipeline.AddStage(osbuild.NewSystemdUnitCreateStageOptions(createMountpointService(serviceName, p.CustomFileSystems)))
```
thozza, line 365: You need to call
```
<stage_object>.MountOSTree(p.osName, ref, 0)
```
before adding it to the pipeline. Have a look at other lines around your added code.

say-paul: updated

pkg/distro/rhel9/images.go
say-paul: modified  inside Description now

say-paul:
converted to draft as the functionality is not working as expected

say-paul:
osbuild satge can now be customized to store the system unit file in /etc too, which shall persist after `rpm-ostree upgrade` , modification done accordingly to accommodate the customization.

say-paul:
re-applied 0e4425222a17e3b7bbf26f4e396a1c3e52a7a0f1

achilleas-k (changes requested):
Looks good in general.  The git history needs some work though.
The first commit adds a config to the config-map (`ostree-filesystem-customizations-installer.json`) that doesn't get added until the fourth commit.
Also the commit message in the fourth commit doesn't match the patch anymore.  It mentions disabling the edge-ami boot test which it doesn't do.

pkg/osbuild/systemd_unit_create_stage_test.go
```diff
@@ -0,0 +1,69 @@
...
+	var service = Service{
+		Type:            Oneshot,
+		RemainAfterExit: true,
+		ExecStartPre:    []string{"echo creating_files"},
+		ExecStopPost:    []string{"echo done_creating_files"},
+		ExecStart:       []string{"mkdir -p /etc/mydir", "touch /etc/myfiles"},
+	}
+
+	var install = Install{
+		RequiredBy: []string{"multi-user.tager", "boot-complete.target"},
```
miabbott, line 26: I think this should be `multi-user.target`

pkg/osbuild/systemd_unit_create_stage_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func createSystemdUnit() SystemdServiceUnit {
+
+	var unit = Unit{
+		Description:         "Create directory and files",
+		ConditionPathExists: []string{"!/etc/mufiles"},
```
miabbott, line 13: I think this should be `/etc/myfiles`

pkg/osbuild/systemd_unit_create_stage_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func createSystemdUnit() SystemdServiceUnit {
+
+	var unit = Unit{
+		Description:         "Create directory and files",
+		ConditionPathExists: []string{"!/etc/myfile"},
```
miabbott, line 13: Should be `/etc/myfiles`...currently missing the trailing `s`

pkg/osbuild/systemd_unit_create_stage_test.go
```diff
@@ -0,0 +1,69 @@
...
+	var service = Service{
+		Type:            Oneshot,
+		RemainAfterExit: true,
+		ExecStartPre:    []string{"echo creating_files"},
+		ExecStopPost:    []string{"echo done_creating_files"},
+		ExecStart:       []string{"mkdir -p /etc/mydir", "touch /etc/myfiles"},
+	}
+
+	var install = Install{
+		RequiredBy: []string{"multi-user.tagert", "boot-complete.target"},
```
miabbott, line 26: Should be `multi-user.target`....misspelled as `tagert`

pkg/osbuild/systemd_unit_create_stage_test.go
miabbott, line 13: Should there also be a `ConditionPathExists: []string{"!/etc/mydir"}` here?

achilleas-k (changes requested):

pkg/osbuild/systemd_unit_create_stage_test.go
say-paul: I though of using `/mydir` at first, on reading the systemd.ConditionPathExits [definition](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html)  considered using file instead.

miabbott: You are right; you would need support for `ConditionPathIsDirectory`

say-paul: ah! it works with `ConditionPathExists`  unless we have user trying to fool the system by creating a file with same name. I can block this PR until we have that support in osbuild- I will create PR there to add the required support.

say-paul: now have a PR: https://github.com/osbuild/osbuild/pull/1666/ for the support.

say-paul:
blocked on : https://github.com/osbuild/osbuild/pull/1666/

achilleas-k (approved):

PR #405, opened 2024-01-26
Test: rebuild disks from bootable containers when the BIB container changes

achilleas-k:
Write the [bootc-image-builder](https://quay.io/repository/centos-bootc/bootc-image-builder) container ID that we used to build a disk image and boot test bootable containers.  We use this ID to retest bootable container images when the bib container changes to ensure we maintain compatibility.

### DIFF
--- a/test/scripts/test_imgtestlib.py
+++ b/test/scripts/test_imgtestlib.py
@@ -1,6 +1,20 @@
 import os
+import subprocess as sp
+import tempfile
+
+import pytest
 
 import imgtestlib as testlib
+
+TEST_ARCHES = ["amd64", "arm64"]
+
+
+def can_sudo_nopw() -> bool:
+    """
+    Check if we can run sudo without a password.
+    """
+    job = sp.run(["sudo", "-n", "true"], capture_output=True, check=False)
+    return job.returncode == 0
 
 
 def test_runcmd():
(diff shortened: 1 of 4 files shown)


ondrejbudai (dismissed):
Yay!

achilleas-k:
Requested review from @mvo5 because tests ^_^

achilleas-k:
Okay, the bib test failed.  This is good :)

test/scripts/test_imgtestlib.py
```diff
@@ -21,3 +33,49 @@ def test_read_seed():
...
+
+# image IDs for
+#  registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/manifest-list-test:latest
+image_ids = {
+    "amd64": "sha256:dbb63178dc9157068107961f11397df3fb62c02fa64f697d571bf84aad71cb99",
+    "arm64": "sha256:62d2a7b3bf9e0b4f3aba22553d6971227b5a39f7f408d46347b1ee74eb97cb20",
+}
+
+
+@pytest.mark.parametrize("arch", ["amd64", "arm64"])
```
mvo5, line 55: (nipitck) given that `["amd64", "arm64"]` is repeated maybe `TEST_ARCHES = ["amd64", "arm64"]` or something? Then if we add s390x or riscv64 we only need to change one place?

achilleas-k:
Current status: The image doesn't build with BIB since we switched to using bootupd.  The metadata isn't part of our base image.  This was expected (and the reason I added the test in the first place).  Working on adding this now.

achilleas-k:
Rebased.  Should be good to go!

test/scripts/test_imgtestlib.py
```diff
@@ -32,3 +46,48 @@ def test_path_generators():
...
+    assert testlib.skopeo_inspect_id(f"{transport}{test_container}@{manifest_id}", arch) == image_id
+
+
+@pytest.mark.skipif(not can_sudo_nopw(), reason="requires passwordless sudo")
+@pytest.mark.parametrize("arch", TEST_ARCHES)
+def test_skopeo_inspect_localstore(arch):
+    transport = "containers-storage:"
+    image = "registry.gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/manifest-list-test:latest"
+    with tempfile.TemporaryDirectory() as tmpdir:
+        testlib.runcmd(["sudo", "podman", "pull", f"--arch={arch}", "--storage-driver=vfs", f"--root={tmpdir}", image])
```
ondrejbudai, line 91: `--storage-driver=vfs` is so we can run the test in a container, right? It might be worth leaving a small note about that here, it might not be obvious to future readers.

EDIT from future Ondrej who read all commits: So it apparently doesn't work in a container... weird...

ondrejbudai (approved):
Awesome! I think we can resolve my comment in a follow-up, the test is disabled nevertheless.

Issue #407, opened 2024-01-29
osbuild fails building FIPS enabled image-installer images.


mmartinv:
```console
Generating manifest for image_installer_fips_true: 2024/01/29 16:12:35 The host building this image is not running in FIPS mode. The image will still be FIPS compliant. If you have custom steps that generate keys or perform cryptographic operations, those must be considered non-compliant.
[WARNING]
The host building this image is not running in FIPS mode. The image will still be FIPS compliant. If you have custom steps that generate keys or perform cryptographic operations, those must be considered non-compliant.
WARN[0000] Failed to load consumer certs: no consumer key found 
No match for group package "dracut-config-rescue"
DONE
Building manifest: /home/mmartinv/output/rhel_9.4-x86_64-image_installer-image_installer_fips_true/manifest.json
Traceback (most recent call last):
  File "/usr/bin/osbuild", line 33, in <module>
    sys.exit(load_entry_point('osbuild==105', 'console_scripts', 'osbuild')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/osbuild/main_cli.py", line 134, in osbuild_cli
    manifest = fmt.load(desc, index)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/osbuild/formats/v2.py", line 371, in load
    load_pipeline(desc, index, manifest, source_refs)
  File "/usr/lib/python3.12/site-packages/osbuild/formats/v2.py", line 350, in load_pipeline
    load_stage(desc, index, pl, manifest, source_refs)
  File "/usr/lib/python3.12/site-packages/osbuild/formats/v2.py", line 319, in load_stage
    load_input(name, desc, index, stage, manifest, source_refs)
  File "/usr/lib/python3.12/site-packages/osbuild/formats/v2.py", line 274, in load_input
    raise ValueError(f"Unknown source reference(s) {unknown_refs}")
ValueError: Unknown source reference(s) {'sha256:03fc1d407d48fd091625350f6482676dd08ef587d24761c8e38a3c99e5497af9'}
running osbuild failed: exit status 1
```

PR #408, opened 2024-01-29
Ensure FIPS files are added to the build

mmartinv:
Make sure the FIPS files are always added to the build
when needed.

Fixes: #407


achilleas-k (changes requested):
We should test for this.  The manifest in the issue should never have been possible to generate.  It wasn't a build failure, it was an invalid manifest because it contained a reference to a file that didn't exist in the sources.
We should add at least one test that enables the options so the manifest gets built when it changes.

achilleas-k:

achilleas-k (approved):
Let's merge it and add the test in a follow-up.

PR #412, opened 2024-01-30
osbuild: add org.osbuild.bootc.install-to-filesystem support

mvo5:
This PR adds support for `org.osbuild.bootc.install-to-filesystem` from https://github.com/osbuild/osbuild/pull/1547.

### DIFF
--- a/pkg/osbuild/bootc_install_to_filesystem_stage.go
+++ b/pkg/osbuild/bootc_install_to_filesystem_stage.go
@@ -1,0 +1,21 @@
+package osbuild
+
+// NewBootcInstallToFilesystem creates a new stage for the
+// org.osbuild.bootc.install-to-filesystem stage.
+//
+// It requires a mount setup so that bootupd can be run by bootc. I.e
+// "/", "/boot" and "/boot/efi" need to be set up so that
+// bootc/bootupd find and install all required bootloader bits.
+//
+// The mounts input should be generated with GenBootupdDevicesMounts.
+func NewBootcInstallToFilesystemStage(devices map[string]Device, mounts []Mount) (*Stage, error) {
+	if err := validateBootupdMounts(mounts); err != nil {
+		return nil, err
+	}
+
+	return &Stage{
+		Type:    "org.osbuild.bootc.install-to-filesystem",
+		Devices: devices,
+		Mounts:  mounts,
+	}, nil
+}
--- a/pkg/osbuild/bootc_install_to_filesystem_stage_test.go
+++ b/pkg/osbuild/bootc_install_to_filesystem_stage_test.go
@@ -1,0 +1,82 @@
+package osbuild_test
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+func TestBootcInstallToFilesystemStageNewHappy(t *testing.T) {
+	devices := makeOsbuildDevices("dev-for-/", "dev-for-/boot", "dev-for-/boot/efi")
+	mounts := makeOsbuildMounts("/", "/boot", "/boot/efi")
+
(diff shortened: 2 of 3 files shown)


pkg/osbuild/bootc_install_to_filesystem_stage_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+func TestBootcInstallToFilesystemStageNewHappy(t *testing.T) {
+	devices := makeOsbuildDevices("dev-/", "dev-/boot", "dev-/boot/efi")
```
bcl, line 14: What is this `dev-` and `mnt-` notation?

achilleas-k:

pkg/osbuild/bootc_install_to_filesystem_stage_test.go
```diff
@@ -0,0 +1,82 @@
...
+  "type": "org.osbuild.bootc.install-to-filesystem",
+  "devices": {
+    "dev-/": {
+      "type": "orgosbuild.loopback"
+    },
+    "dev-/boot": {
+      "type": "orgosbuild.loopback"
+    },
+    "dev-/boot/efi": {
+      "type": "orgosbuild.loopback"
```
cgwalters, line 55: orgosbuild!  It sounds like a battle cry a barbarian warrior might make to get their computer to work or something.

pkg/osbuild/bootc_install_to_filesystem_stage.go
```diff
@@ -0,0 +1,23 @@
+package osbuild
+
+// NewBootcInstallToFilesystem creates a new stage for the
+// org.osbuild.bootc.install-to-filesystem stage.
+//
+// It requires a mount setup of "/", "/boot" and "/boot/efi" right now
+// so that the bootupd that bootc calls finds and installs all
+// required bootloader bits.
+//
+// XXXX: rename GenBootupdDevicesMounts() helper?
```
achilleas-k, line 10: Sure, why not.

pkg/osbuild/bootc_install_to_filesystem_stage_test.go
achilleas-k: I was wondering how the test was passing and now realise that this is a pre-existing typo from 159551483f which I missed!

achilleas-k (dismissed):

pkg/osbuild/bootc_install_to_filesystem_stage_test.go
mvo5: *cough* I will grab a brown paperbag. Sorry for that! Bug is fixed in my next push.

pkg/osbuild/bootc_install_to_filesystem_stage_test.go
mvo5: This is indeed not very intuitive. Those setup the test device names for the mounts. I changed this now to:
```go
	devices := makeOsbuildDevices("dev-for-/", "dev-for-/boot", "dev-for-/boot/efi")
	mounts := makeOsbuildMounts("/", "/boot", "/boot/efi")
```
which hopefully makes it a bit clearer that the devices created are "for" the given mount point. Ideas how to make this nicer are very welcome, I'm still not 100% happy.

pkg/osbuild/bootc_install_to_filesystem_stage.go
mvo5: I looked at this again and decided to update the comment instead. We are actually doing all this so that bootupd can run (via bootc) inside the target and I couldn't think of a better way to describe what we do and validate so for now left it at `GenBootupdDevicesMounts()`. But of course I'm open for ideas :)

ondrejbudai (approved):
Looks good! :)

PR #417, opened 2024-02-01
create ostree user sudoers file

ldjebran:
In the context of compatibility with edge-management when creating image ISO artifact. 
When creating an ostree user, the user is created with username and ssh-key without a password, the user has no possibility to manage the system as has no password to enter when using sudo command.
Create a sudoer file at first boot stage.

FIXES: https://issues.redhat.com/browse/THEEDGE-3837

### DIFF
--- a/pkg/distro/image_config.go
+++ b/pkg/distro/image_config.go
@@ -20,6 +20,8 @@ type ImageConfig struct {
 	DisabledServices    []string
 	DefaultTarget       *string
 	Sysconfig           []*osbuild.SysconfigStageOptions
+	// whether to create sudoer file for wheel group with NOPASSWD option
+	WheelNoPasswd *bool
 
 	// List of files from which to import GPG keys into the RPM database
 	GPGKeyFiles []string
--- a/pkg/distro/rhel9/edge.go
+++ b/pkg/distro/rhel9/edge.go
@@ -99,6 +99,7 @@ var (
 		defaultImageConfig: &distro.ImageConfig{
 			Locale:          common.ToPtr("en_US.UTF-8"),
 			EnabledServices: edgeServices,
+			WheelNoPasswd:   common.ToPtr(true),
 		},
 		rpmOstree:        true,
 		bootISO:          true,
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -26,6 +26,8 @@ type AnacondaInstallerISOTree struct {
 	Remote  string
 	Users   []users.User
 	Groups  []users.Group
+	// whether to create sudoer file for wheel group with NOPASSWD option
+	WheelNoPasswd bool
 
 	PartitionTable *disk.PartitionTable
 
@@ -330,9 +332,16 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 			osbuild.NewOstreePullStageInputs("org.osbuild.source", p.ostreeCommitSpec.Checksum, p.ostreeCommitSpec.Ref),
 		))
 
+		baseksPath := p.KSPath
+		if p.WheelNoPasswd {
+			// move the base kickstart to another file so we can write the
+			// %post kickstart snippet in the default location and include the
+			// base
+			baseksPath = "/osbuild-base.ks"
+		}
 		// Configure the kickstart file with the payload and any user options
 		kickstartOptions, err := osbuild.NewKickstartStageOptionsWithOSTreeCommit(
-			p.KSPath,
+			baseksPath,
 			p.Users,
 			p.Groups,
 			makeISORootPath(p.PayloadPath),
@@ -345,6 +354,28 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 		}
 
 		pipeline.AddStage(osbuild.NewKickstartStage(kickstartOptions))
+
+		if p.WheelNoPasswd {
+			// Because osbuild core only supports a subset of options,
+			// we append to the base here with hardcoded wheel group with NOPASSWD option
+			hardcodedKickstartBits := `
+%include /run/install/repo/osbuild-base.ks
+
+%post
+echo -e "%wheel\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/wheel"
+chmod 0440 /etc/sudoers.d/wheel
+restorecon -rvF /etc/sudoers.d
+%end
+`
+			kickstartFile, err := fsnode.NewFile(p.KSPath, nil, nil, nil, []byte(hardcodedKickstartBits))
+			if err != nil {
+				panic(err)
+			}
+
+			p.Files = []*fsnode.File{kickstartFile}
+
+			pipeline.AddStages(osbuild.GenFileNodesStages(p.Files)...)
+		}
 	}
 
 	if p.containerSpec != nil {
--- a/pkg/distro/image_config_test.go
+++ b/pkg/distro/image_config_test.go
@@ -70,6 +70,7 @@ func TestImageConfigInheritFrom(t *testing.T) {
 					},
 					LeapsecTz: common.ToPtr(""),
 				},
+				WheelNoPasswd: common.ToPtr(true),
 			},
 			expectedConfig: &ImageConfig{
 				Timezone: common.ToPtr("UTC"),
@@ -92,6 +93,7 @@ func TestImageConfigInheritFrom(t *testing.T) {
 				EnabledServices:  []string{"sshd"},
 				DisabledServices: []string{"named"},
 				DefaultTarget:    common.ToPtr("multi-user.target"),
+				WheelNoPasswd:    common.ToPtr(true),
 				Sysconfig: []*osbuild.SysconfigStageOptions{
 					{
 						Kernel: &osbuild.SysconfigKernelOptions{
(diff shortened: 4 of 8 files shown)


cgwalters:
I don't see how this is conceptually specific to ostree.  I think what's a bit confusing here is whether an image comes out of the box with passwordless sudo.

Fedora CoreOS derivatives [hardcode this](https://github.com/coreos/fedora-coreos-config/blob/f1c4441b4a0c52e968a56bfa86bba5d7b71b3e77/overlay.d/05core/etc/sudoers.d/coreos-sudo-group#L2).

I think what happens on cloud-init systems is that cloud-init automatically adds this for the user configured.

Anyways, I think the other way to achieve this is to just document how to inject a drop-in into `/etc/sudoers` to have the effect you describe here.

ldjebran:
@cgwalters many thanks for reviewing this PR, 

> How this is related to ostree? 

For the moment only for OSTree that we are setting users in firstboot stage  
https://github.com/osbuild/images/blob/main/pkg/manifest/os.go#L447

> I think what's a bit confusing here is whether an image comes out of the box with passwordless sudo.

R: Because we do not set a user password for OSTree image and only an ssh-key, and we do not setup a root user.

> I think the other way to achieve this is to just document how to inject a drop-in into /etc/sudoers

it's impossible to login as root, as we do not set a root password, and root password is locked at installation level.  

At edge-management we are already downloading the ISO artifact generated by image-builder, injecting our kickstart file generated in our project from this template https://github.com/RedHatInsights/edge-api/blob/main/templates/templateKickstart.ks#L77
and then upload it again (the ISO artifact) to be lately downloaded by the user.

This PR is in the context to delegate edge-management functionality to image-builder as in the future image-builder has to build the ISO for edge image and save it directly to storage without any external intervention, the ISO generated has to provide the same installation  experience as with edge-management ISO modifications (eg: without any user intervention at installation level).

Please let me know, if there is any idea how we can achieve this goal without this PR or by applying some modifications.

pkg/manifest/os.go
```diff
@@ -806,6 +806,13 @@ func usersFirstBootOptions(users []users.User) *osbuild.FirstBootStageOptions {
 			cmds = append(cmds, fmt.Sprintf("mkdir -p %s", sshdir))
 			cmds = append(cmds, fmt.Sprintf("sh -c 'echo %q >> %q'", *user.Key, filepath.Join(sshdir, "authorized_keys")))
 			cmds = append(cmds, fmt.Sprintf("chown %s:%s -Rc %s", user.Name, user.Name, sshdir))
+			if user.Name != "root" {
+				sudoersPath := filepath.Join("/etc", "sudoers.d")
+				sudoersUserFilePath := filepath.Join(sudoersPath, user.Name)
+				cmds = append(cmds, fmt.Sprintf(`sh -c 'echo -e "%s\tALL=(ALL)\tNOPASSWD: ALL" >> %q'`, user.Name, sudoersUserFilePath))
+				cmds = append(cmds, fmt.Sprintf("chmod 440 %s", sudoersUserFilePath))
+				cmds = append(cmds, fmt.Sprintf("restorecon -rvF %s", sudoersPath))
+			}
```
achilleas-k, line 815: Since we might not want this for all ostree-based images (for example, Fedora IoT), we should probably put this particular functionality behind a config option on the image type and pass it down through the OS pipeline and into the `usersFirstBootOptions()` function.
That way we can enable it for RHEL for Edge and not enable it for IoT.  We could also (in the future) make it a user-facing option.

achilleas-k (changes requested):
Thanks.  This looks good in general but I'd rather have it configurable for different image types (see my comment below).

achilleas-k:
> I don't see how this is conceptually specific to ostree. I think what's a bit confusing here is whether an image comes out of the box with passwordless sudo.
> 
> Fedora CoreOS derivatives [hardcode this](https://github.com/coreos/fedora-coreos-config/blob/f1c4441b4a0c52e968a56bfa86bba5d7b71b3e77/overlay.d/05core/etc/sudoers.d/coreos-sudo-group#L2).
> 

The RHEL for Edge ISO currently adds the NOPASSWD option to sudoers during installation through the kickstart file.  This PR is part of a larger effort to bring the R4E ISO kickstart modifications into image builder itself, either as configuration options or as defaults.  The teams (Edge Fleet and IB) agreed that including this action as a first-boot action rather than install-time was an acceptable functionally equivalent change.

> I think what happens on cloud-init systems is that cloud-init automatically adds this for the user configured.
> 

It does but I'm not sure how that relates.

> Anyways, I think the other way to achieve this is to just document how to inject a drop-in into `/etc/sudoers` to have the effect you describe here.

That's a decision for the Edge Fleet team to make to judge if that's an acceptable change for their users / customers.  Though if it's taken care of at build time, or at install time, and there is no root user nor a password for the regular user, it's not easy to write the file after provisioning.

ldjebran:
@achilleas-k made some changes created a CreateUsersSodoerFile option in ImageConfig, thanks to revisit.

pkg/distro/image_config.go
```diff
@@ -20,7 +20,8 @@ type ImageConfig struct {
 	DisabledServices    []string
 	DefaultTarget       *string
 	Sysconfig           []*osbuild.SysconfigStageOptions
-
+	// whether to create sudoer file for non root users
+	CreateUsersSodoerFile *bool
```
achilleas-k, line 24: The option name has a typo, "sudoers" not "sodoers".
More generally though, I think it should be a bit more descriptive about what it does.  Yes, we're adding a drop-in for users in the `/etc/sudoers.d`, but functionally the purpose of the option is to enable NOPASSWD sudo-ing.

I'd rename it to something like `NopasswdSudo`.

pkg/distro/rhel9/edge.go
```diff
@@ -24,8 +24,9 @@ var (
 			osPkgsKey: edgeCommitPackageSet,
 		},
 		defaultImageConfig: &distro.ImageConfig{
-			EnabledServices: edgeServices,
-			SystemdUnit:     systemdUnits,
+			EnabledServices:       edgeServices,
+			SystemdUnit:           systemdUnits,
+			CreateUsersSodoerFile: common.ToPtr(true),
```
achilleas-k, line 29: I see you've enabled the option for the edge-commit (both here and in RHEL 8).  I'm curious if that's what we want for functional parity with the current edge builds.  Afaik, Edge Fleet Management doesn't build edge commits with users in them but instead adds the user during installation using the kickstart.  So user configurations go into the `edge-installer` build.  If we stay with that workflow, this wont have any effect because no users will be added during the commit build.

We've been trying to get users to not create user accounts in the base ostree commits, because they become harder to manage.  If we're going to change this, we should figure out all the implications and guidance (upgrade commits for example need to include the same user otherwise it will be removed).

If. however, we want to stay to the current flow as much as possible (no users in commits, user creation in the kickstart), then this option (and the associated first-boot service creation) should happen in the ISO, using the kickstart.

I realise I didn't catch this earlier when we were discussing this solution.  I only realised what we were doing now when I tried to test the change myself.

achilleas-k (changes requested):

pkg/manifest/os.go
ldjebran: done

pkg/distro/rhel9/edge.go
ldjebran: done via AnacondaInstallerISOTree as suggested.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -345,6 +347,28 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
...
+%post
+echo -e "%wheel\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/wheel"
+chmod 0440 /etc/sudoers.d/wheel
+restorecon -rvF /etc/sudoers.d
+%end
+`
+			kickstartFile, err := fsnode.NewFile(p.KSPath, nil, nil, nil, []byte(hardcodedKickstartBits))
+			if err != nil {
+				panic(err)
+			}
```
achilleas-k, line 366: When this is true, the kickstart options above (the base kickstart) should also change so that the path is `/osbuild-base.ks`.  The way it's written now in the PR, both the hardcoded kickstart here and the original kickstart will be written to `p.KSPath` (`/osbuild.ks`) and the second one will overwrite the primary.

Because we can't add an `%include` in the base kickstart currently, we need to write the primary one as `osbuild-base.ks` and add the include here, so this is correct, but Line 337 should change to `"osbuild-base.ks"`.

achilleas-k (changes requested):
Nice work thanks.  One small change required as well for the base kickstart path.

pkg/manifest/anaconda_installer_iso_tree.go
ldjebran: done

runcom:
I read the PR as no behavior change introduced with this on the normal edge artifacts as it's opt-in (and also just for anaconda). FWIW, lgtm

achilleas-k:
> I read the PR as no behavior change introduced with this on the normal edge artifacts as it's opt-in (and also just for anaconda). FWIW, lgtm

Thanks for taking a look.  Yes, to confirm and ftr: This PR makes the Anaconda Edge installer put a sudoer file in `/etc/sudoers.d` that enables nopasswd for the wheel group.

runcom:
> However, it will affect on-prem users of the Anaconda ISO.

_if_ they enable the option right? we're not building the edge anaconda iso with this new option by default right? (or we break backwards compatibility)

achilleas-k:
> > However, it will affect on-prem users of the Anaconda ISO.
> 
> _if_ they enable the option right? we're not building the edge anaconda iso with this new option by default right? (or we break backwards compatibility)

Yeah right now it's on by default and you're right.  We should put it behind an option.  I'll add that in a follow-up and expose it in composer.

achilleas-k (approved):

achilleas-k:
Thanks for the work, @ldjebran.  LGTM!

PR #419, opened 2024-02-01
CI: test integration to osbuild-composer as part of PRs

thozza:
It would be useful to explicitly test if the code change proposed in the PR integrates cleanly to osbuild-composer or if it will require additional work there. The cheapest way to test it is to try to replace the `osbuild/images` ref in the `osbuild-composer@main` and run Golang lint and unit tests.

These test cases should not be blocking. The intention is to provide feedback in the PR, that the author (or someone else) should also work on integrating the images change to osbuild-composer, since it won't be "for free".

### DIFF
--- a/.github/workflows/test-osbuild-composer-intergation.yml
+++ b/.github/workflows/test-osbuild-composer-intergation.yml
@@ -1,0 +1,101 @@
+name: "[osbuild-composer integration]"
+
+on:
+  pull_request:
+    branches:
+      - "*"
+
+jobs:
+  unit-tests:
+    name: "🛃 Unit tests"
+    runs-on: ubuntu-20.04
+    container:
+      image: registry.fedoraproject.org/fedora:latest
+
+    steps:
+        # krb5-devel is needed to test internal/upload/koji package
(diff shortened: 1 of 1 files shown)


thozza:
Example of a simulated issue is in https://github.com/osbuild/images/actions/runs/7744410267/job/21118086564?pr=419

bcl (approved):
Nice! It would also be nice if you could run it against an osbuild-composer PR with the needed changes, but I'm not sure how you'd do that in a nice way.

thozza:
> Nice! It would also be nice if you could run it against an osbuild-composer PR with the needed changes, but I'm not sure how you'd do that in a nice way.

I think that the most reasonable way would be to use a bot account, which could create such PR in composer if asked for it. I'll think about it, since we sometimes want to hold-off merging things in `images` until we have a corresponding changes in `osbuild-composer`.

Anyway, my intention with this PR was not to block any new code from landing in `images`, but to explicitly point out cases with the change will most probably generate more work when integrating it into `osbuild-composer`.

achilleas-k:
This is great.  Thank you!!

PR #421, opened 2024-02-02
Schutzfile: update osbuild commit ID

achilleas-k:
Update osbuild test dependency to current main.

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -3,7 +3,7 @@
   "fedora-39": {
     "dependencies": {
       "osbuild": {
-        "commit": "dfd59a38d9ec7458944f9a901a6aa6e1202a7efb"
+        "commit": "477a21043eeb55468b1f75a9574e46e77cb5fdef"
       }
     },
     "repos": [


bcl:
Seems to be an actual failure.

```
FileNotFoundError: [Errno 2] No such file or directory: 'mkfs.fat'
```
In the f39 and f40 iot-bootable-container tests. But not any others.

achilleas-k:
> Seems to be an actual failure.
> 
> ```
> FileNotFoundError: [Errno 2] No such file or directory: 'mkfs.fat'
> ```
> 
> In the f39 and f40 iot-bootable-container tests. But not any others.

I opened this PR to trigger a rebuild of all test image configurations but then realised that the fedora iot-bootable-container would fail because of the issue you mentioned.  I came across this issue but there's another one which makes the way we test it fail.  It's incompatible with the latest version of bootc-image-builder because it doesn't have the boot metadata that the tool requires: https://github.com/osbuild/images/pull/405#issuecomment-1916632191

Still trying to figure out the best way to write the needed metadata.

achilleas-k:
Handled in #427

PR #422, opened 2024-02-03
Add a btrfs partitioning mode

ondrejbudai:
This PR has basically three parts:

1) The first commits fix already existing btrfs code in the library (it was never used before). Just tiny fixes. I tried to add tests where they had been missing before.
2) The "middle section" adds missing parts for btrfs subvolume. The scaffolding was there, but the translation from an abstract partition table to osbuild stages was missing.
3) The final section is the actual addition of a btrfs partitioning mode. When this mode is used, a root partition from the base partition table is converted to a btrfs one with a subvolume. The code for adding more btrfs subvolumes based on blueprint mountpoints was already there.

Weird parts:
- When a btrfs partitioning mode is used, the compression is hard-coded to zstd:1. It would be great if this could be configured per an image type/distro/arch, but the partitioning code currently cannot do such specific things. In the end, this should be probably also user-configurable. However, I think that this default value is alright for the starters, we can always iterate on this.
- Setting `min-size` for extra mountpoints is a weird thing, because the extra mountpoints are translated to subvolumes... Maybe we should just error out if the min size is set with btrfs and depend just on the total size customization... Happy to discuss this!

Note that I specifically didn't disable the new mode in EL distributions... My thinking is that users are running these distros on a custom btrfs-enabled kernel, we shouldn't block them.

**This doesn't have tests for the actual partitioning mode. There are currently just new tests for the lower level code in the disk package.**

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -36,6 +36,9 @@ const (
 	// RawPartitioningMode always creates a raw layout.
 	RawPartitioningMode PartitioningMode = "raw"
 
+	// BtrfsPartitioningMode creates a btrfs layout.
+	BtfrsPartitioningMode PartitioningMode = "btrfs"
+
 	// DefaultPartitioningMode is AutoLVMPartitioningMode and is the empty state
 	DefaultPartitioningMode PartitioningMode = ""
 )
@@ -43,14 +46,15 @@ const (
 func NewPartitionTable(basePT *PartitionTable, mountpoints []blueprint.FilesystemCustomization, imageSize uint64, mode PartitioningMode, requiredSizes map[string]uint64, rng *rand.Rand) (*PartitionTable, error) {
 	newPT := basePT.Clone().(*PartitionTable)
 
-	if basePT.features().LVM && mode == RawPartitioningMode {
-		return nil, fmt.Errorf("raw partitioning mode set for a base partition table with LVM, this is unsupported")
+	if basePT.features().LVM && (mode == RawPartitioningMode || mode == BtfrsPartitioningMode) {
+		return nil, fmt.Errorf("%s partitioning mode set for a base partition table with LVM, this is unsupported", mode)
 	}
 
 	// first pass: enlarge existing mountpoints and collect new ones
 	newMountpoints, _ := newPT.applyCustomization(mountpoints, false)
 
 	var ensureLVM bool
+	ensureBtrfs := false
 	switch mode {
 	case LVMPartitioningMode:
 		ensureLVM = true
@@ -58,11 +62,18 @@ func NewPartitionTable(basePT *PartitionTable, mountpoints []blueprint.Filesyste
 		ensureLVM = false
 	case DefaultPartitioningMode, AutoLVMPartitioningMode:
 		ensureLVM = len(newMountpoints) > 0
+	case BtfrsPartitioningMode:
+		ensureBtrfs = true
 	default:
 		return nil, fmt.Errorf("unsupported partitioning mode %q", mode)
 	}
 	if ensureLVM {
 		err := newPT.ensureLVM()
+		if err != nil {
+			return nil, err
+		}
+	} else if ensureBtrfs {
+		err := newPT.ensureBtrfs()
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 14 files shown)


bcl:
Maybe use the sum of `min-size` to make sure the btrfs root partition is big enough to hold all the requested space?

pkg/disk/partition_table.go
```diff
@@ -678,6 +689,62 @@ func (pt *PartitionTable) ensureLVM() error {
...
+
+	rootPath := entityPath(pt, "/")
+	if rootPath == nil {
+		panic("no root mountpoint for PartitionTable")
+	}
+
+	// we need a /boot partition to boot btrfs, ensure one exists
+	bootPath := entityPath(pt, "/boot")
+	if bootPath == nil {
+		_, err := pt.CreateMountpoint("/boot", 512*common.MiB)
```
mvo5, line 704: (super nitpick) I would avoid the extra `\n` here as the `if err != nil` is directly connected to the previous previous line. We could even considering just doing `if _, err .= ...; err != nil {` to make it slightly more compact

pkg/disk/partition_table.go
```diff
@@ -678,6 +689,62 @@ func (pt *PartitionTable) ensureLVM() error {
...
+
+		if err != nil {
+			return err
+		}
+
+		rootPath = entityPath(pt, "/")
```
mvo5, line 710: Silly question (again) - why do we set the root path here again, didn't we do this in line 696 already? The two calls look identical to me? Or does the mutation of `pt.CreateMountPoint()` above require that rootPath is recalculated?

pkg/disk/partition_table.go
```diff
@@ -678,6 +689,62 @@ func (pt *PartitionTable) ensureLVM() error {
...
+	}
+
+	parent := rootPath[1] // NB: entityPath has reversed order
```
mvo5, line 713: Thanks for the comment!

pkg/disk/partition_table.go
```diff
@@ -678,6 +689,62 @@ func (pt *PartitionTable) ensureLVM() error {
...
+
+	if _, ok := parent.(*Btrfs); ok {
+		return nil
+	} else if part, ok := parent.(*Partition); ok {
```
mvo5, line 717: If I am not misreading we could follow effective go [0] and avoid the not strictly needed nested else here(?) (because the if returns when taken). So something like(?):
```go
@@ -655,7 +651,8 @@ func (pt *PartitionTable) ensureLVM() error {
 
        if _, ok := parent.(*LVMLogicalVolume); ok {
                return nil
-       } else if part, ok := parent.(*Partition); ok {
+       }
+       if part, ok := parent.(*Partition); ok {
                filesystem := part.Payload
 
                vg := &LVMVolumeGroup{
```

[0] https://go.dev/doc/effective_go#if

pkg/disk/partition_table.go
```diff
@@ -678,6 +689,62 @@ func (pt *PartitionTable) ensureLVM() error {
...
+		if err != nil {
+			return err
+		}
+
+		rootPath = entityPath(pt, "/")
+	}
+
+	parent := rootPath[1] // NB: entityPath has reversed order
+
+	if _, ok := parent.(*Btrfs); ok {
```
mvo5, line 715: I wonder if this could be a go type switch? i.e. 
```go
	switch part := parent.(type) {
	case *LVMLogicalVolume:
		return nil
	case *Partition:
		filesystem := part.Payload
...
	default:
		return fmt.Errorf("Unsupported parent for LVM")
	}
```

pkg/disk/partition_table.go
```diff
@@ -678,6 +689,62 @@ func (pt *PartitionTable) ensureLVM() error {
...
+		// replace the top-level partition payload with a new btrfs filesystem
+		part.Payload = btrfs
+
+		// reset the btrfs partition size - it will be grown later
+		part.Size = 0
+
+		if pt.Type == "gpt" {
+			part.Type = FilesystemDataGUID
+		} else {
+			part.Type = "83"
```
mvo5, line 738: Would it make sense to have a symbolic name for this constant?

pkg/disk/partition_table.go
```diff
@@ -36,28 +36,34 @@ const (
...
 	DefaultPartitioningMode PartitioningMode = ""
 )
 
 func NewPartitionTable(basePT *PartitionTable, mountpoints []blueprint.FilesystemCustomization, imageSize uint64, mode PartitioningMode, requiredSizes map[string]uint64, rng *rand.Rand) (*PartitionTable, error) {
 	newPT := basePT.Clone().(*PartitionTable)
 
-	if basePT.features().LVM && mode == RawPartitioningMode {
-		return nil, fmt.Errorf("raw partitioning mode set for a base partition table with LVM, this is unsupported")
+	if basePT.features().LVM && (mode == RawPartitioningMode || mode == BtfrsPartitioningMode) {
+		return nil, fmt.Errorf("%s partitioning mode set for a base partition table with LVM, this is unsupported", mode)
```
mvo5, line 50: Thanks for including the mode here!

mvo5 (commented):
Thank you for this branch. It looks excellent. I left a lot of comments/suggestion but that is not because there is anything wrong with the branch (the contrary!) but only because a) my default mode is `-Wpedantic -Wextra-pedantic` b) this is really good stuff so I enjoyed reviewing it c) many are just style/preferences and can just be ignored (or considered)

I only managed to review half because some other review came up but I hope this is still helpful

Conan-Kudo:
> Maybe use the sum of `min-size` to make sure the btrfs root partition is big enough to hold all the requested space?

Btrfs needs a minimum size of 16MiB when using `mkfs.btrfs --mixed` or 109MiB when using normal mkfs mode. My suggestion is to automatically use `--mixed` when the partition is below 109MiB.

PR #423, opened 2024-02-04
osbuild: remove the unused NewZiplStageOptions()

mvo5:
The function is not used in the code and it is also not needed as it will just be equivalent to `new(ZiplStageOptions)` or `&ZiplStageOption{}`. If we can always reintroduce it if we need it with a more meaningful init of the options.


thozza (approved):
Thanks!

PR #427, opened 2024-02-06
Enable bootupd for bootable containers

achilleas-k:
When building ostree bootable containers, always enable bootupd.


achilleas-k:
Cherry picked commit from https://github.com/osbuild/images/pull/405 that includes partitioning tools in the base image.

thozza (commented):
LGTM :+1:

achilleas-k:
osbuild PR merged.  Updated commit ID and marked as ready!

pkg/manifest/os.go
```diff
@@ -152,6 +152,12 @@ type OS struct {
 	// OSTreeParent source spec (optional). If nil the new commit (if
 	// applicable) will have no parent
 	OSTreeParent *ostree.SourceSpec
+
+	// Enabling Bootupd runs bootupctl generate-update-metadata in the tree to
+	// transform /usr/lib/ostree-boot into a bootupd-compatible update
+	// payload. Only works with ostree-based images.
+	Bootupd bool
```
mvo5, line 159: Silly question, should this be `Bootupd` or `BootupdGenMetadata` (or something similar more expressive)? I guess I ask if it's unambiguous enough (and it's a honest question, idk if in this context there could be other bootupd things).

pkg/manifest/os.go
```diff
@@ -152,6 +152,12 @@ type OS struct {
 	// OSTreeParent source spec (optional). If nil the new commit (if
 	// applicable) will have no parent
 	OSTreeParent *ostree.SourceSpec
+
+	// Enabling Bootupd runs bootupctl generate-update-metadata in the tree to
+	// transform /usr/lib/ostree-boot into a bootupd-compatible update
+	// payload. Only works with ostree-based images.
```
mvo5, line 158: It says `Only works with ostree-based images` - should we enforce this in code and error `if p.OStreeRef == "" && p.Bootupd`? it seems right now we just silently ignore the option? (again, honest question, maybe we never validate like this and then it would be weird to do it here only).

thozza (approved):
Looks good :+1:

pkg/manifest/os.go
```diff
@@ -768,6 +774,9 @@ func (p *OS) serialize() osbuild.Pipeline {
 				"wheel", "docker",
 			},
 		}))
+		if p.Bootupd {
```
mvo5, line 777: I would love to have a test for this just to make sure that if this option is specified we do not regress in generating the stage. I know it sounds a bit silly but I can see a future in which we refactor this function a bit and then having tests that check the high level behavior really works are useful.

pkg/osbuild/bootupd_gen_metadata_stage.go
```diff
@@ -0,0 +1,7 @@
+package osbuild
+
+func NewBootupdGenMetadataStage() *Stage {
+	return &Stage{
```
mvo5, line 4: Again, incredible silly, but I would love to see a test here if nothing else then to ensure that if we change the way we do stuff (e.g. auto-generate stages from python directly) then the tests will confirm that things are still generated the right way.

mvo5 (commented):
This looks good! I have some suggestions/questions inline and played with the code a bit and attach a small patch for your consideration (based on my suggestions).

[0001-osbuild-manifest-add-tests-for-bootupd-stage.patch.txt](https://github.com/osbuild/images/files/14193724/0001-osbuild-manifest-add-tests-for-bootupd-stage.patch.txt)

pkg/manifest/os.go
achilleas-k: Good idea.

pkg/manifest/os.go
thozza: So, in theory, the manifests that we generate as part of CI should be the way to check that changes / refactoring don't have unintended effect. It is easy to do locally as part of development. It is not so easy to see in the CI itself. In the past, we used to have `diff-manifests` job in the osbuild-composer CI, but the job was probably too noisy and mostly ignored... :confused:

pkg/manifest/os.go
thozza: I liked the option being a bit abstract, since there may be eventually other things that will have to be done differently if the image uses `bootupd`... :thinking:

achilleas-k: Yup, my thinking exactly.

pkg/manifest/os.go
achilleas-k: Yeah, there's no good tradeoff here.  We can't have a test that fails when manifests change because that happens a lot.  We can't have people manually approve manifest changes because that gets too noisy and we just end up approving blindly.
The test we do have now uses bootc-image-builder to make sure that the container image can be built into a bootable disk image and that test will run if anything changes in the image definition (and soon also whenever BIB changes).  This kind of functional test is great IMO, but it doesn't help with catching things during development.  For that, generating manifests locally and checking a diff is the most convenient way.

mvo5: I think at this level it is sufficient to check that the stage is included when the option is set. This is what the `TestBootupdStage` in `[0001-osbuild-manifest-add-tests-for-bootupd-stage.patch.txt](https://github.com/osbuild/images/files/14193724/0001-osbuild-manifest-add-tests-for-bootupd-stage.patch.txt)` is doing. Or am I missing something why we might not want this?

mvo5: Sorry, I'm surely stating the obvious (again) but the way I see it is that the tests are build on top of each others. So here we ensure that the stage is included, elsewhere we ensure the stage is generated and elsewhere that a stage is transformed into correct json. So here we test just that, in the other test `TestNewBootupdGenMetadataStage` we ensure that the right stage is generated and  in stages_test.go we have a test that ensures that `Type` for stages ends up in the right shape for consumption by osbuild. With that we have all we need here and all tests are fast :)

mvo5 (approved):
Thank you

mvo5 (approved):

achilleas-k:
Amended the top commit and repushed to kick CI which seemed to be stuck.

ondrejbudai (approved):
Amazing! :)

PR #437, opened 2024-02-09
README: Add images principles

ochosi:
I have pulled in the principles as discussed previously, condensing some sub-points into the main principles so it doesn't look too convoluted.

I've also harmonized the heading levels with osbuild's README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,21 @@
 Images
 ======
 
-### Project
+## Project
 
  * **Website**: <https://www.osbuild.org>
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
I think this is exactly how we had it written down after the discussion so :+1: from me.
Not merging until others chime in.

teg (approved):

ondrejbudai (commented):
Late approval from me. :)

PR #447, opened 2024-02-16
osbuild: generate bootupd mounts to span a single disk via  partscan=True

mvo5:
The `bootupd` and `bootc install to-filesystem` stages require a
continous disk so that grub can install the bootloader onto the
disk. This means that the new `partscan` feature from [0] must be
used when generating the disk instead of the current approach that
generates a bunch of loopback devices for each partition.

Note that this version of the code does not implement support
for "container" partition tables like LUKS.

[0] https://github.com/osbuild/osbuild/pull/1501

### DIFF
--- a/pkg/osbuild/bootupd_stage.go
+++ b/pkg/osbuild/bootupd_stage.go
@@ -76,17 +76,56 @@ func NewBootupdStage(opts *BootupdStageOptions, devices map[string]Device, mount
 	}, nil
 }
 
-func GenBootupdDevicesMounts(filename string, pt *disk.PartitionTable) (map[string]Device, []Mount) {
-	_, mounts, devices, err := genMountsDevicesFromPt(filename, pt)
-	if err != nil {
-		panic(err)
+func genMountsForBootupd(source string, pt *disk.PartitionTable) ([]Mount, error) {
+	mounts := make([]Mount, 0, len(pt.Partitions))
+	// note that we are not using pt.forEachMountable() here because we
+	// need to keep track of the partition number (even if it's not
+	// mountable)
+	for idx, part := range pt.Partitions {
+		if part.Payload == nil {
+			continue
+		}
+		// TODO: support things like LVM here via supporting "disk.Container"
+		mnt, ok := part.Payload.(disk.Mountable)
+		if !ok {
+			return nil, fmt.Errorf("type %v not supported by bootupd handling yet", mnt)
+		}
+
+		partNum := idx + 1
+		name := fmt.Sprintf("part%v", partNum)
+		mount, err := genOsbuildMount(name, source, mnt)
+		if err != nil {
+			return nil, err
+		}
+		mount.Partition = &partNum
+		mounts = append(mounts, *mount)
 	}
-	devices["disk"] = Device{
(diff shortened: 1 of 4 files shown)


pkg/osbuild/bootupd_stage.go
```diff
@@ -76,17 +77,56 @@ func NewBootupdStage(opts *BootupdStageOptions, devices map[string]Device, mount
...
+func genMountsForBootupd(source string, pt *disk.PartitionTable) ([]Mount, error) {
+	mounts := make([]Mount, 0, len(pt.Partitions))
+	// note that we are not using pt.forEachMountable() here because we
+	// need to keep track of the partition number (even if it's not
+	// mountable)
+	for idx, part := range pt.Partitions {
+		if part.Payload == nil {
+			continue
+		}
+		// TODO: support things like LVM here via supporting "disk.Container"
```
cgwalters, line 89: I'm a bit uncertain longer term about having custom code for stuff like that versus using anaconda to initialize LVM.

cgwalters (commented):
I have no experience with this code, but this looks superficially sane to me at least.

pkg/osbuild/bootupd_stage.go
```diff
@@ -76,17 +77,56 @@ func NewBootupdStage(opts *BootupdStageOptions, devices map[string]Device, mount
...
+			return nil, fmt.Errorf("type %v not supported by bootupd handling yet", mnt)
+		}
+
+		partNum := idx + 1
+		name := fmt.Sprintf("part%v", partNum)
+		mount, err := genOsbuildMount(name, source, mnt)
+		if err != nil {
+			return nil, err
+		}
+		mount.Partition = common.ToPtr(partNum)
```
ondrejbudai, line 101: `mount.Partition = &partNum` would work as well, right?

pkg/osbuild/bootupd_stage_test.go
```diff
@@ -143,3 +145,107 @@ func TestBootupdStageJsonHappy(t *testing.T) {
...
+
+func TestGenBootupdDevicesMountsMissingRoot(t *testing.T) {
+	filename := "fake-disk.img"
+	pt := &disk.PartitionTable{}
+	_, _, err := osbuild.GenBootupdDevicesMounts(filename, pt)
+	assert.EqualError(t, err, "required mounts for bootupd stage [/ /boot /boot/efi] missing")
+}
+
+const (
+	MebiByte = 1024 * 1024        // MiB
```
ondrejbudai, line 157: Can we take these constants from somewhere else? I believe we have them either in `disk`, or `common`.

ondrejbudai (approved):
Two nitpicks, otherwise this looks good. Do you want to fix these now, or as a follow-up?

pkg/osbuild/bootupd_stage_test.go
mvo5: Thanks, yes!

pkg/osbuild/bootupd_stage.go
mvo5: Nice catch, thank you!

ondrejbudai (approved):

PR #450, opened 2024-02-19
Adjust AWS partitioning for RHEL 8 and 9 (COMPOSER-2155)

ondrejbudai:
**distro/rhel9: make /boot even bigger**

We have reports that 600 MiB might still not be enough for certain use-cases,
so let's bump it again. See COMPOSER-2155 for more details.

---

**disk: add PartitionTable.GetMountpointSize**

It's generally useful to be able to get a size of a certain mountpoint. Let's
add a method for this.

In order to cover this method with tests, I extended the fake PT generator to
correctly set partition sizes. Also, I had to move the current content of
partition_table_test to a separate file and package, in order to break
an inclusion cycle.

---

**distro/rhel8: Adjust /boot size again**

It turns out 500 MiB /boot is too small. Thus, we agreed to do the following
changes for 8.10:

- for x86_64, let's not have a separate /boot at all (parity with 8.8 and
  older)
- for aarch64, let's make /boot 1 GiB big

I've removed an old, weird test for EC2 partitioning, and instead implemented
a new one that actually checks whether /boot is correctly sized (or even if it
exists) for a given distro-arch combo.

---

**distro/rhel9: cover the new ec2 partitioning with tests**

Same thing as we did for rhel8

---

### DIFF
--- a/pkg/distro/rhel9/distro_internal_test.go
+++ b/pkg/distro/rhel9/distro_internal_test.go
@@ -1,12 +1,77 @@
 package rhel9
 
 import (
+	"fmt"
+	"math/rand"
+	"strings"
 	"testing"
 
+	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/pkg/blueprint"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/osbuild/images/pkg/distro"
 )
+
+// math/rand is good enough in this case
+/* #nosec G404 */
+var rng = rand.New(rand.NewSource(0))
+
+func TestEC2Partitioning(t *testing.T) {
+	testCases := []struct {
+		distro      string
+		bootSizeMiB uint64
+	}{
+		// x86_64
+		{
+			distro:      "rhel-9.2",
+			bootSizeMiB: 500,
+		},
+		{
+			distro:      "rhel-9.3",
+			bootSizeMiB: 600,
+		},
+		{
+			distro:      "rhel-9.4",
+			bootSizeMiB: 1024,
+		},
+		{
+			distro:      "centos-9",
+			bootSizeMiB: 1024,
+		},
+	}
+
+	for _, tt := range testCases {
+		for _, arch := range []string{"x86_64", "aarch64"} {
+			for _, it := range []string{"ami", "ec2", "ec2-ha", "ec2-sap"} {
+				// skip non-existing combos
+				if strings.HasPrefix(it, "ec2") && strings.HasPrefix(tt.distro, "centos") {
+					continue
+				}
+				if arch == "aarch64" && (it == "ec2-ha" || it == "ec2-sap") {
+					continue
+				}
+				t.Run(fmt.Sprintf("%s/%s/%s", tt.distro, arch, it), func(t *testing.T) {
+					a, err := DistroFactory(tt.distro).GetArch(arch)
+					require.NoError(t, err)
+					i, err := a.GetImageType(it)
+					require.NoError(t, err)
+
+					it := i.(*imageType)
+					pt, err := it.getPartitionTable([]blueprint.FilesystemCustomization{}, distro.ImageOptions{}, rng)
+					require.NoError(t, err)
+
+					bootSize, err := pt.GetMountpointSize("/boot")
+					require.NoError(t, err)
+					require.Equal(t, tt.bootSizeMiB*common.MiB, bootSize)
+				})
+
+			}
+		}
+
+	}
+}
 
 func TestDistroFactory(t *testing.T) {
 	type testCase struct {
--- a/pkg/distro/rhel9/partition_tables.go
+++ b/pkg/distro/rhel9/partition_tables.go
@@ -17,7 +17,7 @@ func defaultBasePartitionTables(t *imageType) (disk.PartitionTable, bool) {
 		bootSize = 600 * common.MebiByte
 	default:
 		// RHEL >= 9.4 needs to have even a bigger /boot, see COMPOSER-2155
-		bootSize = uint64(600) * common.MebiByte
+		bootSize = 1 * common.GibiByte
 	}
 
 	switch t.platform.GetArch() {


pkg/distro/rhel8/ami.go
```diff
@@ -24,18 +24,14 @@ func amiImgTypeX86_64(rd distribution) imageType {
 		buildPipelines:      []string{"build"},
 		payloadPipelines:    []string{"os", "image"},
 		exports:             []string{"image"},
-		basePartitionTables: ec2BasePartitionTables,
+		basePartitionTables: getEc2PartitionTables(rd.osVersion, rd.isRHEL()),
```
thozza, line 27: So previously, the `ami` would always use the latest partitioning scheme, regardless of the release version. This won't be true any more.

pkg/distro/rhel8/ami.go
```diff
@@ -102,17 +94,13 @@ func amiImgTypeAarch64(rd distribution) imageType {
...
```
thozza, line 97: ditto: using the latest partitioning scheme, regardless of the release version.

thozza (commented):
I added a few comments, but in general the change looks good to me. Thanks for adding the tests.

However, I'm wondering if it is worth keeping the separate `/boot` partition for RHEL-8.9 x86_64 case, since it turned out to be very problematic? I would revert the separate `/boot` for all RHEL-8 x86_64...

pkg/disk/partition_table.go
```diff
@@ -750,3 +750,20 @@ func (pt *PartitionTable) GetBuildPackages() []string {
...
+		return 0, fmt.Errorf("cannot find mountpoint %s", mountpoint)
+	}
+
+	for _, ent := range path {
+		if sizeable, ok := ent.(Sizeable); ok {
+			return sizeable.GetSize(), nil
+		}
+	}
+
+	panic(fmt.Sprintf("no sizeable of the entity path for mountpoint %s", mountpoint))
```
mvo5, line 768: I wonder if we could return an internal error here? I guess I'm just not a fan of panics :)

pkg/disk/partition_table_test.go
```diff
@@ -1,26 +1,24 @@
...
-	}
+func TestPartitionTable_GetMountpointSize(t *testing.T) {
+	pt := testdisk.MakeFakePartitionTable("/", "/app")
 
-	for _, tc := range testCases {
-		pt := testPartitionTables[tc.partitionType]
-		assert.Equal(t, tc.expectedFeatures, pt.features())
+	size, err := pt.GetMountpointSize("/")
+	assert.NoError(t, err)
+	assert.Equal(t, testdisk.FakePartitionSize, size)
```
mvo5, line 15: Nice!

pkg/disk/partition_table_test.go
```diff
@@ -1,26 +1,24 @@
...
 
-	}
+	size, err = pt.GetMountpointSize("/app")
+	assert.NoError(t, err)
+	assert.Equal(t, testdisk.FakePartitionSize, size)
+
+	// non-existing
+	_, err = pt.GetMountpointSize("/custom")
+	assert.Error(t, err)
```
mvo5, line 23: I wonder if it makes sense to test for (parts of?) the specific error, I find this usually a nice pattern to also check if the error message is good or if it's one of the: `cannot find foo: foo not found: when trying to read foo it was not found` kind of situation that is so easy to hit in go :)

pkg/distro/rhel8/ami.go
```diff
@@ -136,11 +124,7 @@ func ec2ImgTypeAarch64(rd distribution) imageType {
 }
 
 func ec2SapImgTypeX86_64(rd distribution) imageType {
-	basePartitionTables := ec2BasePartitionTables
-	// use legacy partition tables for RHEL 8.8 and older
-	if common.VersionLessThan(rd.osVersion, "8.9") {
-		basePartitionTables = ec2LegacyBasePartitionTables
-	}
+	basePartitionTables := getEc2PartitionTables(rd.osVersion, rd.isRHEL())
```
mvo5, line 127: really nice to see this simplified in so many places!

pkg/distro/rhel8/distro_internal_test.go
```diff
@@ -66,21 +62,84 @@ func TestDistro_DefaultPartitionTables(t *testing.T) {
...
-		testEc2ImageType.arch = &architecture{
-			name: archName,
-		}
-		pt, err := testEc2ImageType.getPartitionTable(mountpoints, distro.ImageOptions{}, rng)
-		if _, exists := testEc2ImageType.basePartitionTables[archName]; exists {
-			require.Nil(t, err)
-			for _, m := range mountpoints {
-				assert.True(t, pt.ContainsMountpoint(m.Mountpoint))
+func TestEC2Partitioning(t *testing.T) {
+	testCases := []struct {
```
mvo5, line 66: \o/ beautilful

pkg/distro/rhel8/partition_tables.go
```diff
@@ -423,3 +247,188 @@ var edgeBasePartitionTables = distro.BasePartitionTableMap{
 		},
 	},
 }
+
+func getEc2PartitionTables(osVersion string, isRHEL bool) distro.BasePartitionTableMap {
+	if osVersion == "8.9" {
+		// 8.9: has 500 MiB /boot
+		return distro.BasePartitionTableMap{
```
mvo5, line 254: (nitpick) I would probably extract this into a `getEc2PartitionTablesRH8_9_only()` helper to make this function a bit smaller. Something like:
```go
if osVersion == "8.9" {
		// 8.9: has 500 MiB /boot
		return getEc2PartitionTablesRH8_9_only()
}
```

pkg/distro/rhel8/partition_tables.go
```diff
@@ -423,3 +247,188 @@ var edgeBasePartitionTables = distro.BasePartitionTableMap{
...
+							FSTabFreq:    0,
+							FSTabPassNo:  0,
+						},
+					},
+				},
+			},
+		}
+	}
+
+	// All other RHEL 8 minor versions have /boot-less x86_64 layout.
```
mvo5, line 356: Thanks for the comment - now I'm curious - so 8.0 to 8.8 have a /boot-less layout, 8.9 as 500mb and 8.10 is /boot-less again? Or is it more complex and I'm just misreading this?

pkg/distro/rhel9/distro_internal_test.go
```diff
@@ -1,13 +1,68 @@
...
+	"github.com/stretchr/testify/require"
 
 	"github.com/osbuild/images/pkg/distro"
 )
 
+// math/rand is good enough in this case
+/* #nosec G404 */
+var rng = rand.New(rand.NewSource(0))
+
+func TestEC2Partitioning(t *testing.T) {
```
mvo5, line 20: \o/

pkg/distro/rhel9/distro_internal_test.go
```diff
@@ -1,13 +1,68 @@
...
+	testCases := []struct {
+		distro      string
+		bootSizeMiB int
+	}{
+		// x86_64
```
mvo5, line 25: it says `x86_64` here but below it checks both x86_64 and aarch64?

pkg/distro/rhel9/partition_tables.go
```diff
@@ -7,10 +7,12 @@ import (
 )
 
 func defaultBasePartitionTables(t *imageType) (disk.PartitionTable, bool) {
-	// RHEL >= 9.3 needs to have a bigger /boot, see RHEL-7999
-	bootSize := uint64(600) * common.MebiByte
+	// RHEL >= 9.3 needs to have a bigger /boot, see RHEL-7999 and COMPOSER-2155
+	bootSize := uint64(1024) * common.MebiByte
```
mvo5, line 11: I was playing with switch/case here but I don't think we win anything:
```go
	var bootSize uint64
	switch {
	case common.VersionLessThan(t.arch.distro.osVersion, "9.3") && t.arch.distro.isRHEL():
		bootSize = 500 * common.MebiByte
	case common.VersionLessThan(t.arch.distro.osVersion, "9.4") && t.arch.distro.isRHEL():
		bootSize = 600 * common.MebiByte
	default:
		// RHEL >= 9.3 needs to have a bigger /boot, see RHEL-7999
		bootSize = uint64(600) * common.MebiByte
	}
```
wdyt? probably nested if is fine here

mvo5 (commented):
This looks very nice! A few idea/suggestion. Also - why is it draft :) ?

pkg/disk/partition_table.go
thozza: In general, panic should be used for cases which are programming errors. This is IMO such case, so it should be acceptable. Using an error would most probably have the same end result, but the idea behind using panic is to not continue the execution when we hit an undefined state.

ondrejbudai:
OK, I simplified the code in RHEL 8:

1) the ami type now behaves the same way as ec2 types: it now respect the given minor versions. Thus amis built for older RHEL 8 versions will have different partitioning than for 8.10. The rationale behind this change is that we now have only one code-path to maintain, and if the customers want an older image with a bigger /boot, they can just use a customization.
2) 8.9 for x86_64 now doesn't have a separate /boot for x86_64. The original version of this PR had a separate one for the sake of backward compatibility, but since a separate /boot turned out to be basically a bug that broke some users, we should treat it as a bug and revert the faulty behaviour. This once again led to a simplification.

pkg/distro/rhel9/partition_tables.go
ondrejbudai: Done, thanks!

pkg/distro/rhel9/distro_internal_test.go
ondrejbudai: Removed. :)

pkg/distro/rhel8/partition_tables.go
ondrejbudai: See my latest comment, I simplified this in the end.

pkg/disk/partition_table.go
mvo5: I agree here about that this is the indented usage of panics and fine to use this way. I also don't want to argue this specific case (I have not looked deep into this).

I do think we are a bit too liberal with panics. When I compare how many panics are used in e.g. the go standard library to images I see there is a difference:
```
mvogt@top:/usr/share/go/src$ grep -r panic\( io/ os/ net/ text/ |grep -v _test.go|wc -l
169
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Go                             583          16642          24314         127429

mvogt@top:~/devel/osbuild/images$ git grep panic\(|grep -v _test.go| wc -l
1549
mvogt@top:~/devel/osbuild/images$ cloc ./pkg/ ./internal/
...
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Go                             481           7686          11931          52681
```

For me the difference is that in go a panic aborts the program so when writing code that is reused something that panics a lot is harder to (re)use. And because (unlike python) there is generally no `recover` around codeblock the panic will bring a (potentially many many go routines spawning) program down entirely.

mvo5: And just to be clear, totally fine to leave as is, just wanted to raise awareness and maybe a tiny discussion/re-evaluation of the pattern :)

pkg/disk/partition_table_test.go
ondrejbudai: Added a check! :)

pkg/disk/partition_table.go
ondrejbudai: Yeah, this is a clear programming error, and I think that it's fine to `panic` on these - there's something seriously wrong with the partition table, thus I think the safest approach is to get the whole goroutine down, so the errorneous state doesn't spread. I've added a note about the fact that this is a programming error.

pkg/distro/rhel8/distro_internal_test.go
```diff
@@ -66,21 +63,63 @@ func TestDistro_DefaultPartitionTables(t *testing.T) {
...
+					bootSize, err := pt.GetMountpointSize("/boot")
+					require.NoError(t, err)
+					require.Equal(t, tt.aarch64bootSizeMiB*common.MiB, bootSize)
+				})
+
 			}
-		} else {
-			require.EqualError(t, err, fmt.Sprintf("no partition table defined for architecture %q for image type %q", testEc2ImageType.arch.name, testEc2ImageType.name))
 		}
+
```
thozza, line 122: super-nitpick: looks like an unnecessary newline :innocent:

thozza (approved):
LGTM, thanks!

pkg/distro/rhel8/ami.go
thozza: intentional change, resolving...

pkg/distro/rhel8/ami.go
thozza: intentional change, resolving...

PR #451, opened 2024-02-19
container: error if requested architecure and resolved architecture do not match

mvo5:
Should fix https://github.com/osbuild/bootc-image-builder/issues/207 - draft as this needs tests.


ondrejbudai (commented):
So far looks good! :)

PR #453, opened 2024-02-20
RHEL: fix Azure images configuration according to documentation

thozza:
Make sure that the RHEL-8 and RHEL-9 Azure images have the following configuration consistent with the official RHEL documentation:

- kernel commandline 
- modules disabled via modprobe configuration
- grub2 configuration in `/etc/default/grub`

Depends on https://github.com/osbuild/osbuild/pull/1601

Related to https://issues.redhat.com/browse/RHEL-19583

### DIFF
--- a/pkg/distro/rhel8/azure.go
+++ b/pkg/distro/rhel8/azure.go
@@ -11,7 +11,8 @@ import (
 	"github.com/osbuild/images/pkg/subscription"
 )
 
-const defaultAzureKernelOptions = "ro crashkernel=auto console=tty1 console=ttyS0 earlyprintk=ttyS0 rootdelay=300"
+// use loglevel=3 as described in the RHEL documentation and used in existing RHEL images built by MSFT
+const defaultAzureKernelOptions = "ro loglevel=3 crashkernel=auto console=tty1 console=ttyS0 earlyprintk=ttyS0 rootdelay=300"
 
 func azureRhuiImgType() imageType {
 	return imageType{
@@ -473,6 +474,7 @@ var azureRhuiBasePartitionTables = distro.BasePartitionTableMap{
 	},
 }
 
+// based on https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_rhel_8_on_microsoft_azure/assembly_deploying-a-rhel-image-as-a-virtual-machine-on-microsoft-azure_cloud-content-azure#making-configuration-changes_configure-the-image-azure
 var defaultAzureImageConfig = &distro.ImageConfig{
 	Timezone: common.ToPtr("Etc/UTC"),
 	Locale:   common.ToPtr("en_US.UTF-8"),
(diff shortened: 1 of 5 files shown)


pkg/distro/rhel9/azure.go
```diff
@@ -414,7 +414,7 @@ func azureRhuiBasePartitionTables(t *imageType) (disk.PartitionTable, bool) {
 	}
 }
 
-var defaultAzureKernelOptions = "ro console=tty1 console=ttyS0 earlyprintk=ttyS0 rootdelay=300"
+var defaultAzureKernelOptions = "ro loglevel=3 console=tty1 console=ttyS0 earlyprintk=ttyS0 rootdelay=300"
```
thozza, line 417: I'm not 100% sure if adding `loglevel=3` is a good idea, but it is used in RHEL documentation and RHEL images built by MSFT in their marketplace have it set... :thinking:

pkg/distro/rhel8/azure.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/osbuild/images/pkg/subscription"
 )
 
-const defaultAzureKernelOptions = "ro crashkernel=auto console=tty1 console=ttyS0 earlyprintk=ttyS0 rootdelay=300"
+const defaultAzureKernelOptions = "ro loglevel=3 crashkernel=auto console=tty1 console=ttyS0 earlyprintk=ttyS0 rootdelay=300"
```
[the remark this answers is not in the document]
mvo5: Maybe worth a comment in the code then? something like: `// loglevel=3 as described in the RHEL documentation` or `// loglevel=3 to be consistent with existing MSFT images` maybe (or both :) ?

pkg/distro/rhel9/azure.go
```diff
@@ -456,6 +456,12 @@ var defaultAzureImageConfig = &distro.ImageConfig{
 				osbuild.NewModprobeConfigCmdBlacklist("amdgpu"),
 			},
 		},
+		{
+			Filename: "blacklist-intel-cstate.conf",
```
mvo5, line 460: Maybe a small comment in this codeblock that adds the blacklists that include the references from the commit message? (https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/deploying_rhel_9_on_microsoft_azure/assembly_deploying-a-rhel-image-as-a-virtual-machine-on-microsoft-azure_cloud-content-azure#making-configuration-changes_configure-the-image-azure) ? Or just somewhere in the file where we customize things (as it's more than just blacklisting that this describes)

mvo5 (approved):
Thanks, this looks fine, some inline comments but feel free to ignore.

pkg/distro/rhel9/azure.go
thozza: I added a comment at the beginning of the image config.

pkg/distro/rhel8/azure.go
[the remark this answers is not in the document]
thozza: I added the comment.

ondrejbudai (approved):
:heart_eyes: Thanks!

PR #454, opened 2024-02-20
Replace IRC link in README.md with a Matrix link

AdamWill:


README.md
```diff
@@ -5,7 +5,7 @@ Images
 
  * **Website**: <https://www.osbuild.org>
  * **Bug Tracker**: <https://github.com/osbuild/images/issues>
- * **IRC**: #osbuild on [Libera.Chat](https://libera.chat/)
+ * **Matrix (chat)**: [Image Builder channel on Fedora Chat](https://matrix.to/#/#image-builder:fedoraproject.org?web-instance[element.io]=chat.fedoraproject.org)
```
bcl, line 8: The url we use in the docs is `https://matrix.to/#/#image-builder:fedoraproject.org` we should probably be consistent.

bcl (changes requested):

README.md
AdamWill: so, what the additional bit of the URL does is cue clients to go via the Fedora homeserver, which is probably what we want in this case. If you don't add that it will send you through element.io , I think, so if you don't happen to already have an account there you can't just login with FAS and start chatting.

bcl: Ah, nice, didn't know you could use FAS to log in. This assumes they have a FAS account, or want to use the fedora matrix server, which might be true. Or not. I don't use matrix so I guess I'm not the right one to make a decision.
But I do think it should be consistent across all our documentation, so either use the current plain url, or switch all of them to use chat.fedoraproject.org
@ochosi what do you think?

supakeen: I'm in favor of setting the default web-instance that people get if they don't already have a Matrix client to `chat.fedoraproject.org`.

So, keeping this PR as-is and adjusting elsewhere to stay consistent.

As an aside; we might want to look into using `matrix://`-protocol handlers which seem to exist nowadays.

bcl: Might be a bit soon to do that, I can't find any obvious documentation on what browsers support it as a protocol.

bcl (approved):
We'll change the other references to match this.

PR #457, opened 2024-02-21
Service to create custom mount-points using fsnode

say-paul:
#399  for creating service is not surviving upgrade, even when  [org.osbuild.systemd.unit.create](https://github.com/osbuild/osbuild/blob/main/stages/org.osbuild.systemd.unit.create) is modified to save file in /etc/systemd/system.

Adding service using this methods works as expected, and can be a workaround by the time we investigate the issue and fix it accordingly.


say-paul:
closed as its a work around and we pulled the feature out of 9.4

PR #468, opened 2024-02-23
fedora: remove webui kernel arguments

supakeen:
Anaconda auto-detects the webui being available in recent versions instead of using the kernel arguments. See the codepath in pyanaconda here: https://github.com/rhinstaller/anaconda/blob/8b5f5920d62bcebfadc06285e9b681d4d36f45d1/pyanaconda/anaconda.py#L209

### DIFF
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -369,8 +369,6 @@ func imageInstallerImage(workload workload.Workload,
 			// time since they might be running headless and a UI is
 			// unnecessary.
 			img.AdditionalKernelOpts = []string{"inst.text", "inst.noninteractive"}
-		} else {
-			img.AdditionalKernelOpts = []string{"inst.webui", "inst.webui.remote"}
 		}
 	}
 	img.AdditionalAnacondaModules = append(img.AdditionalAnacondaModules, "org.fedoraproject.Anaconda.Modules.Users")


supakeen:
@jkonecny12 this also disables the `webui.remote` however I'd like to double check with you if we should *keep* that enabled for the image installer (payload installer) type?

I know it was used for the previews that were supplied initially but I don't think we still build those :)

supakeen:
Got confirmation that we can remove the `webui.remote` as well :)

AdamWill:
There is still the comment `// Enable anaconda-webui for Fedora > 38` here, but now it heads a block that does something else entirely, which is confusing. Additionally, if the conditional means `>= 38`, it can probably be dropped entirely, as F37 is now EOL. (I'm not clear what the conditional means exactly, because the name would seem to suggest it's equivalent to `>= 38`, but the comment suggests it means `> 38`.)

supakeen:
@AdamWill The comment and version check are addressed in #469. As for keeping around version checks for currently unsupported versions of Fedora; @achilleas-k / @ondrejbudai should we prune these?

ondrejbudai:
Pruning code-paths related to EOL distros is encouraged! See the 4th principle: https://github.com/osbuild/images/blob/main/README.md

supakeen:
Perfect, then I'll be pruning codepaths after #469 is in in a follow-up.

ondrejbudai (approved):
Thanks!

PR #469, opened 2024-02-24
More consistent and readable version checks.

supakeen:
Two idioms were being used in `fedora/package_sets.go` and there was a mismatch between which version of Fedora enables the webui.

To improve readability and since it's such a common idiom in our codebase I've also introduced `VersionGreaterThanOrEqual` in `common`.

### DIFF
--- a/internal/common/distro.go
+++ b/internal/common/distro.go
@@ -28,14 +28,5 @@ func VersionLessThan(a, b string) bool {
 }
 
 func VersionGreaterThanOrEqual(a, b string) bool {
-	aV, err := version.NewVersion(a)
-	if err != nil {
-		panic(err)
-	}
-	bV, err := version.NewVersion(b)
-	if err != nil {
-		panic(err)
-	}
-
-	return aV.GreaterThanOrEqual(bV)
+	return !VersionLessThan(a, b)
 }
--- a/internal/common/distro.go
+++ b/internal/common/distro.go
@@ -26,3 +26,16 @@ func VersionLessThan(a, b string) bool {
 
 	return aV.LessThan(bV)
 }
+
+func VersionGreaterThanOrEqual(a, b string) bool {
+	aV, err := version.NewVersion(a)
+	if err != nil {
+		panic(err)
+	}
+	bV, err := version.NewVersion(b)
+	if err != nil {
+		panic(err)
+	}
+
+	return aV.GreaterThanOrEqual(bV)
+}
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -353,9 +353,9 @@ func imageInstallerImage(workload workload.Workload,
 		img.UnattendedKickstart = instCust.Unattended
 	}
 
-	// Enable anaconda-webui for Fedora > 38
+	// Enable anaconda-webui for Fedora >= 39
 	distro := t.Arch().Distro()
-	if !common.VersionLessThan(distro.Releasever(), "38") {
+	if !common.VersionLessThan(distro.Releasever(), "39") {
 		img.AdditionalAnacondaModules = []string{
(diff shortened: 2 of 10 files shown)


pkg/distro/fedora/images.go
```diff
@@ -353,9 +353,9 @@ func imageInstallerImage(workload workload.Workload,
 		img.UnattendedKickstart = instCust.Unattended
 	}
 
-	// Enable anaconda-webui for Fedora > 38
+	// Enable anaconda-webui for Fedora > 39
 	distro := t.Arch().Distro()
-	if !common.VersionLessThan(distro.Releasever(), "38") {
+	if !common.VersionLessThan(distro.Releasever(), "39") {
```
AdamWill, line 358: so the comment claims this means `> 39`, but the name doesn't seem to match. "not less than X" should be equivalent to "X or greater", no? 39 is not less than 39. I would think these should all be 40, not 39, at least on a naive reading of the name of the function.

supakeen: Thank you, yes. The function name is not ideal and you are correct.

pkg/distro/fedora/package_sets.go
```diff
@@ -498,13 +497,7 @@ func iotInstallerPackageSet(t *imageType) rpmmd.PackageSet {
...
 	ps := anacondaPackageSet(t)
 
-	releasever := t.Arch().Distro().Releasever()
-	version, err := strconv.Atoi(releasever)
-	if err != nil {
-		panic("cannot convert releasever to int: " + err.Error())
-	}
-
-	if version >= 38 {
+	if common.VersionLessThan(t.arch.distro.osVersion, "39") {
```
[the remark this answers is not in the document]
supakeen: Thank you, fixed.

supakeen:
@AdamWill I've fixed up the review issues and introduced a new function for the idiom that should make it all less error prone and more readable.

pkg/distro/fedora/images.go
```diff
@@ -353,9 +353,9 @@ func imageInstallerImage(workload workload.Workload,
 		img.UnattendedKickstart = instCust.Unattended
 	}
 
-	// Enable anaconda-webui for Fedora > 38
+	// Enable anaconda-webui for Fedora >= 39
```
AdamWill, line 356: this comment still should not be here at all, since we are removing the lines that do anything about webui here.

supakeen: I believe the enabling of the `AdditionalAnacondaModules` is still related to the webui but let me verify if that is still necessary.

internal/common/distro.go
```diff
@@ -26,3 +26,16 @@ func VersionLessThan(a, b string) bool {
...
+	aV, err := version.NewVersion(a)
+	if err != nil {
+		panic(err)
+	}
+	bV, err := version.NewVersion(b)
+	if err != nil {
+		panic(err)
+	}
+
+	return aV.GreaterThanOrEqual(bV)
```
ondrejbudai, line 40: Can't we just code this as `return !VersionLessThan(a, b)`? That would save us a few lines of code.

ondrejbudai (commented):
Love the new helper and the clean-ups! Just one question in-line. :)

internal/common/distro.go
supakeen: We can, I've added a commit that changes it into this suggestion.

AdamWill:
I think this PR should no longer have the "fedora: remove webui kernel arguments" commit, since it was merged separately as https://github.com/osbuild/images/pull/468 .

supakeen:
@AdamWill I've rebased this PR on `main` so that commit is now gone from this PR :)

ondrejbudai (approved):
Thanks, this is so cleaner now! :heart:

Issue #470, opened 2024-02-24
Prune Fedora codepaths that relate to Fedora < 38


supakeen:
Fedora 37 is EOL, see: https://docs.fedoraproject.org/en-US/releases/eol/#_unsupported_fedora_linux_releases this was mentioned in #468. With #469 it's easy to see that there's quite a few checks for `>= 38` that can be pruned after it lands.

supakeen:
This was completed in #471.

PR #471, opened 2024-02-24
Remove Fedora 37 codepaths

supakeen:
This removes any code paths relating to checks that assume Fedora 37 exists, see #470.


pkg/distro/fedora/package_sets.go
```diff
@@ -498,13 +497,7 @@ func iotInstallerPackageSet(t *imageType) rpmmd.PackageSet {
...
 	ps := anacondaPackageSet(t)
 
-	releasever := t.Arch().Distro().Releasever()
-	version, err := strconv.Atoi(releasever)
-	if err != nil {
-		panic("cannot convert releasever to int: " + err.Error())
-	}
-
-	if version >= 38 {
+	if !common.VersionLessThan(t.arch.distro.osVersion, "39") {
```
thozza: Just making sure that the intention here is to basically also change the condition to `version >= 39`, which is consistent with the previous commit and the rest of this file, but not mentioned explicitly in the commit message with this commit.

thozza (approved):
LGTM. I love the addition of `VersionGreaterThanOrEqual`, which makes reading the code much easier... :heart_eyes: I added one question, but I'm 99% sure that it is expected and intentional.

pkg/distro/fedora/package_sets.go
supakeen: It is the intention to move this to make it consistent with the other version checks which were all `>= 39`.

supakeen:
@thozza since #469 has landed I've rebased this PR and it should now be ready for merging :)

pkg/distro/rhel7/distro_test.go
```diff
@@ -445,7 +445,7 @@ func TestDistroFactory(t *testing.T) {
 			expected: rhel7.DistroFactory("rhel-7.9"),
 		},
 		{
-			strID:    "fedora-37",
+			strID:    "fedora-38",
```
thozza, line 448: There's a very little value in making these changes going forward, but it does not hurt for the sake of consistency... :slightly_smiling_face:

thozza (approved):
Thanks!

PR #472, opened 2024-02-24
RHEL-8: fix hybrid boot by bringing back /boot/efi

thozza:
PR#450 [1] reverted the separate `/boot` partition for x86_64 images, since it was causing issues. However, it also removed the `/boot/efi` partition. These images are supposed to support UEFI boot and are marked as such. This is causing failures when booting these images with UEFI in osbuild-composer CI.

Fix this regression by bringing back `/boot/efi`.

[1] https://github.com/osbuild/images/pull/450


supakeen (approved):
Good find.

PR #476, opened 2024-02-27
Rename CI jobs testing the integration with osbuild-composer

thozza:
It turns out that the required jobs to pass to auto-merge a PR are based only on job names. Therefore the fact that the actions testing the integration with osbuild-composer had the same job names as those that are used to test the images code (specifically the "Unit tests") made them required. There is no other way to distinguish jobs other than to rename them. So let's do it.


kingsleyzissou (approved):
Cool thanks for sorting this out

PR #477, opened 2024-02-27
osbuild: add containers-storage source

mvo5:
[split out from https://github.com/osbuild/images/pull/448 to make it easier to review]

New containers-storage source.
See https://github.com/osbuild/osbuild/pull/1550

### DIFF
--- a/pkg/osbuild/containers_storage_source_test.go
+++ b/pkg/osbuild/containers_storage_source_test.go
@@ -1,0 +1,36 @@
+package osbuild
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewContainersStorageSource(t *testing.T) {
+	imageID := "sha256:c2ecf25cf190e76b12b07436ad5140d4ba53d8a136d498705e57a006837a720f"
+
+	source := NewContainersStorageSource()
+
+	source.AddItem(imageID)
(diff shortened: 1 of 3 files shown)


pkg/osbuild/containers_storage_source_test.go
```diff
@@ -34,3 +35,29 @@ func TestNewContainersStorageSource(t *testing.T) {
...
+}`)
+
+	// add a second
+	imageID = "sha256:aabbcc5cf190e76b12b07436ad5140d4ba53d8a136d498705e57a006837a720f"
+	source.AddItem(imageID)
+	jsonOutput, err = json.MarshalIndent(source, "", "  ")
+	assert.NoError(t, err)
+	assert.Equal(t, string(jsonOutput), `{
+  "items": {
+    "sha256:aabbcc5cf190e76b12b07436ad5140d4ba53d8a136d498705e57a006837a720f": {},
```
ondrejbudai, line 59: The map ordering in Go isn't stable, right? Won't this test be flakey?

mvo5: This is true in general but for json fortunately the output is sorted, see https://pkg.go.dev/encoding/json#Marshal: 
```
The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:

    keys of any string type are used directly
...
```

PR #478, opened 2024-02-27
container: simplify resolver

mvo5:
[split out from https://github.com/osbuild/images/pull/448 to make it easier to review]

his commit simplifies how local containers are resolved. We can read the
storage-path from system config, which makes it possible to remove the
storage-path argument from the blueprints in a later commit.

---

Minor fix to the local container resolver, using `target.Name()` removes
the container's tag, which is undesirable if we are trying to resolve a
container that is not tagged with `latest`.

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -342,21 +342,11 @@ func (m RawManifest) Digest() (digest.Digest, error) {
 }
 
 func getImageRef(target reference.Named, local bool) (types.ImageReference, error) {
-
-	if !local {
-		ref, err := docker.NewReference(target)
-		if err != nil {
-			return nil, err
-		}
-		return ref, nil
-	} else {
-		ref, err := alltransports.ParseImageName(fmt.Sprintf("containers-storage:%s", target))
-		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/container/client.go
```diff
@@ -344,32 +341,17 @@ func (m RawManifest) Digest() (digest.Digest, error) {
...
-			storage = fmt.Sprintf("[overlay@%s]", storagePath)
-		}
-		ref, err := alltransports.ParseImageName(fmt.Sprintf("%s:%s%s", transport, storage, target.Name()))
-		if err != nil {
-			return nil, err
-		}
-		return ref, nil
-	default:
-		return nil, fmt.Errorf("Unknown containers-transport: %s", transport)
+func getImageRef(target reference.Named, local bool) (types.ImageReference, error) {
```
mvo5, line 344: Fwiw, I'm not a fan of booleans in function args as it's hard to reason about when reading the code on the "call site", i.e. `getImageRef(ref, false)` is less obvious than say `getImageRef(ref, UseLocal)` but #448 uses it this way and we could always cleanup later.

PR #484, opened 2024-02-29
fedora: anaconda cleanups

supakeen:
I saw some of this code while removing the Fedora 37 version gates and code paths and revisited it today. Turns out we don't need any of it anymore to produce working `image-installer`s.


achilleas-k (approved):
Thanks for the cleanup!

PR #479, opened 2024-02-27
Fedora package sets: bump anaconda-webui cutoff to Fedora 41

AdamWill:
The Change to include anaconda-webui was deferred to Fedora 41 by FESCo. This conditional should match that cutoff.

PR #485, opened 2024-02-29
fedora: webui only on rawhide

supakeen:
The webui for Fedora has been delayed until Fedora 41, current rawhide. Let our version gates reflect that.


supakeen:
Speaking of, do we want a `const RAWHIDE` somewhere in `distro/fedora`?

supakeen:
Duplicate of #479.

PR #488, opened 2024-03-02
Allow ISO labels to be predefined instead of templated

supakeen:
Sometimes templating ISO labels is a bit of a chore, they are different between multiple distributions. This PR allows for a new property on installer pipelines to set a non-templated ISO label.

This allows us to move up the generation of the ISO label thus allowing us to override it per image type without being dependent on whatever is already in the format string and the `fmt.Sprintf` call later on.

I also considered using `text/template` here but it was a bit wild. In the future we might want to move all ISO label generation into the specific distro packages and remove the `ISOLabelTmpl` from installer pipelines entirely as it's harder to share.

### DIFF
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -326,7 +326,7 @@ func liveInstallerImage(workload workload.Workload,
 
 	d := t.arch.distro
 
-	img.ISOLabelTempl = d.isolabelTmpl
+	img.ISOLabelTmpl = d.isolabelTmpl
 	img.Product = d.product
 	img.OSName = "fedora"
 	img.OSVersion = d.osVersion
@@ -375,7 +375,7 @@ func imageInstallerImage(workload workload.Workload,
 
 	d := t.arch.distro
 
-	img.ISOLabelTempl = d.isolabelTmpl
+	img.ISOLabelTmpl = d.isolabelTmpl
 	img.Product = d.product
 	img.OSName = "fedora"
 	img.OSVersion = d.osVersion
@@ -537,7 +537,7 @@ func iotInstallerImage(workload workload.Workload,
 
 	img.SquashfsCompression = "lz4"
 
-	img.ISOLabelTempl = d.isolabelTmpl
+	img.ISOLabelTmpl = d.isolabelTmpl
 	img.Product = d.product
 	img.Variant = "IoT"
 	img.OSName = "fedora-iot"
@@ -694,7 +694,7 @@ func iotSimplifiedInstallerImage(workload workload.Workload,
 	}
 
 	d := t.arch.distro
-	img.ISOLabelTempl = d.isolabelTmpl
+	img.ISOLabelTmpl = d.isolabelTmpl
 	img.Product = d.product
 	img.Variant = "iot"
 	img.OSName = "fedora"
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -696,7 +696,7 @@ func iotSimplifiedInstallerImage(workload workload.Workload,
 	d := t.arch.distro
 	img.ISOLabelTmpl = d.isolabelTmpl
 	img.Product = d.product
-	img.Variant = "iot"
+	img.Variant = "IoT"
 	img.OSName = "fedora"
 	img.OSVersion = d.osVersion
 
--- a/pkg/distro/rhel8/images.go
+++ b/pkg/distro/rhel8/images.go
@@ -344,7 +344,7 @@ func imageInstallerImage(workload workload.Workload,
 
 	d := t.arch.distro
 
-	img.ISOLabelTempl = d.isolabelTmpl
+	img.ISOLabelTmpl = d.isolabelTmpl
 	img.Product = d.product
 	img.OSName = "redhat"
 	img.OSVersion = d.osVersion
@@ -461,7 +461,7 @@ func edgeInstallerImage(workload workload.Workload,
 		img.AdditionalAnacondaModules = []string{"org.fedoraproject.Anaconda.Modules.Users"}
 	}
 
-	img.ISOLabelTempl = d.isolabelTmpl
+	img.ISOLabelTmpl = d.isolabelTmpl
 	img.Product = d.product
 	img.Variant = "edge"
 	img.OSName = "rhel"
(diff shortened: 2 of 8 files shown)


Conan-Kudo:
> This also drops the ISO label templates from the Fedora distro as they are no longer being used. I'm open to reverting that last commit as it's a deviation from other distributions.

I would suggest to go further and drop it everywhere. There is no good reason to have this locked in at the code level.

supakeen:
> > This also drops the ISO label templates from the Fedora distro as they are no longer being used. I'm open to reverting that last commit as it's a deviation from other distributions.
> 
> I would suggest to go further and drop it everywhere. There is no good reason to have this locked in at the code level.

It'd still be locked in in code as all our definitions are currently code, however it does move it up so it's much closer to where the distribution gets defined.

I'm open to moving it for RHEL as well but let's see what people think.

AdamWill:
I'm kinda with @Conan-Kudo if it's not too much work; it seems to make more sense to just have one way to do it, and this way seems better on the whole. This split approach feels a bit confusing. At least in theory it raises the risk that somebody might get confused and not set ISOLabel *or* ISOLabelTmpl , in which case I guess we'd wind up with an empty label? It just feels nicer if it's clearer that there's one thing and it's needed...

pkg/distro/fedora/distro.go
```diff
@@ -397,7 +397,6 @@ type distribution struct {
 	releaseVersion     string
 	modulePlatformID   string
 	ostreeRefTmpl      string
-	isolabelTmpl       string
```
thozza, line 400: I'm wondering if it wouldn't be good to make the same changes in RHEL distros, especially for consistency and not diverging in structs used by each distro.

pkg/distro/fedora/images.go
```diff
@@ -375,12 +378,14 @@ func imageInstallerImage(workload workload.Workload,
...
 	d := t.arch.distro
 
-	img.ISOLabelTempl = d.isolabelTmpl
 	img.Product = d.product
 	img.OSName = "fedora"
 	img.OSVersion = d.osVersion
 	img.Release = fmt.Sprintf("%s %s", d.product, d.osVersion)
 
+	// We don't know the variant of the OS pipeline being installed
+	img.ISOLabel = fmt.Sprintf(ISO_LABEL, img.Product, img.OSVersion, "Unknown", img.Platform.GetArch())
```
thozza, line 387: I'm wondering if it would not make sense to instead set the `img.Variant = "Unknown"` and then construct `ISOLabel` in the same way as in other image types.

thozza (dismissed):
I added a few comments, but in general the changes look good.

pkg/distro/fedora/distro.go
supakeen: I can make the same change in RHEL, just wanted to make sure before I touched RHEL :)

pkg/distro/fedora/images.go
supakeen: Agreed, better than the above.

supakeen: Added a commit that resolves this.

pkg/distro/fedora/distro.go
supakeen: Do we want to do RHEL in this PR or in a separate PR?

thozza: Up to you. I don't have any strong preference, as long as it is eventually done.

thozza (approved):

achilleas-k:
> At least in theory it raises the risk that somebody might get confused and not set ISOLabel _or_ ISOLabelTmpl , in which case I guess we'd wind up with an empty label?

The way it's done now, this can only happen if someone modifies the code to drop the label template and the label on the image type.

achilleas-k:
> > This also drops the ISO label templates from the Fedora distro as they are no longer being used. I'm open to reverting that last commit as it's a deviation from other distributions.
> 
> I would suggest to go further and drop it everywhere. There is no good reason to have this locked in at the code level.

It's still in the code.  None of these can be set externally right now.

achilleas-k (approved):

AdamWill:
> > At least in theory it raises the risk that somebody might get confused and not set ISOLabel _or_ ISOLabelTmpl , in which case I guess we'd wind up with an empty label?
> 
> The way it's done now, this can only happen if someone modifies the code to drop the label template and the label on the image type.

I meant if someone was, for instance, making a *new* image type.

supakeen:
> > > At least in theory it raises the risk that somebody might get confused and not set ISOLabel _or_ ISOLabelTmpl , in which case I guess we'd wind up with an empty label?
> > 
> > 
> > The way it's done now, this can only happen if someone modifies the code to drop the label template and the label on the image type.
> 
> I meant if someone was, for instance, making a _new_ image type. Since none of this can be changed outside of code, the only way to customize it is to add new code, so it's important for the implementation to at least be as clean and straightforward as possible. Even just to _understand_ how images are built and what's in them, the only way is to read this code.
> 
> Three people said "shouldn't we just do RHEL the same as Fedora?" but this still got merged with the divergent approach in place...seems weird.

In one of the comments I mentioned I asked if RHEL can be done in a separate followup. I didn't specify *why*. The *why* is because in the RHEL distro(s) we have logic switching between RHEL/CentOS names and currently no predefined fields. I've created a separate issue to track this: https://github.com/osbuild/images/issues/491

achilleas-k:
> > > At least in theory it raises the risk that somebody might get confused and not set ISOLabel _or_ ISOLabelTmpl , in which case I guess we'd wind up with an empty label?
> > 
> > 
> > The way it's done now, this can only happen if someone modifies the code to drop the label template and the label on the image type.
> 
> I meant if someone was, for instance, making a _new_ image type. Since none of this can be changed outside of code, the only way to customize it is to add new code, so it's important for the implementation to at least be as clean and straightforward as possible. Even just to _understand_ how images are built and what's in them, the only way is to read this code.
> 

That's fair and I didn't mean to dismiss your concern.

> Three people said "shouldn't we just do RHEL the same as Fedora?" but this still got merged with the divergent approach in place...seems weird.

With some of these details and the ImageType interface implementation in general, there's generally divergence between Fedora and EL for various reasons.  Sometimes it's necessary (they behave differently or require different configurations at different levels of the pipeline), other times it's just a matter of time or urgency to fix something where it's needed then unify implementations for easier maintenance.  I considered this an instance of the latter (though perhaps not very urgent) in that we can fix the bug in an isolated way, verify that the images changed in the way we expect them to, then refactor separately to make implementations consistent without affecting any definitions.

PR #492, opened 2024-03-04
Change filename for the minimal-raw image type in all distros

achilleas-k:
Fedora's Koji requires specific file extensions for images and `img.xz` is not supported (see [1,2]).  Raw compressed images in Fedora are expected to have the extension `.raw.xz`.
Rename the file and update the filenames in RHEL for the equivalent image type for consistency.

This is on request and approved by the image owners (@nullr0ute).

[1] https://koji.fedoraproject.org/koji/taskinfo?taskID=114454340
[2] https://kojipkgs.fedoraproject.org//work/tasks/4340/114454340/koji-import.log.json


supakeen (approved):
Change makes sense, wonder if the docs commit should be separate.

nullr0ute:
LGTM!

PR #502, opened 2024-03-05
test: run oscap checks on images with oscap customizations (HMS-3710)

achilleas-k:
Add oscap test configs for RHEL 8 and RHEL 9 with the appropriate datastream option enabled.
These replicate the test configs from the oscap test in osbuild-composer [1].

To fully replace the test in osbuild-composer, this PR also adds support for running the boot test script with the config file as argument.  When an oscap config is detected, it runs some extra checks on the score of the system.

_NOTE: This is a bit quick and dirty but should soon be replaced with a more comprehensive test of all customizations on images during boot_

[1] https://github.com/osbuild/osbuild-composer/blob/073e304978acade8bfa059f00a005402aa037c99/test/cases/oscap.sh

### DIFF
--- a/test/configs/oscap-rhel8.json
+++ b/test/configs/oscap-rhel8.json
@@ -1,0 +1,16 @@
+{
+  "name": "oscap-rhel8",
+  "blueprint": {
+    "customizations": {
+      "openscap": {
+        "profile_id": "xccdf_org.ssgproject.content_profile_cis",
+        "datastream": "/usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml",
+        "tailoring": {
+          "unselected": [
+            "grub2_password"
+          ]
+        }
+      }
+    }
+  }
+}
--- a/test/configs/oscap-rhel8.json
+++ b/test/configs/oscap-rhel8.json
@@ -1,6 +1,17 @@
 {
   "name": "oscap-rhel8",
   "blueprint": {
+    "packages": [
+      {
+        "name": "xmlstarlet"
+      },
+      {
+        "name": "openscap-utils"
+      },
+      {
+        "name": "jq"
+      }
+    ],
     "customizations": {
       "openscap": {
         "profile_id": "xccdf_org.ssgproject.content_profile_cis",
--- a/cmd/boot-aws/main.go
+++ b/cmd/boot-aws/main.go
@@ -358,7 +358,7 @@ func teardown(cmd *cobra.Command, args []string) {
 	fnerr = doTeardown(a, res)
 }
 
-func doRunExec(a *awscloud.AWS, filename string, flags *pflag.FlagSet, res *resources) error {
+func doRunExec(a *awscloud.AWS, command []string, flags *pflag.FlagSet, res *resources) error {
 	privKey, err := flags.GetString("ssh-privkey")
 	if err != nil {
 		return err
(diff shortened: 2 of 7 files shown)


achilleas-k:
RHEL 8.10 boot test failing with:
```
systemd-remount-fs[563]: mount: /dev/shm: mount point not mounted or bad option.
```

achilleas-k:
Comparing the oscap version of the image with the base one, `/etc/fstab` gets an extra line:
```
tmpfs /dev/shm tmpfs defaults,nosuid,relatime,mode=755,inode64,nodev,noexec 0 0
```
which I assume is what causes the issue.  Is this a bug in the oscap profile, adding an fstab option that the OS doesn't support, or did I configure it wrong?
I copied the blueprint from the test in osbuild-composer, but we don't check for the system status there so this might have always been happening.

@kingsleyzissou do you know anything about this (from your oscap experience)?

test/configs/oscap-rhel8.json
```diff
@@ -1,6 +1,20 @@
 {
   "name": "oscap-rhel8",
   "blueprint": {
+    "packages": [
+      {
+        "name": "scap-security-guide"
```
kingsleyzissou, line 6: So I have a couple of thoughts on this. But this is probably just a nitpick at the end of the day.
If the `oscap` customization is specified, this package should be installed to the image by default.
So, maybe we want to leave this package out and the test will fail if it somehow doesn't get added.
See:
https://osbuild.org/docs/user-guide/oscap-remediation#openscap

kingsleyzissou:
> which I assume is what causes the issue. Is this a bug in the oscap profile, adding an fstab option that the OS doesn't support, or did I configure it wrong?
So this is technically a required mountpoint in the `rhel-8` datastream for the `cis` profile. So maybe it's adding this as a remediation and breaking the build. It might be an outdated datastream, because I think they may have removed this. I'm double checking

kingsleyzissou:
I have a strong suspicion that we are using an outdated datastream file from an older `scap-security-guides` package.

test/configs/oscap-rhel8.json
achilleas-k: Thanks.  I'll drop it!

achilleas-k:
> I have a strong suspicion that we are using an outdated datastream file from an older `scap-security-guides` package.

Interesting.  Could it be an 8.10 issue since it's not released yet?  I'm playing with the fstab on a live system and trying to get it to work but even with no options (just defaults) `systemd-remount-fs.service` fails.
Manually mounting it works only if I remove the `inode64` option.

Not sure if any of this helps.

test/configs/oscap-rhel8.json
achilleas-k: Dropped.

kingsleyzissou:
> Interesting. Could it be an 8.10 issue since it's not released yet?

It's just a hunch. But up until last year the blueprints generated by the oscap tool were also trying to configure filesystem options for `/dev/shm` which was causing builds to fail. I think they revised the datastream since then and that has been removed. 

But I haven't seen an entry being added to the `/etc/fstab` before. That's interesting. We can maybe reach out to the openscap team about this @evgenyz

achilleas-k:
Disabling the test on RHEL 8.10 for now.

kingsleyzissou (approved):
🙏

evgenyz:
That's very strange. I don't think we force `inode64` anywhere. As to the record itself: it should not be a problem, as `/etc/fstab` is the recommended way (acc. to systemd docs) for mount units options customization.

ondrejbudai (approved):

kingsleyzissou:
@evgenyz it's really weird, if I disabled `partition_for_dev_shm` things work and no entry is added into the `/etc/fstab`.
But when I check the remediation logs:
`FIX FOR THIS RULE 'xccdf_org.ssgproject.content_rule_partition_for_dev_shm' IS MISSING!`

PR #504, opened 2024-03-07
bootc_disk: Add support for vmdk, ova

cgwalters:
BIB PR using this: https://github.com/osbuild/bootc-image-builder/pull/251

---

We want to support generating VMDK for bootc containers too.

### DIFF
--- a/pkg/image/bootc_disk.go
+++ b/pkg/image/bootc_disk.go
@@ -3,10 +3,13 @@ package image
 import (
 	"fmt"
 	"math/rand"
+	"path/filepath"
+	"strings"
 
 	"github.com/osbuild/images/pkg/artifact"
 	"github.com/osbuild/images/pkg/container"
 	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/platform"
 	"github.com/osbuild/images/pkg/runner"
 )
@@ -47,15 +50,38 @@ func (img *BootcDiskImage) InstantiateManifestFromContainers(m *manifest.Manifes
 		panic(fmt.Sprintf("no compression is allowed with %q format for %q", imgFormat, img.name))
 	}
 
+	// In the bootc flow, we reuse the host container context for tools;
+	// this is signified by passing nil to the below pipelines.
+	var hostPipeline manifest.Build
+
 	opts := &baseRawOstreeImageOpts{useBootupd: true}
 	baseImage := baseRawOstreeImage(img.OSTreeDiskImage, buildPipeline, opts)
 	switch imgFormat {
 	case platform.FORMAT_QCOW2:
-		// qcow2 runs without a build pipeline directly from "bib"
-		qcow2Pipeline := manifest.NewQCOW2(nil, baseImage)
+		qcow2Pipeline := manifest.NewQCOW2(hostPipeline, baseImage)
 		qcow2Pipeline.Compat = img.Platform.GetQCOW2Compat()
 		qcow2Pipeline.SetFilename(img.Filename)
 		return qcow2Pipeline.Export(), nil
+	// TODO: refactor to share this with disk.go; note here the build pipeline runs
+	// on the host (that's the nil)
+	case platform.FORMAT_VMDK:
+		vmdkPipeline := manifest.NewVMDK(hostPipeline, baseImage)
+		vmdkPipeline.SetFilename(img.Filename)
+		return vmdkPipeline.Export(), nil
+	case platform.FORMAT_OVA:
+		vmdkPipeline := manifest.NewVMDK(hostPipeline, baseImage)
+		ovfPipeline := manifest.NewOVF(hostPipeline, vmdkPipeline)
+		tarPipeline := manifest.NewTar(hostPipeline, ovfPipeline, "archive")
+		tarPipeline.Format = osbuild.TarArchiveFormatUstar
+		tarPipeline.SetFilename(img.Filename)
+		extLess := strings.TrimSuffix(img.Filename, filepath.Ext(img.Filename))
+		// The .ovf descriptor needs to be the first file in the archive
+		tarPipeline.Paths = []string{
+			fmt.Sprintf("%s.ovf", extLess),
+			fmt.Sprintf("%s.mf", extLess),
+			fmt.Sprintf("%s.vmdk", extLess),
+		}
+		return tarPipeline.Export(), nil
 	}
 
 	switch img.Compression {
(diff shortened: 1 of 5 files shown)


cgwalters:
So...I tried this out along with this patch for bib

```
commit b9431b0f058a4bfbe1a60f394708a559978d0cfc
Author:     Colin Walters <walters@verbum.org>
AuthorDate: Wed Mar 6 18:04:21 2024 -0500
Commit:     Colin Walters <walters@verbum.org>
CommitDate: Thu Mar 7 08:55:23 2024 -0500

    wip

diff --git a/bib/cmd/bootc-image-builder/image.go b/bib/cmd/bootc-image-builder/image.go
index d80c38a..de9d9b4 100644
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -48,7 +48,7 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
 	rng := createRand()
 
 	switch c.ImgType {
-	case "ami", "qcow2", "raw":
+	case "ami", "qcow2", "raw", "vmdk":
 		return manifestForDiskImage(c, rng)
 	case "anaconda-iso", "iso":
 		return manifestForISO(c, rng)
@@ -95,6 +95,9 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 	case "ami", "raw":
 		imageFormat = platform.FORMAT_RAW
 		filename = "disk.raw"
+	case "vmdk":
+		imageFormat = platform.FORMAT_VMDK
+		filename = "disk.vmdk"
 	}
 
 	switch c.Architecture {
diff --git a/bib/cmd/bootc-image-builder/main.go b/bib/cmd/bootc-image-builder/main.go
index 9423e48..3de343f 100644
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -280,12 +280,12 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	switch imgType {
 	case "qcow2":
 		exports = []string{"qcow2"}
-	case "ami", "raw":
+	case "ami", "raw", "vmdk":
 		exports = []string{"image"}
 	case "anaconda-iso", "iso":
 		exports = []string{"bootiso"}
 	default:
-		return fmt.Errorf("valid types are 'qcow2', 'ami', 'raw', 'anaconda-iso', not: '%s'", imgType)
+		return fmt.Errorf("valid types are 'qcow2', 'ami', 'raw', 'vmdk', 'anaconda-iso', not: '%s'", imgType)
 	}
 
 	manifestPath := filepath.Join(outputDir, manifest_fname)
```

and...I see `vmdk` in the manifest, yet I still get a raw disk image for output :confused:

pkg/image/bootc_disk.go
```diff
@@ -56,6 +56,10 @@ func (img *BootcDiskImage) InstantiateManifestFromContainers(m *manifest.Manifes
 		qcow2Pipeline.Compat = img.Platform.GetQCOW2Compat()
 		qcow2Pipeline.SetFilename(img.Filename)
 		return qcow2Pipeline.Export(), nil
+	case platform.FORMAT_VMDK:
+		vmdkPipeline := manifest.NewVMDK(buildPipeline, baseImage)
```
mvo5, line 60: This is nice! I would love to see a minimal test, maybe something like the attached diff?
[0001-image-add-minimal-test-for-vmdk.patch.txt](https://github.com/osbuild/images/files/14526554/0001-image-add-minimal-test-for-vmdk.patch.txt)

Ideally it would check a bit more but it might be a reasonable baseline

cgwalters: Thanks, I've rolled that in!

cgwalters:
OK, if the remaining work is bib side and you think this is right - then lifting draft here.

pkg/image/bootc_disk.go
```diff
@@ -56,6 +59,25 @@ func (img *BootcDiskImage) InstantiateManifestFromContainers(m *manifest.Manifes
...
 		return qcow2Pipeline.Export(), nil
+	// TODO: refactor to share this with disk.go
+	case platform.FORMAT_VMDK:
+		vmdkPipeline := manifest.NewVMDK(buildPipeline, baseImage)
+		vmdkPipeline.SetFilename(img.Filename)
+		return vmdkPipeline.Export(), nil
+	case platform.FORMAT_OVA:
+		vmdkPipeline := manifest.NewVMDK(buildPipeline, baseImage)
+		ovfPipeline := manifest.NewOVF(buildPipeline, vmdkPipeline)
+		tarPipeline := manifest.NewTar(buildPipeline, ovfPipeline, "archive")
```
achilleas-k, line 70: Specifying `nil` instead of the `buildPipeline` as a first argument to each `manifest.New...()` makes them run outside the build pipeline, so on the host.  See above:
```go
		// qcow2 runs without a build pipeline directly from "bib"
		qcow2Pipeline := manifest.NewQCOW2(nil, baseImage)
```

achilleas-k:
> * Add that into bib (edit: it's already there! thankfully), and rework the build pipeline to use the container as buildroot (I have no idea about doing this)

See my in-line comment https://github.com/osbuild/images/pull/504/files#r1516881294

achilleas-k (dismissed):
Make pipelines run outside the build root.

achilleas-k:
> OK, if the remaining work is bib side and you think this is right - then lifting draft here.

Sorry for the confusion.  I looked at the code before noticing the comment and it looked good but then realised the issue with the build root :)

pkg/image/bootc_disk.go
cgwalters: I get a Go nil pointer exception when I try that with `manifest.NewVMDK` because it always calls `buildPipeline.addDependent(p)`.  Though the docs...ah, I see now that I look at the equivalent code in `NewQCOW2`; adjusting.

cgwalters:
OK yep updated, this generates an image!  No idea if it works in vsphere yet, will ping about that.

pkg/image/bootc_disk.go
```diff
@@ -47,15 +50,38 @@ func (img *BootcDiskImage) InstantiateManifestFromContainers(m *manifest.Manifes
...
 	baseImage := baseRawOstreeImage(img.OSTreeDiskImage, buildPipeline, opts)
 	switch imgFormat {
 	case platform.FORMAT_QCOW2:
-		// qcow2 runs without a build pipeline directly from "bib"
-		qcow2Pipeline := manifest.NewQCOW2(nil, baseImage)
+		qcow2Pipeline := manifest.NewQCOW2(hostPipeline, baseImage)
 		qcow2Pipeline.Compat = img.Platform.GetQCOW2Compat()
 		qcow2Pipeline.SetFilename(img.Filename)
 		return qcow2Pipeline.Export(), nil
+	// TODO: refactor to share this with disk.go; note here the build pipeline runs
```
mvo5, line 65: Thanks for this TODO

mvo5 (approved):
Thank you! Let's land it as is and we (images team) will look into the suggested refactor (which makes a ton of sense) in a followup

PR #505, opened 2024-03-07
image: always set filename for bootc raw pipeline

achilleas-k:
In BIB, we intend to export multiple images from the same pipeline and this is the expected filename for the raw images.  Set it so that it's always disk.raw even when we're building a qcow2 or other image type.


achilleas-k:
The comment (and commit message) is a bit wrong.  We're not exporting from the same pipeline but from the same manifest.  It's probably fine though.

cgwalters (approved):

PR #507, opened 2024-03-11
Support user-defined entries for sudo NOPASSWD

achilleas-k:
Replace the wheel-sudo-nopasswd (bool) installer customization with sudo-nopasswd ([]string).  Users can now specify a list of users or groups (groups must be prefixed with %).  For each element in the array, the kickstart file will create a file that enables sudo with NOPASSWD for that entry.  Each entry is added as a separate file in the sudoers.d drop-in directory.

### DIFF
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -3,6 +3,8 @@ package manifest
 import (
 	"fmt"
 	"path"
+	"sort"
+	"strings"
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/container"
@@ -32,8 +34,9 @@ type AnacondaInstallerISOTree struct {
 	Keyboard *string
 	Timezone *string
 
-	// Create a sudoers drop-in file for wheel group with NOPASSWD option
-	WheelNoPasswd bool
+	// Create a sudoers drop-in file for each user or group to enable the
+	// NOPASSWD option
+	NoPasswd []string
 
 	// Add kickstart options to make the installation fully unattended
 	UnattendedKickstart bool
@@ -532,16 +535,10 @@ func (p *AnacondaInstallerISOTree) makeKickstartStages(kickstartOptions *osbuild
 
 	stages = append(stages, osbuild.NewKickstartStage(kickstartOptions))
 
-	if p.WheelNoPasswd {
+	hardcodedKickstartBits := makeKickstartSudoersPost(p.NoPasswd)
+	if hardcodedKickstartBits != "" {
 		// Because osbuild core only supports a subset of options,
 		// we append to the base here with hardcoded wheel group with NOPASSWD option
-		hardcodedKickstartBits := `
-%post
-echo -e "%wheel\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/wheel"
-chmod 0440 /etc/sudoers.d/wheel
-restorecon -rvF /etc/sudoers.d
-%end
-`
 		kickstartFile, err := kickstartOptions.IncludeRaw(hardcodedKickstartBits)
 		if err != nil {
 			panic(err)
@@ -559,4 +556,32 @@ restorecon -rvF /etc/sudoers.d
 func makeISORootPath(p string) string {
 	fullpath := path.Join("/run/install/repo", p)
 	return fmt.Sprintf("file://%s", fullpath)
+}
+
+func makeKickstartSudoersPost(names []string) string {
+	if len(names) == 0 {
+		return ""
+	}
+	echoLineFmt := `echo -e "%[1]s\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/%[1]s"
+chmod 0440 /etc/sudoers.d/%[1]s`
+
+	filenames := make(map[string]bool)
+	sort.Strings(names)
+	entries := make([]string, 0, len(names))
+	for _, name := range names {
+		if filenames[name] {
+			continue
+		}
+		entries = append(entries, fmt.Sprintf(echoLineFmt, name))
+		filenames[name] = true
+	}
+
+	kickstartSudoersPost := `
+%%post
+%s
+restorecon -rvF /etc/sudoers.d
+%%end
+`
+	return fmt.Sprintf(kickstartSudoersPost, strings.Join(entries, "\n"))
+
 }
--- a/pkg/blueprint/installer_customizations.go
+++ b/pkg/blueprint/installer_customizations.go
@@ -1,6 +1,6 @@
 package blueprint
 
 type InstallerCustomization struct {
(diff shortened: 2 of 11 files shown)


bcl (approved):
Looks good to me

supakeen (approved):
Neat :)

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -560,3 +556,28 @@ func makeISORootPath(p string) string {
 	fullpath := path.Join("/run/install/repo", p)
 	return fmt.Sprintf("file://%s", fullpath)
 }
+
+func makeKickstartSudoersPost(names []string) string {
```
mvo5, line 560: Having a test for this would be nice, maybe something like?
```go
func TestMakeKickstartSudoersPostEmpty(t *testing.T) {
	assert.Equal(t, "", makeKickstartSudoersPost(nil))
}

func TestMakeKickstartSudoersPost(t *testing.T) {
	assert.Equal(t, "...", makeKickstartSudoersPost([]string{"foo", "foo", "bar"}))
}
```
which makes me wonder if makeKickstartSudoersPost() for the empty list should just return "" and wit hthat I wonder if we should do something like:
```go
diff --git a/pkg/manifest/anaconda_installer_iso_tree.go b/pkg/manifest/anaconda_installer_iso_tree.go
index ba0ae9db3..ae59a9936 100644
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -534,10 +534,10 @@ func (p *AnacondaInstallerISOTree) makeKickstartStages(kickstartOptions *osbuild
 
        stages = append(stages, osbuild.NewKickstartStage(kickstartOptions))
 
-       if len(p.NoPasswd) > 0 {
+       hardcodedKickstartBits := makeKickstartSudoersPost(p.NoPasswd)
+       if hardcodedKickstartBits != "" {
                // Because osbuild core only supports a subset of options,
                // we append to the base here with hardcoded wheel group with NOPASSWD option
-               hardcodedKickstartBits := makeKickstartSudoersPost(p.NoPasswd)
                kickstartFile, err := kickstartOptions.IncludeRaw(hardcodedKickstartBits)
                if err != nil {
                        panic(err)

```
but that is mostly idle wondering - I still think a test for the helper is easy and nice :)

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -532,16 +534,10 @@ func (p *AnacondaInstallerISOTree) makeKickstartStages(kickstartOptions *osbuild
 
 	stages = append(stages, osbuild.NewKickstartStage(kickstartOptions))
 
-	if p.WheelNoPasswd {
+	if len(p.NoPasswd) > 0 {
 		// Because osbuild core only supports a subset of options,
 		// we append to the base here with hardcoded wheel group with NOPASSWD option
```
mvo5, line 539: (super nitpick) The comment is now slightly outdated

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -560,3 +556,28 @@ func makeISORootPath(p string) string {
 	fullpath := path.Join("/run/install/repo", p)
 	return fmt.Sprintf("file://%s", fullpath)
 }
+
+func makeKickstartSudoersPost(names []string) string {
+	echoLineFmt := `echo -e "%[1]s\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/%[1]s"
+chmod 0440 /etc/sudoers.d/%[1]s`
+
+	filenames := make(map[string]bool)
```
mvo5, line 564: (sooooory super nitpick) this is totally fine as it is, while reading it I was wondering:
```
sort.Strings(names)
names = clies.Compact(names)
```
might be a simpler way to remove duplicates from the inputs than the map but I guess it's really fine either way :)

mvo5 (approved):
I like this a lot, thank you! This is fine as is, I would love a test and have more ideas/suggestions inline but really more about me thinking out loud than that anything needs to be changed :)

pkg/manifest/anaconda_installer_iso_tree.go
achilleas-k: I'm adding a line to the existing test that checks for the expected inline ID, which is a hash of the content.

pkg/manifest/anaconda_installer_iso_tree.go
achilleas-k: If we weren't already looping through the elements, this would be nice, but since we are, it feels simpler to just keep track.  But I don't mind either way.

pkg/manifest/anaconda_installer_iso_tree.go
mvo5: Maybe I'm missing something, I would love to see a test for the generated kickstart file content but maybe that is covered already and I don't see it (sorry!). As an experiment I did:
```go
diff --git a/pkg/manifest/anaconda_installer_iso_tree.go b/pkg/manifest/anaconda_installer_iso_tree.go
index ba0ae9db3..9189dc72e 100644
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -558,6 +558,8 @@ func makeISORootPath(p string) string {
 }
 
 func makeKickstartSudoersPost(names []string) string {
+       return ""
+
        echoLineFmt := `echo -e "%[1]s\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/%[1]s"
 chmod 0440 /etc/sudoers.d/%[1]s`
 
```
and no (unit) test fails afaict. 

But again, not a blocker, it's what I would love to see and it's fine if I don't see it :)

pkg/manifest/anaconda_installer_iso_tree.go
mvo5: totally fine

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -560,3 +557,27 @@ func makeISORootPath(p string) string {
...
 }
+
+func makeKickstartSudoersPost(names []string) string {
+	if len(names) == 0 {
+		return ""
+	}
+	echoLineFmt := `echo -e "%[1]s\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/%[1]s"
+chmod 0440 /etc/sudoers.d/%[1]s`
+
+	slices.Sort(names)
```
bcl, line 568: I really like this, but I don't think we can use it yet, can we? It was added in got 1.21 according to the docs and AFAIK we need to stuck with 1.19 features for now.

achilleas-k: Good catch.  Fixed.  I should add some unit test runs in CI for this.

mvo5 (approved):
Thank you!

PR #506, opened 2024-03-08
Disk image compress pipelines refactor

mvo5:
Draft as a followup for https://github.com/osbuild/images/pull/504 - as a starting point for a discussion how to avoid duplicating the compression and image stages, they are extremely similar right now but also subtly different (like edge images set no ImageFormat right now). The individual commits also need some splitting and more tests

PR #508, opened 2024-03-11
platform: fix panic() when trying to print an unset ImageFormat

mvo5:
During the development of the image pipeline refactor I noticed that the platform code will panic for unset image types. This is a bug and I added a small test to ensure it won't happen again.

[0] https://github.com/osbuild/images/pull/506

### DIFF
--- a/pkg/platform/platform.go
+++ b/pkg/platform/platform.go
@@ -1,6 +1,8 @@
 package platform
 
 import (
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


supakeen:
Do you mean that a call to `panic` is considered a bug or the message not including any details?

supakeen (approved):
Then I understand :)

PR #510, opened 2024-03-11
github: run unit tests on CentOS Stream 9 and 8

achilleas-k:
Run unit tests on CentOS Stream 9 and 8.  This will catch scenarios where we use Go features or libraries not supported by those distributions.

I intend to mark these as required in the repo config when this PR is merged.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,6 +43,76 @@ jobs:
       - name: Send coverage to codecov.io
         run: bash <(curl -s https://codecov.io/bash)
 
+  unit-tests-c9s:
+    name: "🛃 Unit tests (CentOS Stream 9)"
+    runs-on: ubuntu-20.04
+    container:
+      image: quay.io/centos/centos:stream9
+    env:
+      GOFLAGS: "-tags=exclude_graphdriver_btrfs"
+
+    steps:
+      - name: Install dnf plugins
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -43,6 +43,74 @@ jobs:
...
+      - name: Mark the working directory as safe for git
+        run: git config --global --add safe.directory "$(pwd)"
+
+      - name: Run unit tests
+        run: go test -race -covermode=atomic -coverprofile=coverage.txt -coverpkg=$(go list ./... | grep -v rpmmd/test$ | tr "\n" ",") ./...
+
+      - name: Run depsolver tests with force-dnf to make sure it's not skipped for any reason
+        run: go test -race ./pkg/dnfjson/... -force-dnf
+
+      - name: Send coverage to codecov.io
```
supakeen, line 111: Do we still use `codecov.io`?

achilleas-k: https://app.codecov.io/gh/osbuild/images exists but it hasn't been updated since the split.  I'll drop it and we can discuss if we want a coverage report later.

supakeen (approved):

achilleas-k:
Had to fix the env setting in the config yaml and the CS8 repo is called `powertools`, not `crb`.

supakeen (approved):

PR #512, opened 2024-03-12
Generate the ALL-MANIFEST for bootc disk images

achilleas-k:
The bootc-disk image manifest now always contains all pipelines needed to generate all image types.  This way we can build any combination of disk images depending on what we need to export without needing to conditionally add or remove pipelines.

The Platform no longer needs to be read, since we're generating all of the formats and the architecture depends on the source container.  However, the Platform is still required for specifying the boot mode (enabling or disabling BIOS boot).

The first return value of the InstantiateManifestFromContainers() function, the artifact, was not used and has been removed now.

### DIFF
--- a/pkg/image/bootc_disk.go
+++ b/pkg/image/bootc_disk.go
@@ -3,14 +3,10 @@ package image
 import (
 	"fmt"
 	"math/rand"
-	"path/filepath"
-	"strings"
 
-	"github.com/osbuild/images/pkg/artifact"
 	"github.com/osbuild/images/pkg/container"
 	"github.com/osbuild/images/pkg/manifest"
 	"github.com/osbuild/images/pkg/osbuild"
-	"github.com/osbuild/images/pkg/platform"
 	"github.com/osbuild/images/pkg/runner"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/image/bootc_disk.go
```diff
@@ -35,56 +31,41 @@ func NewBootcDiskImage(container container.SourceSpec) *BootcDiskImage {
 func (img *BootcDiskImage) InstantiateManifestFromContainers(m *manifest.Manifest,
 	containers []container.SourceSpec,
 	runner runner.Runner,
-	rng *rand.Rand) (*artifact.Artifact, error) {
```
ondrejbudai, line 38: Interestingly, the `error` is always `nil`. I suggest keeping it as it is, because we want to use less panics in favor of errors, so I'm fairly sure we will have some errors to return in the future. :)

ondrejbudai (commented):
This is amazing, THANK YOU!

(needs a linter fix)

mvo5 (approved):
Thank you!

kingsleyzissou (approved):
Neat!

PR #517, opened 2024-03-13
Install minimal depenencies for the check-build-coverage job

achilleas-k:
The check-build-coverage script only needs python3 and the aws cli.
Speed up the job by avoiding installing all the dependencies.


achilleas-k:
@kingsleyzissou thanks for the review but I see now that we're actually compiling and running `./cmd/list-images`, which needs most of the dependencies.  It basically just doesn't need osbuild but I don't think we're going to get the speed-up from this that I thought we would.  Closing.

PR #519, opened 2024-03-14
Fedora: add partition offsets to IoT images

pcdubs:
Add the same offsets used for the Fedora Minimal images to increase IoT hardware support.


bcl (approved):
Looks ok to me, but could you update the commit message to either say the same thing as commit f04bae57ae6f597b1a2e3d381eacd8a5a4141362 or point to it? I wasn't sure why such a large offset was needed without looking.

supakeen (approved):

nullr0ute (approved):
LGTM

Issue #520, opened 2024-03-14
Fedora Minimal fails to boot on Raspberry Pi


supakeen:
Downstream reported an issue with osbuild produced images in Fedora koji being unbootable for the Fedora Minimal images. This is due to not copying the appropriate files into `/boot` like we do for IoT.

supakeen:
Merged in #521.

PR #521, opened 2024-03-14
manifest/raw: apply BootFiles

supakeen:
BootFiles need to be applied on the raw images as well. See #520, this is blocking Fedora.

This is duplicated code from `manifest/raw_ostree`, let's apply it to all images in a followup and properly abstract it out :)

### DIFF
--- a/pkg/manifest/raw.go
+++ b/pkg/manifest/raw.go
@@ -1,6 +1,8 @@
 package manifest
 
 import (
+	"fmt"
+
 	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/artifact"
 	"github.com/osbuild/images/pkg/osbuild"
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):

pkg/manifest/raw.go
```diff
@@ -59,6 +61,38 @@ func (p *RawImage) serialize() osbuild.Pipeline {
...
+		var fsRootMntName string
+		for _, mnt := range bootCopyMounts {
+			if mnt.Target == "/" {
+				fsRootMntName = mnt.Name
+				break
+			}
+		}
+
+		if fsRootMntName == "" {
+			panic("no mount found for the filesystem root")
```
bcl, line 83: Shouldn't panic in a library.

supakeen: Yea, this was a quick copy and paste to unblock (from `raw_ostree`), we'll be removing all of these panics.

PR #528, opened 2024-03-19
Add /usr/local/{bin,sbin} to CustomFilesPolicies

lzap:
This pull request adds the paths /usr/local/bin and /var/local/bin to the CustomFilesPolicies in the policies.go file. This allows the specified directories to be accessed according to the defined policies.

### DIFF
--- a/pkg/policies/policies.go
+++ b/pkg/policies/policies.go
@@ -42,13 +42,15 @@ var CustomDirectoriesPolicies = pathpolicy.NewPathPolicies(map[string]pathpolicy
 
 // CustomFilesPolicies is a set of default policies for custom files
 var CustomFilesPolicies = pathpolicy.NewPathPolicies(map[string]pathpolicy.PathPolicy{
-	"/":           {Deny: true},
-	"/etc":        {},
-	"/root":       {},
-	"/etc/fstab":  {Deny: true},
-	"/etc/shadow": {Deny: true},
-	"/etc/passwd": {Deny: true},
-	"/etc/group":  {Deny: true},
+	"/":               {Deny: true},
+	"/etc":            {},
+	"/root":           {},
+	"/usr/local/bin":  {},
+	"/usr/local/sbin": {},
+	"/etc/fstab":      {Deny: true},
+	"/etc/shadow":     {Deny: true},
+	"/etc/passwd":     {Deny: true},
+	"/etc/group":      {Deny: true},
 })
 
 // MountpointPolicies for ostree


pkg/policies/policies.go
```diff
@@ -42,13 +42,15 @@ var CustomDirectoriesPolicies = pathpolicy.NewPathPolicies(map[string]pathpolicy
...
-	"/root":       {},
-	"/etc/fstab":  {Deny: true},
-	"/etc/shadow": {Deny: true},
-	"/etc/passwd": {Deny: true},
-	"/etc/group":  {Deny: true},
+	"/":              {Deny: true},
+	"/etc":           {},
+	"/root":          {},
+	"/usr/local/bin": {},
+	"/var/local/bin": {},
```
thozza, line 49: I know that I initially agreed with this path, but checking the FHS now, it does not seem that having `bin` under `/var/local` is standardized. Can you please provide more reasoning behind adding this path?

pkg/policies/policies.go
```diff
@@ -42,13 +42,15 @@ var CustomDirectoriesPolicies = pathpolicy.NewPathPolicies(map[string]pathpolicy
...
-	"/etc":        {},
-	"/root":       {},
-	"/etc/fstab":  {Deny: true},
-	"/etc/shadow": {Deny: true},
-	"/etc/passwd": {Deny: true},
-	"/etc/group":  {Deny: true},
+	"/":              {Deny: true},
+	"/etc":           {},
+	"/root":          {},
+	"/usr/local/bin": {},
```
thozza, line 48: Would it make sense to also add `/usr/local/sbin`?

pkg/policies/policies.go
lzap: Sure, we are working on a new feature for the hosted image builder service - firstboot support. Users will be able to add an executable (a script, ansible or python script, anything with shebang) that will be executed during firstboot.

Now, in order to do that, we need to drop two files - a systemd unit and an executable file. We need to put the executable file somewhere. It cannot be `/etc`, however, because files there have `exec_t` SELinux context and there is a transition from `systemd_unit_t` leading to SELinux denials.

Now, to your point, `/var/local/bin` is a typo. Indeed `/var` is meant for data, this was a mistake. Let me fix this.

lzap:
How about this:

* /usr/local/(s)bin - I think we all agree this is a good place for custom scripts
* /var/tmp - if user needs to add some (temporary) data for the script there is no good place at the moment (/etc or /root are not good) so maybe a non-volatile temporary dir is fine as a "workaround"

thozza:
> How about this:
> 
> * /usr/local/(s)bin - I think we all agree this is a good place for custom scripts

yes

> * /var/tmp - if user needs to add some (temporary) data for the script there is no good place at the moment (/etc or /root are not good) so maybe a non-volatile temporary dir is fine as a "workaround"

I thought about it more and I disagree with this use for multiple reasons:

- Having a script in `/usr/local/(s)bin`, which depends on file in `/var/tmp` is a problem waiting to happen. The path is by definition intended for temporary files. So a script should not make any assumptions about its content.
- The use of guessable filenames is highly discouraged for `(/var)/tmp`, as it may lead to service DoS or vulnerabilities.
- Systemd will apply "ageing" to `/var/tmp` and will by default clean any files older than 30 days. So booting an old image may have unexpected consequences.

I think that a better place would be `/var/local/`. The problem is that we currently do not allow directory creation in this path and it is empty (at least on my Fedora. We can however relax the policy in this regard.

CC @supakeen @mvo5 for additional ideas.

thozza (approved):
LGTM, thanks!

supakeen (approved):
I'm fine with adding these paths, note that we might relax this a lot more in the future.

Issue #491, opened 2024-03-04
Remove `isolabelTmpl` from RHEL.


supakeen:
In #488 I removed `isolabelTmpl` from Fedora. In RHEL this is a bit more work as we don't have the values of the label readily available on the image *and* we do switching between the RHEL/CentOS name in the distribution struct.

Let's fix it there as well *and* remove the fallback in the anaconda pipelines.

PR #529, opened 2024-03-19
Factor out common RHEL parts and RHEL-10 from RHEL-9 (COMPOSER-2185)

thozza:
This is the first step in defining RHEL distros by reusing most of the code which is not release-specific. Each release will then define its package sets, partition tables, image configurations. The rest will be shared among distros.

I factored out the "common" RHEL code into `rhel` package, which is now used by `rhel9`. Subsequently, I factored out `rhel10` out of `rhel9` and reverted the conditionals in `rhel9` distro code which were necessary to incorporate `rhel10`.

I verified that no el9 / c9s / el10 / c10s manifests have been changed by these changes.


supakeen:
Very nice, I believe this will also fix https://github.com/osbuild/images/issues/491? I'll do a review tomorrow.

thozza:
> Very nice, I believe this will also fix #491? I'll do a review tomorrow.

Not sure if I fully understand your intention with #491. But I'm happy to amend the PR to incorporate it.

thozza:
Snyk is complaining about an existing thing in `rhel9` distro. But due to moving the whole distro to a new directory, it thinks that it is a newly introduced code.

pkg/distro/fedora/images.go
```diff
@@ -333,7 +331,7 @@ func liveInstallerImage(workload workload.Workload,
 	img.Variant = "Workstation"
 	img.OSVersion = d.osVersion
 	img.Release = fmt.Sprintf("%s %s", d.product, d.osVersion)
-	img.ISOLabel = fmt.Sprintf(ISO_LABEL, img.Product, img.OSVersion, img.Variant, img.Platform.GetArch())
+	img.ISOLabel = fmt.Sprintf(t.arch.distro.ISOLabel(), img.Product, img.OSVersion, img.Variant, img.Platform.GetArch())
```
bcl, line 334: This seems awkward. distro.ISOLabel() is the template, so it is really what should decide what info from img it wants to use to populate it instead of passing it all in the right order. But passing img to it also seems wrong :)
The iso label has a stronger relation to the image type than the distro so maybe it would make sense to move the template and string creation into there and add a function to populate the string, or get it when needed?

thozza: I mostly followed the style that we already have for OSTree ref template https://github.com/osbuild/images/blob/5538c287fad5c16f6cc9fbf8afecc73450164b79/pkg/distro/distro.go#L52-L53

I'm open to suggestions and don't have strong preference here. The reality in RHEL distros has been so far that we had the template defined on distro level. For Fedora, the template was also technically the same for the whole distro.

I think that the problem here is that the method name suggests that it will return the actual label and not the template. I can rename it.

bcl: It's not just the name, it's using a template from a place that doesn't have the info to populate it with. The template should be next to the img.Product, etc. so that it can be changed if needed. Really, it ought to be a function that is called to get the ISOLabel -- since I also see there is code checking to see if the string has been set, overriding yet another template.

thozza: So, that was actually introduced by https://github.com/osbuild/images/pull/488.

In Fedora land, this is simple, but it gets complicated in RHEL due to reusing the same image type definitions for RHEL and CS.

I now understand that my PR does not really resolves https://github.com/osbuild/images/issues/491, it is keeping the status quo and just making it possible to get the template value from the `distro.Distro`, so that it becomes obtainable from `rhel_common` structs in other packages.

I'll think about this a bit.

thozza: I've submitted https://github.com/osbuild/images/pull/542

thozza: This code change is no longer part of this PR, thus resolving.

thozza:
The PR is again ready for review.

pkg/distro/rhel/rhel9/optioins.go
```diff
@@ -0,0 +1,192 @@
+package rhel9
```
achilleas-k, line 1: There's a typo in the filename here.

pkg/distro/rhel/rhel_common/imagetype.go
```diff
@@ -0,0 +1,346 @@
...
+		name:             name,
+		filename:         filename,
+		mimeType:         mimeType,
+		packageSets:      pkgSets,
+		image:            imgFunc,
+		buildPipelines:   buildPipelines,
+		payloadPipelines: payloadPipelines,
+		exports:          exports,
+	}
+}
```
achilleas-k, line 346: Is this function worth having?  I understand we might have it do more in the future, but right now, since the arguments map directly to the struct fields, it feels a bit redundant.  Even worse, it makes it easier to make mistakes with the position of args when calling the function (e.g. mixing up the last three args that are all `[]string`).

I suppose this way it makes it easier to control when adding new fields that are mandatory.  We could add a new argument and any caller that doesn't set it wouldn't compile.  Not sure what's better honestly.

This function is also consistent with our pipeline and stage generation code.  We add mandatory fields in the constructor and leave optional ones to be set by the caller.  So I guess it's better to keep and if we rethink that convention we can consider this one too.

achilleas-k (changes requested):
Looks good overall.  I didn't verify every single copy of the image types line-by-line since the test manifests haven't changed.

Two minor comments in-line: a file with a mistyped name and a question about the `NewImageType()` function.

SUPER NITPICK: One more thought: Could the common parts be moved to `pkg/distro/rhel/` instead of being under `pkg/distro/rhel/rhel_common`?  I'm not sure I like the repetition in the path (though, unavoidable I guess), the name with the `_`, and I kinda like the idea of the common bits being one level above the version-specific code.

So instead of:
```
pkg/distro/
└── rhel
    ├── rhel10
    ├── rhel9
    └── rhel_common
```

we'd have:
```
pkg/distro
└── rhel
    ├── rhel10
    └── rhel9
```

pkg/distro/rhel/rhel_common/imagetype.go
thozza: This is a very reasonable comment, and I was considering the same thing (having this function vs. having all struct fields exported and set directly).

You correctly deduced that one of the reasons for having this function is to make the mandatory fields required. IOW one must use the function to set them and must set all of them. There is exactly one image type which only calls this function and does not need to set any additional fields - `tar` image. Lastly, I also considered the consistency aspect with the pipeline and stage generation code. :slightly_smiling_face: 

But the main reason for this approach was the fact, that some struct fields can't be really exported with their current names due to the `distro.ImageType` interface methods names. This would mean that I would not name `name` as `Name`, because there is a `Name()` method in the interface. Unfortunately, the `buildPipelines`, `payloadPipelines` and `exports` have the same problem :disappointed:.  As a result, I would have to name some of the struct fields in a weird non-conflicting way, which didn't seem like the best idea.

Nevertheless, I'm open to suggestions.

thozza:
> Looks good overall. I didn't verify every single copy of the image types line-by-line since the test manifests haven't changed.
> 
> Two minor comments in-line: a file with a mistyped name and a question about the `NewImageType()` function.

Thanks a lot for the review!

> SUPER NITPICK: One more thought: Could the common parts be moved to `pkg/distro/rhel/` instead of being under `pkg/distro/rhel/rhel_common`? I'm not sure I like the repetition in the path (though, unavoidable I guess), the name with the `_`, and I kinda like the idea of the common bits being one level above the version-specific code.
> 
> So instead of:
> 
> ```
> pkg/distro/
> └── rhel
>     ├── rhel10
>     ├── rhel9
>     └── rhel_common
> ```
> 
> we'd have:
> 
> ```
> pkg/distro
> └── rhel
>     ├── rhel10
>     └── rhel9
> ```

I don't have a strong opinion on this. I'm happy to move the common code to `pkg/distro/rhel`...

thozza:
I fixed typos in filenames and moved `rhel_common` under `pkg/distro/rhel` and renamed it to `rhel`.

pkg/distro/rhel/rhel_common/imagetype.go
achilleas-k: Ah yeah, I didn't consider the method names on the interface.  That's a good reason.  Let's leave it.  At least we have the consistency.

achilleas-k (dismissed):
LGTM!

bcl (dismissed):
Looks good to me!

achilleas-k:
@thozza Tiny merge conflict :(

achilleas-k (approved):

PR #536, opened 2024-03-20
fedora: set grub2 timeout to 5 seconds

martinezjavier:
The org.osbuild.grub2 stage has a default timeout of 0 seconds, but Fedora images set this value to 5 instead. Let's do the same for osbuild images.

### DIFF
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -377,6 +377,10 @@ var (
 			// NOTE: temporary workaround for a bug in initial-setup that
 			// requires a kickstart file in the root directory.
 			Files: []*fsnode.File{initialSetupKickstart()},
+			Grub2Config: &osbuild.GRUB2Config{
+				// Overwrite the default Grub2 timeout value.
+				Timeout: 5,
+			},
 		},
 		rpmOstree:           false,
 		kernelOptions:       defaultKernelOptions,


nullr0ute:
I think this makes sense if it's the Fedora mainline defaults, I think we also need it for the RHEL minimal-raw as well.

supakeen (dismissed):
Makes sense. 

If this PR is to include the RHEL bits as well I'd like it done *only* for the minimal image type there.

nullr0ute:
> Makes sense.
> 
> If this PR is to include the RHEL bits as well I'd like it done _only_ for the minimal image type there.

Yep, agreed, I think we can deal with Fedora here and I'll follow up for RHEL Minimal

pkg/distro/fedora/images.go
```diff
@@ -237,6 +237,10 @@ func osCustomizations(
 	osc.ShellInit = imageConfig.ShellInit
 
 	osc.Grub2Config = imageConfig.Grub2Config
+
+	// overwrite the default Grub2 timeout value
+	osc.Grub2Config.Timeout = 5
```
supakeen, line 242: This needs an `if osc.Grub2Config != nil` guard around it.

supakeen:
I'm actually unsure now if this is the right place to do so by unconditionally setting it; I think `Grub2Config` might always be nil there and it should instead be set in the `defaultImageConfig`.

martinezjavier:
> I'm actually unsure now if this is the right place to do so by unconditionally setting it; I think `Grub2Config` might always be nil there and it should instead be set in the `GetDefaultImageConfig()` for the distribution.
> 
> Currently we seem to use the `defaultImageConfig`s that are defined on the images.
> 
> cc @achilleas-k @ondrejbudai

I see. I'm not really familiar with this project nor Go :) but just tracked down with @nullr0ute an issue of the Grub menu not being shown due `/etc/default/grub` having a `GRUB_TIMEOUT=0`.

supakeen (approved):
Built images locally and this works. Thanks!

PR #539, opened 2024-03-21
awscloud: implement retrieval of bucket and permissions (HMS-3099)

schuellerf:
Implements
* function to list available AWS buckets and 
* function to check bucket permissions

This is a prerequisite for https://github.com/osbuild/bootc-image-builder/issues/220

…not sure why there is no type similar to `type S3Permission string` defined in the upstream library...

### DIFF
--- a/pkg/cloud/awscloud/awscloud.go
+++ b/pkg/cloud/awscloud/awscloud.go
@@ -26,6 +26,26 @@ type AWS struct {
 	s3       *s3.S3
 }
 
+// S3Permission Implementing an "enum type" for aws-sdk-go permission constants
+type S3Permission string
+
+const (
+	S3PermissionRead        S3Permission = s3.PermissionRead
+	S3PermissionWrite       S3Permission = s3.PermissionWrite
+	S3PermissionFullControl S3Permission = s3.PermissionFullControl
+	S3PermissionReadAcp     S3Permission = s3.PermissionReadAcp
+	S3PermissionWriteAcp    S3Permission = s3.PermissionWriteAcp
+)
+
+// PermissionsMatrix Maps a requested permission to all permissions that are sufficient for the requested one
+var PermissionsMatrix = map[S3Permission][]S3Permission{
+	S3PermissionRead:        {S3PermissionRead, S3PermissionWrite, S3PermissionFullControl},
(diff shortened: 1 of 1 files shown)


pkg/cloud/awscloud/awscloud.go
```diff
@@ -26,6 +26,24 @@ type AWS struct {
...
+	S3PermissionWriteAcp    S3Permission = s3.PermissionWriteAcp
+)
+
+var PermissionsMatrix = map[S3Permission][]S3Permission{
+	S3PermissionRead:        {S3PermissionRead, S3PermissionWrite, S3PermissionFullControl},
+	S3PermissionWrite:       {S3PermissionWrite, S3PermissionFullControl},
+	S3PermissionFullControl: {S3PermissionFullControl},
+	S3PermissionReadAcp:     {S3PermissionReadAcp, S3PermissionWriteAcp},
+	S3PermissionWriteAcp:    {S3PermissionWriteAcp},
+}
```
achilleas-k, line 45: Took me a minute to realise what this mapping was for.  A comment here or a docstring on the public method would help.

pkg/cloud/awscloud/awscloud.go
```diff
@@ -624,9 +642,53 @@ func (a *AWS) Regions() ([]string, error) {
...
+	}
+
+	for _, permission := range requiredPermissions {
+		if permission == currentPermission {
+			return true
+		}
+	}
+	return false
+}
+func (a *AWS) CheckBucketPermission(bucketName string, permission S3Permission) (bool, error) {
```
achilleas-k, line 676: Docstring would be nice explaning that it checks for "the requested permission or higher", so that the matrix is easier to understand.

Not super important though.

achilleas-k (approved):
Thanks.  Tiny nitpick about comments or a docstring but not super important.  Approving.

pkg/cloud/awscloud/awscloud.go
schuellerf: I thought so too. There have been so few comments in the whole file but that's probably subject to change :-)

pkg/cloud/awscloud/awscloud.go
schuellerf: I like good/short doc strings anyway

achilleas-k (approved):

PR #540, opened 2024-03-22
policies: widen allowed filepaths

supakeen:
Allow writing files into `/home`, `/srv`, `/opt`. Allow writing directories into `/srv` and `/opt` but not `/home`.


ondrejbudai (changes requested):
I think we should apply this only to non-ostree artifacts for the starters. With ostree, it gets weird, because IIRC `/var` is problematic in commits. Also, we should probably discuss this with the edge folks before committing to any changes there.

supakeen:
@ondrejbudai @achilleas-k do we want the ostree-specific allowed paths to be what we had or do we have a widened set for those as well?

achilleas-k:
Let's keep the old ones for now and widen then in a separate PR

supakeen:
Should we allow `/usr/local/s?bin` for the ostree policies? They were recently added for non-ostree.

ondrejbudai (dismissed):
Splendid! :heart:

achilleas-k:
Merge conflicts :/
Good to merge once that's resolved though :+1:

supakeen:
@achilleas-k Rebased.

ezr-ondrej:
Have we reverted https://github.com/osbuild/images/pull/528 by this PR?
I think the goal there was to allow `/usr/local/sbin` for firstboot scripts, was it an intention here to disable it again?

lzap:
Yeah: https://github.com/osbuild/images/pull/587

PR #542, opened 2024-03-22
Remove ISOLabelTmpl from all installer image kinds

thozza:
Do not define ISO label templates on distro / image type levels, which
are then passed to the installer image kind. Instead, allow each image
type to define its own function to generate the final ISO Label to be
used for the image type. Pass the final ISO label to the installer image
kinds to be used as is.

Fix https://github.com/osbuild/images/issues/491


supakeen (approved):
<3

achilleas-k (approved):

PR #544, opened 2024-03-25
Distro/rhel8: fix EC2 x86_64 RHEL < 8.9 partition table

thozza:
RHEL-8 EC2 x86_64 images support only BIOS boot until 8.9. The partition table used to reflect that. Since 8.9, the image gained /boot/efi partition and supports hybrid boot mode.

However, I recently noticed that RHEL-8 EC2 x86_64 images for older releases than 8.9 also contain the /boot/efi partition. It turns out, that https://github.com/osbuild/images/commit/7acf235f8349552b705232773ed429336e42e066 unintentionally introduced /boot/efi partition to all RHEL-8 x86_64 EC2 images, regardless of the minor release version.

Fix this issue by ensuring that RHEL-8 x86_64 EC2 images prior to 8.9 do not contain /boot/efi partition.

### DIFF
--- a/pkg/distro/rhel8/partition_tables.go
+++ b/pkg/distro/rhel8/partition_tables.go
@@ -259,8 +259,47 @@ func getEc2PartitionTables(osVersion string, isRHEL bool) distro.BasePartitionTa
 		aarch64BootSize = 1 * common.GibiByte
 	}
 
-	return distro.BasePartitionTableMap{
-		arch.ARCH_X86_64.String(): disk.PartitionTable{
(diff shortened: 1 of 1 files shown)


supakeen (approved):
Looks OK, could maybe use less duplication but I don't see an easy way to insert the extra partition otherwise.

achilleas-k (approved):

thozza:
> Looks OK, could maybe use less duplication but I don't see an easy way to insert the extra partition otherwise.

I agree. I had the same thought and unfortunately came to the same conclusion :confused:

PR #546, opened 2024-03-26
Schutzfile: update osbuild dependency

achilleas-k:
Commit ID for upstream `main` as of right now.

PR #462 adds the new option we included in `org.osbuild.selinux` to the stage options.  This isn't used in any image definition right now except the one for BIB, which we don't attach to any distros so it's not built here or in osbuild-composer.

Let's update osbuild's commit ID separately here so we're consistent (and it's a good opportunity for a mass rebuild).

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -8,7 +8,7 @@
   "fedora-39": {
     "dependencies": {
       "osbuild": {
-        "commit": "6b4bb850a77feade79d839d9ea075e29ca210ae9"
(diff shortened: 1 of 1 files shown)


achilleas-k:
See https://github.com/osbuild/images/pull/547#issuecomment-2021485339 for explanation of RHEL 9.4 oscap test failure.

achilleas-k:
Will be fixed by #551

kingsleyzissou (approved):
Thanks

PR #547, opened 2024-03-26
test/config-maps: re-enable oscap-rhel8 config

kingsleyzissou:
Added additional tailoring rule for `partition_for_dev_shm` which appears to be the rule causing the issues with the `RHEL8.10` boot tests for OpenSCAP configs.

### DIFF
--- a/test/config-map.json
+++ b/test/config-map.json
@@ -198,6 +198,14 @@
       "qcow2"
     ]
   },
+  "./configs/oscap-rhel8.json": {
+    "distros": [
+      "rhel-8.10"
+    ],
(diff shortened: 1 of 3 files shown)


achilleas-k:
We're going to need that `chown` commit in main because of a mixup with the previous oscap PR.  The boot test succeeded at some point while I was writing #502 and then I somehow dropped the `chown` line (bad rebase or switching computers), but the RHEL 9.4 images weren't retested because they were marked as passed (and didn't change after that).

achilleas-k (approved):
Thanks for figuring this out!

PR #548, opened 2024-03-26
Remove internal/boot, cmd/osbuild-image-tests, and cmd/osbuild-dnf-json-tests

achilleas-k:
Delete test code carried over from osbuild-composer that's not used in this repository.

We might end up bringing some parts of these back if we need upload + boot code here for testing, but we can always do it on a case by case basis.  For now, it's better to clean out dead code.


bcl (approved):
Seems like a good idea to me

supakeen (approved):
Less code is more better in this case.

PR #551, opened 2024-03-27
test/scripts: chown `results.xml`

kingsleyzissou:
The `results.xml` file is owned by root, chown the file so we can read the results.

Cherry-picked this commit from #547 into it's own PR. See comments in #546


achilleas-k (approved):
Thank you!

PR #561, opened 2024-04-05
many: revert PR#462 to unblock images temporarily

mvo5:
It turns out that we need user customizations that go beyond what PR#462 provided. So to unblock the other work around the iso installer revert this PR temporarily and once the groundwork in osbuild is ready this commit here will be reverted and the pieces added that enable user customizations with `bootc install to-filesystem`.

[Opening this to have it available if we need it, maybe we can be quick and get osbuild/customzations ready early enough]

[edit: https://github.com/osbuild/images/compare/main...mvo5:bootc-disk-customize-again?expand=1 contains the WIP work on making the user customization work again, it needs https://github.com/osbuild/osbuild/pull/1706 too]


mvo5:
It looks like we won't need this.

PR #566, opened 2024-04-08
Minor fix in check-build-coverage PR reporting

achilleas-k:
**test/check-build-coverage: fix branch comparison**

The PR number is stored in the build info without the "PR-" prefix that
we set in GitLab CI.  We need to remove the prefix for the comparison.

---

**test/upload-results: use removeprefix() instead of lstrip()**

lstrip() is for stripping a class of characters, not removing a string
of text.

---

### DIFF
--- a/test/scripts/check-build-coverage
+++ b/test/scripts/check-build-coverage
@@ -14,7 +14,7 @@ SKIPS = [
 
 def check_build_coverage(cachedir):
     ci_commit = os.environ.get("CI_COMMIT_SHA")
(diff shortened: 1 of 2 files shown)


bcl (approved):
Looks good to me, but the tests seem annoyed by something

achilleas-k:
> Looks good to me, but the tests seem annoyed by something

It's the two non-required tests that check if there are breaking changes in this repo that haven't been made in osbuild-composer yet.

PR #572, opened 2024-04-10
Local container manifest-lists

kingsleyzissou:
This would unblock:
- https://github.com/osbuild/bootc-image-builder/issues/327

But is blocked by:
- https://github.com/containers/image/issues/2374


bcl:
Looks ok, although I worry a bit about changing the signature of a public function.

kingsleyzissou:
> Looks ok, although I worry a bit about changing the signature of a public function.

Yeah valid. I can handle it inside the function rather. 

Still need to re-think how we handle the second commit since that route no longer seems feasible.

kingsleyzissou:
Closing this in favour of #595

PR #574, opened 2024-04-10
dnfjson: support dnf base proxy configuration

croissanne:


achilleas-k:
I generally like (and try to stick to) the pattern of having _required_ properties in the constructor and methods for setting non-required ones.  So instead of:
```go
solver := dnfjson.NewSolver(d.ModulePlatformID(), d.Releasever(), arch, d.Name(), cacheDir, "")
```
and
```go
solver := dnfjson.NewSolver(d.ModulePlatformID(), d.Releasever(), arch, d.Name(), cacheDir, proxy)
```

we would have
```go
solver := dnfjson.NewSolver(d.ModulePlatformID(), d.Releasever(), arch, d.Name(), cacheDir)
solver.SetProxy(proxy)
```
or just
```go
solver := dnfjson.NewSolver(d.ModulePlatformID(), d.Releasever(), arch, d.Name(), cacheDir)
solver.Proxy = proxy
```

I think it's nicer than having an empty string or nil argument in most calls and only having a value for it in one or two places where it's needed.

thozza:
I agree with @achilleas-k. We tend to stick to this approach for optional properties.

pkg/dnfjson/dnfjson.go
```diff
@@ -160,6 +163,10 @@ func (s *Solver) GetCacheDir() string {
 	return filepath.Join(s.cache.root, b)
 }
 
+func (s *Solver) SetProxy(proxy string) {
```
croissanne, line 166: should this validate the proxy string? Make sure it's a valid URL or not?

Does it need some doc comment above?

achilleas-k: We do have URL validation in some places: https://github.com/osbuild/images/blob/d1d5119e1b98d1d365f0a91e168384f7f59e78b4/pkg/ostree/ostree.go#L111-L113

That's pretty simple and should be enough.

A comment would be good too.

croissanne: will do

achilleas-k:
Should this be draft until `osbuild-depsolve-dnf` gets proxy support?  Did we add it already?

achilleas-k:
> Should this be draft until `osbuild-depsolve-dnf` gets proxy support? Did we add it already?

Speaking of, if the PR hasn't been prepped yet, do you mind basing it on https://github.com/osbuild/osbuild/pull/1674?  There's a bunch of changes in the depsolve job handling there and it'd be easier to avoid conflicts.

achilleas-k:
> wait I don't follow, the depsolve job is defined in composer and the depsolver we use is still the python dnf-json in composer?

The depsolver now is osbuild-depsolve-dnf, lives in the osbuild repo, and is packaged as osbuild-depsolve-dnf (separate rpm, subpackage of osbuild).

There's a `dnf-json` in osbuild-composer still but I thought we removed that.  I think (hope) osbuild-composer isn't using that anymore.

achilleas-k:
> There's a `dnf-json` in osbuild-composer still but I thought we removed that. I think (hope) osbuild-composer isn't using that anymore.

https://github.com/osbuild/osbuild-composer/blob/41a532291cb739a7b2e5186541b4668b2dd7d84f/cmd/osbuild-composer/config.go#L130

achilleas-k (approved):

PR #576, opened 2024-04-10
test: inspect manifests in manifest-only jobs

achilleas-k:
We don't build ppc64le and s390x in gitlab CI, we only generate the manifests to ensure there's no issues with manifest generation and package selection.  However, we can also validate them using 'osbuild --inspect' to make sure they're at least valid manifests.


bcl (approved):
Looks good

PR #577, opened 2024-04-10
Reduce GitLab CI pipeline depth

achilleas-k:
Previously, our top-level gitlab-ci.yml included a call to `./test/scripts/configure-generators` which was used to generate a dynamic pipeline with jobs that would generate manifests for each distro/arch/image-type configuration.  Then, each of those manifest generation stages in the dynamic pipeline would generate another level of dynamic pipelines based on the manifest changes.

This PR reduces the depth of dynamic pipelines by 1 (to 1).  The gitlab-ci.yml in the repository is replaced by the configuration that was previously dynamically generated at the first level, with the other previous top-level jobs added (`init`, `verify`, `finish`).  A CI job makes sure that the generated config is up to date.

This change has two benefits:
- It reduces the total CI runtime *in all cases*.  The `configure-generators` step would take 3-4 minutes and would run every time.  Now it's done offline and a local run takes a few seconds.  The runtime of the job in CI was mostly spent setting up the runner.
- It makes the CI pipelines in GitLab easier to navigate since there is now only *one* level of dynamic generation.

Includes #576


bcl:
Source prep is hitting that btrfs dev problem I've seen mentioned somewhere else...

thozza (approved):
Makes sense, thanks!

.github/workflows/tests.yml
```diff
@@ -138,7 +138,7 @@ jobs:
 
   prepare:
     name: "🔍 Check source preparation"
```
thozza, line 140: You probably need to install the build deps here, now that this step also generates the list of images which compiles `images`.

So probably `sudo apt install -y libgpgme-dev libbtrfs-dev libdevmapper-dev`

achilleas-k: Oops, yes.  Absolutely right.

achilleas-k:
> It would be great to not run Gitlab CI at all if `Check source preparation` step fails.

This is a good suggestion.  I wanted to get away from what we do in osbuild-composer which is gate gitlab on all the unit tests, which can take some time.  But the source-prep stage, *especially* now that it's generating the CI config, should be in the same workflow as the GitLab trigger now.  I'll move some stuff around.

achilleas-k:
Done and done.

thozza (approved):
Thanks.

PR #580, opened 2024-04-11
fedora: we're building RCs for F40

supakeen:
In #522 I implemented a naive way to show/not-show pre-releases in installers. Since Fedora 40 is currently building RCs and Finals it should be turned off for 40 now.


thozza (approved):
LGTM

croissanne:
I'm removing this from the queue so we can merge https://github.com/osbuild/images/pull/586 first, it'll fix ci caching.

PR #581, opened 2024-04-11
Port RHEL-7 to the common `rhel` package

thozza:
Port RHEL-7 as the last distro to the common `rhel` package.

Notable changes:

- Add `NoBLS` to `ImageConfig`
- Add `Codename()` to the `distro.Distro` interface
- Pass YUM config from the `ImageConfig` to `OSCustomizations` in distro/rhel.
 - Pass `SELinuxForceRelabel` from the `ImageConfig` to `OSCustomizations` in distro/rhel.
 - Add `DiskImageVPCForceSize` and `DiskImagePartTool` to `rhel.ImageType`, since these needs to be set to non-default values for RHEL-7.

I compared **all** manifests generated on `main` and `HEAD` of this PR and there were no changes.


pkg/distro/distro.go
```diff
@@ -41,6 +41,9 @@ type Distro interface {
 	// Returns the name of the distro.
 	Name() string
 
+	// Returns the nick of the distro.
```
supakeen, line 44: What is the nick? A nickname?

supakeen (dismissed):
Looks very nice just a question to clarify what a nick is :)

pkg/distro/distro.go
achilleas-k: I know we had this already but I think a more correct term for it is "codename".
> Codename: Ootpa (is based on Fedora 28, upstream Linux kernel 4.18, systemd 239, and GNOME 3.28.)
https://access.redhat.com/articles/3078

> VERSION_CODENAME=
> A lower-case string (no spaces or other characters outside of 0–9, a–z, ".", "_" and "-") identifying the operating system release code name
https://www.freedesktop.org/software/systemd/man/latest/os-release.html

Let's rename it to `codename`.  I think it'll be clearer.

thozza: I've submitted my comment without refreshing the page. Will rename it :)

achilleas-k (approved):
I initially thought we might not want this because RHEL 7 is "too different" from 8 and 9 to be merged cleanly but the actual changes here are quite small.  That last commit has a large diff but it's all mostly structural changes.

Really nice that this gets unified too.  Clean.  Thank you!!

LGTM

achilleas-k:
Let's hold off merging this until #586 gets through otherwise we'll get a lot of unnecessary builds.

achilleas-k (changes requested):
Since this isn't critical and we're not in a hurry, I'm marking it with a change request for the Nick -> Codename rename.

pkg/distro/distro.go
thozza: I renamed `Nick` -> `Codename`

achilleas-k (approved):

PR #585, opened 2024-04-12
container: include the architecture of the image in container.Spec

mvo5:
This is a different attempt to resolve the issue that the resolver consider the requested architecture more of a "hint". The initial attempt is in https://github.com/osbuild/images/pull/451 but it caused some issues as this seems to be expected behavior of the library in tests at least.

So this commit is more conservative and just includes the the architecture in the `container.Spec` so that the consumer of the library can inspect and error on mismatch of requested arch and resolved arch.

With that in bib we can check the arch, I outlined this in https://github.com/osbuild/bootc-image-builder/compare/main...mvo5:check-resolved-image-arches?expand=1 what we could have to do there (plus some other changes because the image API has changed quite a bit and a go.mod change to make it trivial to test locally but that will need adjustment of course).

### DIFF
--- a/pkg/arch/arch.go
+++ b/pkg/arch/arch.go
@@ -7,7 +7,8 @@ import (
 type Arch uint64
 
 const ( // architecture enum
-	ARCH_AARCH64 Arch = iota
+	ARCH_UNSET Arch = iota
+	ARCH_AARCH64
 	ARCH_PPC64LE
 	ARCH_S390X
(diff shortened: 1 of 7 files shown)


kingsleyzissou:
I like it :)

kingsleyzissou (approved):
Thanks for taking care of this!

achilleas-k (approved):

PR #595, opened 2024-04-15
Resolve local container manifests

kingsleyzissou:
This is a POC to resolve local containers by their ID rather than by their name, the podman desktop team
asked if this was possible since local manifests are not.

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -351,6 +351,20 @@ func getImageRef(target reference.Named, local bool) (types.ImageReference, erro
 	return docker.NewReference(target)
 }
 
+func (cl *Client) resolveContainerImageArch(ctx context.Context, ref types.ImageReference) (*arch.Arch, error) {
+	img, err := ref.NewImage(ctx, cl.sysCtx)
+	if err != nil {
+		return nil, err
+	}
+	defer img.Close()
+	info, err := img.Inspect(ctx)
+	if err != nil {
+		return nil, err
+	}
+	a := arch.FromString(info.Architecture)
+	return &a, nil
+}
+
 // GetManifest fetches the raw manifest data from the server. If digest is not empty
 // it will override any given tag for the Client's Target.
 func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest, local bool) (r RawManifest, err error) {
@@ -360,17 +374,6 @@ func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest,
 	if err != nil {
 		return
 	}
-	img, err := ref.NewImage(ctx, cl.sysCtx)
-	if err != nil {
-		return r, err
-	}
-	defer img.Close()
-
-	info, err := img.Inspect(ctx)
-	if err != nil {
-		return r, err
-	}
-	r.Arch = arch.FromString(info.Architecture)
 
 	src, err := ref.NewImageSource(ctx, cl.sysCtx)
 	if err != nil {
@@ -398,7 +401,22 @@ func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest,
 			secondaryDigest = &instanceDigest
 		}
 		r.Data, r.MimeType, err = src.GetManifest(ctx, secondaryDigest)
-		return err
+		if err != nil {
+			return err
+		}
+
+		// getting the container image arch doesn't work with local manifest lists
+		if local && (r.MimeType == imgspecv1.MediaTypeImageIndex || r.MimeType == manifest.DockerV2ListMediaType) {
+			return nil
+		}
+
+		imageArch, err := cl.resolveContainerImageArch(ctx, ref)
+		if err != nil {
+			return err
+		}
+		r.Arch = *imageArch
+
+		return nil
 	}, &retryOpts); err != nil {
 		return
 	}
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -108,6 +108,8 @@ type Client struct {
 
 	UserAgent string // user agent string to use for requests, defaults to DefaultUserAgent
 
+	Store string // another store location other than the main one, useful for testing
+
 	// internal state
 	policy *signature.Policy
 	sysCtx *types.SystemContext
@@ -167,6 +169,17 @@ func (cl *Client) SetAuthFilePath(path string) {
 // GetAuthFilePath gets the location of the `containers-auth.json(5)` file.
 func (cl *Client) GetAuthFilePath() string {
 	return cl.sysCtx.AuthFilePath
+}
+
+func (cl *Client) SetContainersStore(store string) {
+	cl.Store = store
+}
+
+func (cl *Client) GetContainersStore() string {
+	if cl.Store == "" {
+		return "/var/lib/containers/storage"
+	}
+	return cl.Store
 }
 
 func (cl *Client) SetArchitectureChoice(arch string) {
@@ -344,13 +357,14 @@ func (m RawManifest) Digest() (digest.Digest, error) {
 	return manifest.Digest(m.Data)
 }
 
-func getImageRef(target reference.Named, id string, local bool) (types.ImageReference, error) {
+func (cl *Client) getImageRef(target reference.Named, id string, local bool) (types.ImageReference, error) {
 	if local {
 		imageName := target.String()
 		if id != "" {
 			imageName = id
 		}
-		return alltransports.ParseImageName(fmt.Sprintf("containers-storage:%s", imageName))
+		options := fmt.Sprintf("containers-storage:[overlay@%s+/run/containers/storage]%s", cl.GetContainersStore(), imageName)
+		return alltransports.ParseImageName(options)
 	}
 
 	if target == nil {
@@ -375,7 +389,7 @@ func (cl *Client) resolveContainerImageArch(ctx context.Context, ref types.Image
 }
 
 func (cl *Client) getLocalImageIDFromDigest(instance digest.Digest) (string, error) {
-	store, err := storage.GetStore(storage.StoreOptions{})
+	store, err := storage.GetStore(storage.StoreOptions{GraphRoot: cl.GetContainersStore()})
 	if err != nil {
 		return "", err
 	}
@@ -406,7 +420,7 @@ func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest,
 		}
 	}
 
-	ref, err := getImageRef(target, id, local)
+	ref, err := cl.getImageRef(target, id, local)
 	if err != nil {
 		return
 	}
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -353,18 +353,8 @@ func getImageRef(target reference.Named, local bool) (types.ImageReference, erro
 
 // GetManifest fetches the raw manifest data from the server. If digest is not empty
 // it will override any given tag for the Client's Target.
-func (cl *Client) GetManifest(ctx context.Context, digest digest.Digest, local bool) (r RawManifest, err error) {
+func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest, local bool) (r RawManifest, err error) {
 	target := cl.Target
-
-	if digest != "" {
-		t := reference.TrimNamed(cl.Target)
-		t, err = reference.WithDigest(t, digest)
-		if err != nil {
-			return
-		}
-
-		target = t
-	}
 
 	ref, err := getImageRef(target, local)
 	if err != nil {
@@ -398,7 +388,16 @@ func (cl *Client) GetManifest(ctx context.Context, digest digest.Digest, local b
 	}
 
 	if err = retry.RetryIfNecessary(ctx, func() error {
-		r.Data, r.MimeType, err = src.GetManifest(ctx, nil)
+		var secondaryDigest *digest.Digest
+		if instanceDigest != "" {
+			// We can pass the instance digest, if it is nil, then this is the primary manifest.
+			// If it is not nil, then this is the instance digest and the primary manifest is a
+			// manifest list. The `GetManifest` call will then retrieve the manifest for the
+			// desired instance, see:
+			// https://github.com/containers/image/blob/cdb2f596a95018444f4dee0993e321d3d8bc328d/types/types.go#L252C2-L253C121
+			secondaryDigest = &instanceDigest
+		}
+		r.Data, r.MimeType, err = src.GetManifest(ctx, secondaryDigest)
 		return err
 	}, &retryOpts); err != nil {
 		return
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -14,6 +14,7 @@ import (
 	_ "github.com/containers/image/v5/docker/archive"
 	_ "github.com/containers/image/v5/oci/archive"
 	_ "github.com/containers/image/v5/oci/layout"
+	"github.com/containers/storage"
 	"golang.org/x/sys/unix"
 
 	"github.com/containers/common/pkg/retry"
@@ -343,9 +344,17 @@ func (m RawManifest) Digest() (digest.Digest, error) {
 	return manifest.Digest(m.Data)
 }
 
-func getImageRef(target reference.Named, local bool) (types.ImageReference, error) {
+func getImageRef(target reference.Named, id string, local bool) (types.ImageReference, error) {
 	if local {
-		return alltransports.ParseImageName(fmt.Sprintf("containers-storage:%s", target))
+		imageName := target.String()
+		if id != "" {
+			imageName = id
+		}
+		return alltransports.ParseImageName(fmt.Sprintf("containers-storage:%s", imageName))
+	}
+
+	if target == nil {
+		return nil, fmt.Errorf("target cannot be nil for remote image")
 	}
 
 	return docker.NewReference(target)
@@ -365,12 +374,39 @@ func (cl *Client) resolveContainerImageArch(ctx context.Context, ref types.Image
 	return &a, nil
 }
 
+func (cl *Client) getLocalImageIDFromDigest(instance digest.Digest) (string, error) {
+	store, err := storage.GetStore(storage.StoreOptions{})
+	if err != nil {
+		return "", err
+	}
+	images, err := store.ImagesByDigest(instance)
+	if err != nil {
+		return "", err
+	}
+	if len(images) == 0 {
+		return "", fmt.Errorf("Unable to find image id for digest: %v", instance)
+	}
+	// the `ImagesByDigest` function always returns a list
+	// of images. The list could be larger than 1 in the case
+	// since it searches all stores for the matching digest,
+	// so it is okay to return the first item.
+	return images[0].ID, nil
+}
+
 // GetManifest fetches the raw manifest data from the server. If digest is not empty
 // it will override any given tag for the Client's Target.
 func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest, local bool) (r RawManifest, err error) {
 	target := cl.Target
 
-	ref, err := getImageRef(target, local)
+	var id string
+	if instanceDigest != "" && local {
+		id, err = cl.getLocalImageIDFromDigest(instanceDigest)
+		if err != nil {
+			return r, err
+		}
+	}
+
+	ref, err := getImageRef(target, id, local)
 	if err != nil {
 		return
 	}
@@ -392,7 +428,8 @@ func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest,
 
 	if err = retry.RetryIfNecessary(ctx, func() error {
 		var secondaryDigest *digest.Digest
-		if instanceDigest != "" {
+		// passing in a digest for a local manifest list will break
+		if instanceDigest != "" && !local {
 			// We can pass the instance digest, if it is nil, then this is the primary manifest.
 			// If it is not nil, then this is the instance digest and the primary manifest is a
 			// manifest list. The `GetManifest` call will then retrieve the manifest for the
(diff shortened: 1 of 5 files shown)


pkg/container/client.go
```diff
@@ -111,9 +115,59 @@ type Client struct {
...
+
+	if targetId != nil {
+		// TODO: the line below is going to create a named
+		// reference of `docker.io/library/{targetid}`.
+		// Some ideas on how to handle this:
+		// 1. Allow `docker.io/library/{targetid}`
+		// 2. Set it to `localhost/unknown`
+		// 3. Set the name to `<none>` which causes an error on the
+		//    `ParsedNormalizedNamed` function call
+		target = UnknownContainerName
```
kingsleyzissou, line 169: This is pretty janky... but we rely heavily on named references and I didn't want to break that.
So this was a work around.

achilleas-k:
I'm losing the thread a bit here (totally my fault, sorry) so let me see if I can summarise a few things:
- Our container resolver expects a ref with tag (`:latest` by default) or a manifest list ID.
- When resolving local containers, manifest lists aren't supported.
- Our resolver has trouble resolving containers from the local store (sometimes?).
- In the most common case, the resolver resolves a container image ID as `ref -> manifest list -> image ID for selected architecture`.


> the podman desktop team asked if this was possible since local manifests are not.

Is this sentence missing something? :smile: 

Generally, I'm wondering if we could "simply" have separate resolve paths for local and remote.  How is this done in places like skopeo?  I think I might still be missing some context around what exactly isn't working.

kingsleyzissou:
> I'm losing the thread a bit here (totally my fault, sorry) so let me see if I can summarise a few things:

Sorry I think I should have given a bit more context to this. But was still playing around with this.  I had a call with @cdrage last week about this and trying to resolve the manifests. 


>     - Our container resolver expects a ref with tag (`:latest` by default) or a manifest list ID.
>     - When resolving local containers, manifest lists aren't supported.
>     - Our resolver has trouble resolving containers from the local store (sometimes?).
>     - In the most common case, the resolver resolves a container image ID as `ref -> manifest list -> image ID for selected architecture`.

So this is more or less correct. Here is a little more context. The podman desktop team has built a local manifest list, here is an example of the output:

```
$ podman images
REPOSITORY                                 TAG         IMAGE ID      CREATED         SIZE
localhost/multi                            latest      603da7b1a137  25 minutes ago  1.05 kB
<none>                                     <none>      d5e5bd2654d3  25 minutes ago  138 MB
<none>                                     <none>      7eebab5c88a2  25 minutes ago  93.1 MB
```

The images have a name of `<none>` but the manifest list is named. So at the moment you have to inspect the manifest, get the digest for the arch you want and then match it against the list to get the id which is all quite manual.

The only thing to get this working is to then tag the image for the arch you want. So for example I would have done `localhost/multi:arm64` and `localhost/multi:amd64`. This works but feels a bit janky having to tag images in a manifest list. What they are asking for, is for us to resolve by id so that we don't have to go through that last step.

> 
> > the podman desktop team asked if this was possible since local manifests are not.
> 
> Is this sentence missing something? 😄

It is 🤦‍♂️ I was trying to say since manifest lists aren't working locally

 
> Generally, I'm wondering if we could "simply" have separate resolve paths for local and remote. How is this done in places like skopeo? I think I might still be missing some context around what exactly isn't working.

A second path for local containers could be a solid option. I will look into that.

cdrage:
> I'm losing the thread a bit here (totally my fault, sorry) so let me see if I can summarise a few things:
> 
> * Our container resolver expects a ref with tag (`:latest` by default) or a manifest list ID.
> * When resolving local containers, manifest lists aren't supported.
> * Our resolver has trouble resolving containers from the local store (sometimes?).
> * In the most common case, the resolver resolves a container image ID as `ref -> manifest list -> image ID for selected architecture`.
> 
> > the podman desktop team asked if this was possible since local manifests are not.
> 
> Is this sentence missing something? 😄
> 
> Generally, I'm wondering if we could "simply" have separate resolve paths for local and remote. How is this done in places like skopeo? I think I might still be missing some context around what exactly isn't working.

Hi! We should of had an issue open (@kingsleyzissou are you able to open up an issue?).

Basically in summary, it's this issue: https://github.com/containers/podman-desktop-extension-bootc/issues/324

When building a manifest, you create <none> images which are the actual data, with the manifest being the only one named.

We cannot pass in a "<none>" image to bootc-image-builder at the moment and we are forced to tag / give the image a name.

That's why we requested if we can just pass in the container ID instead.

achilleas-k:
> We cannot pass in a "" image to bootc-image-builder at the moment and we are forced to tag / give the image a name.
> 
> That's why we requested if we can just pass in the container ID instead.

Right, I see.  Thanks for clarifying.

kingsleyzissou:
So I figured out that we can query the local store and get the image by its digest. This is also a lot cleaner than trying to resolve by the container ID. Steps to verify locally are listed below.

Create a multi-arch manifest list locally, either with the commands below or with an existing manifest list:
```
amd=$(buildah from --arch=amd64 scratch)
arm=$(buildah from --arch=arm64 scratch)
buildah copy "${amd}" README.md
buildah copy "${arm}" README.md
amdid=$(buildah commit --format=docker --rm "${amd}")
armid=$(buildah commit --format=docker --rm "${arm}")
name="localhost/multi-arch"
buildah manifest create "${name}" "${amdid}" "${armid}"
```

Create a config file:
```
{
  "name": "local-container-config",
  "blueprint": {
    "containers": [
      {
        "source": "localhost/multi-arch",
        "local-storage": true
      }
    ]
  }
}

```

And then:
```
go build -o gen-manifests ./cmd/gen-manifests \
        && sudo ./gen-manifests \
        --metadata=false \
        --distros "fedora-39" \
        --arches x86_64,aarch64 \
        --output ./manifests \
        --images qcow2 \
        --config ./config.json
```

achilleas-k:
Neat!  I'll have a closer look later but this seems to have worked out great.

cdrage:
Hey all! So I did a bunch of troubleshooting and this works great!

Steps I did:

```
# Created the test bib
podman build -t bib --build-context=osbuild=../osbuild --build-context=images=../images -f devel/Containerfile .

# Created a test manifest against httpd bootc container
podman build --platform linux/arm64,linux/amd64 --log-level trace --manifest manifest-test .
```

I then tested the following combinations:

- .raw ARM64 🟢 
- .raw AMD64 🟢 
- .qcow2 AMD64 🟢 

Did not check any other ones, but they all built correct, I tested with `qemu-system-x86_64` as well as UTM on a mac M1 and confirmed that bib built the image corresponding to what I passed in:

```
podman run \
    --rm \
    -it \
    --privileged \
    --pull=newer \
    --security-opt label=type:unconfined_t \
    -v $(pwd)/123:/output \
    -v /var/lib/containers/storage:/var/lib/containers/storage \
    bib:latest \
    --type raw --target-arch amd64 \
    --local \
    localhost/manifest-test
```


Screenshots:

achilleas-k:
> Hey all! So I did a bunch of troubleshooting and this works great!

Thanks!!  Appreciate the double (triple) checking! :)

achilleas-k (commented):
LGTM.  Can we set up a quick local unit test for this?  I'm thinking something that quickly creates a local manifest list (completely offline) and resolves it.

I played around a bit and this should work:

**Containerfile**
```Containerfile
FROM scratch

ARG arch
COPY Containerfile /Containerfile-$arch
```

**Build with**
```
› podman build . --build-arg=arch=arm64 --arch=arm64 -t ll:arm64
STEP 1/3: FROM scratch
STEP 2/3: ARG arch
--> Using cache e4ce523588b509350da05865b6341925a5bf1e9b872eeb71f73766674f493cac
--> e4ce523588b5
STEP 3/3: COPY Containerfile /Containerfile-$arch
--> Using cache c7e9281003121f9c29c52cf10bb1f59637ee63da78b97d6f4350dfb4cdb92e2b
COMMIT ll:arm64
--> c7e928100312
Successfully tagged localhost/ll:arm64
c7e9281003121f9c29c52cf10bb1f59637ee63da78b97d6f4350dfb4cdb92e2b

› podman build . --build-arg=arch=amd64 --arch=amd64 -t ll:amd64
STEP 1/3: FROM scratch
STEP 2/3: ARG arch
--> Using cache e4ce523588b509350da05865b6341925a5bf1e9b872eeb71f73766674f493cac
--> e4ce523588b5
STEP 3/3: COPY Containerfile /Containerfile-$arch
--> Using cache 9968956c205ab0e05244f127b441567701db14c51d01692ec5e02ca007a4848f
COMMIT ll:amd64
--> 9968956c205a
Successfully tagged localhost/ll:amd64
Successfully tagged localhost/ll:arm
9968956c205ab0e05244f127b441567701db14c51d01692ec5e02ca007a4848f
```

**Make manifest list**
```
› podman manifest create -a list:local containers-storage:localhost/ll:amd64
3062ceed6afe65e8729f419d7388a6023cdb452734c6790fedc1e65fc9f7118e

› podman manifest create -a list:local containers-storage:localhost/ll:arm64
3062ceed6afe65e8729f419d7388a6023cdb452734c6790fedc1e65fc9f7118e

› podman manifest inspect localhost/list:local
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 495,
            "digest": "sha256:58b7d23d23cff14368a1ba698986248d6833f90c1d024314cd8f68acd8a81362",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 495,
            "digest": "sha256:057a6d89ea41bcc65029887d90b94004f84f8629aef9ba5729b37e37785494f2",
            "platform": {
                "architecture": "arm64",
                "os": "linux"
            }
        }
    ]
}
```

and then run the resolver on `localhost/list:local`.

achilleas-k:
We don't really depend on `podman` in the project but we do for integration tests so maybe we could skip the particular test when podman isn't installed.

cdrage:
> podman manifest create

Could the test also have `podman build --platform linux/arm64,linux/amd64 --manifest manifest-test .` too?

This would be the "main" test that Podman Desktop also uses. With your above test, each image is labelled. However, how it's built and recommended via `podman build --platform` the resulting children images will have `<none>` on repo tags, etc.

So would be best to also have a test for that too.

pkg/container/resolver_test.go
```diff
@@ -114,3 +117,82 @@ func TestResolverFail(t *testing.T) {
...
+	containerFile, err := os.CreateTemp("", "Containerfile")
+	if err != nil {
+		t.Fatalf("Failed to create tempfile: %v", err)
+	}
+	defer os.Remove(containerFile.Name())
+
+	// we need something more than scratch otherwise
+	// two containers with different arches will share
+	// the same underlying container image
+	_, err = containerFile.Write([]byte("FROM fedora-minimal\nRUN uname -m"))
```
kingsleyzissou, line 145: To do the whole `podman build --platform ...` thing I had to use a `fedora-minimal` base because if we just used `scratch` it would only build a single container that could be used for both arches.

kingsleyzissou:
@achilleas-k added a unit test but having some permissions issues with the store in the github runner.

pkg/container/resolver.go
```diff
@@ -28,6 +28,7 @@ type SourceSpec struct {
 	Digest    *string
 	TLSVerify *bool
 	Local     bool
+	Store     *string
```
kingsleyzissou, line 31: So this breaks composer, but it makes testing a little easier

cdrage:
Hi @achilleas-k @kingsleyzissou any way we could get a preliminary review so that when e2e tests pass we can get this merged in / updated in bootc-image-builder? We're anticipating that this feature get in before end-of-week so we can go a bootc release before summit 🙏

achilleas-k:
Yeah, that's on me.  I didn't mean to delay this.  Going through the last bits now.

achilleas-k (dismissed):

pkg/container/resolver_test.go
```diff
@@ -104,8 +111,88 @@ func TestResolverFail(t *testing.T) {
...
+		nil,
 	})
 	specs, err = resolver.Finish()
 	assert.Error(t, err)
 	assert.Len(t, specs, 0)
 }
+
+func TestResolverLocalManifest(t *testing.T) {
+	currentUser, err := user.Current()
+	if err != nil {
```
mvo5, line 123: Why not just `assert.NoError(t, err)` here? It seems to be the style of the other tests and is slightly shorter :) (here and in some places below)

pkg/container/resolver_test.go
achilleas-k: I was thinking we could use scratch and add an argument that changes the two based on the architecture, like I described in https://github.com/osbuild/images/pull/595#pullrequestreview-2015155890, but we can do that in a followup.

pkg/container/resolver_test.go
```diff
@@ -104,8 +111,88 @@ func TestResolverFail(t *testing.T) {
...
+	if currentUser.Uid != "0" {
+		t.Skip("User is not root, skipping test")
+	}
+
+	_, err = exec.LookPath("podman")
+	if err != nil {
+		t.Skip("Podman not available, skipping test")
+	}
+
+	containerFile, err := os.CreateTemp("", "Containerfile")
```
mvo5, line 136: (nitpick) there is `t.TempDir()`  in go now so `os.CreateTemp(t.TempDir(), "Containerfile")` could be used to avoid avoid to do the defer os.Remove as golang will auto-cleanup the dir

pkg/container/resolver_test.go
```diff
@@ -104,8 +111,88 @@ func TestResolverFail(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to create tempfile: %v", err)
+	}
+	defer os.Remove(containerFile.Name())
+
+	tmpStorage, err := os.MkdirTemp("", "storage")
```
mvo5, line 142: Same as above, `t.TempDir()` could be used here

pkg/container/resolver_test.go
```diff
@@ -104,8 +111,88 @@ func TestResolverFail(t *testing.T) {
...
+		"podman",
+		"--root", tmpStorage, // don't dirty the default store
+		"build",
+		"--platform", "linux/amd64,linux/arm64",
+		"--manifest", "multi-arch",
+		"-f", containerFile.Name(),
+		".",
+	)
+
+	err = cmd.Run()
```
mvo5, line 166: (niptick) this could have `cmd.Stdout = os.Stdout` and `cmd.Stderr = os.Stderr` here for more visiblity of what is going on during the test or use `output, err := cmd.CombinedOutput()` so that if anything goes wrong there is more debug output.

pkg/container/resolver_test.go
```diff
@@ -104,8 +111,88 @@ func TestResolverFail(t *testing.T) {
...
+	}
+
+	// try resolve an x86_64 container using a local manifest list
+	resolver := container.NewResolver("amd64")
+	resolver.Add(container.SourceSpec{
+		"localhost/multi-arch",
+		"",
+		common.ToPtr(""),
+		common.ToPtr(false),
+		true,
```
mvo5, line 178: (niptick) it feels slightly easier to read to have a local var `local:=true` and pass it here instead of `true` - this is my personal pet-peeve about APIs that take booleans as inputs, it's hard to know what the "true"/"false" stand for without additional context :)

pkg/container/resolver_test.go
```diff
@@ -104,8 +111,88 @@ func TestResolverFail(t *testing.T) {
...
+		&tmpStorage,
+	})
+	specs, err := resolver.Finish()
+	assert.NoError(t, err)
+	assert.Len(t, specs, 1)
```
mvo5, line 183: Is there a way to check if the spec matches the expected tag or id?

pkg/container/resolver_test.go
```diff
@@ -104,8 +111,88 @@ func TestResolverFail(t *testing.T) {
...
+
+	// try resolve an x86_64 container using a local manifest list
```
mvo5, line 185: this should read "try resolve an **aarch64** ..." now, yes?

pkg/container/client.go
```diff
@@ -343,44 +357,70 @@ func (m RawManifest) Digest() (digest.Digest, error) {
...
+	images, err := store.ImagesByDigest(instance)
+	if err != nil {
+		return "", err
+	}
+	for _, image := range images {
+		// return first result
+		return image.ID, nil
+	}
+	// no id found
+	return "", nil
```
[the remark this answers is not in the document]
kingsleyzissou: Yeah actually this makes more sense

pkg/container/resolver_test.go
kingsleyzissou: This was a lot trickier than expected. I added a test case for the arch name.
And I found that the `arch` wasn't being set - so I put the fix in the follow up pr #635

pkg/container/resolver_test.go
kingsleyzissou: I added a comment... because of overlay storage `t.TempDir` doesn't get cleaned up properly and so the test fails.
So I left it as is

mvo5: Interesting! I (strongly) suspect our own cleanup below (defer os.Remove()) also fails but because we ignore the error we just never noticed.

kingsleyzissou: I just tested and you are correct

kingsleyzissou: Okay pushing a fix for this.

achilleas-k:

pkg/container/client.go
```diff
@@ -398,8 +441,33 @@ func (cl *Client) GetManifest(ctx context.Context, digest digest.Digest, local b
 	}
 
 	if err = retry.RetryIfNecessary(ctx, func() error {
-		r.Data, r.MimeType, err = src.GetManifest(ctx, nil)
-		return err
+		var secondaryDigest *digest.Digest
+		// passing in a digest for a local manifest list will break
+		if instanceDigest != "" && !local {
```
mvo5, line 446: Thanks for the comment here, I am struggling a bit here and wonder if we can somehow tweak it to make it clearer to me but probably followup material.

mvo5 (approved):
Thank you, I think this is fine, I have some suggestions inline but we probably want this in and we can do followups, the only thing I'm struggling with a bit is the logic in the secondaryArch piece, I'm sure it's right but I wonder if we can find a way to make it a bit easier to follow, not a blocker but something I would love to come back to. It's a really nice feature and people want it :)

pkg/container/client.go
```diff
@@ -343,44 +357,73 @@ func (m RawManifest) Digest() (digest.Digest, error) {
 	return manifest.Digest(m.Data)
 }
 
-func getImageRef(target reference.Named, local bool) (types.ImageReference, error) {
+func (cl *Client) getImageRef(target reference.Named, id string, local bool) (types.ImageReference, error) {
```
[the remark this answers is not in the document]
kingsleyzissou: Oh yeah of course, this is already a lot better!

pkg/container/client.go
```diff
@@ -107,6 +108,8 @@ type Client struct {
 
 	UserAgent string // user agent string to use for requests, defaults to DefaultUserAgent
 
+	Store string // another store location other than the main one, useful for testing
```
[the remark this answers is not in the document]
kingsleyzissou: This is a very good point which was a bit of an oversight from me. Happy to move this change to the follow up

pkg/container/client.go
kingsleyzissou: Yeah it might also be a naming thing. It was a bit tricky because this function is called recursively, so if we are getting this info for a single image then it will be empty but if we are calling the function on a manifest list we pass in the child digest. If that makes sense?

mvo5: Yeah, probably followup material at this point (sorry, I really don't want slow things down here). I did play a little bit and was thinking maye:
```go
index d2476f656..42e0b140e 100644
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -396,15 +396,26 @@ func (cl *Client) getLocalImageIDFromDigest(instance digest.Digest) (string, err
 // GetManifest fetches the raw manifest data from the server. If digest is not empty
 // it will override any given tag for the Client's Target.
 func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest, local bool) (r RawManifest, err error) {
-       var id string
-       if instanceDigest != "" && local {
-               id, err = cl.getLocalImageIDFromDigest(instanceDigest)
-               if err != nil {
-                       return r, err
+       var localID string
+       var overrideDigest *digest.Digest
+
+       if instanceDigest != "" {
+               if local {
+                       localID, err = cl.getLocalImageIDFromDigest(instanceDigest)
+                       if err != nil {
+                               return r, err
+                       }
+               } else {
+                       // We can pass the instance digest, if it is nil, then this is the primary manifest.
+                       // If it is not nil, then this is the instance digest and the primary manifest is a
+                       // manifest list. The `GetManifest` call will then retrieve the manifest for the
+                       // desired instance, see:
+                       // https://github.com/containers/image/blob/cdb2f596a95018444f4dee0993e321d3d8bc328d/types/types.go#L252C2-L253C121
+                       overrideDigest = &instanceDigest
                }
        }
 
-       ref, err := cl.getImageRef(id, local)
+       ref, err := cl.getImageRef(localID, local)
        if err != nil {
                return
        }
@@ -425,17 +436,7 @@ func (cl *Client) GetManifest(ctx context.Context, instanceDigest digest.Digest,
        }
 
        if err = retry.RetryIfNecessary(ctx, func() error {
-               var secondaryDigest *digest.Digest
-               // passing in a digest for a local manifest list will break
-               if instanceDigest != "" && !local {
-                       // We can pass the instance digest, if it is nil, then this is the primary manifest.
-                       // If it is not nil, then this is the instance digest and the primary manifest is a
-                       // manifest list. The `GetManifest` call will then retrieve the manifest for the
-                       // desired instance, see:
-                       // https://github.com/containers/image/blob/cdb2f596a95018444f4dee0993e321d3d8bc328d/types/types.go#L252C2-L253C121
-                       secondaryDigest = &instanceDigest
-               }
-               r.Data, r.MimeType, err = src.GetManifest(ctx, secondaryDigest)
+               r.Data, r.MimeType, err = src.GetManifest(ctx, overrideDigest)
                if err != nil {
                        return err
                }
```
would help but I'm not sure. WDYT? I will try to mediate a bit more later (but again, not a blocker, this is really just me)

kingsleyzissou: Yeah I can have a think about it too. But I love the detail and the quality of the review. Thanks again for this. I'll get this in and then start adding some of this to the follow up

PR #599, opened 2024-04-15
Multi bib cache

### LINKED ISSUE #594: CI test caching only stores most recently tested BIB hash causing repeat builds
The CI cache is used to decide if an image should be rebuilt in testing based on whether the image definition or one of the test dependencies has changed.  The following items are considered:
1. Manifest ID (hash of image definition + content).
2. osbuild version.
3. bootc-image-builder container ID.
4. Boot test success (this will rebuild an image if we add it to the image types that can boot test and no information about a previous boot test exists).

The decision is made in the `check_for_build()` function: https://github.com/osbuild/images/blob/5ecb18e4ef9cb086507cc4292e4b9b99df85c981/test/scripts/imgtestlib.py#L246

The manifest ID and osbuild version are part of the path of the cached results, so any change in one of those elements creates a new entry in the cache directory tree, keeping the old result intact.
However, the bootc-image-builder container ID is only part of the metadata in the `info.json` file for each entry.  This means that the following sequence of events can occur and cause unnecessary rebuilds:
1. Cache entry exists at path `<distro>/<arch>/<osbuild-version>/<manifest-id>` and the `info.json` contains `"bib-id": "sha256:aaaa"`.
2. PR#A to update bib ID is created (automatically, daily).  The CI run finds the cache entry, sees that the new bib ID, `sha256:bbbb` doesn't match the one in `info.json` and rebuilds the image and boot tests it.
3. `<distro>/<arch>/<osbuild-version>/<manifest-id>/info.json` is updated and now contains `"bib-id": "sha256:bbbb"`.
4. PR#B is created (before the previous one is merged) with the bib ID defined in the `main` branch.  The CI run finds the cache entry, sees that the bib ID, `sha256:aaaa` doesn't match the one in `info.json` and rebuilds the image and boot tests it.
5. `<distro>/<arch>/<osbuild-version>/<manifest-id>/info.json` is updated and now contains `"bib-id": "sha256:aaaa"` again.
6. PR#A is added to the merge queue and rebuilds the image again like in step 2, changing the bib ID in the cache back to `sha256:bbbb`.

Slightly different sequences of events and more PRs can cause more rebuilds, but this demonstrates a simple version of the issue.
A quick solution would be to include the bib ID in the path like we do with the osbuild version, but only a single image type depends on this in testing, so it wouldn't make sense in other cases.
Another solution would be to separate the bib metadata so we could have:
- `<distro>/<arch>/<osbuild-version>/<manifest-id>/info.json`
- `<distro>/<arch>/<osbuild-version>/<manifest-id>/bib-aaaa.json`: with content `{"boot-success": true}`
- `<distro>/<arch>/<osbuild-version>/<manifest-id>/bib-bbbb.json`: with content `{"boot-success": true}`

Also, we could be a bit smarter with what we test.  If the manifest ID and osbuild version haven't changed, and all we need to do is boot test the image, we could download the image itself from the cache instead of rebuilding it and only run the boot test.

achilleas-k:
On successful boot of an image type supported by bootc-image-builder (Fedora iot-bootable-container), save the bootc-image-builder image ID as a separate (empty) file so we can save the boot success of multiple images and avoid unnecessary rebuilds.

Fixes #594

### DIFF
--- a/test/scripts/imgtestlib.py
+++ b/test/scripts/imgtestlib.py
@@ -4,6 +4,7 @@ import os
 import pathlib
 import subprocess as sp
 import sys
+from glob import glob
 
 TEST_CACHE_ROOT = ".cache/osbuild-images"
 CONFIGS_PATH = "./test/configs"
@@ -243,7 +244,8 @@ def read_manifests(path):
     return manifests
 
 
-def check_for_build(manifest_fname, build_info_path, errors):
+def check_for_build(manifest_fname, build_info_dir, errors):
+    build_info_path = os.path.join(build_info_dir, "info.json")
     # rebuild if matching build info is not found
     if not os.path.exists(build_info_path):
         print(f"Build info not found: {build_info_path}")
@@ -282,11 +284,20 @@ def check_for_build(manifest_fname, build_info_path, errors):
 
         # check if it's a BIB type and compare image IDs
         if image_type in BIB_TYPES:
-            booted_id = dl_config.get("bib-id", None)
+            # Successful boot tests with BIB add a file to the directory as bib-<image ID>. Collect them and compare.
+            bib_ids = glob("bib-*", root_dir=build_info_dir)
+            # add the _old_ bib ID that we used to keep in the info.json
+            config_bib_id = dl_config.get("bib-id")
+            if config_bib_id:
+                bib_ids.append(f"bib-{config_bib_id}")
             bib_ref = get_bib_ref()
             current_id = skopeo_inspect_id(f"docker://{bib_ref}", host_container_arch())
-            if booted_id != current_id:
-                print(f"Container disk image was built with bootc-image-builder {booted_id}")
+            if f"bib-{current_id}" not in bib_ids:
+                if bib_ids:
+                    print("  Container disk image was built with the following bootc-image-builder images:")
+                    print("    - " + "\n    -".join(bib_ids))
+                else:
+                    print("  No bib IDs found.")
                 print(f"  Testing {current_id}")
                 print("  Adding config to build pipeline.")
                 return True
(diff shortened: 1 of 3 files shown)


achilleas-k:
https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/6629880784#L2399
```
🖼️ Manifest fedora_39-x86_64-iot_bootable_container-empty.json was successfully built in commit 647bb751f04dcd02ded4785f6ebc3851d95e66d5
  https://github.com/osbuild/images/commit/647bb751f04dcd02ded4785f6ebc3851d95e66d5
  PR-593: https://github.com/osbuild/images/pull/593
  This image was successfully boot tested
  Container disk image was built with the following bootc-image-builder images:
    - bib-sha256:942465b131205df03298e62b010d3d16723861adc9c6878b980a5f463b52a127
  Testing sha256:942465b131205df03298e62b010d3d16723861adc9c6878b980a5f463b52a127
  Adding config to build pipeline.
```
not good

ondrejbudai (approved):

test/scripts/imgtestlib.py
```diff
@@ -282,11 +284,19 @@ def check_for_build(manifest_fname, build_info_path, errors):
 
         # check if it's a BIB type and compare image IDs
         if image_type in BIB_TYPES:
-            booted_id = dl_config.get("bib-id", None)
+            # Successful boot tests with BIB add a file to the directory as bib-<image ID>. Collect them and compare.
+            bib_ids = glob("bib-*", root_dir=build_info_dir)
+            # add the _old_ bib ID that we used to keep in the info.json
+            if config_bib_id := dl_config.get("bib-id"):
```
ondrejbudai, line 290: Python 3.6 incompatible :/

achilleas-k: I could've sworn I noticed we use this already elsewhere.  But probably in none of the code we run on the CS8 test.

achilleas-k: Fixed.

ondrejbudai (approved):
Thanks!

PR #601, opened 2024-04-15
Distro/RHEL-7: support EC2 (RHUI) image (COMPOSER-2060)

thozza:
Add support for building EC2 (RHUI) image on RHEL-7. It is based on the internal KS previously used for building it using ImageFactory.

Depends on https://github.com/osbuild/osbuild/pull/1729

Notable differences:

- Boot entry: the original image used a nice title as the boot entry ID. osbuild grub2.legacy stage instead uses a generic entry ID as the boot entry and then specifies the same "nice" title in grub.cfg in the menuentry.
- Cloud-init default user: the original image modified the main cloud-init config, while in the ported image, we use a drop-in config.
- Dracut config: original image placed config files in `/etc/dracut.conf.d`, while osbuild stage puts it in `/usr/lib/dracut/dracut.conf.d`.
- Firewall config: original image installed firewalld and then uninstalled it in the %post section. The FW config in the original image was a result of this. Since we do not install firewalld at all in the image, there is no side-effect.
- Modprobe config: original image placed config files in `/etc/modprobe.d`, while osbuild stage puts it in `/usr/lib/modprobe.d`.
- Package set: original image contained `qemu-guest-agent`, which is a side-effect of building the image using Anaconda. Ported image does not contain it. Ported image contains `linux-firmware`, which is a direct dependency of `kernel` and can't be excluded. The original image did a hack in %post, which uninstalled the package, but it would get pulled in on future updates. Ported image also has GPG keys imported.
- Services: `qemu-guest-agent.service` is not enabled on the ported image, because the package is not installed at all.
- Sudoers: original image added `ec2-user` to the `/etc/sudoers`, while the ported image adds a dropin config `/etc/sudoers.d/ec2-user`.
- SELinux ctx mismatch: there are many incorrectly labeled files in the ported image. This is workarounded by forcing the auto-relabel on first boot, which is the case for all RHEL-7 images.

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -8,21 +8,21 @@
   "centos-8": {
     "dependencies": {
       "osbuild": {
-        "commit": "f26e62b23f547fc5ad665cb08cc52e6431657779"
+        "commit": "5b75592fefc4ee2d9e240a15b7002f2537de31b0"
       }
     }
   },
   "centos-9": {
     "dependencies": {
       "osbuild": {
-        "commit": "f26e62b23f547fc5ad665cb08cc52e6431657779"
+        "commit": "5b75592fefc4ee2d9e240a15b7002f2537de31b0"
       }
     }
   },
   "fedora-39": {
     "dependencies": {
       "osbuild": {
-        "commit": "f26e62b23f547fc5ad665cb08cc52e6431657779"
+        "commit": "5b75592fefc4ee2d9e240a15b7002f2537de31b0"
       }
     },
     "repos": [
--- a/pkg/distro/rhel/rhel7/ami.go
+++ b/pkg/distro/rhel/rhel7/ami.go
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	ec2KernelOptions = "console=ttyS0,115200n8 console=tty0 net.ifnames=0 rd.blacklist=nouveau nvme_core.io_timeout=4294967295"
+	ec2KernelOptions = "ro console=ttyS0,115200n8 console=tty0 net.ifnames=0 rd.blacklist=nouveau nvme_core.io_timeout=4294967295 crashkernel=auto LANG=en_US.UTF-8"
 )
 
 func mkEc2ImgTypeX86_64() *rhel.ImageType {
--- a/pkg/distro/rhel/rhel7/ami.go
+++ b/pkg/distro/rhel/rhel7/ami.go
@@ -1,0 +1,289 @@
+package rhel7
+
+import (
+	"os"
+
+	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/pkg/arch"
+	"github.com/osbuild/images/pkg/customizations/fsnode"
+	"github.com/osbuild/images/pkg/disk"
+	"github.com/osbuild/images/pkg/distro"
+	"github.com/osbuild/images/pkg/distro/rhel"
+	"github.com/osbuild/images/pkg/osbuild"
+	"github.com/osbuild/images/pkg/rpmmd"
+)
+
+const (
+	ec2KernelOptions = "console=ttyS0,115200n8 console=tty0 net.ifnames=0 rd.blacklist=nouveau nvme_core.io_timeout=4294967295"
+)
+
+func mkEc2ImgTypeX86_64() *rhel.ImageType {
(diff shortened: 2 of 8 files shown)


thozza:
[rhel-server-ec2-7.9-image-builder-port.img-info.json](https://github.com/osbuild/images/files/14984073/rhel-server-ec2-7.9-image-builder-port.img-info.json)
[rhel-server-ec2-7.9-160.x86_64.img-info.json](https://github.com/osbuild/images/files/14984074/rhel-server-ec2-7.9-160.x86_64.img-info.json)

achilleas-k:
> Add support for building EC2 (RHUI) image on RHEL-7. It is based on the internal KS previously used for building it using ImageFactory.
> 
> Depends on [osbuild/osbuild#1729](https://github.com/osbuild/osbuild/pull/1729)
> 
> Notable differences:
> 
> * Boot entry: the original image used a nice title as the boot entry ID. osbuild grub2.legacy stage instead uses a generic entry ID as the boot entry and then specifies the same "nice" title in grub.cfg in the menuentry.

If this was for any other distro I'd say it might be nice to do it the way the original image does but I think it's fine for this case.  I wouldn't want to change grub2.legacy for this.

> * Cloud-init default user: the original image modified the main cloud-init config, while in the ported image, we use a drop-in config.

Drop-in config is the "correct" way in this case anyway, so I'm counting this as an improvement.

> * Dracut config: original image placed config files in `/etc/dracut.conf.d`, while osbuild stage puts it in `/usr/lib/dracut/dracut.conf.d`.
> * Modprobe config: original image placed config files in `/etc/modprobe.d`, while osbuild stage puts it in `/usr/lib/modprobe.d`.

:+1:

> * Firewall config: original image installed firewalld and then uninstalled it in the %post section. The FW config in the original image was a result of this. Since we do not install firewalld at all in the image, there is no side-effect.

Another improvement.

> * Package set: original image contained `qemu-guest-agent`, which is a side-effect of building the image using Anaconda. Ported image does not contain it. Ported image contains `linux-firmware`, which is a direct dependency of `kernel` and can't be excluded. The original image did a hack in %post, which uninstalled the package, but it would get pulled in on future updates. Ported image also has GPG keys imported.
> * Services: `qemu-guest-agent.service` is not enabled on the ported image, because the package is not installed at all.

More improvements.

> * Sudoers: original image added `ec2-user` to the `/etc/sudoers`, while the ported image adds a dropin config `/etc/sudoers.d/ec2-user`.

Drop-in good.

> * SELinux ctx mismatch: there are many incorrectly labeled files in the ported image. This is workarounded by forcing the auto-relabel on first boot, which is the case for all RHEL-7 images.

:(


> [rhel-server-ec2-7.9-160.x86_64.img-info.json](https://github.com/osbuild/images/files/14984074/rhel-server-ec2-7.9-160.x86_64.img-info.json) [rhel-server-ec2-7.9-image-builder-port.img-info.json](https://github.com/osbuild/images/files/14994218/rhel-server-ec2-7.9-image-builder-port.img-info.json)

Awesome.  Thanks for these!!

achilleas-k (dismissed):
Love the comments for the weird bits in the image config!  TY!
LGTM

kingsleyzissou (approved):

achilleas-k:
EC2 boot test failing on kdump service not starting :(
```
● kdump.service                                                           loaded failed failed    Crash recovery kernel arming
```

achilleas-k:
We've had similar issues before, solved by using larger instances: https://issues.redhat.com/browse/COMPOSER-2050
Might have to check the exact message on a running system though.

ondrejbudai (approved):
Looks good! :)

thozza:
> EC2 boot test failing on kdump service not starting :(
> 
> ```
> ● kdump.service                                                           loaded failed failed    Crash recovery kernel arming
> ```

So the "funny" thing is that the KS has `kdump` service disabled, but then it is enabled in the final image after the build :woozy_face: So I can disable it, but then it will be really disabled in the final image, which would be technically a difference :shrug:

achilleas-k:
Three options, I guess:
1. Disable kdump in our image
2. Disable kdump in our test config and investigate later
3. Investigate now and delay the PR

achilleas-k (approved):
Oh cool!!

PR #604, opened 2024-04-16
image: bring back the pre-PR#462 `BootcDiskImage`

mvo5:
This commit brings back the pre-PR#462 version of BootcDiskImage as BootcLegacyDiskImage. The rational is that right now we cannot use `bootc install to-filesystem` for cross arch building the FITRIM ioctl is not supported by qemu-user

To fix this is requires any of the following:
1. ensure https://github.com/containers/bootc/pull/462 is available everywhere we care about
2. ensure https://www.mail-archive.com/qemu-devel@nongnu.org/msg1034508.html is available everywhere we care about
3. bind mount /usr/bin/true over /usr/bin/fsfreeze inside the buildroot. This bind mount can be done via `org.osbuild.bind` but it leaks into the manifest and that does not feel great.

So the alternative is to keep using the "old" pipeline that does not yet use `bootc install to-filesystem` to for cross-arch builds and the new pipeline for everything else. The selection will happen on the level of `bib`.

This is all not nice and fragile and needs to be removed/reverted again but short term might be the only thing we can do.

With that we can do in bib:
```
index 1692a03..eb3b1cb 100644
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -126,7 +126,13 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
        mf := manifest.New()
        mf.Distro = manifest.DISTRO_FEDORA
        runner := &runner.Linux{}
-       err = img.InstantiateManifestFromContainers(&mf, []container.SourceSpec{containerSource}, runner, rng)
+
+       if c.Architecture != arch.Current() {
+               legacyImg := image.NewBootcLegacyDiskImage(img)
+               err = legacyImg.InstantiateManifestFromContainers(&mf, []container.SourceSpec{containerSource}, runner, rng)
+       } else {
+               err = img.InstantiateManifestFromContainers(&mf, []container.SourceSpec{containerSource}, runner, rng)
+       }
 
        return &mf, err
 }

```

### DIFF
--- a/pkg/image/bootc_disk_legacy.go
+++ b/pkg/image/bootc_disk_legacy.go
@@ -1,0 +1,86 @@
+package image
+
+import (
+	"fmt"
+	"math/rand"
+
+	"github.com/osbuild/images/pkg/container"
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/images/pkg/osbuild"
+	"github.com/osbuild/images/pkg/runner"
+)
+
+// Legacy pipeline to create bootc images. This can be removed once
+//
+//	https://github.com/containers/bootc/pull/462
+//
+// or
+//
+//	https://www.mail-archive.com/qemu-devel@nongnu.org/msg1034508.html
+//
+// is available everyhwere
+type BootcLegacyDiskImage struct {
+	bootcImg *BootcDiskImage
+}
+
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):
Thanks! :)

PR #608, opened 2024-04-17
stages/systemd.unit.create: fix DefaultDependencies option

say-paul:
Added a function to pass the bool as pointer.

DefaultDependencies in systemd  is by default set as true and needs to be exluscively set as false which is oppositeof  golangs syntax omitempty. Leaving out omitempty causes the default dependencies to be set as false by default which will cause the service to behave abnormally as its by default set to true in systemd world.

This PR will fix the behavior 

Also refactored the bash scripts for osbuild-ostree-mountpoints.service requird for custom-file-system feature


pkg/manifest/ostree_deployment.go
```diff
@@ -513,7 +513,7 @@ func createMountpointService(serviceName string, mountpoints []string) *osbuild.
 	}
 	unit := osbuild.Unit{
 		Description:              "Ensure custom filesystem mountpoints exist",
-		DefaultDependencies:      false,
+		DefaultDependencies:      osbuild.SetServiceDefaultDependencies(false),
```
achilleas-k, line 516: We have a common function for this: `common.ToPtr()`

achilleas-k (changes requested):

achilleas-k:
Thanks for spotting this and fixing it.

pkg/manifest/ostree_deployment.go
say-paul: fixed in ddeca965c3dccefb544b2df99be78d68f3637f46

achilleas-k (dismissed):

achilleas-k:
Something's not right with the change in the systemd service's exec line changes because the boot test is showing that service is failing to start now.

achilleas-k:
I opened #613 that prints failed unit status outputs in the boot test so cases like these are easier to troubleshoot without reproducing by hand.

achilleas-k:
@say-paul `edge-ami` boot tests are failing with:
```
❌ Status for all failed units
× osbuild-ostree-mountpoints.service - Ensure custom filesystem mountpoints exist
     Loaded: loaded (/etc/systemd/system/osbuild-ostree-mountpoints.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Tue 2024-04-23 15:10:35 EDT; 55s ago
    Process: 800 ExecStartPre=/bin/sh -c grep -Uqs composefs /run/ostree-booted || chattr -i / (code=exited, status=0/SUCCESS)
    Process: 807 ExecStart=mkdir -p /foo /foo/bar /root /mnt /srv /opt /var/mydata (code=exited, status=1/FAILURE)
    Process: 808 ExecStopPost=/bin/sh -c grep -Uqs composefs /run/ostree-booted || chattr +i / (code=exited, status=0/SUCCESS)
   Main PID: 807 (code=exited, status=1/FAILURE)
        CPU: 21ms
Apr 23 15:10:35 localhost mkdir[807]: mkdir: cannot create directory ‘/root’: File exists
Apr 23 15:10:35 localhost mkdir[807]: mkdir: cannot create directory ‘/srv’: File exists
Apr 23 15:10:35 localhost mkdir[807]: mkdir: cannot create directory ‘/opt’: File exists
terminating instance i-018c92151576f4400
deleting security group sg-0e5b3f74ac8458cca
deleting EC2 image ami-06894585993d267c6 and snapshot snap-070e520d98107490f
exit status 3
exit status 1
```
https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/6699182166

say-paul:
hmmh..!! thought -p would take care of it

achilleas-k:
Yeah it's strange.  I wonder if the change in the way the `chattr` is handled isn't working properly so the immutable flag isn't being unset.

achilleas-k:

achilleas-k:
Replaced by #668

PR #610, opened 2024-04-17
test/imgtestlib: download bib-id files with info files

achilleas-k:
We can't check if a bootable container was tested with a version of bootc-image-builder if we don't download the files.


ochosi (approved):
LGTM

croissanne (approved):

PR #613, opened 2024-04-18
test/base-host-check: run systemctl status --failed

achilleas-k:
Run systemctl status --failed if the is-system-running call fails.  This will print the status output of all failed units, which can save a lot of time troubleshooting when the failure cause isn't obvious.


say-paul:
cc @mcattamoredhat

PR #614, opened 2024-04-18
image,manifest,osbuild: support kickstart kernel arguments for the iso container installer

mvo5:
Support for osbuild for this option landed in https://github.com/osbuild/osbuild/pull/1742

The bib bits are in https://github.com/osbuild/bootc-image-builder/pull/370

### DIFF
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -34,6 +34,10 @@ type AnacondaInstallerISOTree struct {
 	Keyboard *string
 	Timezone *string
 
+	// Kernel options that will be apended to the installed system
+	// (not the iso)
+	KickstartKernelOptionsAppend []string
+
 	// Create a sudoers drop-in file for each user or group to enable the
 	// NOPASSWD option
 	NoPasswd []string
@@ -67,6 +71,7 @@ type AnacondaInstallerISOTree struct {
 	ContainerSource  *container.SourceSpec
 	containerSpec    *container.Spec
 
+	// Kernel options for the ISO image
 	KernelOpts []string
 
 	// Enable ISOLinux stage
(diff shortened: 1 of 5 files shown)


pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -34,6 +34,8 @@ type AnacondaInstallerISOTree struct {
 	Keyboard *string
 	Timezone *string
 
+	KernelOptionsAppend []string
```
mvo5, line 37: This should have a comment because there is also "KernelOpts" below that appears (85% confidence) about kernel opts for the iso itself not the target system

achilleas-k: The separation of options between the ISO and the image it's deploying is a frequent source of confusion.
I would say for now call it `ImageKernelOptionsAppend` or rename the existing one `ISOKernelOpts`.  In the long run (or here, if you're feeling up to it and have time), we might want to separate them into substructs so the isotree pipeline has, for example:
```go
isoTree.ISO.Timezone = "utc"
isoTree.ISO.KernelOptions = stuff
...
isoTree.Kickstart.Timezone = "utc"
isoTree.Kickstart.KernelOptions = stuff + "some image option"
```

It's often the case that you want a lot of these things to be the same.  If a user is building an ISO to install a system and wants a CET timezone to be set in their image through the kickstart (because it's deploying a generic ostree payload), or needs a kernel arg to make it boot on specific hardware, it makes sense that the ISO should have the same timezone and kernel args.  But separating them makes it clear for things that _might_ be different, like specific kernel args for the ISO or the image that would differ.

mvo5: Thanks! I like this suggestion, I went with `KickstartKernelOptionsAppend` and added some (light) testing. I think this is good enough for now (I still want a simpler stratey for testing when we pass options into an image and then pass that option on to a manifest but that is something I will think about another day :)

ondrejbudai (approved):

achilleas-k (approved):

PR #627, opened 2024-04-23
Remove RHC credentials from the system after registration (HMS-3814)

achilleas-k:
This PR swaps out the `org.osbuild.first-boot` stage with an `org.osbuild.systemd.unit.create` stage that runs the same commands and explicitly adds the `rm` command for the condition file as the last command.  The file name and service description are changed to describe the purpose of the service.

The stage is functionally identical with the previous one with one main difference: the organisation ID and the activation key are not included in the service file but in a separate environment file in `/etc` which also acts as the condition file and gets removed after the service runs for the first time.

Not that on ostree-based systems, putting the environment file in `/etc` prior to building the commit and then `rm`ing it on boot means that the file remains in `/usr/etc`, in the base system, until a system update removes it.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -37,8 +37,8 @@ func check(err error) {
 
 type BuildConfig struct {
 	Name      string               `json:"name"`
-	OSTree    *ostree.ImageOptions `json:"ostree,omitempty"`
 	Blueprint *blueprint.Blueprint `json:"blueprint,omitempty"`
+	Options   distro.ImageOptions  `json:"options"`
 	Depends   interface{}          `json:"depends,omitempty"` // ignored
 }
 
@@ -61,8 +61,7 @@ func loadConfig(path string) BuildConfig {
 func makeManifest(imgType distro.ImageType, config BuildConfig, distribution distro.Distro, repos []rpmmd.RepoConfig, archName string, seedArg int64, cacheRoot string) (manifest.OSBuildManifest, error) {
 	cacheDir := filepath.Join(cacheRoot, archName+distribution.Name())
 
-	options := distro.ImageOptions{Size: 0}
-	options.OSTree = config.OSTree
+	options := config.Options
 
 	// add RHSM fact to detect changes
 	options.Facts = &facts.ImageOptions{
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -558,17 +558,27 @@ func (p *OS) serialize() osbuild.Pipeline {
 	// - Register with subscription-manager, no Insights or rhc
 	// - Register with subscription-manager and enable Insights, no rhc
 	if p.Subscription != nil {
-		subkeyFilepath := "/etc/osbuild-first-boot"
+		// Write a key file that will contain the org ID and activation key to be sourced in the systemd service.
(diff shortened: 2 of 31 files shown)


cmd/build/main.go
```diff
@@ -36,10 +37,11 @@ func check(err error) {
...
 
 type BuildConfig struct {
-	Name      string               `json:"name"`
-	OSTree    *ostree.ImageOptions `json:"ostree,omitempty"`
-	Blueprint *blueprint.Blueprint `json:"blueprint,omitempty"`
-	Depends   interface{}          `json:"depends,omitempty"` // ignored
+	Name         string                     `json:"name"`
+	OSTree       *ostree.ImageOptions       `json:"ostree,omitempty"`
+	Blueprint    *blueprint.Blueprint       `json:"blueprint,omitempty"`
+	Subscription *subscription.ImageOptions `json:"subscription,omitempty"`
```
thozza, line 43: Wouldn't it make sense to add the the whole `distro.ImageOptions` instead?

thozza (dismissed):
LGTM. I've added one suggestion.

cmd/build/main.go
achilleas-k: Probably a good idea yeah.  Will need to change the test configs for ostree.  But good idea.

achilleas-k: Done!

pkg/manifest/os.go
```diff
@@ -558,16 +558,27 @@ func (p *OS) serialize() osbuild.Pipeline {
...
-			commands = []string{fmt.Sprintf("/usr/bin/rhc connect -o=%s -a=%s --server %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl)}
+			commands = []string{fmt.Sprintf("/usr/bin/rhc connect --organization=${ORG_ID} --activation-key=${ACTIVATION_KEY} --server %s", p.Subscription.ServerUrl)}
 			// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
 			commands = append(commands, "restorecon -R /root/.gnupg")
 			// execute the rhc post install script as the selinuxenabled check doesn't work in the buildroot container
 			commands = append(commands, "/usr/sbin/semanage permissive --add rhcd_t")
 		} else {
-			commands = []string{fmt.Sprintf("/usr/sbin/subscription-manager register --org=%s --activationkey=%s --serverurl %s --baseurl %s", p.Subscription.Organization, p.Subscription.ActivationKey, p.Subscription.ServerUrl, p.Subscription.BaseUrl)}
+			commands = []string{fmt.Sprintf("/usr/sbin/subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY} --serverurl %s --baseurl %s", p.Subscription.ServerUrl, p.Subscription.BaseUrl)}
```
loadtheaccumulator, line 581: @achilleas-k as you pointed out, the creds vars will remain in /usr/etc/osbuild-first-boot due to the way ostree works with /etc. I was originally using a vars file in /root for Edge Management, and that might work here.

thozza (approved):
LGTM :+1:

achilleas-k:

PR #631, opened 2024-04-24
Add support for arbitrary kickstart file injection into ISOs

achilleas-k:
Adds a new blueprint customization `[installer.kickstart.contents]` that will allow users to inject their own kickstart file into an ISO build.

This PR also includes a partial implementation of a recommendation I made in a previous PR: https://github.com/osbuild/images/pull/614#discussion_r1572325795

A new internal customization struct called `kickstart.Options` holds all kickstart options that we can set in pipelines and is attached to Anaconda pipelines and image kinds that support them.  This is similar to other internal options structs like users.User, users.Group, fsnode.File, etc.
There's a bit more we can do with this (for example a `ks.GetPath()` method that returns the default when it's not set, even with `ks == nil`, instead of checking all the time) but I wanted to get this PR open and merged to get this into bootc-image-builder.

Note that our initial idea was to make any other kickstart option illegal when this option is used.  Leaving it wide open can cause all sorts of issues for users who don't know what we're adding to the kickstart when they append their own, but from what I understand, it should be safe in the general case where they add some `%post` bits or enable options we don't support.

I'm fine with restricting it if others disagree.


achilleas-k:

thozza (commented):
Nice work. Overall this looks good, but one commit needs cleanup.

https://github.com/osbuild/images/pull/631/commits/bd6234b82de5c6f2e82dd023586c2a133281d4d3 contains a syntax error, which is then fixed by https://github.com/osbuild/images/pull/631/commits/d6399503140677ddc226c27a6aafe361938780bd.

pkg/manifest/anaconda_installer_iso_tree_test.go
```diff
@@ -121,16 +122,48 @@ func findRawKickstartFileStage(stages []*osbuild.Stage) *osbuild.CopyStageOption
...
+
+	content += extra
+
+func calculateInlineFileChecksum(parts ...string) string {
+	content := "%include /run/install/repo/test-base.ks\n"
+	for _, part := range parts {
+		content += "\n" + part
+	}
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(content)))
+}
```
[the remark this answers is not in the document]
achilleas-k: wow.  Rebase madness.  Thanks!

thozza (approved):
:+1:

pkg/customizations/kickstart/kickstart.go
```diff
@@ -0,0 +1,46 @@
...
+	KernelOptionsAppend []string
+
+	// Enable networking on on boot in the installed system
+	NetworkOnBoot bool
+
+	Language *string
+	Keyboard *string
+	Timezone *string
+
+	OSName string
```
bcl, line 36: These are for OSTree, right? Why not use the OSTree struct defined above?

bcl:
I wonder if we should run `ksvalidator` on the user blob at some point, either here or in osbuild? As long as you have a newer version of pykickstart installed it can validate other supported versions by passing it `--version RHEL9` for example.

pkg/customizations/kickstart/kickstart.go
achilleas-k: I am looking at this and scratching my head because I'm sure that was the intention but I somehow went a different way halfway through.

achilleas-k: Alright, fixed.  Thanks for catching this!!

achilleas-k:
> I wonder if we should run `ksvalidator` on the user blob at some point, either here or in osbuild?

Could be a good idea, yeah.  Though, it gets tricky with cross-distro builds doesn't it?  If you're building a RHEL N image on RHEL N+1, you'd want to run it through the validator for RHEL N, not the newer one.  So it should probably happen in the buildroot in osbuild I guess?

bcl:
As long as it knows the release name that pykickstart recoginizes (RHEL9, RHEL10, etc.) and is running on the same or newer release it will work. No need to try and work this into this PR though, it can come later.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -473,6 +473,10 @@ bootc switch --mutate-in-place --transport %s %s
 %%end
 `, targetContainerTransport, p.containerSpec.LocalName)
 
+	if p.Kickstart.UserFile != nil {
+		hardcodedKickstartBits += "\n" + p.Kickstart.UserFile.Contents
```
ondrejbudai, line 477: Hmm, doesn't that mean that the user cannot override partitioning? Does it mean that we need to use just the user-provided kickstart ig given and skip our parts?

achilleas-k: Well, _it's complicated and it depends_.

I see a few ways out of this:
1. Document what we add to the kickstart and how the user's is added so they can do their own bits and figure it out themselves.
2. Document what we add to the kickstart and tell users that if they add a kickstart they need to add or replace the bits we add because their file will replace ours.
3. Use the same options we do in the other image types (like edge installer) so the user can control what gets added by enabling or disabling the `Unattended` option and choose if they want to have that or not.

achilleas-k:
After discussing with @ondrejbudai and @mvo5, we decided that user kickstarts combined with any other kickstart customization would not be supported and produce errors.  So now, if a user adds a kickstart file/contents, we only add the install line for the payload and then have the user's kickstart file include ours.

For that last part, it would be nicer if our kickstart had an `%include` to the user's, but that requires osbuild stage changes so we'll do it later.

achilleas-k:
The test config file, `test/configs/interactive-iso-userkickstart.json`, creates the following in the root of the ISO:
```
› head -n-0 *.ks
==> osbuild-base.ks <==
liveimg --url file:///run/install/repo/liveimg.tar.gz

==> osbuild.ks <==
%include /run/install/repo/osbuild-base.ks
%post

echo "=== Custom user kickstart ==="

echo -e "%sudo\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/%sudo"
chmod 0440 /etc/sudoers.d/%sudo
echo -e "%wheel\tALL=(ALL)\tNOPASSWD: ALL" > "/etc/sudoers.d/%wheel"
chmod 0440 /etc/sudoers.d/%wheel
restorecon -rvF /etc/sudoers.d
%end
```

pkg/distro/fedora/imagetype.go
```diff
@@ -418,11 +418,24 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
...
 
-	if customizations.GetInstaller() != nil {
+	if instCust, err := customizations.GetInstaller(); instCust != nil {
 		// only supported by the Anaconda installer
 		if slices.Index([]string{"iot-installer"}, t.name) == -1 {
-			return nil, fmt.Errorf("installer customizations are not supported for %q", t.name)
+			return nil, fmt.Errorf("installer customizations are not supported for %q", t.Name())
 		}
+
+		// NOTE: the image type check is redundant with the check above, but
```
mvo5, line 427: Pardon my ignorance, what redundant with what check above exactly? It's not the check from GetInstaller(), iirc/aiui it does not check for user cusomizations.

pkg/distro/rhel/images.go
```diff
@@ -487,17 +488,28 @@ func EdgeInstallerImage(workload workload.Workload,
 
 	img.Platform = t.platform
 	img.ExtraBasePackages = packageSets[InstallerPkgsKey]
-	img.Users = users.UsersFromBP(customizations.GetUsers())
-	img.Groups = users.GroupsFromBP(customizations.GetGroups())
+	img.Kickstart = &kickstart.Options{
```
mvo5, line 491: (idle wondering) This makes me wonder if we should have a "kickstart.New(customizations *blueprint.Customizations)" ? But maybe they are all too diffrent.

pkg/distro/rhel/images.go
```diff
@@ -487,17 +488,28 @@ func EdgeInstallerImage(workload workload.Workload,
...
+	img.Kickstart.Language, img.Kickstart.Keyboard = customizations.GetPrimaryLocale()
 	// ignore ntp servers - we don't currently support setting these in the
 	// kickstart though kickstart does support setting them
-	img.Timezone, _ = customizations.GetTimezoneSettings()
+	img.Kickstart.Timezone, _ = customizations.GetTimezoneSettings()
 
-	if instCust := customizations.GetInstaller(); instCust != nil {
-		img.NoPasswd = instCust.SudoNopasswd
-		img.UnattendedKickstart = instCust.Unattended
+	if instCust, err := customizations.GetInstaller(); err != nil {
```
mvo5, line 505: Another candidate for a helper or kickstart.Merge() :)

pkg/distro/rhel/images.go
```diff
@@ -676,8 +687,22 @@ func ImageInstallerImage(workload workload.Workload,
...
-	img.Users = users.UsersFromBP(customizations.GetUsers())
-	img.Groups = users.GroupsFromBP(customizations.GetGroups())
+	img.Kickstart = &kickstart.Options{
+		Users:    users.UsersFromBP(customizations.GetUsers()),
+		Groups:   users.GroupsFromBP(customizations.GetGroups()),
+		Language: &img.OSCustomizations.Language,
+		Keyboard: img.OSCustomizations.Keyboard,
+		Timezone: &img.OSCustomizations.Timezone,
+	}
+	if instCust, err := customizations.GetInstaller(); err != nil {
```
mvo5, line 697: Potential candidate for `img.Kickstart.Merge()`

pkg/distro/rhel/rhel10/options.go
```diff
@@ -81,11 +81,13 @@ func checkOptions(t *rhel.ImageType, bp *blueprint.Blueprint, options distro.Ima
 		warnings = append(warnings, w)
 	}
 
-	if customizations.GetInstaller() != nil {
+	if instCust, err := customizations.GetInstaller(); instCust != nil {
```
mvo5, line 84: Same idea/suggestion as before, I would do the error check first here to make it more linear.

pkg/distro/rhel/rhel8/options.go
```diff
@@ -175,11 +175,20 @@ func checkOptions(t *rhel.ImageType, bp *blueprint.Blueprint, options distro.Ima
 		warnings = append(warnings, w)
 	}
 
-	if customizations.GetInstaller() != nil {
+	if instCust, err := customizations.GetInstaller(); instCust != nil {
```
mvo5, line 178: And here, I would go linear(-b)

pkg/distro/rhel/rhel9/options.go
```diff
@@ -194,11 +194,20 @@ func checkOptions(t *rhel.ImageType, bp *blueprint.Blueprint, options distro.Ima
 		warnings = append(warnings, w)
 	}
 
-	if customizations.GetInstaller() != nil {
+	if instCust, err := customizations.GetInstaller(); instCust != nil {
```
mvo5, line 197: Also more linear here maybe?

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -415,11 +397,24 @@ func (p *AnacondaInstallerISOTree) ostreeContainerStages() []*osbuild.Stage {
...
 
+	stages = append(stages, p.bootcInstallerKickstartStages()...)
+	return stages
+}
+
+// bootcInstallerKickstartStages sets up kickstart-related stages for Anaconda
+// ISOs that install a bootc bootable container.
+func (p *AnacondaInstallerISOTree) bootcInstallerKickstartStages() []*osbuild.Stage {
+	if p.Kickstart == nil {
+		panic(fmt.Sprintf("Kickstart options not set for %s pipeline", p.name))
```
mvo5, line 408: slightly sad but I guess there is no other way right now - maybe `fmt.Errorf()` so that we prepare for a time when have error instead of panics (but actually ignore me, that is trivial enough to change once we add more errors)

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -522,47 +541,86 @@ func (p *AnacondaInstallerISOTree) tarPayloadStages() []*osbuild.Stage {
...
+		kickstartOptions = new(kickstart.Options)
+	}
+
 	stages := make([]*osbuild.Stage, 0)
-	if p.UnattendedKickstart {
+
+	// users, groups, and other kickstart options are not allowed when users
+	// add their own kickstarts
+	if kickstartOptions.UserFile != nil {
+		// check if any other option is set and panic - these combinations
```
mvo5, line 555: Maybe worthwhile to move the validation policy into `kickstart.go` ? This might be helpful in other places too where we need to validate that the kickstart options work(?). Something like:
```go
diff --git a/pkg/customizations/kickstart/kickstart.go b/pkg/customizations/kickstart/kickstart.go
index 70918e3f9..bf20e700b 100644
--- a/pkg/customizations/kickstart/kickstart.go
+++ b/pkg/customizations/kickstart/kickstart.go
@@ -1,6 +1,10 @@
 package kickstart
 
-import "github.com/osbuild/images/pkg/customizations/users"
+import (
+       "fmt"
+
+       "github.com/osbuild/images/pkg/customizations/users"
+)
 
 type File struct {
        Contents string
@@ -45,3 +49,27 @@ type Options struct {
        // User-defined kickstart files that will be added to the ISO
        UserFile *File
 }
+
+func (kickstartOptions *Options) Validate() error {
+       // users, groups, and other kickstart options are not allowed when users
+       // add their own kickstarts
+       if kickstartOptions.UserFile != nil {
+               // check if any other option is set and panic - these combinations
+               // should be verified by the caller
+               if kickstartOptions.Unattended {
+                       return fmt.Errorf("kickstart unattended options are not compatible with user-supplied kickstart content")
+               }
+
+               if len(kickstartOptions.SudoNopasswd) > 0 {
+                       return fmt.Errorf("kickstart sudo nopasswd %q drop-in file creation is not compatible with user-supplied kickstart content", kickstartOptions.SudoNopasswd)
+               }
+
+               // options are usually already initialised from outside this function
+               // with the payload options (ostree commit or tarball), but might also
+               // have Users and Groups added
+               if len(kickstartOptions.Users)+len(kickstartOptions.Groups) > 0 {
+                       return fmt.Errorf("kickstart users %v and/or groups %v are not compatible with user-supplied kickstart content", kickstartOptions.Users, kickstartOptions.Groups)
+               }
+       }
+       return nil
+}
diff --git a/pkg/manifest/anaconda_installer_iso_tree.go b/pkg/manifest/anaconda_installer_iso_tree.go
index 10816bfcc..51e0d588c 100644
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -424,16 +424,9 @@ func (p *AnacondaInstallerISOTree) bootcInstallerKickstartStages() []*osbuild.St
        }
 
        if p.Kickstart.UserFile != nil {
-               // when a user defines their own kickstart, we create a kickstart that
-               // takes care of the installation and let the user kickstart handle
-               // everything else
-
-               // users and groups are NOT allowed when users add their own kickstarts
-               if len(kickstartOptions.Users)+len(kickstartOptions.Groups) > 0 {
-                       // this is a programming error - the combinations should have been verified already
-                       panic("kickstart users and/or groups are not compatible
-               // when a user defines their own kickstart, we create a kickstart that
-               // takes care of the installation and let the user kickstart handle
-               // everything else
-
-               // users and groups are NOT allowed when users add their own kickstarts
-               if len(kickstartOptions.Users)+len(kickstartOptions.Groups) > 0 {
-                       // this is a programming error - the combinations should have been verified already
-                       panic("kickstart users and/or groups are not compatible with user-supplied kickstart content")
+               if err := p.Kickstart.Validate(); err != nil {
+                       panic(err)
                }
-
                stages = append(stages, osbuild.NewKickstartStage(kickstartOptions))
                kickstartFile, err := kickstartOptions.IncludeRaw(p.Kickstart.UserFile.Contents)
                if err != nil {

```

pkg/distro/fedora/imagetype.go
achilleas-k: The redundancy is between:
```go
if slices.Index([]string{"iot-installer"}, t.name) == -1 { ... }
```
and the first part in the `&&` chain below
```
if t.Name() == "iot-installer" && ...
```

The first one should list all Anaconda installers but there's an open issue about the `image-installer` (which was buried in a closed PR and I transferred to an issue now #672).
The second check should only be for the `iot-installer`.

pkg/distro/rhel/images.go
achilleas-k: I think that might become conditioned on the image type.  It would help clarify the behaviour.

Actually, thinking about it now I realise this would be a good way to centralise all the interactions and consider why they differ if they do.  I wouldn't mind blocking on this.

mvo5 (changes requested):
Setting to "request changes" as requested by @achilleas-k

pkg/manifest/anaconda_installer_iso_tree.go
achilleas-k: Every other panic in the function is a string, so I'd rather keep it like this for consistency there.

noelmiller:
One major issue I see for this is the User Module is disabled in Anaconda, meaning if I don't specify a user in the kickstart, I cannot create a user in Anaconda: 
image

pkg/customizations/kickstart/kickstart.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+	}
+
+	if err := options.Validate(); err != nil {
+		return nil, err
+	}
+	return options, nil
+}
+
+func (options Options) Validate() error {
```
mvo5, line 78: It's a much bigger change and I'm not suggesting to do it but I want to mention it. This might be a case where having getters/setters is quite useful because then the state of options can never be invalid and the user does not need to remember to call validate() before using the options. But not now, just an idle thought.

pkg/customizations/kickstart/kickstart.go
```diff
@@ -0,0 +1,93 @@
...
+	Groups []users.Group
+
+	// ostree-related kickstart options
+	OSTree *OSTree
+
+	// User-defined kickstart files that will be added to the ISO
+	UserFile *File
+}
+
+func New(customizations *blueprint.Customizations) (*Options, error) {
```
mvo5, line 54: I really like how this turned out

pkg/distro/fedora/imagetype.go
```diff
@@ -418,10 +418,25 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
...
-			return nil, fmt.Errorf("installer customizations are not supported for %q", t.name)
+			return nil, fmt.Errorf("installer customizations are not supported for %q", t.Name())
+		}
+
+		// NOTE: the image type check is redundant with the check above, but
+		// let's keep it explicit in case one of the two changes.
+		// The kickstart contents is incompatible with the users and groups
+		// customization only for the iot-installer.
+		if t.Name() == "iot-installer" &&
+			instCust.Kickstart != nil &&
```
mvo5, line 436: This check here (and the other similar ones) seem to reimplement kickstart.Options.Validate() (in slightly different words) (also in customization.GetInstaller() similar validation is done) - I wonder if we can consolidate the validation maybe into a single place ideally? I poked around a little bit and one way would be to just use `kickstart.New().Validate()` here which seems to be equivalent to "GetInstaller()" (needs some small tweaks to kickstart.New() that are probably useful anyway, i.e. return nil if nothing is to kickstart but this might also be dangerous, needs some more thought).

Something like:
```go
diff --git a/pkg/customizations/kickstart/kickstart.go b/pkg/customizations/kickstart/kickstart.go
index 4b5897d59..6a92d0baa 100644
--- a/pkg/customizations/kickstart/kickstart.go
+++ b/pkg/customizations/kickstart/kickstart.go
@@ -52,23 +52,23 @@ type Options struct {
 }
 
 func New(customizations *blueprint.Customizations) (*Options, error) {
-       options := &Options{
-               Users:  users.UsersFromBP(customizations.GetUsers()),
-               Groups: users.GroupsFromBP(customizations.GetGroups()),
-       }
-
        instCust, err := customizations.GetInstaller()
        if err != nil {
                return nil, err
        }
-       if instCust != nil {
-               options.SudoNopasswd = instCust.SudoNopasswd
-               options.Unattended = instCust.Unattended
-               if instCust.Kickstart != nil {
-                       options.UserFile = &File{Contents: instCust.Kickstart.Contents}
-               }
+       if instCust == nil {
+               return nil, nil
        }
 
+       options := &Options{
+               Users:  users.UsersFromBP(customizations.GetUsers()),
+               Groups: users.GroupsFromBP(customizations.GetGroups()),
+       }
+       options.SudoNopasswd = instCust.SudoNopasswd
+       options.Unattended = instCust.Unattended
+       if instCust.Kickstart != nil {
+               options.UserFile = &File{Contents: instCust.Kickstart.Contents}
+       }
        if err := options.Validate(); err != nil {
                return nil, err
        }
diff --git a/pkg/distro/fedora/imagetype.go b/pkg/distro/fedora/imagetype.go
index ba1cc926c..4244b9835 100644
--- a/pkg/distro/fedora/imagetype.go
+++ b/pkg/distro/fedora/imagetype.go
@@ -10,6 +10,7 @@ import (
        "github.com/osbuild/images/internal/workload"
        "github.com/osbuild/images/pkg/blueprint"
        "github.com/osbuild/images/pkg/container"
+       "github.com/osbuild/images/pkg/customizations/kickstart"
        "github.com/osbuild/images/pkg/customizations/oscap"
        "github.com/osbuild/images/pkg/disk"
@@ -418,25 +419,20 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
                return []string{w}, nil
        }
 
-       instCust, err := customizations.GetInstaller()
+       ks, err := kickstart.New(customizations)
        if err != nil {
                return nil, err
        }
-       if instCust != nil {
+       if ks != nil {
                // only supported by the Anaconda installer
                if slices.Index([]string{"iot-installer"}, t.name) == -1 {
                        return nil, fmt.Errorf("installer customizations are not supported for %q", t.Name())
                }
 
-               // NOTE: the image type check is redundant with the check above, but
-               // let's keep it explicit in case one of the two changes.
-               // The kickstart contents is incompatible with the users and groups
-               // customization only for the iot-installer.
-               if t.Name() == "iot-installer" &&
-                       instCust.Kickstart != nil &&
-                       len(instCust.Kickstart.Contents) > 0 &&
-                       (customizations.GetUsers() != nil || customizations.GetGroups() != nil) {
-                       return nil, fmt.Errorf("iot-installer installer.kickstart.contents are not supported in combination with users or groups")
+               if t.Name() == "iot-installer" {
+                       if err := ks.Validate(); err != nil {
+                               return nil, err
+                       }
                }
        }
````
(and keep the comment, my diff is bad for this) - here and in the various other cases.

But the downside of this approach is that the API is less symetric now, "customizations.GetInstaller()" seems to be not really needed anymore with this change and that is not too nice in the context of the other customizations.Get* - anyway, rambling but maybe something interesting is in this :)
```

mvo5 (commented):
Some quick drive by comments but I like it and nothing blocks it, really just optimizing :)

pkg/customizations/kickstart/kickstart.go
achilleas-k: I agree, that's a nice change and it would save us from doing the same nil check everywhere too

achilleas-k:
@mvo5 @noelmiller I created an issue with your comments to look into later (#691).  The PR is already long enough and does the basics.  We can polish the code in separate PRs (for Michael's comments).  We need to have a discussion about the mechanics of Anaconda Kickstart modules (for Noel's comments) and I think that might block this for too long and make the PR even harder to review.

I propose we merge this as is unless there are other blockers.

pkg/manifest/anaconda_installer_iso_tree.go
mvo5: Indeed makes sense! Consistency is more important!

mvo5 (dismissed):
Thank you! This looks great, I had ideas in the previous comment but they should not be blockers, could be a followup as well (or ignored if I missed something there of course :)

mvo5 (approved):
Still looking good! Sorry, I accidentally pushed to this branch but fixed it right away.

PR #633, opened 2024-04-24
Fixing osbuild-depsolve-dnf5 test errors

bcl:


achilleas-k:
Oh, I see the issue.
```
RuntimeError: Input value "f38" not allowed, allowed values for this option are defined by regular expression ".+:.+"
```
In tests I (foolishly) use `f38` for the module platform ID, which is invalid.  It should be `platform:f38`.  Looks like dnf5 validates this whereas dnf didn.t

bcl:
@achilleas-k  Looks like this needs a newer osbuild, it is testing against something that doesn't support `root_dir`

achilleas-k:
Schutzfile has it pinned to https://github.com/osbuild/osbuild/commit/5b75592fefc4ee2d9e240a15b7002f2537de31b0

That should be new enough right?

bcl:
This osbuild PR needs to (pass and) be merged - https://github.com/osbuild/osbuild/pull/1758 and then the Schutzfile f40 commit id needs to be updated to point to it, and THEN this PR should pass. Un-drafting it since it is fine, except for all the moving parts under the hood.

supakeen (approved):
Looks good.

.github/workflows/tests.yml
```diff
@@ -13,8 +13,8 @@ on:
   merge_group:
 
 jobs:
-  unit-tests:
-    name: "🛃 Unit tests"
+  unit-tests-f39:
+    name: "🛃 Unit tests (Fedora 39)"
```
achilleas-k, line 17: The reason I kept it id because github configuration on which test is required is based on name, which is annoying to do when a PR is open.  If I mark `Unit tests (Fedora 39)` as required now, other PRs wont be allowed to merge before this one because they wont have the F39 config.

achilleas-k: Changed the repo config to require the new name *and* Fedora 40.

achilleas-k (approved):

PR #632, opened 2024-04-24
Run unit tests on Fedora 40

achilleas-k:
Fedora 40 has been released and the fedora:latest container changed to
F40, breaking our unit tests because the 'setup-osbuild-repo' wasn't
setting up a pinned repository for osbuild on F40.

Let's run unit tests on both F39 and F40.

PR #635, opened 2024-04-25
Fix local manifest arch not being set

kingsleyzissou:
This is a follow up to #595. Local manifest lists resulted in the `arch` being `unset`. This pr adds
the fix for this as well as a check in the unit test.

### DIFF
--- a/pkg/container/resolver_test.go
+++ b/pkg/container/resolver_test.go
@@ -51,7 +51,6 @@ func TestResolver(t *testing.T) {
 			Digest:    common.ToPtr(""),
 			TLSVerify: common.ToPtr(false),
 			Local:     false,
-			Store:     nil,
 		})
 	}
 
@@ -80,7 +79,6 @@ func TestResolverFail(t *testing.T) {
 		Digest:    common.ToPtr(""),
 		TLSVerify: common.ToPtr(false),
 		Local:     false,
-		Store:     nil,
 	})
 	specs, err := resolver.Finish()
 	assert.Error(t, err)
@@ -95,7 +93,6 @@ func TestResolverFail(t *testing.T) {
 		Digest:    common.ToPtr(""),
 		TLSVerify: common.ToPtr(false),
 		Local:     false,
-		Store:     nil,
 	})
 	specs, err = resolver.Finish()
 	assert.Error(t, err)
@@ -107,7 +104,6 @@ func TestResolverFail(t *testing.T) {
 		Digest:    common.ToPtr(""),
 		TLSVerify: common.ToPtr(false),
 		Local:     false,
-		Store:     nil,
 	})
 	specs, err = resolver.Finish()
 	assert.Error(t, err)
@@ -119,7 +115,6 @@ func TestResolverFail(t *testing.T) {
 		Digest:    common.ToPtr(""),
 		TLSVerify: common.ToPtr(false),
 		Local:     false,
-		Store:     nil,
 	})
 	specs, err = resolver.Finish()
 	assert.Error(t, err)
@@ -175,14 +170,16 @@ func TestResolverLocalManifest(t *testing.T) {
 	assert.NoError(t, err)
 
 	// try resolve an x86_64 container using a local manifest list
-	resolver := container.NewResolver("amd64")
+	resolver := container.NewResolverWithTestClient("amd64", func(target string) (*container.Client, error) {
+		return container.NewClientWithTestStorage(target, tmpStorage)
+	})
+
 	resolver.Add(container.SourceSpec{
 		Source:    "localhost/multi-arch",
 		Name:      "",
 		Digest:    common.ToPtr(""),
 		TLSVerify: common.ToPtr(false),
 		Local:     true,
-		Store:     &tmpStorage,
 	})
 	specs, err := resolver.Finish()
 	assert.NoError(t, err)
@@ -191,14 +188,16 @@ func TestResolverLocalManifest(t *testing.T) {
 	assert.Equal(t, specs[0].Arch.String(), arch.ARCH_X86_64.String())
 
 	// try resolve an  aarch64 container using a local manifest list
-	resolver = container.NewResolver("arm64")
+	resolver = container.NewResolverWithTestClient("arm64", func(target string) (*container.Client, error) {
+		return container.NewClientWithTestStorage(target, tmpStorage)
+	})
+
 	resolver.Add(container.SourceSpec{
 		Source:    "localhost/multi-arch",
 		Name:      "",
 		Digest:    common.ToPtr(""),
 		TLSVerify: common.ToPtr(false),
 		Local:     true,
-		Store:     &tmpStorage,
 	})
 	specs, err = resolver.Finish()
 	assert.NoError(t, err)
--- a/pkg/container/resolver_test.go
+++ b/pkg/container/resolver_test.go
@@ -176,6 +176,7 @@ func TestResolverLocalManifest(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, specs, 1)
 	assert.Equal(t, specs[0].LocalName, "localhost/multi-arch:latest")
+	assert.Equal(t, specs[0].Arch.String(), arch.ARCH_X86_64.String())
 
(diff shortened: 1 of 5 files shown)


pkg/container/resolver_test.go
```diff
@@ -133,25 +131,18 @@ func TestResolverLocalManifest(t *testing.T) {
...
-		t.Fatalf("Failed to create tempfile: %v", err)
-	}
-	defer os.Remove(containerFile.Name())
+	containerFile, err := os.CreateTemp(t.TempDir(), "Containerfile")
+	assert.NoError(t, err)
 
+	// we can't use `t.TempDir()` here because the cleanup
+	// fails when trying to remove the `<tmpstorage>/overlay`
+	// subdirectory
 	tmpStorage, err := os.MkdirTemp("", "storage")
```
kingsleyzissou, line 140: It wasn't possible to do `t.TempDir` here because it fails to cleanup the `overlay` subdirectory, so I added a note :)

pkg/container/resolver_test.go
```diff
@@ -163,33 +154,35 @@ func TestResolverLocalManifest(t *testing.T) {
...
 		"",
 		common.ToPtr(""),
 		common.ToPtr(false),
-		true,
+		local,
 		&tmpStorage,
 	})
 	specs, err := resolver.Finish()
 	assert.NoError(t, err)
 	assert.Len(t, specs, 1)
```
kingsleyzissou, line 176: It was actually really difficult to try and assert that the expected id matches.

pkg/container/resolver_test.go
```diff
@@ -41,6 +41,7 @@ func TestResolver(t *testing.T) {
 			common.ToPtr(""),
 			common.ToPtr(false),
 			false,
+			nil,
```
achilleas-k, line 47: Since we're cleaning things / tidying up, let's initialise `SourceSpec{}` with named parameters here (and below).

pkg/container/resolver_test.go
achilleas-k: Thanks.  I was wondering about that.

pkg/container/resolver_test.go
```diff
@@ -114,3 +117,77 @@ func TestResolverFail(t *testing.T) {
...
+	tmpStorage, err := os.MkdirTemp("", "storage")
+	assert.NoError(t, err)
+	defer os.Remove(tmpStorage)
+
+	_, err = containerFile.Write([]byte("FROM scratch"))
+	assert.NoError(t, err)
+
+	cmd := exec.Command( //nolint:gosec
+		"podman",
+		"--root", tmpStorage, // don't dirty the default store
```
achilleas-k, line 149: Love this.  Thank you :)

pkg/container/resolver_test.go
```diff
@@ -114,3 +117,77 @@ func TestResolverFail(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	// we can't use `t.TempDir()` here because the cleanup
+	// fails when trying to remove the `<tmpstorage>/overlay`
+	// subdirectory
+	tmpStorage, err := os.MkdirTemp("", "storage")
+	assert.NoError(t, err)
+	defer os.Remove(tmpStorage)
+
+	_, err = containerFile.Write([]byte("FROM scratch"))
```
achilleas-k, line 144: Wasn't there an issue that having the same container for both architectures wasn't testing well?

pkg/container/resolver_test.go
achilleas-k: Does it change?  I wonder what the problem is.  Maybe we can have tiny container archives checked in to the repo so they never change and we can `podman pull` them into the registry to create the manifest with known container IDs.

pkg/container/resolver_test.go
```diff
@@ -104,8 +111,85 @@ func TestResolverFail(t *testing.T) {
...
+	currentUser, err := user.Current()
+	assert.NoError(t, err)
+
+	if currentUser.Uid != "0" {
+		t.Skip("User is not root, skipping test")
+	}
+
+	_, err = exec.LookPath("podman")
+	if err != nil {
+		t.Skip("Podman not available, skipping test")
```
achilleas-k, line 131: For the dnf tests, we have a `--force-dnf` flag to make sure we don't skip the tests when we're certain they should be running.
https://github.com/osbuild/images/blob/a233d256318f70e86f891adb7abd4cf9da90f58f/pkg/dnfjson/dnfjson_test.go#L19-L27

This makes sure we don't skip the tests in CI if somehow something changes and the detection fails.
https://github.com/osbuild/images/blob/a233d256318f70e86f891adb7abd4cf9da90f58f/.github/workflows/tests.yml#L43-L44

Let's add a similar flag here.

achilleas-k (changes requested):

pkg/container/resolver_test.go
kingsleyzissou: Yeah so weirdly if you copy a file in then the id's would be exactly the same... but if you just leave `FROM scratch` then it's fine ¯\_(ツ)_/¯

achilleas-k: ah, probably timestamps.  We could `touch` the file with a known time and it might work.

kingsleyzissou: Yeah it's kind of weird... just using the scratch image on it's own works... so not sure if we want to add an extra step there just to copy in another file :) plus using the scratch is a lot faster than using `fedora-minimal`

achilleas-k:

kingsleyzissou:
Marking this as draft until my previous PR is merged (to avoid more confusion)

kingsleyzissou:
Thanks to @mvo5's suggestion, we won't have to update composer now because the `store` is no longer exported

pkg/container/resolver_test.go
```diff
@@ -15,6 +16,12 @@ import (
 	"github.com/osbuild/images/pkg/container"
 )
 
+var forceLocal = flag.Bool(
```
mvo5, line 19: :heart:

pkg/container/resolver_test.go
```diff
@@ -161,34 +169,39 @@ func TestResolverLocalManifest(t *testing.T) {
...
+		Name:      "",
+		Digest:    common.ToPtr(""),
+		TLSVerify: common.ToPtr(false),
+		Local:     true,
 	})
 	specs, err := resolver.Finish()
 	assert.NoError(t, err)
 	assert.Len(t, specs, 1)
 	assert.Equal(t, specs[0].LocalName, "localhost/multi-arch:latest")
+	assert.Equal(t, specs[0].Arch.String(), arch.ARCH_X86_64.String())
```
mvo5, line 188: :heart_eyes:

mvo5 (approved):
Looks great, thank you!

cdrage (changes requested):
As of right now with testing this, I'm unable to get this to build with bib (I built this entire thing with the fix):

```
podman build -t bib --build-context=osbuild=$SPATH/dev/go/src/osbuild --build-context=images=$SPATH/dev/go/src/images --no-cache -f devel/Containerfile .
```

After building, I try to run it from a test manifest:

```
podman run \
    --rm \
    -it \
    --privileged \
    --pull=newer \
    --security-opt label=type:unconfined_t \
    -v $(pwd)/output:/output \
    -v /var/lib/containers/storage:/var/lib/containers/storage \
    localhost/bib:latest \
    --type qcow2 --target-arch amd64 \
    --local \
    manifest-test
```

And I am getting the error:
```
Generating manifest manifest-raw.json
2024/04/26 13:03:44 error: cannot build manifest: image found is for unexpected architecture "aarch64" (expected "x86_64"), if that is intentional, please make sure --target-arch matches
Error: cannot build manifest: image found is for unexpected architecture "aarch64" (expected "x86_64"), if that is intentional, please make sure --target-arch matches
WARNING: target-arch is experimental and needs an installed 'qemu-user' package
```

Which looks like it was introduced with PR: https://github.com/osbuild/bootc-image-builder/pull/347

So at the moment even with the architecture propagation being set, I'm unable to build.

achilleas-k (approved):
Thank you.  LGTM!

PR #637, opened 2024-04-25
Change dnf-yum to yum in Fedora package sets

AdamWill:
In Fedora, there is no 'dnf-yum' package. dnf-yum is provided by yum (which is a subpackage of dnf, confused yet?)

With the dnf5-replacing-dnf transition, dnf5 includes the file /usr/bin/yum and provides and obsoletes 'yum', but it does not provide 'dnf-yum'. In a test of a Kiwi image build using a config which specified 'dnf-yum', this caused a failure, because dnf wanted to pull in 'yum' to satisfy the 'dnf-yum' specification, but it was obsoleted by dnf5 (which it was pulling in as the provider of 'dnf'), so it gave up. I suspect we'd have the same problem with osbuild builds that specify 'dnf-yum'. I think changing it to just 'yum' will solve the problem while remaining compatible across Fedora releases before and after the dnf5 switchover, because in earlier releases, this will pull in the 'yum' package as before, and in later releases, dnf should decide that the 'dnf5' package provides both it and 'dnf', and be happy installing that.

### DIFF
--- a/pkg/distro/fedora/package_sets.go
+++ b/pkg/distro/fedora/package_sets.go
@@ -582,7 +582,7 @@ func containerPackageSet(t *imageType) rpmmd.PackageSet {
 		Include: []string{
 			"bash",
 			"coreutils",
-			"dnf-yum",
+			"yum",
 			"dnf",
 			"fedora-release-container",
 			"glibc-minimal-langpack",


achilleas-k:
Sounds complicated :laughing: 
Thanks for looking into the whole thing and explaining it.

So this shouldn't change anything in the images and I just checked now and it does seem that test manifests are identical and the CI runs should confirm it.

achilleas-k (approved):
Thank you!

AdamWill:
right, for the pre-transition case it should result in the exact same build (because we're *already* using the package called 'yum' to satisfy the spec 'dnf-yum'). post-transition, specifying 'dnf-yum' will likely cause a conflict (as it did with kiwi), while the specification 'yum' will be satisfied by the package called 'dnf5' (which is indeed the thing meant to provide /usr/bin/yum in a dnf5-as-default world).

AdamWill:
huh. what's the deal with the failed f40 unit tests? I can't really see what's going on there.

achilleas-k:
> huh. what's the deal with the failed f40 unit tests? I can't really see what's going on there.

Sorry, missed this comment.

It's not an issue.  We had some unit tests that were failing on Fedora 40 because of the depsolver changing to use dnf5 (for F40 only) and error messages weren't matching expected ones in tests.  The F40 tests were not required at the time because of this, but it's been fixed now (since yesterday).

PR #647, opened 2024-04-29
manifest: ensure /var/roothome is available when needed

mvo5:
Currently creating a customization for the root user does not work because `/var/roothome` may not be avaialble in the image. This commit ensures it is created if needed just like we do for `/var/home`.

Closes https://github.com/osbuild/bootc-image-builder/issues/143

[edit: to (end-to-end) test I pointed https://github.com/osbuild/bootc-image-builder/pull/363 to this branch]

### DIFF
--- a/pkg/manifest/raw_bootc.go
+++ b/pkg/manifest/raw_bootc.go
@@ -85,6 +85,39 @@ func (p *RawBootcImage) serializeEnd() {
 	p.containerSpecs = nil
 }
 
+func buildHomedirPaths(users []users.User) []osbuild.MkdirStagePath {
+	var containsRootUser, containsNormalUser bool
+
+	for _, user := range users {
+		if user.Name == "root" {
+			containsRootUser = true
+		} else {
+			containsNormalUser = true
+		}
+	}
+
+	rootHomePath := osbuild.MkdirStagePath{
+		Path:    "/var/roothome",
+		Mode:    common.ToPtr(os.FileMode(0700)),
+		ExistOk: true,
+	}
+	userHomePath := osbuild.MkdirStagePath{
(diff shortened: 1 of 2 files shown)


pkg/manifest/raw_bootc.go
```diff
@@ -148,16 +181,12 @@ func (p *RawBootcImage) serialize() osbuild.Pipeline {
 		groupsStage.Devices = devices
 		pipeline.AddStage(groupsStage)
 	}
+
 	if len(p.Users) > 0 {
-		// ensure /var/home is available
+		// ensure home root dir (currently /var/home, /var/roothome) is
+		// available
 		mkdirStage := osbuild.NewMkdirStage(&osbuild.MkdirStageOptions{
```
supakeen, line 188: Can we skip the entire `osbuild.NewMkdirStage` if `buildHomedirPaths(...) == nil`?

mvo5: Thanks, that is an interesting question. In practice `if len(p.Users) > 0` then there will always be a buildHomedirPath result that is also > 0 - but that is true today who knows about tomorrow so maybe worth adding (otoh, tradeoffs, it adds one more level of if nested that is currently not needed and even if we never update it the worst case would be that we have an mkdir stage that just does nothing) - sorry, not very helpful I suppose :)

supakeen: As long as the first premise holds (we always have to create homedirs for users) it's fine to keep as is :)

supakeen (approved):

achilleas-k:
Rebasing for unit test fixes.

achilleas-k (approved):

PR #650, opened 2024-04-30
distro/rhel: unify osname for edge deployments (HMS-4011)

achilleas-k:
Our RHEL Edge deployments (ISOs and disk images) had inconsistent osnames set.  The edge-installer used "rhel" while the simplified installer and disk image used "redhat".

For bonus inconsistency, the RHEL Edge ISOs created by the Edge Service modify the kickstart file to set the osname to "rhel-edge".

Meanwhile, Fedora IoT images all set the osname to "fedora-iot".

To make everything consistent and unify the names, let's set all osnames to "rhel-edge".  This aligns better with Fedora (the equivalent to "fedora-iot" in the RHEL world is "rhel-edge") and with what has already been happening in the service.

Practically, the osname sets the deployment path:

  /ostree/deploy/$OSNAME

so this will affect new deployments, but will have no effect on existing installations and upgrades.

### DIFF
--- a/pkg/distro/rhel/images.go
+++ b/pkg/distro/rhel/images.go
@@ -524,7 +524,7 @@ func EdgeInstallerImage(workload workload.Workload,
 
 	img.Product = t.Arch().Distro().Product()
 	img.Variant = "edge"
-	img.OSName = "rhel"
+	img.OSName = "rhel-edge"
 	img.OSVersion = t.Arch().Distro().OsVersion()
 	img.Release = fmt.Sprintf("%s %s", t.Arch().Distro().Product(), t.Arch().Distro().OsVersion())
 	img.FIPS = customizations.GetFIPS()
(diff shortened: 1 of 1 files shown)


thozza (approved):
Looks reasonable :+1:

miabbott:
There's a weak argument to be made that we should just be using `rhel` as the osname, so that we strengthen the idea that RHEL for Edge is just "RHEL". But given that the Edge Service was already usign `rhel-edge`, might as well just go with that.

achilleas-k:
> There's a weak argument to be made that we should just be using `rhel` as the osname, so that we strengthen the idea that RHEL for Edge is just "RHEL". But given that the Edge Service was already usign `rhel-edge`, might as well just go with that.

Thanks for the comment Micah.  These were my thoughts exactly, with the extra point in `rhel-edge`'s favour for aligning with `fedora-iot` (though I don't know how strong that argument is).

supakeen (approved):

PR #651, opened 2024-04-30
manifest: modify insights-client.service for ostree-based systems (HMS-4031)

achilleas-k:
Insights-client collection must occur at boot time so that the current ostree commit hash can be reflected after upgrade.  Otherwise, the upgrade shows as failed in the console UI.  Add a drop-in file that enables insights-client.service to run on successful boot.

Drop-in files can normally be created with the `org.osbuild.systemd.unit` stage but the stage doesn't support all the options we need.  This is a temporary workaround until we get the stage updated to support everything we need.

This change was previously part of the Edge Fleet Management service.  The drop-in file was created by a `%post` section in the injected kickstart file.  Creating it in the base ostree-commit is cleaner and should be functionally the same, with the major difference that the drop-in will be removed during an upgrade if it's not also included in the new ostree commit.

### DIFF
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -882,3 +882,14 @@ func (p *OS) getInline() []string {
 
 	return inlineData
 }
+
+// Filename and contents for the insights-client service drop-in.
+// This is a temporary workaround until the org.osbuild.systemd.unit stage
+// gains support for all the options we need.
+func insightsClientDropin() (string, string) {
+	return "/etc/systemd/system/insights-client.service.d/override.conf", `[Unit]
+Requisite=greenboot-healthcheck.service
+After=network-online.target greenboot-healthcheck.service osbuild-first-boot.service
+[Install]
+WantedBy=multi-user.target`
+}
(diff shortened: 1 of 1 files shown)


thozza (approved):
LGTM

miabbott:
@achilleas-k Just for my own clarity, this is only injected if an image is built with subscription-manager entitlements?

I'm trying to unwind how those entitlements are provided...doesn't look like there is a blueprint option for that, so I assume osbuild is just doing something smart and figuring out that the osbuild host is already entitled?

loadtheaccumulator:
@miabbott If an activation key is selected, Edge API provides it in its call to Image Builder. This replaces our current injection into a %post on the Edge API side.

achilleas-k:
> I'm trying to unwind how those entitlements are provided...doesn't look like there is a blueprint option for that

In osbuild-composer, subscription options are only available through the API (in other words, image builder on console.redhat.com).  We don't provide a blueprint option for this.
The options are defined here: https://github.com/osbuild/osbuild-composer/blob/37233b7b621b3a614d4e92f9ad47ca133f5453d8/internal/cloudapi/v2/openapi.v2.yml#L1716-L1748

Issue #88, opened 2023-08-07
Minimise code duplication in cmd/ utilities


achilleas-k:
Some utilities in cmd have a lot of common code that requires duplicating maintenance and changes.  We should move common functions into an internal package, or create a single binary with subcommands to share code.

achilleas-k:
This is mostly covered by the `manifestgen` package.

PR #664, opened 2024-05-06
gen-manifests: derive seedArg from filename and env

mvo5:
To ensure that manifests get random(ish) and stable UUIDs we set the rng seed arg based on the filename.

This should fix the issue discovered in
osbuild/manifest-db#124

that duplicated UUIDs for xfs/btrfs can trigger random(ish) and hard to diagnose errors.

This is the same as
https://github.com/osbuild/osbuild-composer/pull/4124 hopefully for the right place this time.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"hash/fnv"
 	"io"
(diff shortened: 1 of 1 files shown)


achilleas-k:
We should also do the same in `cmd/build` for consistency (see also https://github.com/osbuild/images/issues/88).

mvo5:
Closing in favor of #674

PR #671, opened 2024-05-08
github: update bib ref every week

achilleas-k:
With things slowing down a bit, we don't need to have daily updates of the BIB ref here.  Weekly updates should be enough.

I set it to run every Sunday at noon, so CI jobs can run when there's not much other activity and we can deal with failures on Monday.


supakeen (approved):

ondrejbudai (approved):
Thanks!

PR #673, opened 2024-05-08
cmd: remove some duplication in BuldConfig handling

mvo5:
This is a prereq for https://github.com/osbuild/images/pull/664 - not sure if I can extract a common MakeConfig() as they are similar and different but it probably a good idea to try.


bcl (approved):
Looks good to me.

achilleas-k (approved):

PR #674, opened 2024-05-08
cmd: add helper for seed generation

mvo5:
To ensure that manifests get random(ish) and stable UUIDs we set the rng seed arg based on the filename.

This should fix the issue discovered in
osbuild/manifest-db#124

that duplicated UUIDs for xfs/btrfs can trigger random(ish) and hard to diagnose errors.

This is the same as
osbuild/osbuild-composer#4124 hopefully for the right place this time.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -42,7 +42,6 @@ func makeManifest(
 	distribution distro.Distro,
 	repos []rpmmd.RepoConfig,
 	archName string,
-	seedArg int64,
 	cacheRoot string,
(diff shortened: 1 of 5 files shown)


achilleas-k (changes requested):
LGTM but the linter is unhappy (`fakeNamer` is unused).

mvo5:
> LGTM but the linter is unhappy (`fakeNamer` is unused).

Thank you! Silly me, fixed now :)

bcl (approved):
Looks good!

PR #678, opened 2024-05-10
pkg/osbuild: move mount name generation into getOsbuildMount

ondrejbudai:
Mount names are just identifiers. They don't have any other real
meaning.

Prior this commit, callers of genOsbuildMount had to supply the mount
name, which led to inconsistencies: genMountsDevicesFromPt used the
name of the leaf device, genMountsForBootupd used synthetic names
constructed from a partition number.

The issue with the genMountsDevicesFromPt's approch is that multiple
mounts can be created from a simple device. This is not currently
used, but it will come with btrfs (single device) and its subvolumes
(multiple mounts). The genMountsForBootupd's approach has a similar
flaw.

This commit implements the mount name generation directly in
genOsbuildMount. The new name is no longer referring to the source,
but it's rather constructed from the mountpoint (by using pathEscape).
This has two benefits:
- Every mount obviously must have a mountpoint
- Mountpoints are unique(*), so names will be also unique

I also added some docstrings, because these functions signatures are
quite hard to comprehend.

(*) Yeah, you can stack mounts, but this is not something supported
by our disk library.

### DIFF
--- a/pkg/osbuild/bootupd_stage_test.go
+++ b/pkg/osbuild/bootupd_stage_test.go
@@ -222,21 +222,21 @@ func TestGenBootupdDevicesMountsHappy(t *testing.T) {
 	})
 	assert.Equal(t, mounts, []osbuild.Mount{
 		{
-			Name:      "part4",
+			Name:      "-",
 			Type:      "org.osbuild.ext4",
 			Source:    "disk",
 			Target:    "/",
 			Partition: common.ToPtr(4),
 		},
 		{
-			Name:      "part3",
+			Name:      "boot",
 			Type:      "org.osbuild.ext4",
 			Source:    "disk",
 			Target:    "/boot",
 			Partition: common.ToPtr(3),
 		},
 		{
-			Name:      "part2",
+			Name:      "boot-efi",
 			Type:      "org.osbuild.fat",
 			Source:    "disk",
 			Target:    "/boot/efi",
--- a/pkg/osbuild/device.go
+++ b/pkg/osbuild/device.go
@@ -164,6 +164,18 @@ func deviceName(p disk.Entity) string {
 	panic(fmt.Sprintf("unsupported device type in deviceName: '%T'", p))
 }
 
+// getDevices takes an entity path, and returns osbuild devices required before being able to mount the leaf Mountable
+//
+// - path is an entity path as defined by the disk.entityPath function
+// - filename is the name of an underlying image file (which will get loop-mounted)
+// - lockLoopback determines whether the loop device will get locked after creation
+//
+// The device names are created from the payload that they are holding. This is useful to easily visually map e.g.
+// a loopback device and its mount (in the case of ordinary partitions): they should have the same, or similar name.
+//
+// The first returned value is a map of devices for the given path.
+// The second returned value is the name of the last device in the path. This is the device that should be used as the
+// source for the mount.
 func getDevices(path []disk.Entity, filename string, lockLoopback bool) (map[string]Device, string) {
 	var pt *disk.PartitionTable
 
--- a/pkg/osbuild/device.go
+++ b/pkg/osbuild/device.go
@@ -237,8 +237,14 @@ func pathEscape(path string) string {
 	return strings.ReplaceAll(path, "/", "-")
 }
 
-func genOsbuildMount(name, source string, mnt disk.Mountable) (*Mount, error) {
+// genOsbuildMount generates an osbuild mount from Mountable mnt
+//
+// - source is the name of the device that the mount should be created from.
+// The name of the mount is derived from the mountpoint of the mountable, escaped with pathEscape. This shouldn't
+// create any conflicts, as the mountpoint is unique within the partition table.
+func genOsbuildMount(source string, mnt disk.Mountable) (*Mount, error) {
 	mountpoint := mnt.GetMountpoint()
+	name := pathEscape(mountpoint)
 	t := mnt.GetFSType()
 	switch t {
 	case "xfs":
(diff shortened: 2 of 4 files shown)


pkg/osbuild/bootupd_stage_test.go
```diff
@@ -222,21 +222,21 @@ func TestGenBootupdDevicesMountsHappy(t *testing.T) {
 	})
 	assert.Equal(t, mounts, []osbuild.Mount{
 		{
-			Name:      "part4",
+			Name:      "-",
```
bcl, line 225: Can this be named root? I see '-' there and my brain says 'stdout?'

bcl (dismissed):
Other than my question, looks like a nice improvement. And thanks for the docs, we need more of those!

pkg/osbuild/device.go
```diff
@@ -242,21 +260,30 @@ func genOsbuildMount(name, source string, mnt disk.Mountable) (*Mount, error) {
...
+// 2) generated mounts
+// 3) generated devices
+// 4) error if any
 func genMountsDevicesFromPt(filename string, pt *disk.PartitionTable) (string, []Mount, map[string]Device, error) {
 	devices := make(map[string]Device, len(pt.Partitions))
 	mounts := make([]Mount, 0, len(pt.Partitions))
 	var fsRootMntName string
 	genMounts := func(mnt disk.Mountable, path []disk.Entity) error {
-		stageDevices, name := getDevices(path, filename, false)
+		stageDevices, leafDeviceName := getDevices(path, filename, false)
```
mvo5, line 276: I wonder if we should take the opportunity to add a tiny bit more testing around `genMountsDevicesFromPt()` ? Not much just generate a few device names, something like:
```go
diff --git a/pkg/osbuild/device_test.go b/pkg/osbuild/device_test.go
index a8232bd2c..035dac4b2 100644
--- a/pkg/osbuild/device_test.go
+++ b/pkg/osbuild/device_test.go
@@ -188,12 +188,14 @@ func TestMountsDeviceFromPtNoRootErrors(t *testing.T) {
 
 func TestMountsDeviceFromPtHappy(t *testing.T) {
        filename := "fake-disk.img"
-       fakePt := testdisk.MakeFakePartitionTable("/")
+       fakePt := testdisk.MakeFakePartitionTable("/", "/boot", "/boot/efi")
        fsRootMntName, mounts, devices, err := genMountsDevicesFromPt(filename, fakePt)
        require.Nil(t, err)
        assert.Equal(t, fsRootMntName, "-")
        assert.Equal(t, mounts, []Mount{
                {Name: "-", Type: "org.osbuild.ext4", Source: "-", Target: "/"},
+               {Name: "boot", Type: "org.osbuild.ext4", Source: "boot", Target: "/boot"},
+               {Name: "boot-efi", Type: "org.osbuild.fat", Source: "boot-efi", Target: "/boot/efi"},
        })
        assert.Equal(t, devices, map[string]Device{
                "-": {
@@ -203,5 +205,19 @@ func TestMountsDeviceFromPtHappy(t *testing.T) {
                                Size:     testdisk.FakePartitionSize / 512,
                        },
                },
+               "boot": {
+                       Type: "org.osbuild.loopback",
+                       Options: &LoopbackDeviceOptions{
+                               Filename: "fake-disk.img",
+                               Size:     testdisk.FakePartitionSize / 512,
+                       },
+               },
+               "boot-efi": {
+                       Type: "org.osbuild.loopback",
+                       Options: &LoopbackDeviceOptions{
+                               Filename: "fake-disk.img",
+                               Size:     testdisk.FakePartitionSize / 512,
+                       },
+               },
        })
 }
``` 
maybe?

mvo5 (approved):
Thank you, this is a nice cleanup and will improve readability of the generated manifests. One tiny (and optional) idea inline.

pkg/osbuild/bootupd_stage_test.go
ondrejbudai: Yeah, I'm not a fan either, it's just us following the systemd mount convention:

```
$ sudo systemctl | grep '\.mount'
  -.mount                         loaded active mounted   Root Mount
  boot-efi.mount                  loaded active mounted   /boot/efi
  boot.mount                      loaded active mounted   /boot
  dev-hugepages.mount             loaded active mounted   Huge Pages File System
  dev-mqueue.mount                loaded active mounted   POSIX Message Queue File System
  home.mount                      loaded active mounted   /home
  proc-sys-fs-binfmt_misc.mount   loaded active mounted   Arbitrary Executable File Formats File System
  run-user-1000-doc.mount         loaded active mounted   /run/user/1000/doc
  run-user-1000-gvfs.mount        loaded active mounted   /run/user/1000/gvfs
  run-user-1000.mount             loaded active mounted   /run/user/1000
  sys-fs-fuse-connections.mount   loaded active mounted   FUSE Control File System
  sys-kernel-config.mount         loaded active mounted   Kernel Configuration File System
  sys-kernel-debug.mount          loaded active mounted   Kernel Debug File System
  sys-kernel-tracing.mount        loaded active mounted   Kernel Trace File System
  tmp-.mount_jetbraXWOO5X.mount   loaded active mounted   /tmp/.mount_jetbraXWOO5X
  tmp.mount                       loaded active mounted   Temporary Directory /tmp
  var-lib-nfs-rpc_pipefs.mount    loaded active mounted   RPC Pipe File System
```

pkg/osbuild/device.go
ondrejbudai: Done! :)

mvo5 (approved):

pkg/osbuild/bootupd_stage_test.go
bcl: Oh geez.

bcl (approved):
Looks good!

PR #683, opened 2024-05-13
distro/rhel10: replace dhcp-client with dhcpcd

achilleas-k:
The dhcp-client was dropped from RHEL 10.  As a cloud-init dependency, it has been replaced by dhcpcd.

See RHEL-26304


elkoniu (approved):
LGTM

PR #685, opened 2024-05-14
pkg/osbuild: clean up genMountsForBootupd

ondrejbudai:
Instead of the if statement, I switched to switch (pun absolutely intended). Also, I fixed the error message, previously it always printed type <nil>, because mnt is set to nil, if ok == false.


mvo5 (approved):
Thank you! I like switches so I approve of the switch to switch (but let's not switch again).

achilleas-k (approved):

PR #686, opened 2024-05-14
pkg/osbuild: Simplify GenMkfsStages arguments

ondrejbudai:
Prior this commit, the implementation was slightly confusing: GenMkfsStages accepted a generic device as the source of the underlying disk image. However, the function could only operate on a loopback, otherwise it panicked. Moreover, the function didn't actually use the given device: it just used the filename from the given loopback device. The actual device for generated stages is created in getDevices.

Let's simplify this: The function now accepts just the filename, since this is actually the only thing that it needs.

Also added tests. The unhappy path still relies on a panic, but this is a huge rabbit hole that I don't want to explore at this point.


mvo5 (approved):
Thank you! I like this simplification!

PR #702, opened 2024-05-22
dnfjson: enable loading of optional metadata when needed

achilleas-k:
**dnfjson: add OptionalMetadata property to depsolve args**

New property on the arguments object that controls the types of metadata
to download for repositories and use when depsolving.

See https://github.com/osbuild/osbuild/pull/1775

---

**dnfjson: enable filelists for older distro versions**

Enable filelists in the optional metadata for EL7, EL8, EL9, and Fedora 39.

When running osbuild-depsolve-dnf with the libdnf version that's
included in these distro versions, this option will have no effect since
it is enabled by default.  However, when depsolving packages for these
distro versions with newer versions of libdnf, the filelists are
disabled and can cause issues with packages that declare dependencies on
filepaths.

The cutoff point is a little fuzzy for CentOS and Fedora.  Depsolving
all our test manifests for Fedora 39 and CentOS Stream 9 without
filelists works, but since the packaging guidelines of not depending on
filepaths went into effect in EL10 and Fedora 40, it's possible there
are packages in F39 and C9S that don't conform yet, so it's probably
safer to keep filelists enabled for these distro versions.

---

Requires https://github.com/osbuild/osbuild/pull/1775

### DIFF
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -640,6 +640,9 @@ type arguments struct {
 	// Load repository configurations, gpg keys, and vars from an os-root-like
 	// tree.
 	RootDir string `json:"root_dir"`
+
+	// Optional metadata to download for the repositories
+	OptionalMetadata []string `json:"optional-metadata,omitempty"`
 }
 
 type searchArgs struct {
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -460,10 +460,12 @@ func (s *Solver) makeDepsolveRequest(pkgSets []rpmmd.PackageSet) (*Request, map[
 	if err != nil {
 		return nil, nil, err
 	}
+
 	args := arguments{
-		Repos:        dnfRepoMap,
-		RootDir:      s.rootDir,
-		Transactions: transactions,
+		Repos:            dnfRepoMap,
+		RootDir:          s.rootDir,
+		Transactions:     transactions,
+		OptionalMetadata: s.optionalMetadataForDistro(),
 	}
 
 	req := Request{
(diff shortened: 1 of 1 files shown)


bcl:
I guess we don't really have a way to express what version of osbuild-depsolve-dnf this depends on other than making sure to set a new Requires version in osbuild-composer when this is imported, right?

thozza:
I'm wondering what would be the downside of enabling the downloading of filelists explicitly in all cases?

pkg/dnfjson/dnfjson.go
```diff
@@ -479,6 +481,20 @@ func (s *Solver) makeDepsolveRequest(pkgSets []rpmmd.PackageSet) (*Request, map[
...
 
+func (s *Solver) optionalMetadataForDistro() []string {
+	// filelist repo metadata is required when using newer versions of libdnf
+	// with old repositories or packages that specify dependencies on files.
+	// EL10+ and Fedora 40+ packaging guidelines prohibit depending on
+	// filepaths so filelist downloads are disabled by default and are not
+	// required when depsolving for those distros. Explicitly enable the option
+	// for older distro versions in case we are using a newer libdnf.
+	switch s.modulePlatformID {
+	case "platform:f39", "platform:el8", "platform:el9":
```
thozza, line 492: RHEL-7 should be also included, right? What about F40 worker depsolving for F38 image?

achilleas-k:
> I'm wondering what would be the downside of enabling the downloading of filelists explicitly in all cases?

The only downside is the size, which, on a slow connection, can make osbuild-composer very slow to start (building the initial rpmmd cache).  Right now, this switch will make startup and depsolving faster on Fedora 40 and RHEL/CentOS 10, and I think that's nice to have.

pkg/dnfjson/dnfjson.go
achilleas-k: heh, yeah, I forgot about RHEL 7 because it didn't cause any issues, most likely because the package that triggers this issue in our base images is `platform-python`.

Fedora 38 is EOL and I was planning on removing it soon.

achilleas-k:
> I guess we don't really have a way to express what version of osbuild-depsolve-dnf this depends on other than making sure to set a new Requires version in osbuild-composer when this is imported, right?

Yes, exactly.  However, since osbuild-depsolve-dnf's json deserialization isn't strict, adding this new option to the request (or any unknown property) has no effect and just gets ignored.  And on the depsolver side, the property is optional, so this change is non-breaking in any direction:
- New composer adds the option to an old depsolver, the depsolver ignores it.
- Old composer with new depsolver will never specify the option.

Both cases will behave like the change was never made.

pkg/dnfjson/dnfjson.go
achilleas-k: Ok to resolve this now?
(with an F38 removal later this week)

thozza (approved):

bcl (approved):
Looks good!

PR #707, opened 2024-05-23
distro/rhel8: update RHUI client RPM for Azure SAP 8.10 images (COMPOSER-2254)

achilleas-k:
Since 8.10 is the last minor release of RHEL8, and it's not E4S one / has unversioned repos, its RHUI client rpm name is different.

See COMPOSER-2254

---

While the [snapshot job](https://github.com/osbuild/rpmrepo/actions/runs/9317155443/job/25646887891) is still running, the repository we need for this is already done, so let's use it here and we'll update the rest of the repo snapshots later.

### DIFF
--- a/pkg/distro/rhel/rhel8/azure.go
+++ b/pkg/distro/rhel/rhel8/azure.go
@@ -256,10 +256,14 @@ func azureRhuiPackageSet(t *rhel.ImageType) rpmmd.PackageSet {
 // Includes the common azure package set, the common SAP packages, and
 // the azure rhui sap package.
 func azureSapPackageSet(t *rhel.ImageType) rpmmd.PackageSet {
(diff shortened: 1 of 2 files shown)


achilleas-k:
The bug caught by the linter is fixed in #716

thozza (approved):

achilleas-k:
Discovered a fun little edge case race condition in the rules for rebuilding images involving the ostree containers.  The full rebuild in #716 (triggered by changing the osbuild commit ID) changes the commit ID inside an ostree container, but the manifest ID is the same.  The ref of the ostree container is based on the manifest ID.  So in this PR, even though the image definitions didn't change for edge-based images (disks and installers), the CI found a different commit ID so is rebuilding those.  This basically only happens if a pipeline rebuilds an ostree container and another pipeline uses it before the first pipeline manages to build the disk images and ISOs that use it.

PR #708, opened 2024-05-23
[wip] OpenSCAP json tailoring & overrides (HMS-3826)

kingsleyzissou:
- simplify the OpenSCAP stage option creation
- Add ability to override vars

Still to do:
- Add some tests
- Implement the json tailoring

This depends on:
https://github.com/osbuild/osbuild/pull/1777

### DIFF
--- a/pkg/blueprint/openscap_customizations.go
+++ b/pkg/blueprint/openscap_customizations.go
@@ -1,0 +1,78 @@
+package blueprint
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type OpenSCAPCustomization struct {
+	DataStream string                           `json:"datastream,omitempty" toml:"datastream,omitempty"`
+	ProfileID  string                           `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
+	Tailoring  *OpenSCAPTailoringCustomizations `json:"tailoring,omitempty" toml:"tailoring,omitempty"`
+}
+
+type OpenSCAPTailoringCustomizations struct {
+	Selected   []string                    `json:"selected,omitempty" toml:"selected,omitempty"`
+	Unselected []string                    `json:"unselected,omitempty" toml:"unselected,omitempty"`
+	Overrides  []OpenSCAPTailoringOverride `json:"overrides,omitempty" toml:"overrides,omitempty"`
+}
+
+type OpenSCAPTailoringOverride struct {
+	Var   string      `json:"var,omitempty" toml:"var,omitempty"`
+	Value interface{} `json:"value,omitempty" toml:"value,omitempty"`
+}
+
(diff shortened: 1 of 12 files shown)


pkg/blueprint/openscap_customizations.go
```diff
@@ -0,0 +1,78 @@
...
+type OpenSCAPCustomization struct {
+	DataStream string                           `json:"datastream,omitempty" toml:"datastream,omitempty"`
+	ProfileID  string                           `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
+	Tailoring  *OpenSCAPTailoringCustomizations `json:"tailoring,omitempty" toml:"tailoring,omitempty"`
+}
+
+type OpenSCAPTailoringCustomizations struct {
+	Selected   []string                    `json:"selected,omitempty" toml:"selected,omitempty"`
+	Unselected []string                    `json:"unselected,omitempty" toml:"unselected,omitempty"`
+	Overrides  []OpenSCAPTailoringOverride `json:"overrides,omitempty" toml:"overrides,omitempty"`
```
kingsleyzissou, line 17: We could potentially just not add the overrides and just have the json tailoring file instead.
The only thing is the json tailoring file will only be supported by newer versions of `oscap-utils`

pkg/distro/fedora/images.go
```diff
@@ -228,9 +220,14 @@ func osCustomizations(
 			// overwrite the profile id with the new tailoring id
 			oscapStageOptions.ProfileID = newProfile
 			oscapStageOptions.Tailoring = tailoringFilepath
+		}
 
-			// add the parent directory for the tailoring file
-			osc.Directories = append(osc.Directories, tailoringDir)
+		for _, directory := range directoriesToCreate {
```
kingsleyzissou, line 218: Maybe we can simplify this completely and just save the tailoring file to the existing `oscapDataDir`?

pkg/distro/fedora/images.go
```diff
@@ -200,40 +198,39 @@ func osCustomizations(
...
-
-			tailoringOptions := osbuild.OscapAutotailorConfig{
-				NewProfile: newProfile,
-				Datastream: datastream,
-				ProfileID:  oscapConfig.ProfileID,
-				Selected:   oscapConfig.Tailoring.Selected,
-				Unselected: oscapConfig.Tailoring.Unselected,
+		var tailoringConfig oscap.TailoringConfig
+		if tc := oscapConfig.Tailoring; tc != nil {
+			tailoringPath := filepath.Join(oscapDataDir, "tailoring.xml")
```
kingsleyzissou, line 209: This would mean that the tailoring file would get saved to `/oscap_data` rather than `/usr/share/xml/osbuild-oscap-data`. Not sure if we would consider this a breaking change?

Issue #230, opened 2023-10-27
Test: Build edge-commit with cloud-init to use with edge-ami and edge-qcow2


achilleas-k:
> > Also of course, adding it would require it being in the commit, I don't know why I thought it would be included in the `ami`. But for our test, we can build an ostree commit that has it to test against the `edge-ami` specifically.
> > @7flying do you know if adding `cloud-init` to an `edge/iot-commit` conflict will conflict with `initial-setup`?
> 
> @achilleas-k no sorry, I did not test that

I think I'll add this as part of our test configs: Build an edge commit with cloud-init installed and use it as the base image for an edge-ami and then boot test it with cloud-init.

But I'll leave that for a follow-up PR.
For now, here, I'll see if I can figure out what's making the firstboot service fail to start (could be nothing, but it could be a side effect of some bug) and take it from there.

Thanks for the input!

_Originally posted by @achilleas-k in https://github.com/osbuild/images/issues/217#issuecomment-1782884916_

achilleas-k:
The build config should be really simple, but we don't have any support for cloud-init in our boot tests, so this would need a bit of work.  Would be nice to have for non-edge too though.

PR #714, opened 2024-05-27
manifest: wait for user input before rebooting when creating a bootc installer

mvo5:
In issue #230 it was raised that it is undesirable to automatically reboot when the install is finished. So instead halt the install and show a message to the user.

This is a draft because it changes the existing behavior so users who rely on the auto-reboot (like our tests) will fail [0] - but  if everyone is happy and we consider the old behavior a bug I will undraft it.

Closes: https://github.com/osbuild/bootc-image-builder/issues/230

[0] We can fix this via https://github.com/osbuild/bootc-image-builder/compare/main...mvo5:handle-halt-instead-of-reboot?expand=1

### DIFF
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -502,7 +502,7 @@ reqpart --add-boot
 part swap --fstype=swap --size=1024
 part / --fstype=ext4 --grow
 
-reboot --eject
+halt --eject
 `
 
 	// Workaround for lack of --target-imgref in Anaconda, xref https://github.com/osbuild/images/issues/380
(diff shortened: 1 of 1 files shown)


bcl:
I'm not sure of the exact setup where this is expected to be used, but in my experience you do want it to reboot when it is finished installing. In those cases you set things up so that the installer iso is booted once and on reboot it uses the disk that was just installed.

achilleas-k:
The issue could also be solved by putting the choice on the user: https://github.com/osbuild/bootc-image-builder/pull/438

We could expose the `unattended` option in BIB too, which I would argue should imply `reboot`, and exposing all the options will let users stick to fully interactive or add their own full kickstart file to do whatever they want.

cdrage:
Silly question, since this will wait for user input, is there a message that will say "free to remove removable media" or something like that, so that you can easily remove the USB stick / press continue?

I'm assuming you would remove whatever you have plugged in, hit next and now you have your OS.

mvo5:
> Silly question, since this will wait for user input, is there a message that will say "free to remove removable media" or something like that, so that you can easily remove the USB stick / press continue?
> 
> I'm assuming you would remove whatever you have plugged in, hit next and now you have your OS.

This is what it looks like:
Screenshot from 2024-06-04 19-08-15

bcl:
Pulling the plug without a proper shutdown isn't safe. Anaconda keeps the target disk mounted until it shuts down or reboots.

achilleas-k:
What's the decision here?  Do we still want this?

achilleas-k (approved):

PR #717, opened 2024-05-29
Set the chunk size for S3 upload to 64MB

bcrochet:
The default minimum chunk size of 5MB for s3 upload only allows for an upload of about 50Gi. It is not out of the question for a bootc image to get above this threshhold.

For now, this will increase the chunk size to 64MB making it possible to upload images up to ~640Gi. This could in the future become a dynamic value, but for now, this should at least give some head room.


achilleas-k (approved):
TY.  LGTM!

achilleas-k:
There's a small chance the PR might fail in the merge queue because of some flaky testing that started happening recently.  If it does, then #725 should fix it, so I'll re-queue it after that's merged.

PR #723, opened 2024-06-03
dnfjson: allow (optional) field `solver` in the dnfjson output

mvo5:
With the new "dnf5" solver is seems useful to allow the dnfjson code return what solver was actually used in the transaction.

This commit allows an (optional) "solver" field in the json that is returned from the dnfjson helper. It does not do more yet but we should (probably) make it avaialble for the higher layers too so that it can be e.g. logged.

See also https://github.com/osbuild/osbuild/pull/1776

### DIFF
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -690,6 +690,9 @@ type packageSpecs []PackageSpec
 type depsolveResult struct {
 	Packages packageSpecs          `json:"packages"`
 	Repos    map[string]repoConfig `json:"repos"`
+
+	// (optional) contains the solver used, e.g. "dnf5"
+	Solver string `json:"solver"`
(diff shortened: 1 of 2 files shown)


bcl:
Looks good, other than the linting complaints :)

achilleas-k (changes requested):

achilleas-k (approved):
LGTM.  Technically we can merge this before the osbuild PR and it wouldn't be a problem, but let's do it in proper order.

Then we could also run the unit tests in `dnfjson_test.go` against both solvers.

bcl (approved):
Looks good, pylint is apparently befuddled by having a raise in an else before using a variable.

bcl:
Made #729 to ignore the pylint error

thozza (approved):

PR #716, opened 2024-05-29
Run CI build on Fedora 40 runners

achilleas-k:
Update the GitLab CI configurations (static and autogenerated) to use Fedora 40 runners.
Also, pull in a new version of osbuild to test the new optional metadata selection for depsolving [1] and to make sure everything is rebuilt.

[1] https://github.com/osbuild/osbuild/pull/1775

PR #725, opened 2024-06-03
Schutzfile: update osbuild dependency

achilleas-k:
Some strange failures started happening last week in our CI.  The ostree containers that we use to build ostree-based images appear to be randomly failing to start, or return `403 Forbidden` when queried.  This makes our CI pipeline flaky which is blocking any PR from being merged through the merge queue (see for example the failures here https://gitlab.com/redhat/services/products/image-builder/ci/images/-/pipelines/1314481981).

I suspect this might be related to PR #716, which changes the CI configurations to all run on Fedora 40.  In that PR, all ostree containers were rebuilt on F40 and now all subsequent runs (that don't need to update the image) use these new containers.  I'm hoping the issue is related to our Fedora 40 runners being quite old (pre GA), and some already-solved bug is causing this.

I need to look into this further and figure out if (and why) this is happening, but in the meantime, let's try to unblock PRs by rebuilding everything on Fedora 39 again.

This PR updates the osbuild commit ID, which will trigger a rebuild of all image types.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -136,43 +136,6 @@ jobs:
       - name: Run depsolver tests with force-dnf to make sure it's not skipped for any reason
         run: go test -v -race ./pkg/dnfjson/... -force-dnf
 
-  unit-tests-c8s:
-    name: "🛃 Unit tests (CentOS Stream 8)"
-    runs-on: ubuntu-latest
-    container:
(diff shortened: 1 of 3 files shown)


bcl (approved):
Looks ok. Hopefully this stabilizes things.

elkoniu (approved):
LGTM

PR #732, opened 2024-06-05
Partition Table, Stages, Mounts, and Devices generator commands for otk

achilleas-k:
This is a joint PR by @mvo5 and me that adds three new commands:
- otk-gen-partition-table: generates a fully resolved partition table from a base description and modifications.  It's an externalised version of the `disk.NewPartitionTable()` function.
- otk-make-partition-stages: generates stages that can be added to a manifest to create a disk image from a fully resolved partition table.  It's an externalised version of the `osbuild.GenImagePrepareStages()` function.
- otk-make-partition-mounts-devices: generates mounts and devices that can be attached to osbuild stages that need to operate on a partitioned image file.  It's an externalised version of the `osbuild.GenMountsDevicesFromPT()` function.

Some tests can still be added and it might be good to test these with otk directly before merging, but they're mostly functionally complete.

### DIFF
--- a/cmd/otk-gen-partition-table/main.go
+++ b/cmd/otk-gen-partition-table/main.go
@@ -55,6 +55,7 @@ type InputModifications struct {
 	PartitionMode disk.PartitioningMode               `json:"partition_mode"`
 	Filesystems   []blueprint.FilesystemCustomization `json:"filesystems"`
 	MinDiskSize   string                              `json:"min_disk_size"`
+	Filename      string                              `json:"filename"`
 }
 
 type Output = otkdisk.Data
@@ -181,6 +182,10 @@ func genPartitionTable(genPartInput *Input, rng *rand.Rand) (*Output, error) {
 	if err != nil {
 		return nil, err
 	}
+	fname := "disk.img"
+	if genPartInput.Modifications.Filename != "" {
+		fname = genPartInput.Modifications.Filename
+	}
 
 	kernelOptions := osbuild.GenImageKernelOptions(pt)
 	otkPart := &Output{
@@ -190,6 +195,7 @@ func genPartitionTable(genPartInput *Input, rng *rand.Rand) (*Output, error) {
 			},
 			KernelOptsList: kernelOptions,
 			PartitionMap:   makePartMap(pt),
+			Filename:       fname,
 		},
 	}
 
--- a/cmd/otk-gen-partition-table/main.go
+++ b/cmd/otk-gen-partition-table/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/osbuild/images/internal/buildconfig"
 	"github.com/osbuild/images/internal/cmdutil"
 	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/internal/otk"
 	"github.com/osbuild/images/pkg/blueprint"
 	"github.com/osbuild/images/pkg/disk"
 	"github.com/osbuild/images/pkg/osbuild"
@@ -75,9 +76,7 @@ type OutputPartition struct {
 	UUID string `json:"uuid"`
 }
 
-type OutputInternal struct {
-	PartitionTable *disk.PartitionTable `json:"partition-table"`
-}
+type OutputInternal = otk.PartitionInternal
 
 func makePartMap(pt *disk.PartitionTable) map[string]OutputPartition {
 	pm := make(map[string]OutputPartition, len(pt.Partitions))
--- a/cmd/otk-gen-partition-table/main.go
+++ b/cmd/otk-gen-partition-table/main.go
@@ -53,6 +53,7 @@ type InputPartition struct {
 type InputModifications struct {
 	PartitionMode disk.PartitioningMode               `json:"partition_mode"`
 	Filesystems   []blueprint.FilesystemCustomization `json:"filesystems"`
+	MinDiskSize   string                              `json:"min_disk_size"`
 }
 
 type Output struct {
@@ -165,12 +166,38 @@ func makePartitionTableFromOtkInput(input *Input) (*disk.PartitionTable, error) 
 	return pt, nil
 }
 
+func getDiskSizeFrom(input *Input) (diskSize uint64, err error) {
+	var defaultSize, modMinSize uint64
+
+	if input.Properties.DefaultSize != "" {
+		defaultSize, err = common.DataSizeToUint64(input.Properties.DefaultSize)
+		if err != nil {
+			return 0, err
+		}
+	}
+	if input.Modifications.MinDiskSize != "" {
+		modMinSize, err = common.DataSizeToUint64(input.Modifications.MinDiskSize)
+		if err != nil {
+			return 0, err
+		}
+	}
+	// TODO: use max() once we move to go1.21
+	if defaultSize > modMinSize {
+		return defaultSize, nil
+	}
+	return modMinSize, nil
+}
+
 func genPartitionTable(genPartInput *Input, rng *rand.Rand) (*Output, error) {
 	basePt, err := makePartitionTableFromOtkInput(genPartInput)
 	if err != nil {
 		return nil, err
 	}
-	pt, err := disk.NewPartitionTable(basePt, genPartInput.Modifications.Filesystems, 0, genPartInput.Modifications.PartitionMode, nil, rng)
+	diskSize, err := getDiskSizeFrom(genPartInput)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get the disk size: %w", err)
+	}
+	pt, err := disk.NewPartitionTable(basePt, genPartInput.Modifications.Filesystems, diskSize, genPartInput.Modifications.PartitionMode, nil, rng)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/otk-gen-partition-table/main.go
+++ b/cmd/otk-gen-partition-table/main.go
@@ -161,13 +161,8 @@ func makePartitionTableFromOtkInput(input *OtkGenPartitionInput) (*disk.Partitio
 // Missing:
 // 1. customizations^Wmodifications, e.g. extra partiton tables
 // 2. refactor, make this nicer, it sucks a bit right now
-func run(r io.Reader, rng *rand.Rand) (*OtkGenPartitionsOutput, error) {
-	var genPartInput OtkGenPartitionInput
-	if err := json.NewDecoder(r).Decode(&genPartInput); err != nil {
-		return nil, err
-	}
-
-	basePt, err := makePartitionTableFromOtkInput(&genPartInput)
+func genPartitionTable(genPartInput *OtkGenPartitionInput, rng *rand.Rand) (*OtkGenPartitionsOutput, error) {
+	basePt, err := makePartitionTableFromOtkInput(genPartInput)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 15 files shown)


internal/otkdisk/partition.go
```diff
@@ -0,0 +1,36 @@
...
+
+type Const struct {
+	KernelOptsList []string `json:"kernel_opts_list"`
+
+	// PartitionMap is generated for convenient indexing of certain partitions
+	// with predictable names in otk, such as
+	// "filesystem.partition_map.boot.uuid"
+	PartitionMap map[string]Partition `json:"partition_map"`
+
+	// Internal representation of the full partition table.
```
mvo5, line 22: Maybe we could add  here
```
	// Internal representation of the full partition table. The representation is internal to the partition tools and should not be used by otk directly. No external API gurantees about the conent
```
or something like this?

pkg/disk/partition.go
```diff
@@ -18,6 +19,44 @@ type Partition struct {
 	Payload Entity
 }
 
+// XXX: this needs refactoring and tests, it's kinda complicated
```
mvo5, line 22: I will work on this in a followup/parallel PR, this is tested indirectly via the otk-make-* tests but of course a direct test is much better and will get added. Note that this is only used by the otk-make-* tooling right now so the risk to merge in the current state is low.

mvo5 (approved):
Looks good to me (but my vote counts only +0.5 as I wrote parts of this)

internal/otkdisk/partition.go
achilleas-k: Updated the docstring.

pkg/disk/partition.go
mvo5: I started this in https://github.com/osbuild/images/pull/737

mvo5 (dismissed):
Still looks good to me (but only 0.5 because I wrote some of the code)

cmd/otk-gen-partition-table/main.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/osbuild/images/internal/buildconfig"
+	"github.com/osbuild/images/internal/cmdutil"
+	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/internal/otkdisk"
+	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/osbuild/images/pkg/disk"
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+type Input struct {
```
ondrejbudai, line 19: Since this basically becomes our public API, I think that all of these have to be documented with docstrings.

cmd/otk-gen-partition-table/main.go
```diff
@@ -0,0 +1,240 @@
...
+	Properties InputProperties   `json:"properties"`
+	Partitions []*InputPartition `json:"partitions"`
+
+	Modifications InputModifications `json:"modifications"`
+}
+
+type InputProperties struct {
+	UEFI        InputUEFI `json:"uefi"`
+	BIOS        bool      `json:"bios"`
+	Type        string    `json:"type"`
```
ondrejbudai, line 29: Can we make this an enum?

achilleas-k: I considered it but since this is an input data structure I think making it an enum string doesn't give us much benefit.  But we should validate the values somehow so perhaps an enum makes that more explicit.

ondrejbudai (changes requested):
Awesome! I played with this a bit, and found several tiny issues, see inline. :)

pkg/disk/partition.go
```diff
@@ -18,6 +19,44 @@ type Partition struct {
...
+func (p *Partition) UnmarshalJSON(data []byte) error {
+	type partAlias Partition
+	var aux struct {
+		partAlias
+		Payload struct {
```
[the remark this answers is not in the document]
mvo5: Thank you! This part of the code will be replaced with https://github.com/osbuild/images/pull/737 and I added code to handle this case now (it was not doing that before, so a huge *thank you* for providing this example).

cmd/otk-gen-partition-table/main.go
```diff
@@ -0,0 +1,240 @@
...
+	DefaultSize string    `json:"default_size"`
+	UUID        string    `json:"uuid"`
+
+	SectorSize uint64 `json:"sector_size"`
+}
+
+type InputUEFI struct {
+	Size string `json:"size"`
```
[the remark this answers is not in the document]
mvo5: I was thinking about this and the `BIOSBootPartition` and wonder if we should change the structure a little bit as both of those are about automatically creating partitions, maybe something like:
```go
diff --git a/cmd/otk-gen-partition-table/main.go b/cmd/otk-gen-partition-table/main.go
index 496914894..6d78fb2f9 100644
--- a/cmd/otk-gen-partition-table/main.go
+++ b/cmd/otk-gen-partition-table/main.go
@@ -27,17 +27,20 @@ type Input struct {
 
 // InputProperties contains global properties of the partition table
 type InputProperties struct {
-       UEFI        InputUEFI        `json:"uefi"`
-       BIOS        bool             `json:"bios"`
        Type        otkdisk.PartType `json:"type"`
        DefaultSize string           `json:"default_size"`
        UUID        string           `json:"uuid"`
+       Create      InputCreate      `json:"create"`
 
        SectorSize uint64 `json:"sector_size"`
 }
 
-// InputUEFI contains the uefi specific bits of the partition input
-type InputUEFI struct {
+// InputAutogenerate contains details what partitions to auto-generate
+type InputCreate struct {
+       BIOSBootPartition bool   `json:"bios-boot-partition"`
+       EspPartition      bool   `json:"esp-partition"`
+       EspPartitionSize  string `json:"esp-partition-size"`
+
        Size string `json:"size"`
 }
 
@@ -111,7 +114,7 @@ func makePartitionTableFromOtkInput(input *Input) (*disk.PartitionTable, error)
                Type:       string(input.Properties.Type),
                SectorSize: input.Properties.SectorSize,
        }
-       if input.Properties.BIOS {
+       if input.Properties.Create.BIOSBootPartition {
                if len(pt.Partitions) > 0 {
                        return nil, fmt.Errorf("internal error: bios partition *must* go first")
                }
@@ -122,8 +125,13 @@ func makePartitionTableFromOtkInput(input *Input) (*disk.PartitionTable, error)
                        UUID:     disk.BIOSBootPartitionUUID,
                })
        }
-       if input.Properties.UEFI.Size != "" {
-               uintSize, err := common.DataSizeToUint64(input.Properties.UEFI.Size)
+       if input.Properties.Create.EspPartition {
+               size := input.Properties.Create.EspPartitionSize
+               if size == "" {
+                       // XXX: is this a good default?
+                       size = "1 Gb"
+               }
+               uintSize, err := common.DataSizeToUint64(size)
                if err != nil {
                        return nil, err
                }

```
alternative names might be `InputAutoCreate`, `InputAutoGenerate` or similar.

cmd/otk-gen-partition-table/main.go
```diff
@@ -0,0 +1,240 @@
...
+type Input struct {
+	Properties InputProperties   `json:"properties"`
+	Partitions []*InputPartition `json:"partitions"`
+
+	Modifications InputModifications `json:"modifications"`
+}
+
+type InputProperties struct {
+	UEFI        InputUEFI `json:"uefi"`
+	BIOS        bool      `json:"bios"`
```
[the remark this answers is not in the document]
mvo5: See above.

cmd/otk-gen-partition-table/main.go
```diff
@@ -0,0 +1,240 @@
...
+			})
+		}
+	}
+
+	for _, part := range input.Partitions {
+		uintSize, err := common.DataSizeToUint64(part.Size)
+		if err != nil {
+			return nil, err
+		}
+		pt.Partitions = append(pt.Partitions, disk.Partition{
```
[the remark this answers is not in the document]
mvo5: Indeed, good point!

cmd/otk-make-partition-stages/main.go
```diff
@@ -0,0 +1,54 @@
...
+type Input = otkdisk.Data
+
+func makeImagePrepareStages(inp Input, filename string) (stages []*osbuild.Stage, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("cannot generate image prepare stages: %v", r)
+		}
+	}()
+
+	// rhel7 uses PTSgdisk, if we ever need to support this, make this
```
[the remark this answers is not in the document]
mvo5: AFAICT we are safe here, if we need to support another tool we would do something like:
```go
diff --git a/cmd/otk-make-partition-stages/main.go b/cmd/otk-make-partition-stages/main.go
index 66c25349a..6d6f76bff 100644
--- a/cmd/otk-make-partition-stages/main.go
+++ b/cmd/otk-make-partition-stages/main.go
@@ -20,10 +20,7 @@ func makeImagePrepareStages(inp Input, filename string) (stages []*osbuild.Stage
                }
        }()
 
-       // rhel7 uses PTSgdisk, if we ever need to support this, make this
-       // configurable
-       partTool := osbuild.PTSfdisk
-       stages = osbuild.GenImagePrepareStages(inp.Const.Internal.PartitionTable, inp.Const.Filename, partTool)
+       stages = osbuild.GenImagePrepareStages(inp.Const.Internal.PartitionTable, inp.Const.Filename, osbuild.PartTool(inp.Const.Internal.PartTool))
        return stages, nil
 }
 
diff --git a/internal/otkdisk/partition.go b/internal/otkdisk/partition.go
index 37b649c5c..33c199885 100644
--- a/internal/otkdisk/partition.go
+++ b/internal/otkdisk/partition.go
@@ -50,6 +50,7 @@ type Partition struct {
 // "otk-{gen,make}-*" tools for their data exchange.
 type Internal struct {
        PartitionTable *disk.PartitionTable `json:"partition-table"`
+       PartTool       string               `json:"part-tool"`
 }
 
 // PartType represents a partition type
```
and from otk-gen-* we would set it (or move it out of internal if it should be something the user should control). Or am I misunderstanding the comment here?

pkg/disk/partition.go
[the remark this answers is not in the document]
mvo5: Might be easiest to review #737 first and then I can just rebase on top of this and drop this initial and slightly naive implementation for the unmarshal.

mvo5 (approved):
I rebased and removed the old marshal code (now that #775 is in) - this is ready now. Please consider my approval only a +0.5 as I wrote good chunks of this.

achilleas-k (commented):
I reviewed the most recent changes and it all LGTM.  I'd approve but I can't since I'm the author/owner here.

The weak rng usage in tests needs comments for linter skipping.

I think we should go ahead and merge this without requiring further detailed reviews.  The PR isn't in any critical path yet and I think bugs, edge cases, and usability issues will surface once we start using the features in otk.  We can iterate on the implementation then.

cmd/otk-gen-partition-table/main.go
```diff
@@ -0,0 +1,268 @@
...
+			return 0, err
+		}
+	}
+	if input.Modifications.MinDiskSize != "" {
+		modMinSize, err = common.DataSizeToUint64(input.Modifications.MinDiskSize)
+		if err != nil {
+			return 0, err
+		}
+	}
+	// TODO: use max() once we move to go1.21
```
ondrejbudai, line 193: The future is now.

cmd/otk-make-partition-mounts-devices/main.go
```diff
@@ -0,0 +1,50 @@
...
+type Output struct {
+	RootMountName string                    `json:"root_mount_name"`
+	Mounts        []osbuild.Mount           `json:"mounts"`
+	Devices       map[string]osbuild.Device `json:"devices"`
+}
+
+func run(r io.Reader, w io.Writer) error {
+	var inp Input
+	if err := json.NewDecoder(r).Decode(&inp); err != nil {
+		return err
```
ondrejbudai, line 24: Wrap me. :)

cmd/otk-make-partition-stages/main.go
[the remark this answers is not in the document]
ondrejbudai: But is an omnifest author supposed to touch the internal fields? Feels fishy... That's why I proposed adding another layer around `internal`. Alternatively, we can put it into `InputProperties` and pass it from `gen-partition-table` to `make-partition-stages` in `internal`. Might be simpler for the user.

internal/otkdisk/partition.go
```diff
@@ -0,0 +1,70 @@
+package otkdisk
+
+import (
+	"fmt"
+
+	"golang.org/x/exp/slices"
```
ondrejbudai, line 6: The future is indeed now, so we can just use `slices` here. Might be worth doing this in a follow-up and changing it globally. :)

mvo5 (dismissed):
Fwiw, I agree

ondrejbudai (commented):
I'm loving this! :) I have some concerns about field names, missing documentation and hard-to-understand error messages. However, functionality looks alright to me. :) I don't think that anything is a real blocker here, but it might be a good idea to adjust the field names before merging, so we can rely on them from now on.

Also, we will need an fstab stage generator. Once again, not sure if this should be a part of this PR, or a follow-up.

cmd/otk-gen-partition-table/main.go
```diff
@@ -0,0 +1,268 @@
...
+	if input.Properties.Create.EspPartition {
+		size := input.Properties.Create.EspPartitionSize
+		if size == "" {
+			size = "1 GiB"
+		}
+		uintSize, err := common.DataSizeToUint64(size)
+		if err != nil {
+			return nil, err
+		}
+		if uintSize > 0 {
```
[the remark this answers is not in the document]
achilleas-k: Do we want to error or treat 0 as a "don't create"?

mvo5 (dismissed):
Thanks for the extra tweaks/fixes!

ondrejbudai (approved):
Thanks!

supakeen (dismissed):

supakeen:
I've resolved all conversations in this PR so it can go into the merge since it has multiple acks since those conversations were created (a lot of them *should* go into followups!).

mvo5 (dismissed):
I had to rebase as there was a rename in the code that made this branch stop compiling (genMountsDevicesFromPt -> GenMountsDevicesFromPT)

mvo5 (approved):

PR #737, opened 2024-06-07
disk: support json unmarshal for `Partition`

mvo5:
The `disk.Partition` uses interfaces inside the data structure. To support json unmarshal we need a custom `UnmarshalJSON()` that peeks into the marshalled json and detects what concrete type to use.

This commit provides the `disk.Partition.UnmarshalJSON` implementation and a test.

Split out from https://github.com/osbuild/images/pull/732 with extra test.

Draft until I find a little bit more time to make the tests a bit more comprehensive and until the other payloads are supported.

### DIFF
--- a/pkg/disk/btrfs.go
+++ b/pkg/disk/btrfs.go
@@ -3,10 +3,15 @@ package disk
 import (
 	"fmt"
 	"math/rand"
+	"reflect"
 	"strings"
 
 	"github.com/google/uuid"
 )
+
+func init() {
+	entityTypes = append(entityTypes, reflect.TypeOf(Btrfs{}))
+}
 
 type Btrfs struct {
 	UUID       string
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -20,6 +20,7 @@ import (
 	"encoding/hex"
 	"io"
 	"math/rand"
+	"reflect"
 
 	"github.com/google/uuid"
 )
@@ -49,6 +50,10 @@ const (
 	// Extended Boot Loader Partition
 	XBootLDRPartitionGUID = "BC13C2FF-59E6-4262-A352-B275FD6F7172"
 )
+
+// entityTypes must contain all the types that implemnt "Entity"
+// in the package so that json unmarshal works transparently
+var entityTypes []reflect.Type
 
 // Entity is the base interface for all disk-related entities.
 type Entity interface {
--- a/pkg/disk/filesystem.go
+++ b/pkg/disk/filesystem.go
@@ -2,9 +2,14 @@ package disk
 
 import (
 	"math/rand"
+	"reflect"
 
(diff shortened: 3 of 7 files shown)


achilleas-k:
Sorry for taking so long to review this.

In terms of implementation this lgtm, but I wonder if it's the right approach.  I think we can both agree that this isn't the cleanest way to work with these interfaces and types (as you mentioned in a comment).  The advantage is that we get to use our implementation as is, we can serialize it and deserialize it directly and the deserializer needs to do the "dirty work".

Before we merge this I'd like to consider an alternative: Would it be easier to work with these things if we had a serialization format that added extra properties to aide switching instead?  Here's a very rough draft of what I'm imagining:

```go
type partSerializer struct {
  Start       uint64 `json:"start"`
	Size        uint64 `json:"size"`
	Type        string `json:"type"`
	Bootable    bool   `json:"bootable"`
	UUID        string `json:"uuid"`
  PayloadType string `json:"payload_type"`
  Payload     Entity `json:"payload"`
}

func (p *Partition) MarshalJSON() ([]byte, error) {
  // figure out payload type using reflection

  ps := partSerializer{
    ... // copy Partition properties from p
    PayloadType: pType,  // from reflect
  }

  // serialize and return ps
}

func (p *Partition) UnmarshalJSON(data []byte) error {
  ps := partSerializer{}
  json.Unmarshal(data, &ps)  // but do it with DisallowUnknownFields
  p = &Partition{
    ... // copy Partition properties from ps into p
  }

  switch ps.PayloadType {
    ... // one case per known type
    case "btrfs":
      p.Payload = ps.Payload.(Btrfs)
    ...
    default:
      return err  // unknown type
  }
}
```

The advantage of the approach in this PR is that it's less work to add new payload types; you just need to register it with the `entityTypes` and it's good to go, whereas with the approach in this comment we would need to add it to the serializer which is less obvious I think.

The advantage of the approach in this comment though is that it's more readable (in my opinion at least; feel free to disagree).  Also there's the (perhaps questionable) benefit that the `payload_type` string in the serialized output makes the serialized format more explicit and human readable; you can tell what payload type a partition holds from just reading the string instead of inferring from the properties of the payload.  We also get a serialization format (`partSerializer`) that's not necessarily a direct mapping to the actual object, which can be both a benefit or an annoyance.  It's sometimes nice to be able to change internal representations without affecting the serialized format (especially if the serialized format is part of the API) and vice versa, but it also means you need to make changes in two places when you want to add things to the base type.

mvo5:
I did https://github.com/osbuild/images/pull/776 and https://github.com/osbuild/images/pull/775 with various levels of magic - I hope I got all the permuations.

mvo5:
I'm closing this in favor of https://github.com/osbuild/images/pull/775 which implements the suggestions from @achilleas-k

PR #742, opened 2024-06-10
manifest: Add awscli2 to RHEL 9 AMI (CLOUDX-913)

major:
The awscli2 package is included in the upcoming RHEL 9.5.0 release and it would be helpful to include it in AWS AMIs by default.

### DIFF
--- a/pkg/distro/rhel/rhel9/ami.go
+++ b/pkg/distro/rhel/rhel9/ami.go
@@ -240,6 +240,11 @@ func ec2CommonPackageSet(t *rhel.ImageType) rpmmd.PackageSet {
 		},
 	}.Append(distroSpecificPackageSet(t))
 
+	// Include awscli2 on RHEL 9.5+ (CLOUDX-913)
+	if common.VersionGreaterThanOrEqual(t.Arch().Distro().OsVersion(), "9.5") {
+		ps.Include = append(ps.Include, "awscli2")
+	}
+
 	return ps
 }
 


bcl:
If this is only in 9.5 and later it will need a check for `common.VersionGreaterThanOrEqual(t.Arch().Distro().OsVersion(), "9.5")`

major:
> If this is only in 9.5 and later it will need a check for `common.VersionGreaterThanOrEqual(t.Arch().Distro().OsVersion(), "9.5")`

Thanks, @bcl! I missed that even though it was in the section just below the one I edited. 🤦‍♂️

bcl (approved):
Looks ok to me, assuming 2 things:
- we want it on all ec2 based images (I assume we do)
- It is already in the 9.5 repos (otherwise testing will blow up)

major:
> Looks ok to me, assuming 2 things:
> 
>     * we want it on all ec2 based images (I assume we do)

That's the current plan.

>     * It is already in the 9.5 repos (otherwise testing will blow up)

It is in the nightly composes today.

achilleas-k:
> > Looks ok to me, assuming 2 things:
> > ```
> > * we want it on all ec2 based images (I assume we do)
> > ```
> 
> That's the current plan.
> 
> > ```
> > * It is already in the 9.5 repos (otherwise testing will blow up)
> > ```
> 
> It is in the nightly composes today.

Looks like it's already been there for a couple of weeks 'cause it's in our rpmrepo snapshots from 01 June.

achilleas-k (approved):
Thank you!!  LGTM

PR #751, opened 2024-06-18
Improve error descriptions

schuellerf:
adding context to help pin down possible errors

### DIFF
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -196,7 +196,7 @@ func (s *Solver) SetProxy(proxy string) error {
 func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet) ([]rpmmd.PackageSpec, []rpmmd.RepoConfig, error) {
 	req, rhsmMap, err := s.makeDepsolveRequest(pkgSets)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("makeDepsolveRequest failed: %w", err)
 	}
 
 	// get non-exclusive read lock
@@ -205,7 +205,7 @@ func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet) ([]rpmmd.PackageSpec, []rp
 
 	output, err := run(s.dnfJsonCmd, req)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("running osbuild-depsolve-dnf failed:\n%w", err)
 	}
 	// touch repos to now
 	now := time.Now().Local()
@@ -219,7 +219,7 @@ func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet) ([]rpmmd.PackageSpec, []rp
 	dec := json.NewDecoder(bytes.NewReader(output))
 	dec.DisallowUnknownFields()
 	if err := dec.Decode(&result); err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("decoding depsolve result failed: %w", err)
 	}
 
 	packages, repos := result.toRPMMD(rhsmMap)
@@ -721,6 +721,13 @@ func (err Error) Error() string {
 // message.
 func parseError(data []byte, repos []repoConfig) Error {
 	var e Error
+	if len(data) == 0 {
+		return Error{
+			Kind:   "InternalError",
+			Reason: "dnf-json output was empty",
+		}
+	}
+
 	if err := json.Unmarshal(data, &e); err != nil {
 		// dumping the error into the Reason can get noisy, but it's good for troubleshooting
 		return Error{
@@ -776,7 +783,7 @@ func run(dnfJsonCmd []string, req *Request) ([]byte, error) {
 	cmd := exec.Command(ex, args...)
 	stdin, err := cmd.StdinPipe()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("creating stdin pipe for %s failed: %w", ex, err)
 	}
 
 	cmd.Stderr = os.Stderr
@@ -785,12 +792,12 @@ func run(dnfJsonCmd []string, req *Request) ([]byte, error) {
 
 	err = cmd.Start()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("starting %s failed: %w", ex, err)
 	}
 
 	err = json.NewEncoder(stdin).Encode(req)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("encoding request for %s failed: %w", ex, err)
 	}
 	stdin.Close()
 
--- a/pkg/osbuild/osbuild-exec.go
+++ b/pkg/osbuild/osbuild-exec.go
@@ -70,6 +70,9 @@ func RunOSBuild(manifest []byte, store, outputDirectory string, exports, checkpo
 
 	if result {
 		// try to decode the output even though the job could have failed
+		if stdoutBuffer.Len() == 0 {
+			return nil, fmt.Errorf("osbuild did not return any output")
+		}
 		decodeErr := json.Unmarshal(stdoutBuffer.Bytes(), &res)
 		if decodeErr != nil {
 			return nil, fmt.Errorf("error decoding osbuild output: %v\nthe raw output:\n%s", decodeErr, stdoutBuffer.String())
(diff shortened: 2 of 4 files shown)


pkg/dnfjson/dnfjson.go
```diff
@@ -196,7 +196,7 @@ func (s *Solver) SetProxy(proxy string) error {
 func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet) ([]rpmmd.PackageSpec, []rpmmd.RepoConfig, error) {
 	req, rhsmMap, err := s.makeDepsolveRequest(pkgSets)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("makeDepsolveRequest failed: %w", err)
```
supakeen, line 199: I tend to forget our minimal Go version requirement but if it's `>= 1.20` I would prefer `errors.Join(errors.New("makeDepsolveRequest failed"), err)` here.

mvo5: I'm curious why `errors.Join()` is preferred here. My understanding so far was that `%w` is used to provide additional context (which seems to be the case here) and errors.Join() is more common in cases when handling one error results in an additional erorr (the classic example would be an error from `defer os.Close()`). But maybe I'm missing something here or am not fully up-to-date?

pkg/osbuild/osbuild-exec.go
```diff
@@ -70,9 +70,13 @@ func RunOSBuild(manifest []byte, store, outputDirectory string, exports, checkpo
 
 	if result {
 		// try to decode the output even though the job could have failed
-		decodeErr := json.Unmarshal(stdoutBuffer.Bytes(), &res)
-		if decodeErr != nil {
-			return nil, fmt.Errorf("error decoding osbuild output: %v\nthe raw output:\n%s", decodeErr, stdoutBuffer.String())
+		if stdoutBuffer.Len() == 0 {
+			return nil, fmt.Errorf("osbuild did not return any output")
+		} else {
```
mvo5, line 75: The `else` here seems to be not needed because when the `if` is taken it will return and go prefers to avoid the "else" in this case (c.f. https://go.dev/doc/effective_go#if)

pkg/dnfjson/dnfjson.go
```diff
@@ -776,7 +783,7 @@ func run(dnfJsonCmd []string, req *Request) ([]byte, error) {
 	cmd := exec.Command(ex, args...)
 	stdin, err := cmd.StdinPipe()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("calling %s failed: %w", ex, err)
 	}
 
 	cmd.Stderr = os.Stderr
```
mvo5, line 789: (unrelated to the PR) - this is slightly unusual, generally capturing stderr to provide more meaningful error message programmatically is a good idea.

pkg/dnfjson/dnfjson.go
```diff
@@ -205,7 +205,7 @@ func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet) ([]rpmmd.PackageSpec, []rp
 
 	output, err := run(s.dnfJsonCmd, req)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("running osbuild-depsolve-dnf failed:\n%w", err)
```
mvo5, line 208: Why is there an extra `\n` but in the others changes there is not?

pkg/dnfjson/dnfjson.go
```diff
@@ -721,6 +721,13 @@ func (err Error) Error() string {
 // message.
 func parseError(data []byte, repos []repoConfig) Error {
 	var e Error
+	if len(data) == 0 {
```
mvo5, line 724: I would love to see at least a tiny bit of testing, maybe something like:
```go
diff --git a/pkg/dnfjson/dnfjson_test.go b/pkg/dnfjson/dnfjson_test.go
index 20771ce5a..13e2effe0 100644
--- a/pkg/dnfjson/dnfjson_test.go
+++ b/pkg/dnfjson/dnfjson_test.go
@@ -776,3 +776,15 @@ func TestRepoConfigMarshalAlsmostEmpty(t *testing.T) {
        // double check here that anything that uses pointers has "omitempty" set
        assert.Equal(t, string(js), `{"id":"","gpgcheck":false,"repo_gpgcheck":false}`)
 }
+
+func TestRunErrorEmptyOutput(t *testing.T) {
+       fakeDnfJsonPath := filepath.Join(t.TempDir(), "dnfjson")
+       fakeDnfJsonNoOutput := `#!/bin/sh -e
+exit 1
+`
+       err := os.WriteFile(fakeDnfJsonPath, []byte(fakeDnfJsonNoOutput), 0o755)
+       assert.NoError(t, err)
+
+       _, err = run([]string{fakeDnfJsonPath}, &Request{})
+       assert.EqualError(t, err, `DNF error occurred: InternalError: dnf-json output was empty`)
+}
```
so that if we need to tweak our strings we can easily check that they are correct. Go error handling has this slightly annyoing tendency to "stutter" as it is often hard to know how much context the previous layer provided. The only good way to deal with this I found so far was just writing tests :/ But I think it's a bit excessive to test all of the new errors, but having an example might be good to double check for "stuttering" and also that the formating looks nice.

pkg/dnfjson/dnfjson.go
```diff
@@ -776,7 +783,7 @@ func run(dnfJsonCmd []string, req *Request) ([]byte, error) {
 	cmd := exec.Command(ex, args...)
 	stdin, err := cmd.StdinPipe()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("calling %s failed: %w", ex, err)
```
mvo5, line 786: This might be a bit misleading, the error is from `StdinPipe` so maybe instead of `calling %s failed` something like `creating stdin pipe failed` or similar ?

pkg/dnfjson/dnfjson.go
supakeen: Both can do both (starting at 1.20) historically `%w` could only be used once so that probably shoehorned it into the additional context role. It's a preference, I find it to be a bit easier to read (like a slice of errors).

mvo5 (commented):
Thank you! It's great to see the error handling improved. I have some questions inline and some (small) suggestion. Looking forward to see this landed soon as it looks like it will nicely improve things :)

pkg/dnfjson/dnfjson.go
schuellerf: Hmm, maybe I saw that in this case `err` is larger and this looks nicer in the end - but I can also make this symmetric with the rest

pkg/dnfjson/dnfjson.go
schuellerf: > I tend to forget our minimal supported Go version requirement but if it's `>= 1.20` I would prefer `errors.Join(errors.New("makeDepsolveRequest failed"), err)` for this idiom.

who can confirm our minimal supported golang version?

mvo5 (dismissed):
Thank you, this is a nice step forward.

pkg/dnfjson/dnfjson.go
mvo5: I don't care too much, I approved the PR. I personally would still make it symetrical unless there is a good reason not to (having a test that contains a real world example might be nice way to check/document what the error does look like). But I approved, those are details.

pkg/dnfjson/dnfjson.go
mvo5: The go version according to go.mod is 1.20. And yeah, of course %w and errors.Join can do the same, I was mostly wondering what the idiomatic way is. Looking at the go standard library it seems `errors.Join()` is mostly used when there are multiple errors from e.g. multiple go routines or multiple files (e.g. modload/init.go) and it's relatively rare (I counted 6 with a quick grep in go-1.22). And %w is used a lot more (135) and it seems (from a subset of samples I looked at) that it is mostly used to provide extra context (I did not find this for errors.Join())). Based on this it seems the use in this PR is fine (but of course there is an argument that errors.Join is relatively new and the usage has just not caught up yet in the stdlib).

pkg/dnfjson/dnfjson.go
achilleas-k: `osbuild-depsolve-dnf` prints some errors to stderr (check the `printe()` calls in the script) but all meaningful (captured) info and errors are returned to stdout.  Piping stderr straight out is done because libdnf will print a lot of useless warnings that we don't care about.  We _could_ parse them and filter out the ones we expect (`No match for group package <whatever>`) and maybe log anything else.`

PR #752, opened 2024-06-18
Refactor OpenSCAP stage config (HMS-3826)

kingsleyzissou:
This PR is in preparation for a follow up PRs to simplify tailoring to accept an `xml` file.
It's an attempt to simplify the stage creation options using an internal abstraction to create
the options.

This splits up #708, drops the `json` tailoring and the tailoring `overrides`.

### DIFF
--- a/pkg/customizations/oscap/oscap.go
+++ b/pkg/customizations/oscap/oscap.go
@@ -43,8 +43,9 @@ const (
 	defaultRHEL8Datastream   string = "/usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml"
 	defaultRHEL9Datastream   string = "/usr/share/xml/scap/ssg/content/ssg-rhel9-ds.xml"
 
-	// tailoring directory path
-	tailoringDirPath string = "/usr/share/xml/osbuild-openscap-data"
+	// oscap related directories
+	DataDir          string = "/oscap_data"
+	TailoringDirPath string = "/usr/share/xml/osbuild-openscap-data"
 )
 
 func DefaultFedoraDatastream() string {
@@ -83,9 +84,9 @@ func IsProfileAllowed(profile string, allowlist []Profile) bool {
 
 func GetTailoringFile(profile string) (string, string, *fsnode.Directory, error) {
 	newProfile := fmt.Sprintf("%s_osbuild_tailoring", profile)
-	path := filepath.Join(tailoringDirPath, "tailoring.xml")
+	path := filepath.Join(TailoringDirPath, "tailoring.xml")
 
-	tailoringDir, err := fsnode.NewDirectory(tailoringDirPath, nil, nil, nil, true)
+	tailoringDir, err := fsnode.NewDirectory(TailoringDirPath, nil, nil, nil, true)
 	if err != nil {
 		return "", "", nil, err
 	}
--- a/pkg/customizations/oscap/oscap.go
+++ b/pkg/customizations/oscap/oscap.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
-
-	"github.com/osbuild/images/pkg/customizations/fsnode"
 )
 
 type Profile string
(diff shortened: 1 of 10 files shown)


pkg/customizations/oscap/oscap.go
```diff
@@ -40,6 +40,21 @@ const (
...
+	Datastream         string
+	ProfileID          string
+	TailoringPath      string
+	CompressionEnabled bool
+}
+
+type TailoringConfig struct {
+	RemediationConfig
+	Filepath   string
+	NewProfile string
```
thozza, line 53: Nitpick: Is this the new profile ID? If so, would it make sense to call it `NewProfileID` or `TailoredProfileID`? :thinking:

pkg/distro/fedora/images.go
```diff
@@ -200,40 +198,27 @@ func osCustomizations(
...
-
-			// add the parent directory for the tailoring file
-			osc.Directories = append(osc.Directories, tailoringDir)
+			remediationConfig.ProfileID = tailoringConfig.NewProfile
+			remediationConfig.TailoringPath = tailoringConfig.Filepath
 		}
 
-		osc.OpenSCAPConfig = osbuild.NewOscapRemediationStageOptions(oscapDataDir, oscapStageOptions)
+		osc.OpenSCAPTailorConfig = osbuild.NewOscapAutotailorStageOptions(tailoringConfig)
+		osc.OpenSCAPConfig = osbuild.NewOscapRemediationStageOptions(oscapDataDir, &remediationConfig)
```
thozza, line 221: I think that it would be nice to make `osc.OpenSCAPTailorConfig` and `osc.OpenSCAPConfig` be of their respective types from the `oscap` package, instead of `osbuild` stage options. This was the grand plan since the beginning for `OSCustomizations`, but we never finished it. Since you are already refactoring a lot of the code, it may make sense to make this change as well...

thozza (commented):
Nice refactoring. :+1: 

I've added a few comments. On a general note, I think that it would probably make sense to move the `oscapDataDir` definition to the `oscap` package and make it exported. It is the same for all distros.

pkg/customizations/oscap/oscap.go
kingsleyzissou: Yeah good suggestion

pkg/distro/fedora/images.go
kingsleyzissou: Ah cool. Yeah this makes a lot of sense. I'll go ahead and do that

pkg/customizations/oscap/oscap.go
```diff
@@ -43,10 +39,25 @@ const (
...
 
+type RemediationConfig struct {
+	Datastream         string
+	ProfileID          string
+	TailoringPath      string
+	CompressionEnabled bool
+}
+
+type TailoringConfig struct {
+	RemediationConfig
```
thozza, line 54: I just noticed that embedded the `RemediationConfig` makes the `TailoringConfig` have the tailoring file path set twice. Once in `TailoringPath` and also in `Filepath`. Is there some good reason behind this?

thozza (commented):
Thanks for the changes. I've added one comment, the rest looks good. :+1:

pkg/customizations/oscap/oscap.go
kingsleyzissou: Great catch, no I think we can just use the one from the `RemediationConfig`

kingsleyzissou: I've added it to the `RemediationConfig` struct since it is potentially needed for both stage options.

pkg/osbuild/oscap_autotailor_stage.go
```diff
@@ -50,6 +50,12 @@ func NewOscapAutotailorStageOptions(options *oscap.TailoringConfig) *OscapAutota
 		return nil
 	}
 
+	// TODO: don't panic! unfortunately this would involve quite
+	// a big refactor and we still need to be a bit defensive here
+	if options.RemediationConfig.TailoringPath == "" {
```
kingsleyzissou, line 55: This might be a bit controversial. I've added it to its own commit for now, but happy to delete or squash depending on how we feel about it.

kingsleyzissou:
CI errors are a result of the code 🤦

thozza (dismissed):
LGTM :+1:

kingsleyzissou:
I had to add https://github.com/osbuild/images/blob/931eee58dec919afc20aebf69a5cd99018eda9de/test/scripts/base-host-check.sh#L46 to fix the tests

thozza (approved):

PR #753, opened 2024-06-18
pre-commit: introduce pre-commit and fix linters

schuellerf:
not sure if the settings of those pre-commit checks/linters are perfect but this commit would be a start to avoid having the round-trip with failing github tests, hopefully


test/scripts/boot-image
```diff
@@ -66,18 +78,34 @@ def ensure_uncompressed(filepath):
 def cmd_boot_aws(arch, image_name, privkey, pubkey, image_path, script_cmd):
     # pylint: disable=too-many-arguments
     aws_config = get_aws_config()
-    cmd = ["go", "run", "./cmd/boot-aws", "run",
-           "--access-key-id", aws_config["key_id"],
```
mvo5, line 70: Uh, let's please not do this change here, this makes the `cmd` much harder to read IMHO. It's a bit silly that `ruff` cannot be configured to do better here (or can it)?

schuellerf: I saw that too - I'll switch this to draft and dig deeper

schuellerf:
> A bit of a meta comment, would it make sense to introduce PRs with one linter at a time? That would probably make both the review and the commit history a bit easier to read :)

Not sure if this would create an unnecessary high number of PRs - but the straight forward ones and `ruff` can be split, I guess🤔 😁

schuellerf:
@mvo5 autopep8 seems to be ok, fixing lines that are too long but keeping others… please check

docs/developer/cmds.md
```diff
@@ -2,7 +2,7 @@
 
 The following utilities, defined in the `cmd/` directory, are useful for
 development and testing. They **should not** be relied on for production
-purposes. In particular, command line options and default behaviour can change
+purposes. In particular, command line options and default behavior can change
```
kingsleyzissou, line 5: 😢 😭

kingsleyzissou (approved):

docs/developer/cmds.md
schuellerf: Not sure if we should argue about
```
matrix:
    aspell:
      d: en_US
```
😉

kingsleyzissou: Not arguing, just hurts my soul a little 😆😭

PR #758, opened 2024-06-24
pkg/osbuild: Skip EFI partition check for non UEFI platforms

yoheiueda:
Currently, validateBootupdMounts always ensures that `/boot/efi` exists. This requirement is not necessary on non-UEFI platforms such as s390x.

This patch skips the check for `/boot/efi` when UEFIVendor is not set.

This change is required for this issue: https://github.com/osbuild/bootc-image-builder/issues/460


achilleas-k (approved):
Thanks!

supakeen (approved):

PR #760, opened 2024-06-24
github: disable the cron scheduled updates of the BIB ref (COMPOSER-2245)

achilleas-k:
Our bootc containers are incompatible with newer versions of the bootc-image-builder tool because the tool sets options that require composefs, which our images don't support yet.

Temporarily disable the update cronjob until we fix this.

### DIFF
--- a/.github/workflows/update-bootc-image-builder.yml
+++ b/.github/workflows/update-bootc-image-builder.yml
@@ -4,9 +4,10 @@ name: "Update bootc-image-builder ref"
 
 on:
   workflow_dispatch:
-  schedule:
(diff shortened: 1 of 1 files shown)


schuellerf:
Is there already a github or jira issue for this? 🤔 Seems to be an important update to be implemented...

achilleas-k:
> Is there already a github or jira issue for this? 🤔 Seems to be an important update to be implemented...

https://issues.redhat.com/browse/COMPOSER-2245?jql=text%20~%20%22iot-bootable-container%22

schuellerf (approved):
LGTM

PR #764, opened 2024-06-25
Make sure the default kickstart path is used when the option is unset

### LINKED ISSUE #763: unset `AnacondaContainerInstaller.Kickstart.Path` returns an error
`AnacondaContainerInstaller.Kickstart.Path` can be unset, see:

https://github.com/osbuild/images/blob/14f8342f5475b7a2951ee5adf847512a72e94c43/pkg/image/anaconda_container_installer.go#L106

Let's trace it:

It's passed into an iso tree pipeline:
https://github.com/osbuild/images/blob/14f8342f5475b7a2951ee5adf847512a72e94c43/pkg/image/anaconda_container_installer.go#L120

It gets passed into kickstart options:

https://github.com/osbuild/images/blob/14f8342f5475b7a2951ee5adf847512a72e94c43/pkg/manifest/anaconda_installer_iso_tree.go#L413

It's passed into the generator of "raw kickstarts":

https://github.com/osbuild/images/blob/14f8342f5475b7a2951ee5adf847512a72e94c43/pkg/manifest/anaconda_installer_iso_tree.go#L506

It's passed into a method creating "FS nodes":

https://github.com/osbuild/images/blob/14f8342f5475b7a2951ee5adf847512a72e94c43/pkg/osbuild/kickstart_stage.go#L263

But this fails due to an empty path:

https://github.com/osbuild/images/blob/14f8342f5475b7a2951ee5adf847512a72e94c43/pkg/customizations/fsnode/fsnode.go#L78

---

Thus, `AnacondaContainerInstaller.Kickstart.Path` currently must be set in all cases, otherwise we will get an error. Thus, the first linked snippet doesn't make much sense, and its behaviour should probably "inverted": If `AnacondaContainerInstaller.Kickstart.Path` is unset, set it to the default path, so the same path is correctly used in all other places.

achilleas-k:
**image: allow unset kickstart options where possible**

For the AnacondaContainerInstaller, having empty kickstart options when
creating the ImageKind should be possible.  The only option that's
required is the Kickstart.Path, which has a default value we can set.

For the AnacondaOSTreeInstaller, at least the OSTree options are
required, so return an error when they're not set, but also set the
Kickstart.Path to the default when it's not already set from the caller.

---

**image: drop InteractiveDefaultKickstart in AnacondaContainerInstaller**

This isn't used and can cause unnecessary errors when 'img.Kickstart' is
not set.

---

**test: instantiate and serialize anaconda manifests without ks options**

Test that anaconda manifests can be generated without error when
kickstart options are not set (or when only the required ones are set).
This ensures that no errors or panics occur when the kickstart options
are nil or when the kickstart path is empty.

---

Fixes #763

### DIFF
--- a/pkg/image/anaconda_container_installer.go
+++ b/pkg/image/anaconda_container_installer.go
@@ -102,11 +102,14 @@ func (img *AnacondaContainerInstaller) InstantiateManifest(m *manifest.Manifest,
 	bootTreePipeline.UEFIVendor = img.Platform.GetUEFIVendor()
 	bootTreePipeline.ISOLabel = img.ISOLabel
 
-	ksPath := osbuild.KickstartPathOSBuild
-	if img.Kickstart != nil && img.Kickstart.Path != "" {
-		ksPath = img.Kickstart.Path
+	if img.Kickstart == nil {
+		img.Kickstart = &kickstart.Options{}
 	}
-	bootTreePipeline.KernelOpts = []string{fmt.Sprintf("inst.stage2=hd:LABEL=%s", img.ISOLabel), fmt.Sprintf("inst.ks=hd:LABEL=%s:%s", img.ISOLabel, ksPath)}
+	if img.Kickstart.Path == "" {
+		img.Kickstart.Path = osbuild.KickstartPathOSBuild
+	}
+
+	bootTreePipeline.KernelOpts = []string{fmt.Sprintf("inst.stage2=hd:LABEL=%s", img.ISOLabel), fmt.Sprintf("inst.ks=hd:LABEL=%s:%s", img.ISOLabel, img.Kickstart.Path)}
 	if img.FIPS {
 		bootTreePipeline.KernelOpts = append(bootTreePipeline.KernelOpts, "fips=1")
 	}
--- a/pkg/image/anaconda_container_installer.go
+++ b/pkg/image/anaconda_container_installer.go
@@ -76,11 +76,6 @@ func (img *AnacondaContainerInstaller) InstantiateManifest(m *manifest.Manifest,
 	anacondaPipeline.ExtraPackages = img.ExtraBasePackages.Include
 	anacondaPipeline.ExcludePackages = img.ExtraBasePackages.Exclude
 	anacondaPipeline.ExtraRepos = img.ExtraBasePackages.Repositories
-	anacondaPipeline.InteractiveDefaultsKickstart = &kickstart.Options{
-		Users:  img.Kickstart.Users,
-		Groups: img.Kickstart.Groups,
-		Path:   osbuild.KickstartPathOSBuild,
-	}
 	anacondaPipeline.Variant = img.Variant
 	anacondaPipeline.Biosdevname = (img.Platform.GetArch() == arch.ARCH_X86_64)
 	anacondaPipeline.Checkpoint()
--- a/pkg/image/anaconda_ostree_installer.go
+++ b/pkg/image/anaconda_ostree_installer.go
@@ -96,11 +96,13 @@ func (img *AnacondaOSTreeInstaller) InstantiateManifest(m *manifest.Manifest,
 	bootTreePipeline.UEFIVendor = img.Platform.GetUEFIVendor()
 	bootTreePipeline.ISOLabel = img.ISOLabel
 
-	ksPath := osbuild.KickstartPathOSBuild
-	if img.Kickstart != nil && img.Kickstart.Path != "" {
-		ksPath = img.Kickstart.Path
+	if img.Kickstart == nil || img.Kickstart.OSTree == nil {
+		return nil, fmt.Errorf("kickstart options not set for ostree installer")
 	}
-	bootTreePipeline.KernelOpts = []string{fmt.Sprintf("inst.stage2=hd:LABEL=%s", img.ISOLabel), fmt.Sprintf("inst.ks=hd:LABEL=%s:%s", img.ISOLabel, ksPath)}
+	if img.Kickstart.Path == "" {
+		img.Kickstart.Path = osbuild.KickstartPathOSBuild
+	}
+	bootTreePipeline.KernelOpts = []string{fmt.Sprintf("inst.stage2=hd:LABEL=%s", img.ISOLabel), fmt.Sprintf("inst.ks=hd:LABEL=%s:%s", img.ISOLabel, img.Kickstart.Path)}
 	if img.FIPS {
 		bootTreePipeline.KernelOpts = append(bootTreePipeline.KernelOpts, "fips=1")
 	}
--- a/pkg/image/installer_image_test.go
+++ b/pkg/image/installer_image_test.go
@@ -1,0 +1,130 @@
+package image_test
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/osbuild/images/pkg/container"
+	"github.com/osbuild/images/pkg/customizations/kickstart"
+	"github.com/osbuild/images/pkg/image"
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/images/pkg/ostree"
+	"github.com/osbuild/images/pkg/platform"
+	"github.com/osbuild/images/pkg/rpmmd"
+	"github.com/osbuild/images/pkg/runner"
+	"github.com/stretchr/testify/assert"
+)
+
+func mockPackageSets() map[string][]rpmmd.PackageSpec {
+	return map[string][]rpmmd.PackageSpec{
+		"build": {
+			{
+				Name:     "coreutils",
+				Checksum: "sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc",
+			},
+		},
+		"os": {
+			{
+				Name:     "kernel",
+				Checksum: "sha256:eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
+			},
+		},
+		"anaconda-tree": {
+			{
+				Name:     "kernel",
+				Checksum: "sha256:eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
+			},
+		},
+	}
+}
+
+func mockContainerSpecs() map[string][]container.Spec {
+	return map[string][]container.Spec{
+		"bootiso-tree": {
+			{
+				Source:  "repo.example.com/container",
+				Digest:  "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+				ImageID: "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+			},
+		},
+	}
+}
+
+func mockOSTreeCommitSpecs() map[string][]ostree.CommitSpec {
+	return map[string][]ostree.CommitSpec{
+		"bootiso-tree": {
+			{
+				Ref: "test/ostree/3",
+				URL: "http://localhost:8080/repo",
+			},
+		},
+	}
+}
+
+var testPlatform = &platform.X86{
+	BasePlatform: platform.BasePlatform{
+		ImageFormat: platform.FORMAT_ISO,
+	},
+	BIOS:       true,
+	UEFIVendor: "test",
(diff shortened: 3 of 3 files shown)


achilleas-k:
We should have a lot more tests like these and also reconsider what we set in each constructor.  I always liked the idea of having, for example, the `NewAnacondaContainerInstaller()` take as args all the properties it *requires* and exposing all other properties publicly so they can be set optionally from the caller.  An alternative idea, from @mvo5, was to have the optional properties set by an options struct that's always the last argument and can be nil.

Either way, we sort of forgot about this in some places and I see cases where a required property isn't part of the constructor argument list, making the pipeline generator hard to work with; looking at the struct, you'd have to _guess_ which properties are required and which aren't.

A lot of this will probably go away with otk, so it might not be worth investing into cleaning it all up now.

pkg/image/anaconda_container_installer.go
```diff
@@ -76,11 +76,6 @@ func (img *AnacondaContainerInstaller) InstantiateManifest(m *manifest.Manifest,
 	anacondaPipeline.ExtraPackages = img.ExtraBasePackages.Include
 	anacondaPipeline.ExcludePackages = img.ExtraBasePackages.Exclude
 	anacondaPipeline.ExtraRepos = img.ExtraBasePackages.Repositories
-	anacondaPipeline.InteractiveDefaultsKickstart = &kickstart.Options{
```
bcl, line 79: `InteractiveDefaultsKickstart` isn't used anymore?

achilleas-k: Not for the container installer.  There we always use our own, at `/osbuild.ks`.
We use the interactive defaults for the live and image installer sometimes though.

pkg/image/anaconda_ostree_installer.go
```diff
@@ -96,11 +96,13 @@ func (img *AnacondaOSTreeInstaller) InstantiateManifest(m *manifest.Manifest,
...
 
-	ksPath := osbuild.KickstartPathOSBuild
-	if img.Kickstart != nil && img.Kickstart.Path != "" {
-		ksPath = img.Kickstart.Path
+	if img.Kickstart == nil || img.Kickstart.OSTree == nil {
+		return nil, fmt.Errorf("kickstart options not set for ostree installer")
 	}
-	bootTreePipeline.KernelOpts = []string{fmt.Sprintf("inst.stage2=hd:LABEL=%s", img.ISOLabel), fmt.Sprintf("inst.ks=hd:LABEL=%s:%s", img.ISOLabel, ksPath)}
+	if img.Kickstart.Path == "" {
+		img.Kickstart.Path = osbuild.KickstartPathOSBuild
```
mvo5, line 103: I guess for this change we want to test 
1. That it does not crash without a "kickstart" struct
2. That the default kickstart path is set

I wonder if for (2) we could amend the tests slightly, something like:
```go
diff --git a/pkg/image/installer_image_test.go b/pkg/image/installer_image_test.go
index 7fa844806..67d55e157 100644
--- a/pkg/image/installer_image_test.go
+++ b/pkg/image/installer_image_test.go
@@ -71,7 +71,8 @@ func TestContainerInstallerUnsetKSOptions(t *testing.T) {
                BIOS:       true,
                UEFIVendor: "test",
        }
-       instantiateAndSerialize(t, img, mockPackageSets(), mockContainerSpecs(), nil)
+       ser := instantiateAndSerialize(t, img, mockPackageSets(), mockContainerSpecs(), nil)
+       assert.Contains(t, ser, `"inst.ks=hd:LABEL=:/osbuild.ks"`)
 }
 
 func TestContainerInstallerUnsetKSPath(t *testing.T) {
@@ -137,7 +138,7 @@ func TestTarInstallerUnsetKSPath(t *testing.T) {
        instantiateAndSerialize(t, img, mockPackageSets(), nil, nil)
 }
 
-func instantiateAndSerialize(t *testing.T, img image.ImageKind, packages map[string][]rpmmd.PackageSpec, containers map[string][]container.Spec, commits map[string][]ostree.CommitSpec) {
+func instantiateAndSerialize(t *testing.T, img image.ImageKind, packages map[string][]rpmmd.PackageSpec, containers map[string][]container.Spec, commits map[string][]ostree.CommitSpec) string {
        source := rand.NewSource(int64(0))
        // math/rand is good enough in this case
        /* #nosec G404 */
@@ -147,6 +148,7 @@ func instantiateAndSerialize(t *testing.T, img image.ImageKind, packages map[str
        _, err := img.InstantiateManifest(&mf, nil, &runner.CentOS{Version: 9}, rng)
        assert.NoError(t, err)
 
-       _, err = mf.Serialize(packages, containers, commits, nil)
+       ser, err := mf.Serialize(packages, containers, commits, nil)
        assert.NoError(t, err)
+       return string(ser)
 }
```
It's not super nice to just do this string match but it's probably okay.

But this does raise the question (to me, pardon my ignorance) - will this work when there is no ISOLabel set for img?

pkg/image/installer_image_test.go
```diff
@@ -0,0 +1,152 @@
...
+				URL: "http://localhost:8080/repo",
+			},
+		},
+	}
+}
+
+func TestContainerInstallerUnsetKSOptions(t *testing.T) {
+	img := image.NewAnacondaContainerInstaller(container.SourceSpec{}, "")
+	assert.NotNil(t, img)
+	img.Platform = &platform.X86{
```
mvo5, line 67: (niptick) `img.Platform` is set to the same thing in all the tests, maybe worth extracting to make each individual test slightly smaller:
```go
diff --git a/pkg/image/installer_image_test.go b/pkg/image/installer_image_test.
go
index 7fa844806..98310a5c6 100644
--- a/pkg/image/installer_image_test.go
+++ b/pkg/image/installer_image_test.go
@@ -61,29 +61,26 @@ func mockOSTreeCommitSpecs() map[string][]ostree.CommitSpec 
{
        }
 }
 
+var testPlatform = &platform.X86{
+       BasePlatform: platform.BasePlatform{
+               ImageFormat: platform.FORMAT_ISO,
+       },
+       BIOS:       true,
+       UEFIVendor: "test",
+}
+
 func TestContainerInstallerUnsetKSOptions(t *testing.T) {
        img := image.NewAnacondaContainerInstaller(container.SourceSpec{}, "")
        assert.NotNil(t, img)
-       img.Platform = &platform.X86{
-               BasePlatform: platform.BasePlatform{
-                       ImageFormat: platform.FORMAT_ISO,
-               },
-               BIOS:       true,
-               UEFIVendor: "test",
-       }
-       instantiateAndSerialize(t, img, mockPackageSets(), mockContainerSpecs(),
 nil)
+       img.Platform = testPlatform
+       ser := instantiateAndSerialize(t, img, mockPackageSets(), mockContainerS
pecs(), nil)
+       assert.Contains(t, ser, `"inst.ks=hd:LABEL=:/osbuild.ks"`)
 }
 
 func TestContainerInstallerUnsetKSPath(t *testing.T) {
        img := image.NewAnacondaContainerInstaller(container.SourceSpec{}, "")
        assert.NotNil(t, img)
-       img.Platform = &platform.X86{
-               BasePlatform: platform.BasePlatform{
-                       ImageFormat: platform.FORMAT_ISO,
-               },
-               BIOS:       true,
-               UEFIVendor: "test",
-       }
+       img.Platform = testPlatform
        // set empty kickstart options (no path)
        img.Kickstart = &kickstart.Options{}
 
@@ -93,13 +90,7 @@ func TestContainerInstallerUnsetKSPath(t *testing.T) {
 func TestOSTreeInstallerUnsetKSPath(t *testing.T) {
        img := image.NewAnacondaOSTreeInstaller(ostree.SourceSpec{})
        assert.NotNil(t, img)
-       img.Platform = &platform.X86{
-               BasePlatform: platform.BasePlatform{
-                       ImageFormat: platform.FORMAT_ISO,
-               },
-               BIOS:       true,
-               UEFIVendor: "test",
-       }
+       img.Platform = testPlatform
        img.Kickstart = &kickstart.Options{
                // the ostree options must be non-nil
                OSTree: &kickstart.OSTree{},
@@ -111,33 +102,20 @@ func TestOSTreeInstallerUnsetKSPath(t *testing.T) {
 func TestTarInstallerUnsetKSOptions(t *testing.T) {
        img := image.NewAnacondaTarInstaller()
        assert.NotNil(t, img)
-       img.Platform = &platform.X86{
-               BasePlatform: platform.BasePlatform{
-                       ImageFormat: platform.FORMAT_ISO,
-               },
-               BIOS:       true,
-               UEFIVendor: "test",
-       }
-
+       img.Platform = testPlatform
        instantiateAndSerialize(t, img, mockPackageSets(), nil, nil)
 }
 
 func TestTarInstallerUnsetKSPath(t *testing.T) {
        img := image.NewAnacondaTarInstaller()
        assert.NotNil(t, img)
-       img.Platform = &platform.X86{
-               BasePlatform: platform.BasePlatform{
-                       ImageFormat: platform.FORMAT_ISO,
-               },
-               BIOS:       true,
-               UEFIVendor: "test",
-       }
+       img.Platform = testPlatform
        img.Kickstart = &kickstart.Options{}
 
        instantiateAndSerialize(t, img, mockPackageSets(), nil, nil)
 }

```

mvo5 (approved):
Thank you! This is nice, some small nitpicks/suggestions inline for your consideration

pkg/image/anaconda_ostree_installer.go
achilleas-k: > I guess for this change we want to test
> 
> 1. That it does not crash without a "kickstart" struct
> 2. That the default kickstart path is set
> 

1 is tested in the two test functions named `TestXInstallerUnsetKSOptions()` (the kickstart struct is never set).

> I wonder if for (2) we could amend the tests slightly, something like:
> 
> ```go
> diff --git a/pkg/image/installer_image_test.go b/pkg/image/installer_image_test.go
> index 7fa844806..67d55e157 100644
> --- a/pkg/image/installer_image_test.go
> +++ b/pkg/image/installer_image_test.go
> @@ -71,7 +71,8 @@ func TestContainerInstallerUnsetKSOptions(t *testing.T) {
>                 BIOS:       true,
>                 UEFIVendor: "test",
>         }
> -       instantiateAndSerialize(t, img, mockPackageSets(), mockContainerSpecs(), nil)
> +       ser := instantiateAndSerialize(t, img, mockPackageSets(), mockContainerSpecs(), nil)
> +       assert.Contains(t, ser, `"inst.ks=hd:LABEL=:/osbuild.ks"`)
>  }
>  
>  func TestContainerInstallerUnsetKSPath(t *testing.T) {
> @@ -137,7 +138,7 @@ func TestTarInstallerUnsetKSPath(t *testing.T) {
>         instantiateAndSerialize(t, img, mockPackageSets(), nil, nil)
>  }
>  
> -func instantiateAndSerialize(t *testing.T, img image.ImageKind, packages map[string][]rpmmd.PackageSpec, containers map[string][]container.Spec, commits map[string][]ostree.CommitSpec) {
> +func instantiateAndSerialize(t *testing.T, img image.ImageKind, packages map[string][]rpmmd.PackageSpec, containers map[string][]container.Spec, commits map[string][]ostree.CommitSpec) string {
>         source := rand.NewSource(int64(0))
>         // math/rand is good enough in this case
>         /* #nosec G404 */
> @@ -147,6 +148,7 @@ func instantiateAndSerialize(t *testing.T, img image.ImageKind, packages map[str
>         _, err := img.InstantiateManifest(&mf, nil, &runner.CentOS{Version: 9}, rng)
>         assert.NoError(t, err)
>  
> -       _, err = mf.Serialize(packages, containers, commits, nil)
> +       ser, err := mf.Serialize(packages, containers, commits, nil)
>         assert.NoError(t, err)
> +       return string(ser)
>  }
> ```
> 
> It's not super nice to just do this string match but it's probably okay.
> 
> But this does raise the question (to me, pardon my ignorance) - will this work when there is no ISOLabel set for img?

It "works" in the sense that it creates a manifest, but perhaps it shouldn't because it sets the `isolabel` option in the `grub2.iso` stage to an empty string, which is "valid" only in the strictest sense (i.e. it will pass schema validation, but probably break in the build or boot).

For both the isolabel and the default kickstart path, I keep flip-flopping on whether we should test them separately (and fix them up) vs waiting for the big changes coming with otk integration (https://github.com/osbuild/images/pull/764#issuecomment-2189611407).  Looking around the installer pipelines, I think there's some kickstart path handling that's buggy, so I should at least test for that now.

mvo5 (approved):

achilleas-k:
Need to get https://github.com/osbuild/images/pull/779 in first, then will rebase this.

pkg/image/anaconda_ostree_installer.go
achilleas-k: Handled in follow-up (#781).

PR #774, opened 2024-07-02
cmd/build: tweak and replace `check`/`fail`

mvo5:
Use the more idomiatic `if err != nil { return err }` instead of the custom `check/fail` helper. This is first step to making this a standalone binary that can potentially be run inside a container.


achilleas-k (approved):
ty!

PR #776, opened 2024-07-02
disk: support json unmarshal for Partition

mvo5:
The `disk.Partition` uses interfaces inside the data structure. To support json unmarshal we need a custom `{Unm,M}arshalJSON()` that peeks into the marshalled json and detects what concrete type to use.

This commit provides the `disk.Partition.{Unm,M}arshalJSON` implementation and matching tests.

Another alternative version for https://github.com/osbuild/images/pull/737

Split out from https://github.com/osbuild/images/pull/732 with extra test.

Draft until I find a little bit more time to make the tests a bit more comprehensive and until the other payloads are supported.

### DIFF
--- a/pkg/disk/partition.go
+++ b/pkg/disk/partition.go
@@ -1,6 +1,8 @@
 package disk
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 )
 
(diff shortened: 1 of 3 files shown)


mvo5:
The trouble with this approach is that it's too lose, i.e. when a new Payload gets implemented there is nothing indicating that this partition.go file somewhere else also needs to get updated and writing a test that ensures this is also tricky because there really is no connection (as it's written here - so #775 is probably the superior approach as it's more obvious what needs to happen to add a new type and also there is a test that ensures that things do not get out of sync. I leave it open for a bit but if #775 gets merged this should be closed.

PR #777, opened 2024-07-02
test/data: delete old osbuild-composer test data

achilleas-k:
Remove data left over from the split from osbuild-composer.


ondrejbudai (approved):
:broom: Thanks!

PR #780, opened 2024-07-03
bootupd: Don't require /boot mount point

cgwalters:
I'd like to change bib to just drop the separate `/boot` partition by default because it isn't doing anything useful right now.

A separate `/boot` is only useful with things like LUKS or LVM, but since bib doesn't do those now and it's not in the near term roadmap clearly either, let's just not add the complexity of a separate `/boot` by default.


ondrejbudai (approved):
Makes sense, thanks!

ondrejbudai:
Can you please run `gofmt` on `pkg/osbuild/bootupd_stage.go`? It complains about whitespace.

PR #781, opened 2024-07-03
Fix kickstart path handling in AnacondaInstaller

achilleas-k:
Follow-up to #764.

This PR mainly fixes the handling of the kickstart path in AnacondaInstaller.  The fallback logic for unset kickstart paths was a bit buggy, but we never noticed because we never really set the value to anything else.  Tests have now been added and the buggy logic has been fixed.

I also added validation to the grub2.iso stage options, which ensures that certain ISO values are not unset (Product, Version, ISOLabel).

### DIFF
--- a/pkg/osbuild/grub_iso_stage.go
+++ b/pkg/osbuild/grub_iso_stage.go
@@ -1,5 +1,9 @@
 package osbuild
 
+import "fmt"
+
+const grubisoStageType = "org.osbuild.grub2.iso"
+
 type GrubISOStageOptions struct {
 	Product Product `json:"product"`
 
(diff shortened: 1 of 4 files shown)


pkg/osbuild/grub_iso_stage.go
```diff
@@ -23,8 +51,11 @@ type ISOKernel struct {
 
 // Assemble a file system tree for a bootable ISO
 func NewGrubISOStage(options *GrubISOStageOptions) *Stage {
+	if err := options.validate(); err != nil {
+		panic(err)
```
mvo5, line 55: Is this tested? Might be nice to add a tiny test (but given otk is coming maybe the rational is not to overinvest which is fine with me)

achilleas-k: It's tested in the sense that the tests were failing when it was added and the next commit fixes them :D 
I can add negative testing if you like.

achilleas-k: Added panic assertion tests both at the level of the stage itself (in `pkg/osbuild/`) and all the way up at the image serialization level.

mvo5 (approved):
Thank you

PR #783, opened 2024-07-04
OpenSCAP XML tailoring (HMS-4386)

kingsleyzissou:
If the user supplies an xml tailoring file, we can skip the autotailor
stage - since the autotailor stage just generates the xml file. In this
case we can just set the tailoring config to `nil` and ensure that the
remediation configs have the correct values.

### DIFF
--- a/pkg/customizations/oscap/oscap.go
+++ b/pkg/customizations/oscap/oscap.go
@@ -1,7 +1,11 @@
 package oscap
 
 import (
+	"fmt"
+	"path/filepath"
 	"strings"
+
+	"github.com/osbuild/images/pkg/blueprint"
 )
 
 type Profile string
@@ -55,6 +59,53 @@ type TailoringConfig struct {
 	TailoredProfileID string
 	Selected          []string
 	Unselected        []string
+}
+
+func NewConfigs(oscapConfig blueprint.OpenSCAPCustomization, defaultDatastream *string) (*RemediationConfig, *TailoringConfig, error) {
+	var datastream = oscapConfig.DataStream
+	if datastream == "" {
+		if defaultDatastream == nil {
+			return nil, nil, fmt.Errorf("No OSCAP datastream specified and the distro does not have any default set")
+		}
+		datastream = *defaultDatastream
+	}
+
+	remediationConfig := &RemediationConfig{
+		Datastream:         datastream,
+		ProfileID:          oscapConfig.ProfileID,
+		CompressionEnabled: true,
+	}
+
+	tc := oscapConfig.Tailoring
+	if tc == nil {
+		return remediationConfig, nil, nil
+	}
+
+	tailoringPath := filepath.Join(DataDir, "tailoring.xml")
+	tailoredProfileID := fmt.Sprintf("%s_osbuild_tailoring", remediationConfig.ProfileID)
+	if tc.ProfileID != nil && *tc.ProfileID != "" {
+		tailoredProfileID = *tc.ProfileID
+	}
+
+	tailoringConfig := &TailoringConfig{
+		RemediationConfig: RemediationConfig{
+			ProfileID:     remediationConfig.ProfileID,
+			TailoringPath: tailoringPath,
+			Datastream:    datastream,
+		},
+		TailoredProfileID: tailoredProfileID,
+		Selected:          tc.Selected,
+		Unselected:        tc.Unselected,
+	}
+
+	// the reason for changing the remediation config profile
+	// after we create the tailoring configs is that the tailoring
+	// config needs to know about the original base profile id, but
+	// the remediation config needs to know the updated profile id.
+	remediationConfig.ProfileID = tailoredProfileID
+	remediationConfig.TailoringPath = tailoringPath
+
+	return remediationConfig, tailoringConfig, nil
 }
 
 func DefaultFedoraDatastream() string {
--- a/pkg/customizations/oscap/oscap.go
+++ b/pkg/customizations/oscap/oscap.go
@@ -76,6 +76,28 @@ func NewConfigs(oscapConfig blueprint.OpenSCAPCustomization, defaultDatastream *
 		CompressionEnabled: true,
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/customizations/oscap/oscap.go
```diff
@@ -66,6 +66,18 @@ func NewTailoringConfig(tc *blueprint.OpenSCAPTailoringCustomizations, r *Remedi
 		return nil, nil
 	}
 
+	if tc.XMLFilepath != nil {
```
achilleas-k, line 69: Should we also check that the `*tc.XMLFilepath` isn't `""`, or is that not possible?

pkg/distro/fedora/images.go
```diff
@@ -204,19 +203,9 @@ func osCustomizations(
...
-				RemediationConfig: remediationConfig,
-				TailoredProfileID: fmt.Sprintf("%s_osbuild_tailoring", oscapConfig.ProfileID),
-				Selected:          oscapConfig.Tailoring.Selected,
-				Unselected:        oscapConfig.Tailoring.Unselected,
-			}
-			// we need to set this after the tailoring config
-			// since the tailoring config needs to know about both
-			// the base profile id and the tailored profile id
-			remediationConfig.ProfileID = tailoringConfig.TailoredProfileID
+		tailoringConfig, err := oscap.NewTailoringConfig(oscapConfig.Tailoring, &remediationConfig)
```
achilleas-k, line 206: Does it make sense to expand the helper function to include the things above this line and have it generate the `remediationConfig` from scratch?  Comparing the code in Fedora and RHEL, they seem identical (but I haven't looked very closely).

I mean putting
```go
		var datastream = oscapConfig.DataStream
		if datastream == "" {
			if imageConfig.DefaultOSCAPDatastream == nil {
				return manifest.OSCustomizations{}, fmt.Errorf("No OSCAP datastream specified and the distro does not have any default set")
			}
			datastream = *imageConfig.DefaultOSCAPDatastream
		}
		remediationConfig := oscap.RemediationConfig{
			Datastream:         datastream,
			ProfileID:          oscapConfig.ProfileID,
			CompressionEnabled: true,
		}
```
into the helper function (and renaming it), and then
```go
tailoringConfig, remediationConfig, err := oscap.NewConfigs(oscapConfig, imageConfig.DefaultOSCAPDatastream)
```

Or maybe just moving the initial creation of the `remediationConfig` in the helper and having:
```go
tailoringConfig, remediationConfig, err := oscap.NewConfigs(oscapConfig, datastream)
```
whichever makes more sense.

pkg/customizations/oscap/oscap.go
kingsleyzissou: Yeah flip sorry I missed that. Good catch

pkg/distro/fedora/images.go
kingsleyzissou: Yeah solid shout. Kinda felt weird only doing this for tailoring and not the remediation configs. Will sort that out

kingsleyzissou:
Was trying to figure out why the tests were failing and it's because I'm not setting the datastream in the tailoring configs 🤦‍♂️

achilleas-k (approved):
Love the cleanup.  LGTM!

thozza (commented):
Nice work.

I have only one comment. The code does not handle the case when the caller (user) provides both, the XML file and tailoring rules (selected / unselected) for the autotailor. AFAICT, the explicit tailoring rules will be silently ignored, right? I guess that at least an error would probably make sense.

kingsleyzissou:
> Nice work.
> 
> 
> 
> I have only one comment. The code does not handle the case when the caller (user) provides both, the XML file and tailoring rules (selected / unselected) for the autotailor. AFAICT, the explicit tailoring rules will be silently ignored, right? I guess that at least an error would probably make sense.

Yeah that's a great catch, they will be silently ignored. Erroring out would make more sense

kingsleyzissou:
> I have only one comment. The code does not handle the case when the caller (user) provides both, the XML file and tailoring rules (selected / unselected) for the autotailor. AFAICT, the explicit tailoring rules will be silently ignored, right? I guess that at least an error would probably make sense.

I've made the changes, I wonder if we should have a separate struct under OpenSCAP customizations for XML tailoring?
i.e:
```go
type OpenSCAPCustomization struct {
        DataStream   string                              `json:"datastream,omitempty" toml:"datastream,omitempty"`
        ProfileID    string                              `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
        Tailoring    *OpenSCAPTailoringCustomizations    `json:"tailoring,omitempty" toml:"tailoring,omitempty"`
        XMLTailoring *OpenSCAPXMLTailoringCustomizations `json:"xml_tailoring,omitempty" toml:"xml_tailoring,omitempty"`
}
```

thozza:
> I've made the changes, I wonder if we should have a separate struct under OpenSCAP customizations for XML tailoring? i.e:
> 
> ```go
> type OpenSCAPCustomization struct {
>         DataStream   string                              `json:"datastream,omitempty" toml:"datastream,omitempty"`
>         ProfileID    string                              `json:"profile_id,omitempty" toml:"profile_id,omitempty"`
>         Tailoring    *OpenSCAPTailoringCustomizations    `json:"tailoring,omitempty" toml:"tailoring,omitempty"`
>         XMLTailoring *OpenSCAPXMLTailoringCustomizations `json:"xml_tailoring,omitempty" toml:"xml_tailoring,omitempty"`
> }
> ```

Yeah, this LGTM.

I mean, unless it would make sense to actually merge the two ways if provided by the user - take XML file and extend it with the specified tailoring rules. However, that is probably a feature that is not needed, so let's not complicate our (to be precise, "your") lives :sweat_smile:

kingsleyzissou:
> take XML file and extend it with the specified tailoring rules

Ya let's not do that 😅 

I'll make those changes just to add some distinction between the two tailoring types then

thozza (approved):
LGTM :+1:

PR #785, opened 2024-07-05
Schutzfile: move to latest osbuild with curl --parallel support

mvo5:
This commit updates osbuild so that we benefit from osbuild PR#1573 to enable curl --parallel. This should speed up downloading especially on fast connections as less time is spend with the connection setup/handshake etc.

Similar to https://github.com/osbuild/osbuild-composer/pull/4247

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -24,43 +24,7 @@
       "osbuild": {
         "commit": "4697a3fb840a337438bd31b81f78cb8c5bfb14bc"
       }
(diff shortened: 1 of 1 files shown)


achilleas-k (dismissed):

achilleas-k (approved):

mvo5:
This is not ready, see https://github.com/osbuild/osbuild/pull/1811

PR #787, opened 2024-07-08
cmd: add `otk-make-fstab-stage` helper

mvo5:
This helper is similar to the `otk-make-partition-stages` but for
the `fstab` generation. It consumes the variables generated by
`otk-gen-partition-table` and produces a fragment to generate
the `org.osbuild.fstab` stage.

The expected usage is something like:
```yaml
otk.define:
 filesystem:
   otk.external.gen_partition_table:
     modifications: ${modifications.filesystem}
     properties: ...
     partitions: ...

stages:
 ...
 otk.external.make_fstab_stage:
   internal: ${filesystem.internal}
```

Build on top of https://github.com/osbuild/images/pull/732 only the last commit is relevant.

### DIFF
--- a/cmd/otk-make-fstab-stage/main.go
+++ b/cmd/otk-make-fstab-stage/main.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/osbuild/images/pkg/osbuild"
+
+	"github.com/osbuild/images/internal/otkdisk"
+)
+
+type Input = otkdisk.Data
+
+func makeFstabStages(inp Input) (stages *osbuild.Stage, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("cannot generate image prepare stages: %v", r)
+		}
+	}()
+
+	pt := inp.Const.Internal.PartitionTable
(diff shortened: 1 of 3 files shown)


ondrejbudai (commented):
fwiw this looks good :)

cmd/otk-make-fstab-stage/main.go
```diff
@@ -0,0 +1,52 @@
...
+
+func makeFstabStages(inp Input) (stages *osbuild.Stage, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("cannot generate image prepare stages: %v", r)
+		}
+	}()
+
+	pt := inp.Const.Internal.PartitionTable
+	fstabStage := osbuild.NewFSTabStage(osbuild.NewFSTabStageOptions(pt))
```
achilleas-k, line 24: Let's make a note / issue to remove panics from these functions and have them return errors.  I know we've been saying it for a while now, but now that we're using them for these externals, I think we can at least prioritise the ones we want.

achilleas-k (approved):
LGTM.  Love how simple this turned out.

cmd/otk-make-fstab-stage/main.go
mvo5: Sure! i create a reminder for myself and will get to that in a followup.

PR #793, opened 2024-07-12
New github action that regularly updates osbuild commit ID

achilleas-k:
New job that runs weekly (or on-demand) that updates the osbuild commit ID in the Schutzfile to the latest main commit.  This is used to install the osbuild dependency in gitlab CI to build images.  While we often update this manually when needed, this job will make sure we don't miss any updates when no new features required us to update manually.

The job runs on Sundays at noon (UTC).  When the osbuild commit ID changes, our gitlab CI jobs always rebuild every test image configuration, which can take up to three hours so it is good to do it on a Sunday, when there are much fewer CI jobs running.

### DIFF
--- a/.github/workflows/update-osbuild.yml
+++ b/.github/workflows/update-osbuild.yml
@@ -1,0 +1,51 @@
+# This action updates the osbuild ref in the Schutzfile
+---
+name: "Update osbuild commit ID"
+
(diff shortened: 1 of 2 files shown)


thozza (approved):
Good idea, thanks!

PR #796, opened 2024-07-15
distro/rhel9: match azure boot partition with other cloud images

ondrejbudai:
I failed to bump the size of the boot partition for Azure images in
4aff47dcbed000fded0486061be70bb1dfb14fb2.

This commit fixes that.


achilleas-k (approved):
Thank you!

PR #798, opened 2024-07-16
Add OpenSCAP json tailoring (HMS-3826)

kingsleyzissou:
This PR adds the ability to consume an JSON tailoring file. 
- create autotailor stage options where the input is a json tailoring file
- the osbuild stage generates an xml tailoring file which is consumed by the remediation stage.

Depends on: https://github.com/osbuild/osbuild/pull/1794


kingsleyzissou:
For context, Sanne and I had a meeting with compliance and the OpenSCAP team. They are trying to hide and abstract away the `xml` tailoring completely for future versions of RHEL and Fedora. So we decided we do actually need the `json` tailoring.

pkg/customizations/oscap/oscap.go
```diff
@@ -98,33 +107,46 @@ func NewConfigs(oscapConfig blueprint.OpenSCAPCustomization, defaultDatastream *
...
+
+		tailoringConfigs.TailoredProfileID = jsonConfigs.ProfileID
+		tailoringConfigs.JSONFilepath = jsonConfigs.Filepath
+		remediationConfig.ProfileID = tailoringConfigs.TailoredProfileID
+
+		return remediationConfig, tailoringConfigs, nil
+	}
+
+	tailoringConfigs.Selected = tc.Selected
+	tailoringConfigs.Unselected = tc.Unselected
```
thozza, line 140: If I'm reading the code correctly, `tc` could be `nil` at this point, no?

pkg/osbuild/oscap_autotailor_stage.go
```diff
@@ -11,29 +11,57 @@ type OscapAutotailorStageOptions struct {
...
-		return fmt.Errorf("'profile_id' must be specified")
+		return fmt.Errorf("'profile_id' tailoring_file' must be specified")
+	}
+	return nil
+}
+
+type AutotailorJSONConfig struct {
+	TailoredProfileID string `json:"tailored_profile_id,omitempty"`
+	Datastream        string `json:"datastream"`
+	TailoringFile     string `json:"tailoring_file,omitempty"`
```
thozza, line 45: All of these are required by the stage schema, so shouldn't be really `omitempty`.

pkg/osbuild/oscap_autotailor_stage.go
```diff
@@ -11,29 +11,57 @@ type OscapAutotailorStageOptions struct {
...
+func (c AutotailorKeyValueConfig) isAutotailorConfig() {}
 
-func (c OscapAutotailorConfig) validate() error {
+func (c AutotailorKeyValueConfig) validate() error {
 	if c.Datastream == "" {
 		return fmt.Errorf("'datastream' must be specified")
 	}
+	if c.NewProfile == "" {
+		return fmt.Errorf("'new_profile' must be specified")
+	}
```
thozza, line 35: The stage schema does not require it to be specified AFAICT and the struct definitions also marks it as `omitempty`... :thinking:

pkg/osbuild/oscap_autotailor_stage.go
```diff
@@ -11,29 +11,57 @@ type OscapAutotailorStageOptions struct {
...
 	if c.ProfileID == "" {
-		return fmt.Errorf("'profile_id' must be specified")
+		return fmt.Errorf("'profile_id' tailoring_file' must be specified")
```
thozza, line 37: There's a typo in the error message `'profile_id' tailoring_file' ...`

thozza (commented):
Overall this looks good. I've added a few comments... :wink:

pkg/customizations/oscap/oscap.go
achilleas-k: I think no, but your question (and my own thoughts reading this function) make me realise this could use some refactoring.  I think a switch with separate functions for each case (xml vs json) would make things easier to read.

(var names shortened for brevity)
```go
switch {
case xml != nil && json != nil, xml != nil && tc != nil, tc != nil && json != nil:
	// return error multiple configs
case xml != nil:
	return xmlConfigs(xml, remediationConfigs)
case json != nil:
	return jsonConfigs(json, remediationConfigs)
default:
	// do the other stuff
}
```

pkg/osbuild/oscap_autotailor_stage.go
kingsleyzissou: > stage schema does not require it

I think this is a mistake and I'll need to fix it in `osbuild`. It needs to be one of `new_profile` or `tailored_profile_id`

kingsleyzissou: Here's a PR to fix that: https://github.com/osbuild/osbuild/pull/1823

thozza: You should update the osbuild ref in `Schutzfile` once the PR is merged, just to be sure.

croissanne (approved):
lgtm! ty

thozza (dismissed):
Nice work, thanks!

thozza (approved):

PR #799, opened 2024-07-18
Add Anaconda Module customizations to Blueprint (RHEL-49499)

achilleas-k:
This PR generalises the solution to the problem we had in BIB (osbuild/bootc-image-builder#547) 
 where Anaconda Modules were not enabled for custom kickstarts, making it impossible to use certain kickstart features.  For example, in most Anaconda ISO builds, we only enable the Users module if a user is specified in the blueprint.  A custom kickstart with a user creation command would therefore have no effect, since the Users module would be disabled.  The same is true for many other kickstart commands.

However, enabling certain modules unconditionally is also not desirable.  For example, in the case of ostree payloads, enabling the user module can prevent automated/unattended installation, since Anaconda will block and prompt for an admin user to be created.  Users may not want to create a user account during installation if one is already included in the OS payload (ostree commit or container).

This PR makes it possible to specify a list of Anaconda modules to enable or explicitly disable using blueprint customizations.

### DIFF
--- a/pkg/distro/rhel/images.go
+++ b/pkg/distro/rhel/images.go
@@ -7,6 +7,7 @@ import (
 	"github.com/osbuild/images/internal/workload"
 	"github.com/osbuild/images/pkg/blueprint"
 	"github.com/osbuild/images/pkg/container"
+	"github.com/osbuild/images/pkg/customizations/anaconda"
 	"github.com/osbuild/images/pkg/customizations/fdo"
 	"github.com/osbuild/images/pkg/customizations/fsnode"
 	"github.com/osbuild/images/pkg/customizations/ignition"
@@ -482,7 +483,7 @@ func EdgeInstallerImage(workload workload.Workload,
 
 	if len(img.Kickstart.Users)+len(img.Kickstart.Groups) > 0 {
 		// only enable the users module if needed
-		img.AdditionalAnacondaModules = []string{"org.fedoraproject.Anaconda.Modules.Users"}
+		img.AdditionalAnacondaModules = []string{anaconda.ModuleUsers}
 	}
 
 	img.ISOLabel, err = t.ISOLabel()
@@ -662,7 +663,7 @@ func ImageInstallerImage(workload workload.Workload,
 		img.AdditionalDrivers = installerConfig.AdditionalDrivers
 	}
 
-	img.AdditionalAnacondaModules = []string{"org.fedoraproject.Anaconda.Modules.Users"}
+	img.AdditionalAnacondaModules = []string{anaconda.ModuleUsers}
 
 	img.SquashfsCompression = "xz"
 
--- a/pkg/distro/rhel/images.go
+++ b/pkg/distro/rhel/images.go
@@ -481,9 +481,18 @@ func EdgeInstallerImage(workload workload.Workload,
 		img.AdditionalDrivers = installerConfig.AdditionalDrivers
 	}
 
+	instCust, err := customizations.GetInstaller()
+	if err != nil {
+		return nil, err
+	}
+	if instCust != nil && instCust.Modules != nil {
+		img.AdditionalAnacondaModules = append(img.AdditionalAnacondaModules, instCust.Modules.Enable...)
+		img.DisabledAnacondaModules = append(img.DisabledAnacondaModules, instCust.Modules.Disable...)
+	}
+
 	if len(img.Kickstart.Users)+len(img.Kickstart.Groups) > 0 {
 		// only enable the users module if needed
-		img.AdditionalAnacondaModules = []string{anaconda.ModuleUsers}
+		img.AdditionalAnacondaModules = append(img.AdditionalAnacondaModules, anaconda.ModuleUsers)
 	}
 
 	img.ISOLabel, err = t.ISOLabel()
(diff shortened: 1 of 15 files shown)


pkg/osbuild/anaconda_stage.go
```diff
@@ -30,10 +30,13 @@ func defaultModuleStates() map[string]bool {
 	}
 }
 
-func enableModules(states map[string]bool, additional []string) {
-	for _, modname := range additional {
+func setModuleStates(states map[string]bool, enable, disable []string) {
```
thozza, line 38: Should we check that `enable` and `disable` are disjoint?

pkg/image/installer_image_test.go
```diff
@@ -290,3 +290,145 @@ func TestTarInstallerPanics(t *testing.T) {
...
+			assert.NotNil(t, img)
+			img.Platform = testPlatform
+			mfs := instantiateAndSerialize(t, img, mockPackageSets(), nil, mockOSTreeCommitSpecs())
+			modules := findAnacondaStageModules(t, manifest.OSBuildManifest(mfs))
+			assert.NotNil(t, modules)
+			assert.ElementsMatch(t, modules, tc.expected)
+		})
+	}
+}
+func TestTarInstallerModules(t *testing.T) {
```
thozza, line 415: Nitpick: missing newline.

thozza (commented):
Sorry it took me so long to review. :innocent:

The PR looks great and it was a pleasure to review. I've found just one missing newline and added one question.

pkg/osbuild/anaconda_stage.go
achilleas-k: It crossed my mind and I'm not sure.  I thought about this a bit and landed on allowing a module to appear in both lists, making it possible to disable a module that's in the enabled list (the tests even cover this case explicitly).  My reasoning is that in image type defaults, we only ever **enable** modules.  The **disable** list comes straight from the user blueprint.  So by letting a module appear in both lists, we essentially give the user the ability to **disable** a module that is **enabled** by default.

For example, if a user is building an image which **enables** the Users module by default and they want to **disable** it for some reason (automated installation, maybe they want to create a user with initial-setup, etc), or if they want to **disable** the Storage module (which we always enable) to do partitioning in a kickstart `%pre` without having Anaconda block an automated installation, this gives them that control.

pkg/image/installer_image_test.go
achilleas-k: Fixed.

pkg/osbuild/anaconda_stage.go
achilleas-k: One thing we can do to make things a bit saner is to raise an error when the user adds the same module to both lists in the blueprint itself.  That would make things less ambiguous for the user (they can't enable and disable the same module in the same build) while also letting them disable a default module.

thozza: So, I've added the question while going through the beginning of the patch set and ended up with pretty much the same conclusion as you, by the end of it. But I thought it would make sense to have this discussion.

WRT the BP customization. I think that either what you suggested would be good, or explicitly documenting the behavior of it (that disabling a module has precedence).

thozza (dismissed):

pkg/distro/rhel/images.go
```diff
@@ -658,7 +668,15 @@ func ImageInstallerImage(workload workload.Workload,
...
-	img.AdditionalAnacondaModules = []string{"org.fedoraproject.Anaconda.Modules.Users"}
+	instCust, err := customizations.GetInstaller()
+	if err != nil {
+		return nil, err
+	}
+	if instCust != nil && instCust.Modules != nil {
+		img.AdditionalAnacondaModules = append(img.AdditionalAnacondaModules, instCust.Modules.Enable...)
+		img.DisabledAnacondaModules = append(img.DisabledAnacondaModules, instCust.Modules.Disable...)
+	}
+	img.AdditionalAnacondaModules = append(img.AdditionalAnacondaModules, anaconda.ModuleUsers)
```
supakeen, line 679: I'm a bit confused that we add the users module unconditionally for the `ImageInstaller` here; but conditionally for the others (only if kickstart user customizations are enabled). Why is that?

As far as I know one needs to add a user for both image types in Anaconda.

achilleas-k: The image installer's payload never contains a user account, so we enable it to force interactive user creation during installation, otherwise the user will end up with no way to log in to the machine.  This decision was made to accommodate the IB service which doesn't support user creation (in the UI at least) but does have support for building image installer ISOs.

For other image types (ostree-based installers), the user module is only enabled by default when the blueprint contains a user so that it can be created.  It's not required however to create a user for such payloads because it is possible that a user is already created in the payload itself (the ostree commit).  So if a user has an ostree commit with a user account in it, wants to create an ISO to deploy it, and also wants to make that ISO fully unattended (without creating a user at install time), the user module will have to be disabled for that to be possible.

I'm open to discussing these decisions and reevaluating, especially now that users will have control of the modules that are enabled.

supakeen: I think we can keep it as is as we always build the tar that gets deployed by image installer as well.

supakeen (approved):

pkg/osbuild/anaconda_stage.go
achilleas-k: Prepared docs PR: https://github.com/osbuild/osbuild.github.io/pull/100

achilleas-k:
Rebased to fix conflicts.

thozza (approved):

PR #807, opened 2024-07-24
Test utility function for validating Partition Table sizes

achilleas-k:
There is currently a bug that occurs when converting a partition table to LVM with a root partition size that is larger than the original partition size (or larger than the partition table).  @croissanne discovered it and we investigated it together yesterday.  We both have ideas for how to fix it properly, but first I decided to write some tests to catch similar issues.

To keep things easier to follow and review, I decided to first introduce a utility function for validating partition table sizes.  Currently this is a test utility function.  If we discover that we need it outside of testing, we can move it to the `partition_table.go` file.

### DIFF
--- a/pkg/disk/partition_table_internal_test.go
+++ b/pkg/disk/partition_table_internal_test.go
@@ -24,3 +24,49 @@ func TestPartitionTableFeatures(t *testing.T) {
 
 	}
 }
+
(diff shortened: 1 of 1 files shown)


thozza (approved):
LGTM :bug:

croissanne (approved):

PR #813, opened 2024-07-29
Add BP customization for importing RPM GPG keys from the image tree (COMPOSER-2308)

thozza:
As part of https://issues.redhat.com/browse/COMPOSER-2308, I'm moving RHUI client RPMs installation from the image definition to customizations. However, Azure RHUI images import by default RPM GPG keys from the image tree, which are installed by the RHUI client rpm. Which means that the image definition technically still has hard dependency on the RHUI client rpm. Therefore it makes sense to also move the RPM GPG key importing from tree also to the customizations.

Add a BP customization for specifying paths to RPM GPG keys installed in the image, which will be imported to the RPM database during image build.

### DIFF
--- a/pkg/blueprint/customizations.go
+++ b/pkg/blueprint/customizations.go
@@ -27,6 +27,7 @@ type Customizations struct {
 	FIPS               *bool                          `json:"fips,omitempty" toml:"fips,omitempty"`
 	ContainersStorage  *ContainerStorageCustomization `json:"containers-storage,omitempty" toml:"containers-storage,omitempty"`
 	Installer          *InstallerCustomization        `json:"installer,omitempty" toml:"installer,omitempty"`
+	RPM                *RPMCustomization              `json:"rpm,omitempty" toml:"rpm,omitempty"`
 }
 
 type IgnitionCustomization struct {
@@ -411,4 +412,11 @@ func (c *Customizations) GetInstaller() (*InstallerCustomization, error) {
 	}
 
 	return c.Installer, nil
+}
+
+func (c *Customizations) GetRPM() *RPMCustomization {
+	if c == nil {
+		return nil
+	}
+	return c.RPM
 }
--- a/pkg/blueprint/customizations_test.go
+++ b/pkg/blueprint/customizations_test.go
@@ -376,3 +376,20 @@ func TestGetOpenSCAPConfig(t *testing.T) {
 
 	assert.EqualValues(t, expectedOscap, *retOpenSCAPCustomiztions)
 }
+
+func TestGetImportRPMGPGKey(t *testing.T) {
+	expectedRPM := RPMCustomization{
+		ImportKeys: &RPMImportKeys{
+			[]string{
+				"/etc/pki/rpm-gpg/RPM-GPG-KEY",
+			},
+		},
+	}
+
+	TestCustomizations := Customizations{
+		RPM: &expectedRPM,
+	}
+
+	retRPM := TestCustomizations.GetRPM()
+	assert.EqualValues(t, expectedRPM, *retRPM)
+}
--- a/pkg/blueprint/rpm_customizations.go
+++ b/pkg/blueprint/rpm_customizations.go
@@ -1,0 +1,10 @@
+package blueprint
+
+type RPMImportKeys struct {
+	// File paths in the image to import keys from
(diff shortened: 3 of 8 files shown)


pkg/blueprint/customizations.go
```diff
@@ -27,6 +27,11 @@ type Customizations struct {
 	FIPS               *bool                          `json:"fips,omitempty" toml:"fips,omitempty"`
 	ContainersStorage  *ContainerStorageCustomization `json:"containers-storage,omitempty" toml:"containers-storage,omitempty"`
 	Installer          *InstallerCustomization        `json:"installer,omitempty" toml:"installer,omitempty"`
+	ImportRPMGPGKey    *ImportRPMGPGKeyCustomization  `json:"import_rpm_gpg_key,omitempty" toml:"import-rpm-gpg-key,omitempty"`
```
thozza, line 30: Not the best name, I'm open to suggestions :innocent:

thozza: I'm more concerned about the TOML / JSON property name. The struct is hopefully fine. Nevertheless, I appreciate the suggestion.

kingsleyzissou: > I'm more concerned about the TOML / JSON property name

Is there a reason to have one property in snake case and the other in kebab case?

Slightly further up we have both in snake case for `installation_device`:
https://github.com/osbuild/images/blob/eff7b17b216ee87b85a52ce3820a6829c92873d9/pkg/blueprint/customizations.go#L20

achilleas-k: I'd recommend putting it one level down into a category.  I don't have a great idea for this, but I reflexively don't like adding top-level customizations for something this specific.

Now, that said, I don't have a good idea what the category should be.  I wonder if it makes sense to have an `rpm` category.  Is it possible we might need rpm configs in the future?
So given all that, I'm thinking maybe the customization could be, in json:
```json
{
  "rpm": {
    "import_keys": {
      "keys": [
        "actual key strings"
      ],
      "files": [
        "paths to files in the tree"
      ]
    }
  }
}
```
which would be a bit more tidy, enables pasting keys straight into the image (which might be convenient), and leaves room for more rpm-related configurations.

It might have been nice to have a package-related customization block which would also include the Repository customizations.

achilleas-k: A more extreme version:
```json
{
  "rpm": {
    "gpg": {
      "import": {
        "keys": [],
        "paths": []
      }
    }
  }
}
```

thozza: > I'd recommend putting it one level down into a category. I don't have a great idea for this, but I reflexively don't like adding top-level customizations for something this specific.
> 
> Now, that said, I don't have a good idea what the category should be. I wonder if it makes sense to have an `rpm` category. Is it possible we might need rpm configs in the future? So given all that, I'm thinking maybe the customization could be, in json:
> 
> ```json
> {
>   "rpm": {
>     "import_keys": {
>       "keys": [
>         "actual key strings"
>       ],
>       "files": [
>         "paths to files in the tree"
>       ]
>     }
>   }
> }
> ```
> 
> which would be a bit more tidy, enables pasting keys straight into the image (which might be convenient), and leaves room for more rpm-related configurations.

Makes sense...

I've already defined the struct in a way, so that pasting literal key strings would be possible (there is `from_tree` and there could be something else for strings). rpm-related config is not something I foresee us adding, but it won't hurt to keep the options open :wink: 

> It might have been nice to have a package-related customization block which would also include the Repository customizations, but that ship has sailed.

Agreed, I had a similar idea, but adding it to package customization, which is not really a customization but a top level property of a BP... :confused:

thozza: I went with the first "less extreme" version :sweat_smile:

kingsleyzissou: > It might have been nice to have a package-related customization block which would also include the Repository customizations, but that ship has sailed.

Definitely would have been nice, oh well 😅

kingsleyzissou (approved):
Nice!

achilleas-k (approved):
Nice.  LGTM!

pkg/blueprint/customizations.go
achilleas-k: > I went with the first "less extreme" version 😅

Makes sense :)

PR #816, opened 2024-07-29
[RFC] distro: drop hardcoded partiton table UUIDs

mvo5:
The current distro partition tables contain hardcoded UUIDs for various elements like the partition table, the partition and the filesystem UUIDs. Hardcoding these seems slightly unusual, given that a UUID is supposed to be (practically) unique and that the images library will DTRT if no UUID is set.

This commit removes the hardcoded PartitionTable UUID as a starting point to spark a discussion. I have only a limited understanding of the history of this and while I did do a bit of git archeology I could not find a commit with a rational why the UUIDs are hardcoded [0] so I'm asking for a review from some of the people who are around for a long time. I would love to understand if there is a reason to have these hardcoded UUIDs and i'm happy to do a PR that adds comments etc that will explain this reason. If however we do not need/want the hardcoded UUIDs i will continue and remove them from the partition and filesystems too. If we go down this path I'm sure this is a risky change because some people probably rely by now on the fact that the UUID is actually quite predictable (also https://xkcd.com/1172) and because this behavior is here since day 1 (or so it seems). Also @achilleas-k mentioned "Chestertons fence" to me so I'm trying to be extra cautious. Help is greatly appreciated to help me understand this better!

C.f. https://github.com/osbuild/bootc-image-builder/pull/568

[0] I found 
https://github.com/osbuild/osbuild-composer/commit/22254637f810f1cabc95fb616b46d92049135e99#diff-de240f0bca8f03daafeb653ff429aafa12f43b5e0dc8ba327c56108a6ff864f3R115 and https://github.com/osbuild/osbuild-composer/commit/0dd17ae3f707ab9e89f4a289ed461ee0d28fd203

### DIFF
--- a/pkg/distro/fedora/partition_tables.go
+++ b/pkg/distro/fedora/partition_tables.go
@@ -9,7 +9,6 @@ import (
 
 var defaultBasePartitionTables = distro.BasePartitionTableMap{
 	arch.ARCH_X86_64.String(): disk.PartitionTable{
-		UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
 		Type: "gpt",
 		Partitions: []disk.Partition{
 			{
@@ -61,7 +60,6 @@ var defaultBasePartitionTables = distro.BasePartitionTableMap{
 		},
 	},
 	arch.ARCH_AARCH64.String(): disk.PartitionTable{
-		UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
 		Type: "gpt",
 		Partitions: []disk.Partition{
 			{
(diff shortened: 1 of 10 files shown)


cgwalters:
It may make sense to hoist them all to a shared `const mysteriousUuid = ...` first with an explanatory comment?

>  the images library will DTRT if no UUID is set.

What is right thing here? All zeros so tools like systemd-repart will know to init on firstboot?

achilleas-k:
Let's split this PR into a RHEL 10 change (and merge it) and keep the other ones as draft until we finish evaluating the impact.

achilleas-k:
> All zeros so tools like systemd-repart will know to init on firstboot?

Does repart do this automatically or does it need to be configured with drop-ins?  Afaik, it would need drop-in configs, which we don't write, but maybe something to consider for the future.

achilleas-k:
> my gut feeling is that the UUIDs are simply the values that were used by the original images built by imagefactory and used as the base for our own image definitions

This sounds extremely plausible.

mvo5:
Thanks all for the input! Closing in favor of https://github.com/osbuild/images/pull/823

PR #818, opened 2024-07-30
rhel10: drop `net.ifnames=0` from rhel10/centos10

mvo5:
Drop net.ifnames=0 from RHEL 10 and Centos10 disk images and add test.

Fixes: COMPOSER-2289
Fixes: COMPOSER-2323

### DIFF
--- a/pkg/distro/rhel/rhel10/ami.go
+++ b/pkg/distro/rhel/rhel10/ami.go
@@ -9,7 +9,7 @@ import (
 )
 
 // TODO: move these to the EC2 environment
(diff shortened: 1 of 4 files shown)


achilleas-k (approved):
LGTM.  Ship it!

ondrejbudai (approved):
Thanks!

thozza:
@mvo5 you'll need to also somehow fix https://github.com/osbuild/images/blob/6fac856649000f132e172c3f52ced0467c6fee9b/pkg/distro/rhel/rhel10/ami.go#L63-L71, because the if name won't be `eth0` any more...

PR #822, opened 2024-07-31
rhel10: drop `ifcfg-eth0` sysconfig network-scripts file

mvo5:
The AMI for rhel9 and below writes a `ifcfg-eth0` sysconfig file that enables dhcp and disables ipv6. This commit drops this file because:
1. ipcfg-rh is deprecated
2. NetworkManager defaults to dhcp by default for wired ethernet (unless configured otherwise)
3. It seems we do not need to disable ipv6 in 2024 anymore

With the exception of (3) I'm pretty confident in this commit. Feedback welcome (especially on (3)). Fwiw, I did testbuild/ran the centos10 ami and it booted fine and network is available.

Thanks to @thozza and @ondrejbudai

### DIFF
--- a/pkg/distro/rhel/rhel10/ami.go
+++ b/pkg/distro/rhel/rhel10/ami.go
@@ -58,19 +58,6 @@ func baseEc2ImageConfig() *distro.ImageConfig {
 					Networking: true,
 					NoZeroConf: true,
 				},
-				NetworkScripts: &osbuild.NetworkScriptsOptions{
-					IfcfgFiles: map[string]osbuild.IfcfgFile{
-						"eth0": {
-							Device:    "eth0",
-							Bootproto: osbuild.IfcfgBootprotoDHCP,
-							OnBoot:    common.ToPtr(true),
-							Type:      osbuild.IfcfgTypeEthernet,
-							UserCtl:   common.ToPtr(true),
-							PeerDNS:   common.ToPtr(true),
-							IPv6Init:  common.ToPtr(false),
-						},
-					},
-				},
 			},
 		},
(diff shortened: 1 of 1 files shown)


mvo5:
> I'm OK with IPv6, we need to embrace new technologies 😅
> 
> As for the DHCP, the only thing that comes to my mind is the [`NetworkManager-server-config` package, which is by default part of the `@Server` package group](https://gitlab.com/redhat/centos-stream/release-engineering/comps/-/blob/c10s/comps-centos-stream-10.xml.in?ref_type=heads#L949). AFAIK, it disables DHCP by defailt. But I'm not sure if the `ifcfg` file would override the behavior in such case...

Thanks! I quickly double checked and in the AMI the NetworkManager-server-config package is not installed, there is NetworkManager-cloud-setup - we may still need a followup (or this PR?) to configure that though, it seems it needs an opt-in.

thozza:
> Thanks! I quickly double checked and in the AMI the NetworkManager-server-config package is not installed, there is NetworkManager-cloud-setup - we may still need a followup (or this PR?) to configure that though, it seems it needs an opt-in.
> 
> [edit: turns out we already setup nm-cloud-setup correctly in https://github.com/osbuild/images/blob/main/pkg/distro/rhel/rhel10/ami.go#L124 so no further changes should be required]

Yes, we install `@Server` only in the Azure image, but explicitly exclude NetworkManager-server-config. Sorry for not being explicit, but I meant mostly situation when the user specified `@Server` or the NM package as an additional package to install in the image and I was not sure if the ifcfg file would not "save the day" in such a case.

achilleas-k (approved):
Very nice pure red PR :D 
LGTM!

PR #823, opened 2024-07-31
rhel10: drop hardcoded UUIDs from {centos,rhel}-10, fedora

mvo5:
rhel10: drop hardcoded UUIDs from {centos,rhel}-10

Historically the rhel/centos image defintion hardcoded the partition
table, partition and filesystem UUIDs. The exact reasons for this
is unclear and it seems it was done mostly for comaptibility with
the previous generation of image build tools.

However given that a UUIDs is meant to be (practically) unique
from rhel10 on we stop hardcoding them. The images library will
then create a random one for each image build. For the cases
where this is undesirable the `OSBUILD_TESTING_RNG_SEED` env
can be used to control the RNG.

----

fedora: drop hardcoded UUIDs for partitions/filesystems

Similar to the change in the previous commit for centos10/rhel10
the hardcoding of the UUIDs is also dropped on fedora.

This is the result of the discussion in https://github.com/osbuild/images/pull/816

### DIFF
--- a/pkg/distro/fedora/partition_tables.go
+++ b/pkg/distro/fedora/partition_tables.go
@@ -9,22 +9,18 @@ import (
 
 var defaultBasePartitionTables = distro.BasePartitionTableMap{
 	arch.ARCH_X86_64.String(): disk.PartitionTable{
-		UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
 		Type: "gpt",
 		Partitions: []disk.Partition{
 			{
 				Size:     1 * common.MebiByte,
 				Bootable: true,
 				Type:     disk.BIOSBootPartitionGUID,
-				UUID:     disk.BIOSBootPartitionUUID,
 			},
 			{
 				Size: 200 * common.MebiByte,
 				Type: disk.EFISystemPartitionGUID,
-				UUID: disk.EFISystemPartitionUUID,
 				Payload: &disk.Filesystem{
 					Type:         "vfat",
-					UUID:         disk.EFIFilesystemUUID,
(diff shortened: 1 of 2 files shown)


thozza (approved):
Let's see if anyone will complain :sweat_smile:

cgwalters:
The messy thing here is I think it's actually a best practice to do what Fedora CoreOS derivatives have been doing for a while (and what systemd-repart encourages too), which is to regenerate the filesystem UUIDs on firstboot to avoid "UUIDs" actually being the same across multiple VMs instantiated from the single generated disk image.

That's...more difficult to do if the UUIDs are actually random in each disk image as it's harder to know (without using out of band metadata) that they need regeneration.

FCOS has hardcoded "UUIDs" today, but I think it would make sense to use zero.

BTW...there's an "ignition-vs-cloud-init" thing going on here because a lot of cloud VMs use ext4, and *historically* ext4 didn't allow changing the UUID when the filesystem was already mounted...and cloud-init only runs in the real root. (Actually [I may have it backwards](https://marc.info/?l=linux-fsdevel&m=170777799216007&w=2) and ext4 supports it but other FS's don't).

mvo5:
Thanks for the feedback @cgwalters - you raise a very good point here that this maybe only solving half-of-the problem and by doing that actually making it worse. I will look into what we can do, it might involve getting by-in to be able to run systemd-repart.

achilleas-k (approved):
I think this is good to merge now.
Let's follow it up with a systemd-repart configuration and zeroes like Colin suggested.

For that though I'd talk to other teams (for RHEL), because it might be a pretty significant change for cloud images and expectations around marketplace and golden images.

PR #824, opened 2024-07-31
Expose RHSM configuration as BP customization (COMPOSER-2308)

thozza:
RHSM in the image definitions is the last difference between RHEL RHUI images on el9 / el10, which can't be expressed as a BP customization. To enable the unification of RHUI and non-RHUI images and simplification of the image definitions, expose the RHSM configuration as a BP customization.

I considered naming the high-level customization `Subscription`, however [the name is already used by a customization exposed only by the Cloud API in osbuild-composer](https://github.com/osbuild/osbuild-composer/blob/573b349f1654a62a181fa5955a9a4f91440df4d5/internal/cloudapi/v2/openapi.v2.gen.go#L1013-L1022). So using it for the BP customization would make it clash with the existing one and look different for on-prem vs. service scenario.

**Notable changes**

- I introduced an internal implementation of the RHSM configuration and replaced the osbuild stage options with it in the distro and pipeline implementations.
- Simplified the `OSCustomizations` by having only a single instance of the RHSM configuration, which should be applied to the image, instead of having two and deciding which one to use based on the fact if the image should be subscribed or not. The decision logic is now moved to the function which produces `OSCustomization` instance from `ImageConfig`, `ImageOptions` and BP customizations.

**Open questions**

- I'm not sure if we want to simply override the default RHSM configuration from the image definition if the RHSM configuration is provided also as a BP customization, or merge them with the BP customization having precedence. Overriding would mean that values set in the default image configuration, but not set in the BP would disappear from the produced manifest. However, overriding would make the implementation much simpler, because the merging requires quite a lot of additional code, which I don't like.


pkg/subscription/subscription.go
```diff
@@ -18,3 +18,34 @@ const (
...
+type SubManDNFPluginsConfig struct {
+	ProductID           *DNFPluginConfig
+	SubscriptionManager *DNFPluginConfig
+}
+
+type RHSMConfig struct {
+	DnfPlugins *SubManDNFPluginsConfig
+	YumPlugins *SubManDNFPluginsConfig
+	SubMan     *SubManConfig
+}
```
achilleas-k, line 56: Does this make more sense to be a package in `pkg/customizations/` like we do with other internal representations?

I think we should also move the existing stuff into `pkg/customizations/subscription/` for consistency.

pkg/distro/image_config.go
```diff
@@ -39,7 +39,7 @@ type ImageConfig struct {
 
 	// for RHSM configuration, we need to potentially distinguish the case
 	// when the user want the image to be subscribed on first boot and when not
-	RHSMConfig          map[subscription.RHSMStatus]*osbuild.RHSMStageOptions
+	RHSMConfig          map[subscription.RHSMStatus]*subscription.RHSMConfig
```
achilleas-k, line 42: :heart:

pkg/distro/rhel/images.go
```diff
@@ -223,8 +224,17 @@ func osCustomizations(
...
-	osc.ShellInit = imageConfig.ShellInit
+	if options.Subscription != nil {
+		if rhsmConfig, exists := imageConfig.RHSMConfig[subscription.RHSMConfigWithSubscription]; exists {
+			osc.RHSMConfig = rhsmConfig
+		}
+	} else {
+		if rhsmConfig, exists := imageConfig.RHSMConfig[subscription.RHSMConfigNoSubscription]; exists {
+			osc.RHSMConfig = rhsmConfig
+		}
+	}
```
achilleas-k, line 235: Good change.  This makes more sense to me too.

achilleas-k:
> * I introduced an internal implementation of the RHSM configuration and replaced the osbuild stage options with it in the distro and pipeline implementations.
> * Simplified the `OSCustomizations` by having only a single instance of the RHSM configuration, which should be applied to the image, instead of having two and deciding which one to use based on the fact if the image should be subscribed or not. The decision logic is now moved to the function which produces `OSCustomization` instance from `ImageConfig`, `ImageOptions` and BP customizations.
>

:heart: :heart:

> **Open questions**
>
> * I'm not sure if we want to simply override the default RHSM configuration from the image definition if the RHSM configuration is provided also as a BP customization, or merge them with the BP customization having precedence. Overriding would mean that values set in the default image configuration, but not set in the BP would disappear from the produced manifest. However, overriding would make the implementation much simpler, because the merging requires quite a lot of additional code, which I don't like.

Yeah, I'm not sure either.  There's a lot of code in there to accommodate this and I'm wondering if it's worth it.  I think this blueprint customization is mostly for us (I don't expect users to be fiddling with RHSM configs) and the brew build configs will be pretty stable, not changing much if at all.  Given all that, it's probably not worth having a bunch of code to support conveniently changing a small subset of the configuration.  If this was a more user-facing customization, I'd argue for making it easier to tweak individual options without completely changing the config, but if I'm right that it's not, maybe simplifying the code (and the behaviour) is better in the long run.

achilleas-k (changes requested):
Requesting changes for the package move (from `pkg/subscription` to `pkg/customizations/subscription`).
Otherwise LGTM (pending discussion about merging vs overriding the full config).

pkg/subscription/subscription.go
thozza: Yes, it would probably make more sense. I'll move everything under `customizations`.

pkg/subscription/subscription.go
```diff
@@ -49,3 +54,48 @@ type RHSMConfig struct {
...
+		}
+		if plugins.SubscriptionManager != nil {
+			config.DnfPlugins.SubscriptionManager = &DNFPluginConfig{}
+			if plugins.SubscriptionManager.Enabled != nil {
+				config.DnfPlugins.SubscriptionManager.Enabled = common.ToPtr(*plugins.SubscriptionManager.Enabled)
+			}
+		}
+	}
+
+	// NB: YUMPlugins are not exposed to end users as a customization
```
kingsleyzissou, line 231: Thanks for the comment, I was going to ask about this.

pkg/subscription/subscription_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"testing"
+
+	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/stretchr/testify/assert"
+)
+
+// rhsmConfigNotSamePointers is a helper function to check that the pointers in the
+// two RHSMConfig objects are not the same.
+func rhsmConfigNotSamePointers(t *testing.T, c1, c2 *RHSMConfig) {
```
kingsleyzissou, line 14: Nitpick: this is unused here. Maybe it could be moved to the commit that adds the cloning and tests for this?

thozza:
> > * I'm not sure if we want to simply override the default RHSM configuration from the image definition if the RHSM configuration is provided also as a BP customization, or merge them with the BP customization having precedence. Overriding would mean that values set in the default image configuration, but not set in the BP would disappear from the produced manifest. However, overriding would make the implementation much simpler, because the merging requires quite a lot of additional code, which I don't like.
> 
> Yeah, I'm not sure either. There's a lot of code in there to accommodate this and I'm wondering if it's worth it. I think this blueprint customization is mostly for us (I don't expect users to be fiddling with RHSM configs) and the brew build configs will be pretty stable, not changing much if at all. Given all that, it's probably not worth having a bunch of code to support conveniently changing a small subset of the configuration. If this was a more user-facing customization, I'd argue for making it easier to tweak individual options without completely changing the config, but if I'm right that it's not, maybe simplifying the code (and the behaviour) is better in the long run.

I literally felt bad while writing all that code. :sweat_smile: I also don't like the end result and agree that it is too much code for a little benefit.

While I agree that the customization is mostly for us, it will get exposed to end users because Cloud API uses the `blueprint` implementation from osbuild-composer. And it is also used by Weldr API, thus exposed to users. I could add the customization to `ImageOptions`, so that it could be exposed only via Cloud API, but it feels wrong.

I can think of only one relatively good reason to keep the merging behavior. The default image RHSM config may contain configuration for subscription-manager itself and for its DNF plugins. Imagine that one provides only subscription-manager config in the BP, which would then wipe out DNF plugins config if these were part of the default configuration. This felt weird. But we can clearly document it.

There's maybe a middle-ground to do the merging only one level higher. This means that if ones provides DNF plugins config, we override only that part and don't touch the sub-man part. If one provides sub-man config, we override only that part and don't touch DNF plugins config. But this may still mean quite a lot of code to handle the logic...

IOW, I'm happy to simplify this in any way, just want to find out what majority of people think is the best option. :innocent:

pkg/subscription/subscription_test.go
thozza: Thanks, I probably missed the function while splitting the commit. :+1:

pkg/distro/rhel/images.go
mvo5: I wonder if this could be written slightly less nested, maybe something like:
```go
	var subscriptionStatus subscription.RHSMStatus
	if options.Subscription != nil {
		subscriptionStatus = subscription.RHSMConfigWithSubscription
	} else {
		subscriptionStatus = subscription.RHSMConfigNoSubscription
	}
	if rhsmConfig, exists := imageConfig.RHSMConfig[subscriptionStatus]; exists {
		osc.RHSMConfig = rhsmConfig
	}
```

pkg/manifest/os.go
```diff
@@ -648,14 +649,10 @@ func (p *OS) serialize() osbuild.Pipeline {
...
+	}
 
-		if rhsmConfig, exists := p.RHSMConfig[subscription.RHSMConfigWithSubscription]; exists {
-			pipeline.AddStage(osbuild.NewRHSMStage(rhsmConfig))
-		}
-	} else {
-		if rhsmConfig, exists := p.RHSMConfig[subscription.RHSMConfigNoSubscription]; exists {
-			pipeline.AddStage(osbuild.NewRHSMStage(rhsmConfig))
-		}
+	if rhsmConfig := p.RHSMConfig; rhsmConfig != nil {
```
mvo5, line 654: Maybe just:
```go
diff --git a/pkg/manifest/os.go b/pkg/manifest/os.go
index c0d238533..2cb4d3e5d 100644
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -651,8 +651,8 @@ func (p *OS) serialize() osbuild.Pipeline {
                p.EnabledServices = append(p.EnabledServices, subscribeServiceFile)
        }
 
-       if rhsmConfig := p.RHSMConfig; rhsmConfig != nil {
-               pipeline.AddStage(osbuild.NewRHSMStage(osbuild.NewRHSMStageOptions(rhsmConfig)))
+       if p.RHSMConfig != nil {
+               pipeline.AddStage(osbuild.NewRHSMStage(osbuild.NewRHSMStageOptions(p.RHSMConfig)))
        }
 
        if waConfig := p.WAAgentConfig; waConfig != nil {

```

mvo5 (commented):
Thank you! Inline are two quick drive-by ideas

pkg/manifest/os.go
thozza: Yes, I probably just edited one of the original conditions without thinking about it too much :facepalm: Thanks!

pkg/osbuild/rhsm_stage.go
```diff
@@ -52,3 +57,55 @@ type SubManConfigRHSMCERTDSection struct {
 	// Automatic system registration
 	AutoRegistration *bool `json:"auto_registration,omitempty"`
 }
+
+func NewRHSMStageOptions(config *subscription.RHSMConfig) *RHSMStageOptions {
```
mvo5, line 61: A bit of a meta question, we are doing a lot of work here convert and also deal with pointers - what if we would just make those normal structs without points so that clone would work naturally and then detect what sub-part is enabled/not-enabled by some other mechanism? Maybe a bad/unworkable idea but seeing all the scafolding here makes me wonder :)

pkg/subscription/subscription.go
achilleas-k: I'm so sorry
<details>
  <summary>I'm so sorry</summary>
yump lugins
</details>

achilleas-k:
> While I agree that the customization is mostly for us, it will get exposed to end users because Cloud API uses the `blueprint` implementation from osbuild-composer. And it is also used by Weldr API, thus exposed to users.

Right, of course.  I just meant that users *probably* wont be using it much, but I could be wrong.  I don't imagine many users would want to change the configuration for subman.  Or would they?

thozza:
> > While I agree that the customization is mostly for us, it will get exposed to end users because Cloud API uses the `blueprint` implementation from osbuild-composer. And it is also used by Weldr API, thus exposed to users.
> 
> Right, of course. I just meant that users _probably_ wont be using it much, but I could be wrong. I don't imagine many users would want to change the configuration for subman. Or would they?

Right. Hopefully none of the end users will use it :crossed_fingers: They shouldn't need to.

pkg/osbuild/rhsm_stage.go
thozza: The only reason why we use pointers to `bool` is that we actually need to distinguish if the (`false`) value was explicitly set in the BP / internal image configuration vs. being the default type value that Go initialized it to. Otherwise, we could be unintentionally disabling things in the configuration by the osbuild stage.

We need to use pointers to struct in the osbuild stage options struct to not have them serialized if marked as `omitempty`.

But your comment made me think about this more. We probably don't need to use pointers to structs, only to `bool`, in the internal RHSMConfig representation. I'll try to rework it.

schuellerf:
ultimately this will imply an additional section in https://github.com/osbuild/osbuild.github.io/blob/main/docs/user-guide/01-blueprint-reference.md
did I get this right?
probably stating that it's for on-prem only (until it's handed up to the API?)

achilleas-k:
> ultimately this will imply an additional section in https://github.com/osbuild/osbuild.github.io/blob/main/docs/user-guide/01-blueprint-reference.md did I get this right? probably stating that it's for on-prem only (until it's handed up to the API?)

Needs to land in osbuild-composer first but yes.

pkg/manifest/os.go
thozza: I simplified the condition (and a few additional ones that looked the same) in a separate commit.

pkg/subscription/subscription_test.go
thozza: Should be fixed now.

pkg/distro/rhel/images.go
thozza: I reduced the nesting in a separate commit.

pkg/subscription/subscription.go
thozza: Move the whole thing to `customizations`...

thozza:
I implemented all of the suggestions, except for the override behavior. The reason is that removal of pointers from the internal representation of RHSMConfig simplified the code enough that it may be acceptable. Nevertheless, I'm happy to change it.

To summarize the options for BP RHSM customization behavior when interacting with the image default configuration:

1. Merge the image default config with BP customization with the BP customization having a precedence. Merging is done on the lower level of struct properties. **(the current implementation)**
2. Override the image default config with BP customization. Potentially throwing away configuration values specified in the image definition, even if not set in the BP customization.
3. Do merging as in the point (1,), but on a higher level configuration blocks (e.g. "DNF Plugins" and "SubMan configuration").

Summary:
- Moved `subscription` package to `pkg/customizations`.
- Simplified the internal representation of RHSMConfig to use pointers only for primitive types, which simplified the implementation quite a lot.
- Implemented suggestions from @mvo5 to simplify a few conditions and reduce nesting.

pkg/osbuild/rhsm_stage.go
thozza: I reduced the amount of pointers to the bare minimum - only to the primitive types, which simplified the code a lot.

achilleas-k (approved):
Great stuff.
LGTM

achilleas-k:
@mvo5 Can you take a look at your comments and resolve them (if you're happy :) )?

pkg/osbuild/rhsm_stage.go
mvo5: Thank you! Really appreciated

mvo5 (approved):
Thank you!

PR #825, opened 2024-07-31
blueprint: fix filesystem minsize toml key

achilleas-k:
The toml tag for customizations.filesystem.minsize was incorrectly set to size.  In osbuild-composer this was already changed to minsize [1]. The blueprint implementation in this repo is used directly in bootc-image-builder, so the tags affect how the config is parsed by that tool.

Our blueprint reference for bootc-image-builder also uses `minsize` [2]. Let's finally unify the config everywhere.

[1] https://github.com/osbuild/osbuild-composer/pull/3783
[2] https://osbuild.org/docs/user-guide/blueprint-reference/#filesystems


kingsleyzissou (approved):
Thanks for fixing this 🙏

PR #826, opened 2024-07-31
manifest: remove redundant error check

achilleas-k:
The group stage block was checking for an error after the osbuild.GenGroupStage() call, which doesn't return an error.


mvo5 (approved):
Thank you, this looks like a silly copy/paste error by me.

PR #827, opened 2024-08-01
cmd/otk-*: fix all otk external to be wrapped in a "tree" json

mvo5:
When otk calls externals it expects them to be wrapped inside a "tree" json key for both the inputs and the outputs. This is important to make the schema extensible. However the existing otk externals in this repo do not respect that.

This commit fixes it and updates the tests.


achilleas-k (approved):
Neato.  TY!

PR #828, opened 2024-08-01
disk,otk-make-fstab-stage: make `GetFSTabOptions()` return an error and remove `recover()` 

mvo5:
disk: add static checks that interfaces are implemented

This commit adds some static checks that ensure that the disk,
btrfs,filesystem and lvm types implement the expected interfaces.

This is important because when changing e.g. the Mountable
interface by adding an "error" to GetFSTabOptions() without
this change all non-updated types just silently drop implementing
the interface without any error or warning leading to unexpected
test failures.

---

disk: make `GetFSTabOptions()` return an error

The disk package is now used for the `otk` externals. This means
that the inputs we get are much less controlled and instead of
panicing we should return more user-friendly error messages.

This commit starts by making `GetFSTabOptions()` return an error
and then bubbling it up.

Followup to https://github.com/osbuild/images/pull/787/files#r1673752935

----

otk-make-fstab-stage: remove `recover()`

With the changes in the previous commit to return an error from
`osbuild.NewFSTabStageOptions` if something goes wrong there is
no need anymore for the `recover()` that was used before.

Note that it is currently not entirely easy to reason about this
because one needs to manually follow all possible code paths
into `osbuild` and `disk` to see what functions could `panic()`.
For this one it was realtivly small, for the `otk-make-partition-stages`
it is a lot more involved though.

### DIFF
--- a/pkg/disk/filesystem.go
+++ b/pkg/disk/filesystem.go
@@ -76,15 +76,15 @@ func (fs *Filesystem) GetFSSpec() FSSpec {
 	}
 }
 
-func (fs *Filesystem) GetFSTabOptions() FSTabOptions {
+func (fs *Filesystem) GetFSTabOptions() (FSTabOptions, error) {
 	if fs == nil {
-		return FSTabOptions{}
+		return FSTabOptions{}, nil
 	}
 	return FSTabOptions{
 		MntOps: fs.FSTabOptions,
 		Freq:   fs.FSTabFreq,
 		PassNo: fs.FSTabPassNo,
-	}
+	}, nil
 }
 
 func (fs *Filesystem) GenUUID(rng *rand.Rand) {
--- a/cmd/otk-make-fstab-stage/main.go
+++ b/cmd/otk-make-fstab-stage/main.go
@@ -23,7 +23,11 @@ func makeFstabStages(inp Input) (stages *osbuild.Stage, err error) {
 	}()
 
 	pt := inp.Tree.Const.Internal.PartitionTable
-	fstabStage := osbuild.NewFSTabStage(osbuild.NewFSTabStageOptions(pt))
+	opts, err := osbuild.NewFSTabStageOptions(pt)
+	if err != nil {
+		return nil, err
+	}
+	fstabStage := osbuild.NewFSTabStage(opts)
 	return fstabStage, nil
 }
 
--- a/cmd/otk-make-fstab-stage/main.go
+++ b/cmd/otk-make-fstab-stage/main.go
@@ -15,32 +15,17 @@ type Input struct {
 	Tree otkdisk.Data `json:"tree"`
 }
 
-func makeFstabStages(inp Input) (stages *osbuild.Stage, err error) {
-	defer func() {
-		if r := recover(); r != nil {
-			err = fmt.Errorf("cannot generate image prepare stages: %v", r)
-		}
-	}()
-
-	pt := inp.Tree.Const.Internal.PartitionTable
-	opts, err := osbuild.NewFSTabStageOptions(pt)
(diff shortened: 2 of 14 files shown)


achilleas-k (changes requested):
Love this, thank you!

One change request though: Can we move the checks in the first commit (the interface implementation checks) to one or more `init()` functions?  Feels tidier than having "free-floating" vars in the package.

mvo5:
> Love this, thank you!
> 
> One change request though: Can we move the checks in the first commit (the interface implementation checks) to one or more `init()` functions? Feels tidier than having "free-floating" vars in the package.

Sure, happy to change this (although it is a pattern I have seen in a bunch of go projects). How about:
```go
diff --git a/pkg/disk/btrfs.go b/pkg/disk/btrfs.go
index b7ce637f3..15234a61c 100644
--- a/pkg/disk/btrfs.go
+++ b/pkg/disk/btrfs.go
@@ -10,10 +10,6 @@ import (
 
 const DefaultBtrfsCompression = "zstd:1"
 
-// btrfs implements the interface
-var _ = Container(&Btrfs{})
-var _ = UniqueEntity(&Btrfs{})
-
 type Btrfs struct {
        UUID       string
        Label      string
diff --git a/pkg/disk/btrfs_test.go b/pkg/disk/btrfs_test.go
index 3e51def77..0883e8df1 100644
--- a/pkg/disk/btrfs_test.go
+++ b/pkg/disk/btrfs_test.go
@@ -29,3 +29,8 @@ func TestBtrfsSubvolume_GetFSTabOptionsPanics(t *testing.T) {
        })
 
 }
+
+func TestImplementsInterfacesCompileTimeCheck(t *testing.T) {
+       var _ = Container(&Btrfs{})
+       var _ = UniqueEntity(&Btrfs{})
+}
```
(and same for the others) instead of moving it to a "init()" ? I'm slightly hesitant about init() because I would like to double check that the compiler completely optimizes it away (I'm 85% positive it does for the `var _ = ...` pattern). I guess it's a very minor concern (one extra function call per init) so fine to ignore. But somehow it feels nicer to me to move to the tests (if we do not want the `var _ = ` pattern). Wdyt?

achilleas-k:
> I'm slightly hesitant about init() because I would like to double check that the compiler completely optimizes it away (I'm 85% positive it does for the `var _ = ...` pattern). I guess it's a very minor concern (one extra function call per init) so fine to ignore. But somehow it feels nicer to me to move to the tests (if we do not want the `var _ = ` pattern). Wdyt?

Understandable.  I was going off the [description of `init()` in Effective Go](https://go.dev/doc/effective_go#init):
> init is called after all the variable declarations in the package have evaluated their initializers, and those are evaluated only after all the imported packages have been initialized.
>
> Besides initializations that cannot be expressed as declarations, a common use of init functions is to verify or repair correctness of the program state before real execution begins.

Doing it in tests also makes sense, completely removing it from the runtime and not relying on the compiler to optimise it out.

Honestly, I'm fine with any of the approaches.  You didn't even have to change it after my first review really :)

pkg/disk/filesystem.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/google/uuid"
 )
 
+// filesystem implements expected interface
+var _ = Mountable(&Filesystem{})
+var _ = UniqueEntity(&Filesystem{})
```
achilleas-k, line 12: The other checks were moved to tests but these weren't.  Is that accidental?

mvo5: Thanks, silly me - that was an oversight. Will fix :)

achilleas-k (approved):

PR #829, opened 2024-08-01
[RFC] internal: add new `bool3` package to help with tristate values

mvo5:
It seems we are currently using `*bool` a lot to distinguish between `{unset,true,false}`. This commit contains an alternative idea to introduce a new `Bool3` data-type that serves the same purpose but does not require the extra care that pointers require as it's all values.

[draft and only half done (i.e. tests missing and some more elegance in the usnet handling) but want early feedback so that I can stop if this is silly :)]

### DIFF
--- a/internal/bool3/bool3.go
+++ b/internal/bool3/bool3.go
@@ -1,0 +1,105 @@
+package bool3
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type Bool3 int
+
+const (
+	Unset Bool3 = iota
+	True
+	False
+)
+
+func New(v bool) Bool3 {
+	if v {
+		return True
+	}
+	return False
+}
+
+func (b Bool3) String() string {
+	switch b {
+	case True:
+		return "true"
+	case False:
+		return "false"
+	default:
+		return "unset"
+	}
+}
+
+func parseBool3(v interface{}) (Bool3, error) {
+	switch val := v.(type) {
+	case nil:
+		return Unset, nil
+	case bool:
+		switch val {
+		case true:
+			return True, nil
+		case false:
+			return False, nil
+		}
+	case string:
+		switch val {
+		case "true":
+			return True, nil
+		case "false":
+			return False, nil
+		case "unset":
+			return Unset, nil
+		default:
+			return Unset, fmt.Errorf("cannot parse %q as Bool3", val)
+		}
+	}
+	return Unset, fmt.Errorf("cannot unmarshal %T to Bool3", v)
+}
+
+func (b Bool3) MarshalJSON() ([]byte, error) {
+	switch b {
+	case True, False:
+		return json.Marshal(b.String())
+	}
+	// XXX: or should we just remove this special case and use "unset"?
+	return []byte("null"), nil
+}
+
+func (b *Bool3) UnmarshalJSON(data []byte) error {
+	var v interface{}
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	bb, err := parseBool3(v)
+	*b = bb
+	return err
+}
+
+func (b Bool3) MarshalTOML() ([]byte, error) {
+	// TODO: return "null" here for unset too?
+	return []byte(fmt.Sprintf("\"%s\"", b.String())), nil
+}
+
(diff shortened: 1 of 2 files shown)


achilleas-k:
This does look like it might save us from some headaches.  I'd be curious to see how it changes real uses of `*bool` in the code.  I suspect we will be dropping a lot of `b == nil` checks and minimise a lot of accidental nil dereferencing.

cgwalters:
Of note, the [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) discourage bools:

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

But on the naming bikeshed I've used `OptionalBool` in the past in Go for this because I think in Rust where this is `Option<bool>`.

achilleas-k:
> But on the naming bikeshed I've used `OptionalBool` in the past in Go for this because I think in Rust where this is `Option<bool>`.

I like this naming.  It got me thinking we could name the package `optional` and have `optional.Bool` with the choice of adding more in the future (`optional.String`, with helpers for dereferencing safely), but maybe that's going a bit too far.

internal/bool3/bool3.go
achilleas-k, line 1: Would it make sense to put this in the `common` package instead?

thozza:
Note that we have the same situation also for integers, where the default value `0` may be a valid configuration value and therefore we need to distinguish when it was explicitly set and when not. E.g. https://github.com/osbuild/images/blob/a2ab90999698ec2aa8615a9b1414956cf0d3b9d0/pkg/blueprint/repository_customizations.go#L21

And lastly, we have at least one case of the same for a `string` type in https://github.com/osbuild/images/blob/a2ab90999698ec2aa8615a9b1414956cf0d3b9d0/pkg/osbuild/chrony_stage.go#L5.
[Where an empty string in the osbuild stage options will actually remove the option from the chrony config file.](https://github.com/osbuild/osbuild/blob/b516bdca93f732428598967851f9ec7a8b07c9f0/stages/org.osbuild.chrony.meta.json#L90-L93)

thozza:
And a quick search revealed https://github.com/markphelps/optional, which seems to be what we are trying to do here :thinking:

achilleas-k:
> And a quick search revealed https://github.com/markphelps/optional, which seems to be what we are trying to do here 🤔

I was sure (without looking) that something like this existed, but (at the risk of going full-on NIH) I feel this is small enough to just make ourselves.  It's not exactly leftpad territory but it's getting close.

thozza:
> > And a quick search revealed https://github.com/markphelps/optional, which seems to be what we are trying to do here 🤔
> 
> I was sure (without looking) that something like this existed, but (at the risk of going full-on NIH) I feel this is small enough to just make ourselves. It's not exactly leftpad territory but it's getting close.

I don't have a strong preference here. I naturally incline to using a library that seems to be already tested and proven (not saying that this one is), instead of spending time implementing it. If nothing else, this may be a good inspiration.

The project page also links https://github.com/leighmcculloch/go-optional, which uses a different approach.

mvo5:
> > > And a quick search revealed https://github.com/markphelps/optional, which seems to be what we are trying to do here 🤔
> > 
> > 
> > I was sure (without looking) that something like this existed, but (at the risk of going full-on NIH) I feel this is small enough to just make ourselves. It's not exactly leftpad territory but it's getting close.
> 
> I don't have a strong preference here. I naturally incline to using a library that seems to be already tested and proven (not saying that this one is), instead of spending time implementing it. If nothing else, this may be a good inspiration.
> 
> There are also:
> 
>     * https://github.com/leighmcculloch/go-optional
> 
>     * https://github.com/moznion/go-optional

Thanks for looking into this! I like go-optional. It does not have yaml/toml marshalers, do we need those? If not that lib sounds like a fine starting point to explore if this approach makes the exiting blueprint code simpler (I'm happy to explore this fwiw, I don't want to push my minor pet-peeves on anyone)

mvo5:
Closing in favor of going with a library (at least for now until we learn that we need something else) https://github.com/osbuild/images/pull/831

PR #831, opened 2024-08-02
blueprint: use go-optional to avoid points for unset 

mvo5:
It seems we are currently using *bool a lot to distinguish between
{unset,true,false}. This commit instead uses the `go-optional`
library to distinguish between unset/undefined and the value.

It uses generics to works on every type.

[draft as this is incomplete and also lacks tests that check if toml unmarshal is working correctly but so far I really like the result, thanks @thozza ! ]

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/hashicorp/go-version v1.7.0
 	github.com/kolo/xmlrpc v0.0.0-20201022064351-38db28db192b
+	github.com/moznion/go-optional v0.12.0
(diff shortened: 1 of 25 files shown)


mvo5:
As I feared in the oriinal `bool3` idea the toml parsing is a problem and just using stock `go-optional` does not work. Go makes it really cumbersome to extend types from diffrent packages (IMHO) and toml support for unmarshaling complex types (like structs) is very limited so we need a) toml unmarshal b) a way to limit the generic to primitive types we control. I close this one for now but I have ideas how to archive (a)+(b) in https://github.com/osbuild/images/compare/main...mvo5:go-optional-less-ptr-internal?expand=1. I put it a bit on the backburner though as it's not pressing  (also sorry for the back and for forth here).

Issue #837, opened 2024-08-05
Unable to boot after upgrading an image with file system customization 


say-paul:
The issue is that the osbuild-ostree-mountpoints.service cannot create custom directories as seen in journald. 

```
Aug 05 08:31:59 localhost mkdir[1048]: mkdir: cannot create directory ‘/foo’: Operation not permitted
Aug 05 08:31:59 localhost mkdir[1048]: mkdir: cannot create directory ‘/foo’: Operation not permitted
Aug 05 08:31:59 localhost systemd[1]: osbuild-ostree-mountpoints.service: Main process exited, code=exited, status=1/FAILURE
Aug 05 08:31:59 localhost systemd[1]: osbuild-ostree-mountpoints.service: Failed with result 'exit-code'.
Aug 05 08:31:59 localhost systemd[1]: Failed to start Ensure custom filesystem mountpoints exist.
Aug 05 08:31:59 localhost systemd[1]: foo.mount: Failed to check directory /foo: No such file or directory
Aug 05 08:31:59 localhost systemd[1]: Mounting /foo...
Aug 05 08:31:59 localhost mount[1051]: mount: /foo: mount point does not exist.
Aug 05 08:31:59 localhost systemd[1]: foo.mount: Mount process exited, code=exited, status=32/n/a
Aug 05 08:31:59 localhost systemd[1]: foo.mount: Failed with result 'exit-code'.
Aug 05 08:31:59 localhost systemd[1]: Failed to mount /foo.
Aug 05 08:31:59 localhost systemd[1]: Dependency failed for /foo/bar.
Aug 05 08:31:59 localhost systemd[1]: Dependency failed for Local File Systems.
Aug 05 08:31:59 localhost systemd[1]: Dependency failed for Mark the need to relabel after reboot.
```
osbuild-ostree-mountpoints.service:
```
[core@localhost ~]$ sudo cat /etc/systemd/system/osbuild-ostree-mountpoints.service
[Unit]
Description=Ensure custom filesystem mountpoints exist
DefaultDependencies=False
ConditionPathIsDirectory=|!/foo/bar
ConditionPathIsDirectory=|!/foo
After=ostree-remount.service
Before=foo-bar.mount
Before=foo.mount

[Service]
Type=oneshot
ExecStartPre=/bin/sh -c "if grep -Uq composefs /run/ostree-booted; then chattr -i /; fi"
[...]

PR #838, opened 2024-08-06
fix-fs-customization:exec start and stop logic

say-paul:
ExceStart/Stop should set `chattr` when ostree does not have composefs enabled. With composefs `chattr` is not required explicitly but tansient rootfs should be enabled to ensure the service can create the required mountpoints.

fixes: #837

### DIFF
--- a/pkg/manifest/ostree_deployment.go
+++ b/pkg/manifest/ostree_deployment.go
@@ -521,8 +521,8 @@ func createMountpointService(serviceName string, mountpoints []string) *osbuild.
 		Type:            osbuild.Oneshot,
 		RemainAfterExit: false,
 		// compatibility with composefs, will require transient rootfs to be enabled too.
-		ExecStartPre: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted; then chattr -i /; fi\""},
-		ExecStopPost: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted; then chattr +i /; fi\""},
+		ExecStartPre: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted; then echo 'Warning: composefs enabled! ensure transient rootfs is enabled too.'; else chattr -i /; fi\""},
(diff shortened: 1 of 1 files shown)


pkg/manifest/ostree_deployment.go
```diff
@@ -521,8 +521,8 @@ func createMountpointService(serviceName string, mountpoints []string) *osbuild.
 		Type:            osbuild.Oneshot,
 		RemainAfterExit: false,
 		// compatibility with composefs, will require transient rootfs to be enabled too.
-		ExecStartPre: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted; then chattr -i /; fi\""},
-		ExecStopPost: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted; then chattr +i /; fi\""},
+		ExecStartPre: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted;  then echo 'Warning:composefs enabled! ensure transient rootfs is enabled too.'; else chattr -i /; fi\""},
+		ExecStopPost: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted;  then echo 'Warning:composefs enabled! ensure transient rootfs is enabled too.'; else chattr +i /; fi\""},
```
achilleas-k, line 525: Can you fix the spacing here?
```suggestion
		ExecStartPre: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted; then echo 'Warning: composefs enabled! ensure transient rootfs is enabled too.'; else chattr -i /; fi\""},
		ExecStopPost: []string{"/bin/sh -c \"if grep -Uq composefs /run/ostree-booted; then echo 'Warning: composefs enabled! ensure transient rootfs is enabled too.'; else chattr +i /; fi\""},
```

Space after `:` and single space before `then`.

achilleas-k (changes requested):
Oh, did I accidentally flip the condition when I redid the implementation last time?
I should really add a real test for this.

say-paul:
Testing this is bit lengthy as it needs an upgrade to test it. I have asked QE to have a quick test at it considering itm 26. In long run we can add some tests.

pkg/manifest/ostree_deployment.go
say-paul: done

achilleas-k (approved):

PR #839, opened 2024-08-06
manifest: add passwd when users are created

andremarianiello:
org.usbuild.users runs passwd in the os tree using chroot, but this package wasn't guaranteed to be installed


bcl (approved):
Looks ok to me.

achilleas-k (approved):
Thanks for this

PR #840, opened 2024-08-06
Move pathpolicy out of internal/ into pkg/

achilleas-k:
Following up from 41b1d00770061f616bc7defc53a03cb711227e90, we now move the pathpolicy package out of internal as well.  This is mostly necessary because there are public functions and types, that depend on types in pathpolicy.

blueprint.CheckMountpointsPolicy() takes *pathpolicy.PathPolicies as a second argument.
The policies under pkg/policies/ are created using pathpolicy.NewPathPolicies() with maps where the values are pathpolicy.PathPolicy.

Without access to the pathpolicy package, the policies can be checked using the existing ones, but no new (or modified) policies can be created by a consumer that might have other restrictions (like, bootc-image-builder).

https://github.com/osbuild/images/pull/395#issuecomment-1909832368


bcl (approved):
Looks good!

ondrejbudai (dismissed):

thozza (dismissed):

achilleas-k:
`go fmt` part of the source prep script reordered the imports after the package path changed which I didn't do originally.

Serves me right for `sed`ing and leaving it at that.

thozza (approved):

PR #843, opened 2024-08-07
distro/rhel10: drop the openstack image type

ondrejbudai:
OpenStack documentation points at the RHEL KVM Guest Image as the
official way to get RHEL up and running on OpenStack.

See https://docs.redhat.com/en/documentation/red_hat_openstack_platform/17.1/html/creating_and_managing_images/assembly_glance-creating-images_osp#proc_using-a-rhel-kvm-instance-image_glance-creating-rhel-kvm-images

osbuild-composer inherited the openstack image type from lorax-composer.
I'm not sure if something was different in the RHEL 7 times, or why
a separate image type was created back then. However nowadays, qcow2
is apparently the official option even for OpenStack, so we don't need
to carry a separate type forward. Thus, this commit drops it.


achilleas-k (approved):
Nice cleanup!

PR #845, opened 2024-08-08
distro/rhel{9,10}: remove bogus rd.blacklist=nouveau kernel (RHEL-53491)

ondrejbudai:
rd.blacklist is actually not processed by anything. The correct option is called rd.driver.blacklist, see
https://man.archlinux.org/man/dracut.8.en

Hovewer, we can actually safely remove it altogether, because we also blacklist nouveau in a /etc/modprobe.d drop-in config. This is actually the recommended way, see
https://access.redhat.com/solutions/41278

Thus, this commit removes the redundant, and actually bogus argument.

### DIFF
--- a/pkg/distro/rhel/rhel10/ami.go
+++ b/pkg/distro/rhel/rhel10/ami.go
@@ -9,7 +9,7 @@ import (
 )
 
 // TODO: move these to the EC2 environment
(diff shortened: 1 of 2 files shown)


achilleas-k:
FTR, this page mentions `rd.blacklist`, not as a recommendation, but in passing as an alternative.  I don't know where else this is documented.

achilleas-k (approved):
Splendid!

PR #846, opened 2024-08-08
osbuild: only mount the required mountpoints for bootc install (COMPOSER-2301)

achilleas-k:
bootc install to-filesystem requires a clean root to install, with only the required mountpoints present: /, /boot, and /boot/efi.  Do not mount any user-defined mountpoints when running the bootc install stage.

Content in bootc images cannot be separated across filesystems, therefore there cannot be any content under the custom mountpoints, so this change will not cause any issues of content not being on the correct filesystem.

See also the discussion at https://github.com/osbuild/bootc-image-builder/issues/579

### DIFF
--- a/pkg/osbuild/bootc_install_to_filesystem_stage.go
+++ b/pkg/osbuild/bootc_install_to_filesystem_stage.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/osbuild/images/pkg/platform"
+	"golang.org/x/exp/slices"
 )
 
 type BootcInstallToFilesystemOptions struct {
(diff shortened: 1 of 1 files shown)


pkg/osbuild/bootc_install_to_filesystem_stage.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/osbuild/images/pkg/platform"
+	"golang.org/x/exp/slices"
```
ondrejbudai, line 7: We have Go 1.21, `slices` should work just fine.

ondrejbudai (approved):
Thanks!

thozza (approved):

pkg/osbuild/bootc_install_to_filesystem_stage.go
achilleas-k: Gonna merge this 'cause I'd like to have it for the partitioning in BIB and open another PR for the import swap.

achilleas-k: https://github.com/osbuild/images/pull/849

PR #849, opened 2024-08-09
go.mod: replace golang.org/x/exp/slices with slices

achilleas-k:
**go.mod: replace golang.org/x/exp/slices with slices**

Since we updated the go version to 1.21, we can use the stdlib slices package, instead of the x/exp version of it.

Branch is based on #846.


ondrejbudai (approved):
Yay!

achilleas-k:
ugh, automatic rebase doesn't like the first commit.  Fixing.

bcl (approved):
Oh darn, now I will have to remember not to point out that we can't use it yet ;)

PR #854, opened 2024-08-12
pathpolicy: move to generics and unexport pathTrie

mvo5:
pathpolicy: move `PathTrie` to a generic

With generics support in go it seems like a good idea to move the
`PathTrie` to a generic.

---

pathpolicy: unexport `PathTrie` as it's an implementation
 detail

The `pathpolicy` packages is currently exporting the `PathTrie`
data structure. It seems this is not really used outside of the
package itself and more an implementation detail than a public
API. So just unexport it.


ondrejbudai (approved):
Looks sane, thanks!

PR #855, opened 2024-08-12
add support in macOS crypt function

mgold1234:
this commit generates a slat, hashes the password using SHA-512, and encodes the result in base64.
- Replaced the original CryptSHA512 function with a custom implementation using Go's standard `crypto/sha512` library.
- Removed dependency on system-specific cryptographic functions to ensure compatibility across different operating systems,
- Updated the relevant usage of the CryptSHA512 function in the codebase. including macOS

### DIFF
--- a/pkg/crypt/crypt.go
+++ b/pkg/crypt/crypt.go
@@ -2,6 +2,9 @@ package crypt
 
 import (
 	"crypto/rand"
+	"crypto/sha512"
+	"encoding/base64"
+	"fmt"
 	"math/big"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


pkg/crypt/crypt.go
```diff
@@ -18,8 +21,14 @@ func CryptSHA512(phrase string) (string, error) {
 		return "", nil
 	}
 
-	hashSettings := "$6$" + salt
-	return crypt(phrase, hashSettings)
+	hash := sha512.New()
```
ondrejbudai, line 24: `crypt(3)` does more than just hashing the salt and the password with sha512, see https://github.com/besser82/libxcrypt/blob/72f75aa370ae96ccd2cc44ea3cf4182d8679ffbe/lib/crypt-sha512.c#L85.

I feel quite uncomfortable reimplementing this code, because I don't want us to maintain too much crypto-related code.

mvo5: Nice blast from the past for me (c.f. https://github.com/mvo5/sha512crypt-node) - the paper what exactly sha512crypt is doing is in http://www.akkadia.org/drepper/SHA-crypt.txt but I agree with Ondrej that ideally we want to call into libc here for maximum compatibility.

mgold1234: thank you @ondrejbudai maybe do you have an idea how to support mac after allow user customization with password?
https://github.com/osbuild/image-builder/pull/1283

mgold1234:
closing this pr, we will the relevant tests that failing with mac

PR #859, opened 2024-08-14
platform: remove unused X86BootLoader type

ondrejbudai:
I cannot find any places that use this type. Let's just drop it.


mvo5 (approved):
Thanks!

PR #860, opened 2024-08-14
[RFC] many: add `types.Option` generic and use in customizations

mvo5:
For customizations we currently use pointers when we want to distinguish between set or unset value. This works fine and is supported by json/toml marshaling natively. However we lose some safety because there is always the risk of dereferencing a pointer and extra checks for `nil`. 

This PR introduces a new `types.Option` that wraps the primitive types bool,int,string [0] so that they are there is a distinction between "set" (Some) and "unset" (None) [1]. This allows us to do things like:
```go
-       if hostname := c.GetHostname(); hostname != nil {
-               osc.Hostname = *hostname
-       } else {
-               osc.Hostname = "localhost.localdomain"
-       }
+       osc.Hostname = c.GetHostname().TakeOr("localhost.localdomain")
```
or
```go
-       if hostname := c.GetHostname(); hostname != nil {
-               osc.Hostname = *hostname
-       }
+       osc.Hostname = c.GetHostname().Unwrap()
```
And there is overall a much smaller risk of making mistakes while also be more explicit (well, minor as arguably is as `*` is also a strong indicator).

This PR converts a small subset to show what it would look like, it's a lot of busywork so maybe not worth the effort but I like it over the pointers (but I understand if other wouldn't).

[0] Sadly the go type system and toml make it hard to do the same for structs which would be extremely nice, see the commit message/code comments for details.
[1] Of course this is very much inspired by the rust std::Some type and the go-optional libraries.

### DIFF
--- a/internal/types/option.go
+++ b/internal/types/option.go
@@ -53,11 +53,20 @@ func (o Option[T]) IsNone() bool {
 	return o == nil
 }
 
+// IsZero returns True if the Option *doesn't* have a value
+// compat with the new go-1.24 IsZero() protocol
+func (o Option[T]) IsZero() bool {
+	return o.IsNone()
+}
+
 // IsSome returns whether the Option has a value or not.
 func (o Option[T]) IsSome() bool {
 	return o != nil
 }
 
+// XXX: this probably needs a different name to like UnwrapOrDefault()
+// or something
+//
 // Unwrap returns the value regardless of Some/None status.
 // If the Option value is Some, this method returns the actual value.
 // On the other hand, if the Option value is None, this method returns the *default* value according to the type.
--- a/internal/types/option.go
+++ b/internal/types/option.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+
+	"github.com/osbuild/images/internal/common"
 )
 
 // Option is a more constrained subset of the code in
@@ -114,4 +116,8 @@ func (o *Option[T]) UnmarshalTOML(data any) error {
 	}
 	*o = Some(b)
 	return nil
+}
+
+func (o *Option[T]) UnmarshalYAML(unmarshal func(any) error) error {
+	return common.UnmarshalYAMLviaJSON(o, unmarshal)
 }
--- a/internal/types/option.go
+++ b/internal/types/option.go
@@ -1,0 +1,117 @@
+package types
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+// Option is a more constrained subset of the code in
+//     https://github.com/moznion/go-optional
+//
+// It is not using an external go-optional lib directly because none
+// has toml unmarshal support and also because there is no support for
+// complex types in UnmarshalTOML (it only gives a single
+// reflect.Value of type any) so our optional "lib" must come with
+// limitations for this.
+//
+// Unfortunately there is no way I could find to import the code and
+// limit the supported types. So this is a copy of the subset.
+// Fortunatly the code is small, targeted and easy to follow so it
+// should not be too bad.
+
+type OptionTomlTypes interface {
+	~int | ~bool | ~string
+}
+
+// Option is a subset of github.com/moznion/go-optional for use with toml
+
+// Option is a data type that must be Some (i.e. having a value) or None (i.e. doesn't have a value).
+// This type implements database/sql/driver.Valuer and database/sql.Scanner.
+type Option[T OptionTomlTypes] []T
+
+const (
+	value = iota
+)
+
+// Some is a function to make an Option type value with the actual value.
+func Some[T OptionTomlTypes](v T) Option[T] {
+	return Option[T]{
+		value: v,
+	}
+}
+
+// None is a function to make an Option type value that doesn't have a value.
+func None[T OptionTomlTypes]() Option[T] {
+	return nil
+}
+
+// IsNone returns True if the Option *doesn't* have a value
+func (o Option[T]) IsNone() bool {
+	return o == nil
+}
+
+// IsSome returns whether the Option has a value or not.
+func (o Option[T]) IsSome() bool {
+	return o != nil
+}
+
+// Unwrap returns the value regardless of Some/None status.
+// If the Option value is Some, this method returns the actual value.
+// On the other hand, if the Option value is None, this method returns the *default* value according to the type.
+func (o Option[T]) Unwrap() T {
+	if o.IsNone() {
+		var defaultValue T
+		return defaultValue
+	}
+	return o[value]
+}
+
+// TakeOr returns the actual value if the Option has a value.
+// On the other hand, this returns fallbackValue.
+func (o Option[T]) TakeOr(fallbackValue T) T {
+	if o.IsNone() {
+		return fallbackValue
+	}
+	return o[value]
+}
+
+var jsonNull = []byte("null")
+
+func (o Option[T]) MarshalJSON() ([]byte, error) {
+	if o.IsNone() {
+		return jsonNull, nil
+	}
+
+	marshal, err := json.Marshal(o.Unwrap())
+	if err != nil {
+		return nil, err
+	}
+	return marshal, nil
+}
+
+func (o *Option[T]) UnmarshalJSON(data []byte) error {
+	if len(data) <= 0 || bytes.Equal(data, jsonNull) {
+		*o = None[T]()
+		return nil
+	}
+
+	var v T
+	err := json.Unmarshal(data, &v)
+	if err != nil {
+		return err
+	}
+	*o = Some(v)
+
+	return nil
+}
+
+// not part of github.com/moznion/go-optional
+func (o *Option[T]) UnmarshalTOML(data any) error {
+	b, ok := data.(T)
+	if !ok {
+		return fmt.Errorf("cannot use %[1]v (%[1]T) as bool", data)
+	}
+	*o = Some(b)
+	return nil
+}
--- a/pkg/osbuild/users_stage.go
+++ b/pkg/osbuild/users_stage.go
@@ -1,6 +1,7 @@
 package osbuild
 
 import (
+	"github.com/osbuild/images/internal/types"
 	"github.com/osbuild/images/pkg/crypt"
 	"github.com/osbuild/images/pkg/customizations/users"
 )
@@ -12,16 +13,16 @@ type UsersStageOptions struct {
 func (UsersStageOptions) isStageOptions() {}
 
 type UsersStageOptionsUser struct {
-	UID                *int     `json:"uid,omitempty"`
-	GID                *int     `json:"gid,omitempty"`
-	Groups             []string `json:"groups,omitempty"`
-	Description        *string  `json:"description,omitempty"`
-	Home               *string  `json:"home,omitempty"`
-	Shell              *string  `json:"shell,omitempty"`
-	Password           *string  `json:"password,omitempty"`
-	Key                *string  `json:"key,omitempty"`
-	ExpireDate         *int     `json:"expiredate,omitempty"`
-	ForcePasswordReset *bool    `json:"force_password_reset,omitempty"`
+	UID                types.Option[int]    `json:"uid,omitempty"`
+	GID                types.Option[int]    `json:"gid,omitempty"`
+	Groups             []string             `json:"groups,omitempty"`
+	Description        types.Option[string] `json:"description,omitempty"`
+	Home               types.Option[string] `json:"home,omitempty"`
+	Shell              types.Option[string] `json:"shell,omitempty"`
+	Password           types.Option[string] `json:"password,omitempty"`
+	Key                types.Option[string] `json:"key,omitempty"`
+	ExpireDate         types.Option[int]    `json:"expiredate,omitempty"`
+	ForcePasswordReset types.Option[bool]   `json:"force_password_reset,omitempty"`
 }
 
 func NewUsersStage(options *UsersStageOptions) *Stage {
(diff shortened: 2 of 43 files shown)


internal/types/option.go
```diff
@@ -0,0 +1,117 @@
...
+
+type OptionTomlTypes interface {
+	~int | ~bool | ~string
+}
+
+// Option is a subset of github.com/moznion/go-optional for use with toml
+
+// Option is a data type that must be Some (i.e. having a value) or None (i.e. doesn't have a value).
+// This type implements database/sql/driver.Valuer and database/sql.Scanner.
+type Option[T OptionTomlTypes] []T
```
mvo5, line 31: Fwiw, we could call this `Optional` too which I actually slightly prefer but then `Option` is what rust uses and lots of people are familar with this terminology now.

internal/types/option.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	value = iota
+)
+
+// Some is a function to make an Option type value with the actual value.
+func Some[T OptionTomlTypes](v T) Option[T] {
```
mvo5, line 38: Fwiw, we could call this differently, e.g. `types.NewOptional("foo")` or somesuch but again rust is the inspiration and it seems nice to keep close to their terminology as many people will be familiar with it.

supakeen (dismissed):
I quite like this though I don't think it gets us much *benefit* in Go (there's no nice syntax and stdlib to deal with options) there's also no real downside.

It's nice being able to see at the calling side of the code that we're handling an option and not a random pointer.

pkg/osbuild/users_stage.go
```diff
@@ -39,18 +40,18 @@ func NewUsersStageOptions(userCustomizations []users.User, omitKey bool) (*Users
 	users := make(map[string]UsersStageOptionsUser, len(userCustomizations))
 	for _, uc := range userCustomizations {
 		// Don't hash empty passwords, set to nil to lock account
-		if uc.Password != nil && len(*uc.Password) == 0 {
+		if len(uc.Password.Unwrap()) == 0 {
```
supakeen, line 43: This is different behavior from the Rust we're trying to emulate where `.unwrap()` would panic if `uc.Password` is not `Some`.

mvo5: Yeah, I took this from the upstream go-optional lib, but happy to change and call it e.g. "UnwrapWithDefault" or "TakeWithDefault" or just "Take"  to make this clearer. It convinient but I agree that for rusians it might be confusing.

supakeen: What does it do in this library if the value is None? Still panic?

mvo5:
After re-reading I still think it's nice but given that it got no traction since Aug (and the amount of conflicts/maintenance required to keep it up-to-date) I will close it (which makes me slightly sad)

lzap:
As I said in our chat, I very much like the idea and I would like to try to integrate it into UBP, I will share a PR today or early next week so we can see how it looks like there. I would suggest that we would put this into UBP and use the same type here so fields can be easily copied into `images` BP rather than creating two distinct types.

However, I am not so sure about the `Option` type name. What looks a bit better is `Optional` but that is quite long. And don't get me started about `types` package name, that is way too generic. This name probably fits 90% of all packages since they contain types too. Here are few more ideas:

* Optional
* Nullable (okay and also aligns with OpenAPI/JSON Schema too)
* Reference (pointer is in fact reference)
* Ref (very short version of Reference)
* Value (too generic)
* RefValue (maybe too complicated)
* Pointer (does not encapsulate what it does)
* Wrapper (same)

Now, there is already a package called `ptr` in [UBP at the moment which is used](https://github.com/osbuild/blueprint-schema/blob/main/pkg/ptr/any.go) extensively in the convertor (many pointer to value - value to pointer assignments). I know that some projects (e.g. CRC) do have a pointer helpers in `common` eg. `common.Ptr` but I hate all-catch packages named `common` or `utils` this is a bad practice that leads to many functions mixed in a single package.

Thus I recommend to put this name into this package that could be then used across all projects, it can be in UBP for now and maybe we decide to separate it. Thus we should probably choose a name that goes nicely with `ptr`. Thus let me try this:

* `ptr.Option`
* `ptr.Optional`
* `ptr.Nullable` (feels wrong, all pointers can be nil)
* `ptr.Reference`
* `ptr.Ref`
* `ptr.Value`
* `ptr.RefValue`
* `ptr.Pointer` (nope)
* `ptr.Wrapper`
* `ptr.Safe` (maybe)

I am gonna go with `ptr.Ref` for my prototype, since this package is dedicated only to pointer dereferencing or helpers, I think it make sense to even shorten `ptr.NewRef()` to just `ptr.New()`.

Instead of Some/None terms, we should use the newly selected term + zero because it aligns with `omitzero` / `IsZero` features introduced in Go 1.24.

internal/types/option.go
```diff
@@ -0,0 +1,132 @@
...
+// TakeOr returns the actual value if the Option has a value.
+// On the other hand, this returns fallbackValue.
+func (o Option[T]) TakeOr(fallbackValue T) T {
+	if o.IsNone() {
+		return fallbackValue
+	}
+	return o[value]
+}
+
+var jsonNull = []byte("null")
```
lzap, line 90: This is a very dangerous "constant" by the way, slices are always passed by semi-reference and can be modified, this could lead to fatal errors. A copy will be safer.

internal/types/option.go
```diff
@@ -0,0 +1,132 @@
...
+
+	marshal, err := json.Marshal(o.Unwrap())
+	if err != nil {
+		return nil, err
+	}
+	return marshal, nil
+}
+
+func (o *Option[T]) UnmarshalJSON(data []byte) error {
+	if len(data) <= 0 || bytes.Equal(data, jsonNull) {
```
lzap, line 105: Nitpick: this is inefficient, `string(data) == "null"` will be faster.

internal/types/option.go
```diff
@@ -0,0 +1,132 @@
...
+// reflect.Value of type any) so our optional "lib" must come with
+// limitations for this.
+//
+// Unfortunately there is no way I could find to import the code and
+// limit the supported types. So this is a copy of the subset.
+// Fortunatly the code is small, targeted and easy to follow so it
+// should not be too bad.
+
+type OptionTomlTypes interface {
+	~int | ~bool | ~string
```
lzap, line 26: Just for the record, you do not have to fix this here I think it would be better to have a common type in UBP:

```suggestion
	~bool | ~string | ~int | ~float64 | ~float32 | ~int8 | ~int16 | ~int32 | ~int64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64
```

internal/types/option.go
```diff
@@ -0,0 +1,132 @@
...
+	*o = Some(v)
+
+	return nil
+}
+
+// not part of github.com/moznion/go-optional
+func (o *Option[T]) UnmarshalTOML(data any) error {
+	b, ok := data.(T)
+	if !ok {
+		return fmt.Errorf("cannot use %[1]v (%[1]T) as bool", data)
```
lzap, line 124: Just for the record, you do not have to fix this here I think it would be better to have a common type in UBP:

```suggestion
		return fmt.Errorf("cannot use %[1]v (%[1]T) as type %[2]T", data, *new(T))
```

mvo5:
> So I created a prototype in UBP: [osbuild/blueprint-schema#27](https://github.com/osbuild/blueprint-schema/pull/27)
> 
> My conclusion is: this optional type is **not a good solution** because although it does its job, every structure that uses this type looses both [comparability and order](https://go.dev/ref/spec#Comparison_operators). This might look like an inconvenience for now, but it can render as a major blocker in the future.
> 
> In practice, once this type is in use, any comparison like `b1 == b2` will fail with `Optional[string] is not a comparable Go type`. Also, any attempt to order blueprints will end up similarly, like `b1 < b2`. I already did hit blockers with `mergo` library for automatic struct merging and Googles `cmp` helper functions for comparing types and printing nice diffs in tests.
> 
> For this reason, I suggest to avoid this and to bring amount of nil pointer checks down in any code, just remove pointers at places where the are unnecessary. The UBP nicely shows that they can be dropped for most of the "leaf" nodes and with Go 1.24 `omitzero` they can actually be dropped from something like 90% of all fields.

Thanks for exploring this! The lack of comparability is unfortunate indeed, I'm not convinced that this is a show-stopper though, I think the initial example shows that in the common case a concept like this leads to shorter code and less risk of nil pointer dereferncing (and hence panic()s). I'm even less worried about lack of ordering TBH.

lzap:
For the record, while I think this wrapper is not a good fit for UBP where there are not many pointers and after implementing `omitzero` from Go 1.24 for struct types, there will be literally under a dozen of pointers for the whole structure, it might turn useful for images blueprint or other blueprints. My intention was not to block this, tho, I would like to take a moment and explain why I do not like this at all.

First of all, I want to summary the motivation to get us on the same page:

```go
if hostname := c.GetHostname(); hostname != nil {
 osc.Hostname = *hostname
} else {
 osc.Hostname = "localhost.localdomain"
}
```

Okay, the dereference should be done in the method itself:

```go
osc.Hostname = c.GetHostnameOrEmpty()
```

Ideally [without the](https://go.dev/doc/effective_go#Getters) `Get` prefix, something like `HostnameOr` would do.

Now, I am not big fan of using getters for all the fields even for trivial ones if there is no additional data processing. Then this approach cannot be used. Easy help, there is an existing function available in the standard library called `cmp.Or` that does exactly that but for values (not for pointers). It is super easy to write such function, in fact UBP use it all the time, then it becomes:

```go
osc.Hostname = ptr.Or(c.Hostname)
```

The main drawback of this wrapper is that it is just a syntactic sugar around such function like `ptr.Or` with one thing to consider - for nested structs it can be little bit more "readable". Example, instead of:

```go
if e.from.Network == nil || e.from.Network.Firewall == nil || len(e.from.Network.Firewall.Services) == 0 {
 return nil
}
```

One would write something like:

```go
type Struct struct {
	Substruct ptr.Ref[Struct]
	Name      string
}

name := ptr.New(Struct{}).Get().Substruct.Get().Substruct.Get().Name
```

I gotta say I do not like this at all, this is not Go, this is confusing piece of code that goes against Go readability promise. As original author Rob Pike like to say: Go is not neither JavaScript nor Rust. If you want something else, go for it. Now that does not mean that Go is not evolving, but features are carefully considered (iterators, ehm).

And the great news is that `omitzero` from Go 1.24 solves this particular problem, there is no reason to use pointers to structs for unmarshaling and thus have this problem in the first place. One particular reason, to avoid empty fields when marhshalling, will be solved after we upgrade to 1.24.

In fact, for UBP we can see how much work is it to remove all pointers from nested structs and how well it works. If there are not any blockers, why don't we apply the same thing in this repo.

To sum it up:

* Not useful for fields with getters because dereferencing can be done there.
* Can be replaced with a simple function in a `ptr` or `common` package for primitive type fields.
* Nested structs should not be pointers in the first place and Go 1.24 will help with that

Finally, as I said it breaks comparability and ordering which turns out to be blocker for using one of the best assert package `cmp`. Although there is a possibility to implement own comparing function, it cannot be done for this type because `bool` is not a comparable type therefore the `cmp.Compare` interface cannot be even implemented.

mvo5:
@lzap Thanks for your summary - it seems you really do not like this which is fine :) I will not introduce something this controversial then.

Fwiw, looking at https://github.com/osbuild/blueprint-schema/blob/main/pkg/ptr/any.go I'm not sure we are this far apart, we want convenient ways to work with pointers, the `ptr` api works directly on pointers with the risk of de-referening nils and panic()s. The `Option` thing never exposes the underlying pointer eliminating an entire class of errors via the type system. That seems valuable to me on its own. But *shrug* I guess its not go and we have to agree to disagree about the usefulness (and FTR I think for what we do with types a stronger language like rust would help us).

If we go with `ptr` I would like to at least chat about the API a bit, some of the naming feels a bit idiosyncratic and I wonder if its just me (also having tests for that code would be nice, even if the code is simple enough, having tests is also a form of documentation)

lzap:
I just happened to put the optional type into the `ptr` package, maybe it does not belong in there, I just felt like this is just a wrapper around pointer. Not sure why the go-optional project uses more memory heavy slice, I do not see any benefit there.

Blueprint schema needs review, that is to be done. I am okay with renaming.

Let me finish the convertor and then do a pointer removal with `omitempty` to see if that is something that will work. If not, we would resume this idea and give up on ordering or comparibility.

lzap:
For the record, complete pointer-less UBP patch should solve this issue: https://github.com/osbuild/blueprint-schema/pull/47

internal/types/option.go
```diff
@@ -0,0 +1,132 @@
...
+	if o.IsNone() {
+		var defaultValue T
+		return defaultValue
+	}
+	return o[value]
+}
+
+// TakeOr returns the actual value if the Option has a value.
+// On the other hand, this returns fallbackValue.
+func (o Option[T]) TakeOr(fallbackValue T) T {
```
supakeen, line 83: If we want to be 'like Rust' for people to recognize the idioms this should be called `UnwrapOr`.

internal/types/option.go
```diff
@@ -0,0 +1,132 @@
...
+	return o != nil
+}
+
+// XXX: this probably needs a different name to like UnwrapOrDefault()
+// or something
+//
+// Unwrap returns the value regardless of Some/None status.
+// If the Option value is Some, this method returns the actual value.
+// On the other hand, if the Option value is None, this method returns the *default* value according to the type.
+func (o Option[T]) Unwrap() T {
```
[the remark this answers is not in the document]
supakeen: I am not against it, though I do feel that it's not idiomatic for Go code which might make (other) Go programmers a bit confused :)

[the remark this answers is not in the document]
lzap: What is the added value when it panics? Go runtime will panic as well on `nil` dereference and no syntactic sugar is needed for that.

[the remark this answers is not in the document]
supakeen: This PR tries to avoid panicking but is using a Rust-name that would panic; I was suggesting to change the name to the Rust-equivalent that doesn't panic ;)

mvo5:
Closing this (for real now) - in favor of https://github.com/oapi-codegen/nullable which seems to get more buy-in from the group.

PR #861, opened 2024-08-14
Remove vendored dependencies

achilleas-k:
For a lot of our projects, vendoring is necessary for packaging.  This library is never packaged directly, so vendoring is not necessary.

Removing vendored dependencies reduces the size of the repository, speeding up checkouts, especially shallow ones, and makes PRs that include dependency changes in addition to code changes much easier to review.

The main (if not only) downside is that it makes it impossible to build and test in disconnected environments, which is not a requirement we have.


thozza (approved):
:fire: :fire: :fire: :fire: :fire:

mvo5 (approved):
:heart:

supakeen (commented):
🗑️

PR #862, opened 2024-08-16
manifest: extract tomlPkgsFor() helper and fix ordering

mvo5:
The current code will install python3-toml on fedora and also
on rhel11 (once it comes out) because that is the default for
unhandled distros. With `tomlib` in the stdlib this is no
longer needed.

[draft as it has no test(s)]

### DIFF
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -163,7 +163,6 @@ type OS struct {
 	// OSTreeParent source spec (optional). If nil the new commit (if
 	// applicable) will have no parent
 	OSTreeParent *ostree.SourceSpec
-
 	// Enabling Bootupd runs bootupctl generate-update-metadata in the tree to
 	// transform /usr/lib/ostree-boot into a bootupd-compatible update
 	// payload. Only works with ostree-based images.
@@ -289,6 +288,25 @@ func (p *OS) getContainerSources() []container.SourceSpec {
 	return p.OSCustomizations.Containers
 }
 
+func tomlPkgsFor(distro Distro) []string {
+	switch distro {
+	case DISTRO_EL7:
+		// nothing needs toml in rhel7
+		panic("no support for toml on rhel7")
+	case DISTRO_EL8:
+		// deprecated, needed for backwards compatibility (EL8 manifests)
+		return []string{"python3-pytoml"}
+	case DISTRO_EL9:
+		// older unmaintained lib, needed for backwards compatibility
+		return []string{"python3-toml"}
+	default:
+		// No extra package needed for reading, on rhel10 and
+		// fedora as stdlib has "tomlib" but we need tomli-w
+		// for writing
+		return []string{"python3-tomli-w"}
+	}
+}
+
 func (p *OS) getBuildPackages(distro Distro) []string {
 	packages := p.platform.GetBuildPackages()
 	if p.PartitionTable != nil {
@@ -315,14 +333,7 @@ func (p *OS) getBuildPackages(distro Distro) []string {
 
 	if len(p.OSCustomizations.Containers) > 0 {
 		if p.OSCustomizations.ContainersStorage != nil {
-			switch distro {
-			case DISTRO_EL8:
-				packages = append(packages, "python3-pytoml")
-			case DISTRO_EL10:
-				// no extra package needed, stdlib has "tomli"
-			default:
-				packages = append(packages, "python3-toml")
-			}
+			packages = append(packages, tomlPkgsFor(distro)...)
 		}
 		packages = append(packages, "skopeo")
 	}
@@ -332,14 +343,7 @@ func (p *OS) getBuildPackages(distro Distro) []string {
 	}
 
 	if p.BootcConfig != nil {
-		switch distro {
-		case DISTRO_EL8:
-			packages = append(packages, "python3-pytoml")
-		case DISTRO_EL10:
-			// no extra package needed, stdlib has "tomli"
-		default:
-			packages = append(packages, "python3-toml")
-		}
+		packages = append(packages, tomlPkgsFor(distro)...)
 	}
 
 	return packages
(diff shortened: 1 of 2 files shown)


pkg/manifest/os.go
```diff
@@ -287,6 +287,21 @@ func (p *OS) getContainerSources() []container.SourceSpec {
 	return p.OSCustomizations.Containers
 }
 
+func tomlPkgsFor(distro Distro) []string {
+	switch distro {
+	case DISTRO_EL7:
+		// Supported by upstream but superseeded by tomllib in stdlib
+		return []string{"python3-toml"}
```
mvo5, line 294: This is a slightly weird one - I added it because it was caught by the old codes "default" but I'm not actually sure this works.

pkg/manifest/os.go
```diff
@@ -287,6 +287,21 @@ func (p *OS) getContainerSources() []container.SourceSpec {
...
+	case DISTRO_EL8:
+		// old, unmaintained lib
+		return []string{"python3-pytoml"}
+	default:
+		// No extra package needed, on rhel10 and
+		// fedora. stdlib has "tomlib"
+		return nil
+	}
+}
+
```
achilleas-k, line 304: I think we have a pre-existing issue here that was never caught, because we don't cover it with a build config.

Both stages that trigger this package selection, `org.osbuild.containers.storage.conf` and `org.osbuild.bootc.install.config`, need to write toml files.  In EL10, this will require `python3-tomli-w`.

pkg/manifest/os.go
achilleas-k: We never covered the EL7 case with toml explicitly because there's no scenario where it can be needed.  None of the image types we build for RHEL 7 will ever need to write or read toml (`org.osbuild.containers.storage.conf` is only ever used in ostree-based images and `org.osbuild.bootc.install.config` is very new), and we don't support running osbuild on RHEL 7 either.

I checked to make sure and there's no `python3-toml` available for EL7 rel-eng, which is what we use to get python3 and its modules.  It does have `python-pytoml` (python2) in its main repos, which of course is no good to us.

I think in this case it makes more sense to panic (with a note that it should be an error, once we get around to replacing all panics with errors) if any image definition tries to use toml in EL7.

achilleas-k (changes requested):

achilleas-k:
It might be a good idea to bump the osbuild commit ID to `5401e7a628ec2653b33a17053521cfd257b4cc40`, the HEAD of the branch for https://github.com/osbuild/osbuild/pull/1851, in this PR draft or opening another one, as an extra verification that our image definitions actually work with all the toml changes and general madness.

pkg/manifest/os.go
mvo5: I updated this now to reflect reality (hopefully) a bit better. Still need to see if I can write a small test but it should have the right packages and comments now.

pkg/manifest/os.go
mvo5: Thanks, I updated this too now.

achilleas-k (changes requested):
Typo

pkg/manifest/os.go
```diff
@@ -287,6 +287,25 @@ func (p *OS) getContainerSources() []container.SourceSpec {
 	return p.OSCustomizations.Containers
 }
 
+func tomlPkgsFor(distro Distro) []string {
+	switch distro {
+	case DISTRO_EL7:
+		// nothing needs toml in rhel7
+		panic("no supprot for toml on rhel7")
```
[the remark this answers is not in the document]
mvo5: thanks, fixed

achilleas-k (approved):

pkg/manifest/os.go
```diff
@@ -287,6 +286,25 @@ func (p *OS) getContainerSources() []container.SourceSpec {
...
+	case DISTRO_EL7:
+		// nothing needs toml in rhel7
+		panic("no support for toml on rhel7")
+	case DISTRO_EL8:
+		// deprecated, needed for backwards compatibility (EL8 manifests)
+		return []string{"python3-pytoml"}
+	case DISTRO_EL9:
+		// older unmaintained lib, needed for backwards compatibility
+		return []string{"python3-toml"}
+	default:
```
mvo5, line 300: The CI fails right now https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/7618875295 with:
```
  File "/run/osbuild/bin/org.osbuild.bootc.install.config", line 6, in <module>
    import toml
ModuleNotFoundError: No module named 'toml'
During handling of the above exception, another exception occurred:
  File "/run/osbuild/bin/org.osbuild.bootc.install.config", line 8, in <module>
    import pytoml as toml
ModuleNotFoundError: No module named 'pytoml'
⏱  Duration: 0s
```
so this needs to be behaving like the previous code, at least fedora must return "python3-toml" for now

achilleas-k: yeah, let's keep the build roots identical for now.  Once osbuild/osbuild#1851 is merged and released, we can change our package selection and update the osbuild dependency to use the new module.

mvo5: It's been a while and the osbuild PR is merged/released, should we merge this now too?

achilleas-k (approved):
Good stuff.  Thank you!

For reference, I had a look at the package selection diff this PR introduces and the only change is that Fedora iot commits/containers that embed containers now use the `python3-tomli-w` library instead of `python3-toml`.

pkg/manifest/os.go
```diff
@@ -289,6 +288,25 @@ func (p *OS) getContainerSources() []container.SourceSpec {
...
+		// No extra package needed for reading, on rhel10 and
+		// fedora as stdlib has "tomlib" but we need tomli-w
```
supakeen, line 304: `tomllib` (double L) but nitpick :)

achilleas-k:
Some job on the merge queue got stuck, didn't even start, just appears in a paused/waiting state.  Weird!

pkg/manifest/os.go
achilleas-k: Should we merge anyway?

supakeen: Yes please.

PR #866, opened 2024-08-19
distro/rhel10: do not exclude nss package

evgenyz:
The package is required for openscap-scanner (OpenSCAP hardening). OpenSCAP package switched to NSS crypto backend in RHEL10.

See: https://issues.redhat.com/browse/RHEL-49502


thozza (approved):
LGTM

PR #869, opened 2024-08-20
[RHEL-10] Add `gce` and `image-installer` image types (COMPOSER-2193)

thozza:
SSIA.

The image types are based on RHEL-9, with their package set adjusted for RHEL-10.

The GCE image temporarily uses cloud-init, because there are no GCP guest tools builds for el10 / c10s yet.

On top of that I reported a new issue for GPG keys used by Google, which makes them unusable on el10 / c10s - https://issuetracker.google.com/issues/360905189. And while I was at it and was testing things, I reverted the disabled GPG check for Google repos on el9, because the issue that was blocking it there has been resolved.

Cross-tested in composer in https://github.com/osbuild/osbuild-composer/pull/4314. At the time of edit, all newly added tests passed, except for image-installer on c10s, which got stuck. I rescheduled it, but el10 test passed.


achilleas-k (approved):
Splendid!

LGTM

kingsleyzissou (approved):

PR #870, opened 2024-08-20
Bump VERSION_RAWHIDE to 42

AdamWill:
41 branched, so Rawhide's associated version number is now 42.


achilleas-k:
Thanks!

thozza (approved):

AdamWill:
Can we get this deployed? Because it isn't, the Workstation live images built for Fedora 41 still have anaconda-webui, which they should not.

PR #872, opened 2024-08-21
osbuild: add support for `LVM` to `GenBootupdDevicesMounts()`

mvo5:
This is required for `bootc-image-builder` to support LVM. It works by adding a new `Detectpv` option to the `org.osbuild.lvm2.lv` device and use it when generating the bootc/bootupd mounts.

This way all mounts can come from the same underlying loop device and bootc/bootupd can install the bootloader there. See also https://github.com/osbuild/osbuild/pull/1861


mvo5:
Closing in favor of the better `vg_partnum` approach

PR #876, opened 2024-08-26
osbuild: add support for `LVM` to `GenBootupdDevicesMounts()`

mvo5:
This is required for `bootc-image-builder` to support LVM. It works by using the new `vg_partition` option to the `org.osbuild.lvm2.lv` device and use it when generating the bootc/bootupd mounts.

This way all mounts can come from the same underlying loop device and bootc/bootupd can install the bootloader there. See also https://github.com/osbuild/osbuild/pull/1867

### DIFF
--- a/pkg/osbuild/bootupd_stage.go
+++ b/pkg/osbuild/bootupd_stage.go
@@ -89,7 +89,7 @@ func genMountsForBootupd(source string, pt *disk.PartitionTable) ([]Mount, error
 			continue
 		}
 
-		// TODO: support things like LVM here via supporting "disk.Container"
+		// TODO: support things like LUKS here via supporting "disk.Container"?
 		switch payload := part.Payload.(type) {
 		case disk.Mountable:
 			mount, err := genOsbuildMount(source, payload)
@@ -107,6 +107,20 @@ func genMountsForBootupd(source string, pt *disk.PartitionTable) ([]Mount, error
 				mount.Partition = common.ToPtr(idx + 1)
 				mounts = append(mounts, *mount)
 			}
+		case *disk.LVMVolumeGroup:
+			for i := range payload.LogicalVolumes {
+				lv := &payload.LogicalVolumes[i]
+				mountable, ok := lv.Payload.(disk.Mountable)
+				if !ok {
+					return nil, fmt.Errorf("expected LV payload %+[1]v to be mountable, got %[1]T", lv.Payload)
+				}
+				mount, err := genOsbuildMount(lv.Name, mountable)
+				if err != nil {
+					return nil, err
+				}
+				mount.Source = lv.Name
+				mounts = append(mounts, *mount)
+			}
 		default:
 			return nil, fmt.Errorf("type %T not supported by bootupd handling yet", part.Payload)
 		}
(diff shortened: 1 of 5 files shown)


pkg/osbuild/lvm2_lv_device.go
```diff
@@ -5,6 +5,8 @@ package osbuild
 type LVM2LVDeviceOptions struct {
 	// Logical volume to activate
 	Volume string `json:"volume"`
+	// The partitin the volume group is located on
```
achilleas-k, line 8:
```suggestion
	// The partition the volume group is located on
```

pkg/osbuild/lvm2_lv_device.go
```diff
@@ -5,6 +5,8 @@ package osbuild
...
+	VGPartnum *int `json:"vg_partnum,omitempty"`
```
achilleas-k, line 9: Can we get away with a concrete `int` (or `uint`) here since partition numbers are never 0 or would we rather not rely on 0 being "empty"?

achilleas-k (commented):
Can we break this down into multiple commits?  At least 2, but possibly 3 (if tests are done separately):
- Update of options in `lv2_lv_device.go` (with tests)
- Changes to `bootupd_stage.go` for supporting LVM.

It's not really important, but might make it easier to follow the history.

pkg/osbuild/lvm2_lv_device.go
mvo5: Yeah, I think that would work. I'm not sure about it though, I dislike the pointer but I do like that it's explicit about being optional (and here uniform/consistent with the rest of the way we flag things as optional) - I wonder if https://github.com/osbuild/images/pull/860 would help here and make it nicer?

pkg/osbuild/bootupd_stage.go
```diff
@@ -107,6 +107,20 @@ func genMountsForBootupd(source string, pt *disk.PartitionTable) ([]Mount, error
 				mount.Partition = common.ToPtr(idx + 1)
 				mounts = append(mounts, *mount)
 			}
+		case *disk.LVMVolumeGroup:
+			for i := range payload.LogicalVolumes {
+				lv := &payload.LogicalVolumes[i]
+				mountable, ok := lv.Payload.(disk.Mountable)
+				if !ok {
```
ondrejbudai, line 114: I suggest making this an error to prevent surprises if someone actually puts a non-mountable payload into an LV.

pkg/osbuild/bootupd_stage.go
```diff
@@ -130,6 +143,26 @@ func GenBootupdDevicesMounts(filename string, pt *disk.PartitionTable, pltf plat
 			},
 		},
 	}
+	for partNum0, part := range pt.Partitions {
```
ondrejbudai, line 146: I don't understand the `0` suffix... Is that to make it obvious that the index starts at 0? I suggest putting a little comment here (or before `partNum0 + 1`) to make it obvious.

ondrejbudai (commented):
Awesome! Two suggestions inline.

pkg/osbuild/bootupd_stage.go
mvo5: It was meant as a little reminder that zero-indexed but we need a one-indexed result but if it's more confusing than helping I remove it pagain and just call it `idx` :)

ondrejbudai: Then my understanding was correct! :) I suggest adding an explicit reminder via a comment then. :)

pkg/osbuild/lvm2_lv_device.go
achilleas-k: Let's keep the pointer then and push forward with the `types.Option` PR.

achilleas-k (approved):
Ship it!

PR #878, opened 2024-08-27
Create subscription service and activation key file with ISO kickstart for edge installers (HMS-3814)

achilleas-k:
This PR adds the ability to use subscription options with the edge-installer image type.  Subscription activation in the edge-installer works in a similar way to how it worked in the Edge service, by creating a first-boot systemd service during the kickstart `%post`.

In #627 we changed the way we create the service and key file to be able to remove the credentials (activation key) from the system after the system is registered.  However, as noted on that PR:
> [...] on ostree-based systems, putting the environment file in `/etc` prior to building the commit and then `rm`ing it on boot means that the file remains in `/usr/etc`, in the base system, until a system update removes it.

The changes here address that issue by creating the files during installation of the system, so they are not part of the immutable base system.

A new pipeline is added called `subscription`.  The subscription pipeline is a very simple pipeline that creates files (services, keyfile) needed to subscribe a system on boot.  Usually these files are created in an OS tree when the subscription options are used.  The purpose of the new pipeline is to create the files in a clean tree so that other pipelines can copy them out into their own tree.  The bootiso tree pipeline then copies the files onto the ISO under a directory on the root (`/subscription`) and adds commands to the kickstart file so that the files are copied into the OS tree installation time when deploying an ostree-based system.  This avoids making the files (which include activation keys) part of the immutable base system so that they can be removed after activation without needing to update the base system.

This change is mostly for parity with the old Edge management service and to satisfy the requirements in the linked ticket.  I don't expect this will be used a lot outside of that scenario.  This is also another scenario where the base image and installer configuration need to be compatible: the systemd service created at installation requires certain packages to be installed in the base system.  In normal cases we add these packages when subscription options are enabled.  The edge service always adds these packages to the base systems it builds, so for that purpose, there should be no issues.

---

~~**DRAFT**: I'd like to add some unit tests for the new functions, but I'm opening the PR to get reviews for the general idea in the meantime.~~


thozza (commented):
So far things look good from my PoV. The approach is IMHO OK.

I would propose to move all subscription based functions and structs from `pkg/manifest/os.go` to `pkg/manifest/subscription.go`. Technically, it is all part of the same package, but having them in `os.go` is a bit counter-intuitive.

I was thinking if it would make sense to use the `Subscription` pipeline also for non-ostree images. Basically, not adding the stages to the OS pipeline, but optionally modify the tree with the `Subscription` pipeline. But that would probably complicate things a lot, because all subsequent pipelines, which are partitioning / wrapping the tree into the final format, would need to use the correct pipeline name as an input.

achilleas-k:
> So far things look good from my PoV. The approach is IMHO OK.
> 
Yay!

> I would propose to move all subscription based functions and structs from `pkg/manifest/os.go` to `pkg/manifest/subscription.go`. Technically, it is all part of the same package, but having them in `os.go` is a bit counter-intuitive.
> 

Good idea!

> I was thinking if it would make sense to use the `Subscription` pipeline also for non-ostree images. Basically, not adding the stages to the OS pipeline, but optionally modify the tree with the `Subscription` pipeline. But that would probably complicate things a lot, because all subsequent pipelines, which are partitioning / wrapping the tree into the final format, would need to use the correct pipeline name as an input.

The only reason I use a separate pipeline here is because I don't want the files to be created at their standard locations.  Technically, it wouldn't hurt to create them on the ISO under `/etc` and copy them from there, which is what would happen if we ran the stages in the `bootiso-tree` pipeline, but I considered that having a directory called `/subscription` on the ISO would be less confusing for anyone poking around than a directory called `/etc`.

Now that I think about it, it might not be a bad idea to add a `README` file in there explaining what the directory is for.

bcl (commented):
Looking good to me

achilleas-k:
Tests and README added.  PR is ready!

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -638,6 +636,19 @@ func (p *AnacondaInstallerISOTree) makeKickstartStages(stageOptions *osbuild.Kic
 
 		}
 		hardcodedKickstartBits += makeKickstartSubscriptionPost(subscriptionPath, systemPath)
+
+		// include a readme file on the ISO in the subscription path to explain what it's for
+		subscriptionReadme, err := fsnode.NewFile(
+			filepath.Join(subscriptionPath, "README"),
+			nil, nil, nil,
+			[]byte(`Subscription services and credentials
```
bcl, line 643: I'm reluctant to suggest this :) But embedding a README's text in the code seems messy. But images really doesn't have a good common place to pull such a file from since it's a library and could end up anywhere.

But a README file could be added to the source tree, and it could be embedded in the build using https://pkg.go.dev/embed, allowing for easy editing of the file as a file.

But maybe I'm overthinking things...

achilleas-k: Yeah, you're not wrong.  It's quite weird how I added it here.
We do similar things with comments in some config files, but that's a bit different.

achilleas-k: @bcl Thought about this a bit and I'd like to use embed but I think that should be part of a bigger change that includes a lot of similar changes, like pulling out hardcoded kickstart blobs.

If you don't mind, can we merge this as-is and make a PR with embedding as a follow-up?

pkg/manifest/subscription.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		p.Directories = append(p.Directories, subDirs...)
+		p.Files = append(p.Files, subFiles...)
+
+		for _, stage := range append(osbuild.GenDirectoryNodesStages(p.Directories), osbuild.GenFileNodesStages(p.Files)...) {
+			if stage.Type == "org.osbuild.chown" {
+				// chown requires chrooting into the tree to run it, which
+				// wont work for this pipeline, so skip it
+				continue
+			}
```
thozza, line 58: I have mixed feelings about this.

I have two suggestions:

1. Not set the `user` and `group` on files created by `subscriptionService()`, which will mean that there are no `chown` stages generated later. Keeping them `nil` will make them be owned by `root` implicitly AFAIK.
2. Extend the `Gen*NodesStages()` to optionally skip generation of `chown` stage.

I guess I could learn to live with the code as is in case you'd see some downside in both of my suggestions.

thozza (commented):
I've added one minor comment. The rest look good.

pkg/manifest/subscription.go
achilleas-k: Done.  Thanks!

achilleas-k:
I did some rebase shuffling to make the latest changes easier to see (without the rebase on `main`).
These are the changes I made today: https://github.com/osbuild/images/compare/628493db1d4c4fbf3eedc92f7ea733419ae53f57..3e70a2867d39679ba33d7f6373113d044024bdb9

pkg/manifest/export_test.go
```diff
@@ -1,3 +1,12 @@
...
+var (
+	FindStage               = findStage
+	SubscriptionService     = subscriptionService
+	RunInsightsClientOnBoot = runInsightsClientOnBoot
+	InsightsClientDropin    = insightsClientDropin
+)
+
+type (
+	SubscriptionServiceOptions = subscriptionServiceOptions
+)
```
thozza, line 12: Why not just use `package manifest` in the tests? That way, the unit test will be able to use unexported structs and functions.

pkg/manifest/subscription_test.go
```diff
@@ -0,0 +1,425 @@
...
+
+	for name := range testCases {
+		tc := testCases[name]
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			stage, dirs, files, services, err := manifest.SubscriptionService(tc.subOpts, tc.srvcOpts)
+			assert.NoError(err)
+			assert.Equal(stage, tc.expectedStage)
+			assert.Equal(dirs, tc.expectedDirs)
+			assert.Equal(files, tc.expectedFiles)
```
thozza, line 390: It may be good to extend the test to check that none of the files and dirs have `User` and `Group` set to `nil`.

thozza (commented):
Thanks for the changes, the code looks much better now. Unfortunately, I found some more things to comment on :innocent:

pkg/manifest/subscription.go
```diff
@@ -0,0 +1,218 @@
...
+	// NOTE(akoutsou): drop-in files can normally be created with the
+	// org.osbuild.systemd.unit stage but the stage doesn't support
+	// all the options we need. This is a temporary workaround
+	// until we get the stage updated to support everything we need.
+	icDropinFilepath, icDropinContents := insightsClientDropin()
+	icDropinDirectory, err := fsnode.NewDirectory(filepath.Dir(icDropinFilepath), nil, "root", "root", true)
+	if err != nil {
+		return nil, nil, err
+	}
+	icDropinFile, err := fsnode.NewFile(icDropinFilepath, nil, "root", "root", []byte(icDropinContents))
```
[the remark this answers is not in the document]
achilleas-k: Ugh.  Of course.  Thanks!

pkg/manifest/subscription_test.go
achilleas-k: Good idea!

pkg/manifest/export_test.go
achilleas-k: yeah, I keep flip-flopping on this.  Sometimes it's nice to isolate the things that are being tested by exporting them, other times I think it doesn't give us much.  I think I'm going to revert in this case.

achilleas-k: Amended!

thozza (approved):
:rocket:

PR #879, opened 2024-08-27
Rename image(s) cmd flags to type(s) and add checkpoints to build

achilleas-k:
These are a couple of changes I've been fiddling with while working on other stuff.
1. I think `--type` is a better name for the image type flag than `--image` for `cmd/build` and the same for the plural form for `cmd/gen-manifests` and `cmd/list-images`.
2. I caught myself doing checkpointing by editing the command or generating manifests and running them separately sometimes when experimenting with builds.  Let's just add a `--checkpoints` flag that gets passed down to osbuild.

We might add an `--osbuild-options` flag instead and use that to pass stuff down to osbuild in the future, but let's keep this short for now.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -191,7 +191,7 @@ func run() error {
 	// image selection args
 	var distroName, imgTypeName, configFile string
 	flag.StringVar(&distroName, "distro", "", "distribution (required)")
-	flag.StringVar(&imgTypeName, "image", "", "image type name (required)")
+	flag.StringVar(&imgTypeName, "type", "", "image type name (required)")
 	flag.StringVar(&configFile, "config", "", "build config file (required)")
 
 	flag.Parse()
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -188,6 +188,10 @@ func run() error {
 	flag.StringVar(&osbuildStore, "store", ".osbuild", "osbuild store for intermediate pipeline trees")
 	flag.StringVar(&rpmCacheRoot, "rpmmd", "/tmp/rpmmd", "rpm metadata cache directory")
 
+	// osbuild checkpoint arg
+	var checkpoints cmdutil.MultiValue
+	flag.Var(&checkpoints, "checkpoints", "comma-separated list of pipeline names to checkpoint (passed to osbuild --checkpoint)")
+
 	// image selection args
 	var distroName, imgTypeName, configFile string
 	flag.StringVar(&distroName, "distro", "", "distribution (required)")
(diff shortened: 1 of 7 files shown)


kingsleyzissou:
> or generating manifests and running them separately sometimes when experimenting with builds

I definitely don't do this /s

cmd/build/main.go
```diff
@@ -188,6 +188,10 @@ func run() error {
 	flag.StringVar(&osbuildStore, "store", ".osbuild", "osbuild store for intermediate pipeline trees")
 	flag.StringVar(&rpmCacheRoot, "rpmmd", "/tmp/rpmmd", "rpm metadata cache directory")
 
+	// osbuild checkpoint arg
+	var checkpoints cmdutil.MultiValue
+	flag.Var(&checkpoints, "checkpoints", "comma-separated list of pipeline names to checkpoint (passed to osbuild --checkpoint)")
```
kingsleyzissou, line 193: Love this

internal/cmdutil/multivalue.go
```diff
@@ -0,0 +1,47 @@
...
+// be used to validate the values and resolve globs.
+type MultiValue []string
+
+func (mv *MultiValue) String() string {
+	return strings.Join(*mv, ", ")
+}
+
+func (mv *MultiValue) Set(v string) error {
+	split := strings.Split(v, ",")
+	*mv = split
```
kingsleyzissou, line 20: Genuine question, what is the benefit of splitting this into two lines and not assigning directly?

kingsleyzissou (commented):
I think we need to update `./cmd/list-images` too

internal/cmdutil/multivalue.go
achilleas-k: I don't think there's a difference... might've been left-over from when I was first writing it and it just stayed there.

achilleas-k:
> I think we need to update `./cmd/list-images` too

Oh right ,forgot about that one.  Thanks!

achilleas-k:
Changed list-images too and updated the commit message.

internal/cmdutil/multivalue.go
achilleas-k: Squished.

kingsleyzissou (approved):
Sweet!

thozza (approved):
So you want me to learn to use new flag? No way! :exploding_head: 

Just kidding, all looks fine :sweat_smile:

schuellerf:
Would it make sense to return an error or deprecation warning for people using the "old" argument?

achilleas-k:
> Would it make sense to return an error or deprecation warning for people using the "old" argument?

I considered it but it's just too much fluff for something that's only used internally.

PR #880, opened 2024-08-28
disk: drop Entity.IsContainer()

ondrejbudai:
This method is actually only used in tests, and it's actually not
needed as we can just use a type assertion like everywhere else.

### DIFF
--- a/pkg/disk/btrfs.go
+++ b/pkg/disk/btrfs.go
@@ -25,10 +25,6 @@ func (b *Btrfs) EntityName() string {
 	return "btrfs"
 }
 
-func (b *Btrfs) IsContainer() bool {
-	return true
-}
-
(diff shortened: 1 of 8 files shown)


mvo5 (approved):
Thank you, this looks correct to me, I do not know the history here that well so would also an additional review from somene experienced like @achilleas-k

achilleas-k:
The historical context here was that at some point the Entity interface was only defined by this one method, before `Clone()` was added (very shortly after).  It was a bit of a useless method, but I think we kept it thinking it might be useful for quick checks, but as Ondrej mentions (and as we can see from the patch), checking if an Entity implements the `Container` interface is equivalent and doesn't require implementing a method that might incorrectly disagree with the actual implementation.

achilleas-k (approved):

PR #883, opened 2024-08-29
fix / growth when converting a pt to btrfs and cover it with a test

ondrejbudai:
The first commit ensures that when converting the root partition to btrfs, the enlarged size from a blueprint mountpoint is carried over to the created btrfs subvolume.

The second commit adds a test to check whether the blueprint customizations are correctly applied to the partition table. It verifies that the correct mountable is created and that all sizeables above the mountable have at least the minimum size specified in the blueprint.


bcl (approved):
Looks good to me, and the tests even pass!

mvo5 (approved):
Thank you

PR #885, opened 2024-08-30
blueprint: fix filesystem customization parsing

ondrejbudai:
In the filesystem customization, the toml tags clearly sets the `MinSize` field's key to `minsize`, this is also reflected in `bootc-image-builder` docs. However, the custom unmarshallers are actually wrong, they still refer to the old `size` name (see #825).

This PR fixes the unmarshallers, and adds more test coverage.

### DIFF
--- a/pkg/blueprint/blueprint_test.go
+++ b/pkg/blueprint/blueprint_test.go
@@ -23,16 +23,16 @@ version = "2.4.*"
 
 [[customizations.filesystem]]
 mountpoint = "/var"
-size = 2147483648
+minsize = 2147483648
 
 [[customizations.filesystem]]
 mountpoint = "/opt"
-size = "20 GB"
+minsize = "20 GB"
 `
 
 	var bp Blueprint
 	err := toml.Unmarshal([]byte(blueprint), &bp)
-	require.Nil(t, err)
+	require.NoError(t, err)
 	assert.Equal(t, bp.Name, "test")
 	assert.Equal(t, "/var", bp.Customizations.Filesystem[0].Mountpoint)
 	assert.Equal(t, uint64(2147483648), bp.Customizations.Filesystem[0].MinSize)
@@ -49,7 +49,7 @@ size = "20 GB"
 		}
 	  }`
 	err = json.Unmarshal([]byte(blueprint), &bp)
-	require.Nil(t, err)
+	require.NoError(t, err)
 	assert.Equal(t, bp.Name, "test")
 	assert.Equal(t, "/opt", bp.Customizations.Filesystem[0].Mountpoint)
 	assert.Equal(t, uint64(20*common.GiB), bp.Customizations.Filesystem[0].MinSize)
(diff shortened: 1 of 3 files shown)


pkg/blueprint/blueprint_test.go
```diff
@@ -23,16 +23,16 @@ version = "2.4.*"
 
 [[customizations.filesystem]]
 mountpoint = "/var"
-size = 2147483648
+minsize = 2147483648
```
mvo5, line 26: I assume there is no risk here that there are "wrong" blueprint tomls out there that we would break with this change?

pkg/blueprint/blueprint_test.go
```diff
@@ -23,16 +23,16 @@ version = "2.4.*"
...
 
 [[customizations.filesystem]]
 mountpoint = "/opt"
-size = "20 GB"
+minsize = "20 GB"
```
mvo5, line 30: Maybe we can also organize the test a little bit differently so that there is less risk of "drift" between json/toml parsing? I.e. make the blueprint in json/toml the same and after parsing compare that the two are identical.

I attached 
[0001-blueprint-extend-TestBlueprintParse-to-make-drift-le.patch.txt](https://github.com/user-attachments/files/16834790/0001-blueprint-extend-TestBlueprintParse-to-make-drift-le.patch.txt) with the idea.

pkg/blueprint/filesystem_customizations.go
```diff
@@ -24,17 +24,17 @@ func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
 		return fmt.Errorf("TOML unmarshal: mountpoint must be string, got %v of type %T", d["mountpoint"], d["mountpoint"])
 	}
 
-	switch d["size"].(type) {
+	switch d["minsize"].(type) {
```
mvo5, line 27: It seems part of the issue why this happend is that the `size`->`minsize` change happend in `FilesystemCustomization` but not in the unmarshal code. By adding a test that checks that marshal/unmarshal are symetric this can be prevented. I added a patch 
[0001-blueprint-add-marshal-unmarshal-symetry-test.patch.txt](https://github.com/user-attachments/files/16834969/0001-blueprint-add-marshal-unmarshal-symetry-test.patch.txt) with this idea.

pkg/blueprint/filesystem_customizations_test.go
```diff
@@ -1,13 +1,83 @@
...
 	"testing"
 
+	"github.com/BurntSushi/toml"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/osbuild/images/pkg/pathpolicy"
 )
 
+// Happy tests for unmarshallers are in blueprint_test.go
+func TestFilesystemCustomizationUnmarshalTOMLUnhappy(t *testing.T) {
```
mvo5, line 14: Thanks for adding those!

mvo5 (commented):
Thank you! This looks fine, thanks for fixing this. I was about to approve but there is the open question about backward compatiblity (I assume we do not need that as this is a bug but wanted to double check).

I would also encourage two small test tweaks that I added, happy to push them directly to this branch if people are okay with or alternatively push them as a followup :)

achilleas-k (approved):
LGTM.  I think we should merge this, rebase #886 to get it moving along, and add test improvements separately.  @mvo5 wdyt?

PR #886, opened 2024-08-30
add support for swap partitions and logical volumes

ondrejbudai:
```toml
[[customizations.filesystem]]
type = "swap"
minsize = "10 GiB"
```

now creates a swap partition (or a LV, the usual decision rules between a regular partition an LV apply).

---

After some pondering, I decided that swap is just `Filesystem` with type `swap` and a mountpoint forced/faked to `none`. Thus, I added a new `type` field into the filesystem customization. For now, it supports unset and swap only, but extending it should not be hard. ;)

However, swap is a weird mountpoint:
- It's not possible nor desirable to mount it during build time, so I have to filter it out when generating mounts.
- It's not possible to create a swap mountpoint as a btrfs subvolume, so I had to invent a new mechanism that finds a different container if the firstly selected fails. Example: By default, we create new filesystem as close as possible to the root mountpoint. So, if / is a btrfs subvolume, we make all extra mountpoints sibling subvolumes. However, it's not possible to create a swap subvolume, so in this case the code goes up and tries to create the swap partition as a sibling to the btrfs partition. So swap is created in this case as either a top-level partition, or as an LV (if you have btrfs inside an LV for some reason).

In other cases, it shares properties of other mountpoints:
- It must have an mkswap stage.
- It must be inserted into fstab.

So is `swap` actually a `Filesystem`? It's murky. I'm open to suggestions. The other alternative would be to define a new `Swap` entity that does not implement `Mountable`, but then we would need to have special cases for swap when generating `mkfs` stages and `fstab`. Also, the `MountpointCreator` mechanism would have to be separate (`SwapCreator`?). This sounds more convoluted to me.

One can also argue that it's not needed to define a new `type` in the filesystem customization, `mountpoint == swap` would also do the trick. However, I think that swap is rather a filesystem type (see e.g. fstab) than a mountpoint path.

TODO list (nothing feels blocking, though):
- The code generates an LV device for swap for the bootc stage. I ran out of time while investigating this, but I'm worried that we will need to restructure the code quite a bit to remove this. Nevertheless, it's not harmful, just pointless.
- I would love to see integration tests for this. Maybe Achilleas can help me with adding it into one of the test blueprints. (I ran out of time to do this before my PTO).
- I relaxed the unmarshal validation rules for both TOML and JSON a bit. I think they should be more strict again, e.g. a non-swap mountpoint must not have an empty path. However, this validation should probably live outside the unmarshaller.

Depends on:
- #885 
- osbuild/osbuild#1872

### DIFF
--- a/internal/testdisk/partition.go
+++ b/internal/testdisk/partition.go
@@ -22,6 +22,8 @@ func MakeFakePartitionTable(mntPoints ...string) *disk.PartitionTable {
 		case "/boot/efi":
 			payload.UUID = disk.EFIFilesystemUUID
 			payload.Type = "vfat"
+		case "swap":
+			payload.Type = "swap"
 		default:
 			payload.UUID = disk.FilesystemDataUUID
 		}
@@ -70,6 +72,15 @@ func MakeFakeBtrfsPartitionTable(mntPoints ...string) *disk.PartitionTable {
 				},
 			})
 			size += 100 * common.MiB
+		case "swap":
+			pt.Partitions = append(pt.Partitions, disk.Partition{
+				Start: size,
+				Size:  1 * common.GiB,
+				Payload: &disk.Filesystem{
+					Type: "swap",
+				},
+			})
+			size += 1 * common.GiB
 		default:
 			name := mntPoint
 			if name == "/" {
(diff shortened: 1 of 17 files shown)


achilleas-k:
Rebased on main since the first few commits were merged in #885.

achilleas-k:
Given how special and weird a swap partition is, I think trying to coerce it into the current model of a filesystem will cause a lot of headaches in the long run.  It will be something we'll have to constantly make exceptions for in bits of the code that deal with mountables, for example.  The more accurately we model the real world here, the lower the mental load will be needed when working with these parts of the code and the lower the chance for introducing bugs or allowing impossible scenarios.  A mountable is a mountable for a reason and if something doesn't fit the definition, it shouldn't implement the interface with exceptions.  I don't have the full answer here, but I think a special type for swap partitions will probably make our lives easier.

PR #889, opened 2024-09-01
blueprint: return nil instead of empty user slice in GetUsers()

pastalian:
Callers of GetUser() do not expect to receive an empty slice, which is causing unexpected condition handling.

https://github.com/osbuild/images/blob/e734c0c716d5d93d6daca1433491eb6e67d74945/pkg/distro/rhel/rhel9/options.go#L106-L107
https://github.com/osbuild/images/blob/e734c0c716d5d93d6daca1433491eb6e67d74945/pkg/distro/rhel/rhel9/options.go#L210-L211


achilleas-k (approved):
Thanks for this.  LGTM!

PR #890, opened 2024-09-02
cmd/otk-gen-partition-table: fix to be wrapped in a "tree" json

mvo5:
This is similar to https://github.com/osbuild/images/pull/827 but this time for the `otk-gen-partition-table` which was forgoten in this other PR :(


supakeen (approved):
Awesome, thanks.

achilleas-k (approved):

PR #892, opened 2024-09-02
cmdutil: print rng `TEST MODE...` warning to stderr

mvo5:
Printing on stdout interferes with the json output from `otk-gen-partition-table`.


achilleas-k (approved):
Thanks!  LGTM

Maybe it would be better if utility functions weren't printing at all, or were using a logger with warning/error levels.  But for now, let's merge this as is.

Issue #624, opened 2024-04-22
Consider excluding dracut-config-rescue in ostree-based images


achilleas-k:
Installing dracut-config-rescue will inevitably generate a secondary rescue kernel, which will make `rpm-ostree compose` fail with something like:
```
Preparing kernel
error: Finalizing rootfs: During kernel processing: Multiple vmlinuz- in usr/lib/ostree-boot, occurrences 'usr/lib/ostree-boot/vmlinuz-5.14.0-427.13.1.el9_4.aarch64' and 'usr/lib/ostree-boot/vmlinuz-0-rescue-ffffffffffffffffffffffffffffffff'
Traceback (most recent call last):
  File "/run/osbuild/bin/org.osbuild.ostree.preptree", line 176, in <module>
    r = main(args["tree"], args["options"])
  File "/run/osbuild/bin/org.osbuild.ostree.preptree", line 169, in main
    subprocess.run(["rpm-ostree", "compose", "postprocess",
  File "/usr/lib64/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['rpm-ostree', 'compose', 'postprocess', '/run/osbuild/tree', '/tmp/tmpx7i5hvm5.json']' returned non-zero exit status 1.
```

One solution would be to disable dracut in the rpm stage (which is an option that exists in osbuild's `org.osbuild.rpm` stage).
Alternatively, we can exclude `dracut-config-rescue` in ostree-based builds, but our excludes aren't considered when depsolving user packages.

PR #895, opened 2024-09-03
os: disable dracut for ostree based systems on rpm installation

mmartinv:
Installing the `minimal-environment` RPM group implies the installation
of `dracut-config-rescue` which makes `org.osbuild.ostree.preptree` to
fail because `rpm-ostree` detects multiple kernels within the tree
(the usual one and the rescue one). Disabling dracut during the rpm
installation prevents the rescue kernel/initrd to be generated and
makes `rpm-ostree` to finish the compose succesfully.

`rpm-ostree` runs dracut on his own during the compose so this doesn't
seem to affect the final result.

Resolves: #624

### DIFF
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -424,6 +424,12 @@ func (p *OS) serialize() osbuild.Pipeline {
 	if p.OSTreeRef != "" {
 		rpmOptions.OSTreeBooted = common.ToPtr(true)
 		rpmOptions.DBPath = "/usr/share/rpm"
+		// The dracut-config-rescue package will create a rescue kernel when
+		// installed. This creates an issue with ostree-based images because
+		// rpm-ostree requires that only one kernel exists in the image.
+		// Disabling dracut for ostree-based systems resolves this issue.
+		// Dracut will be run by rpm-ostree itself while composing the image.
+		// https://github.com/osbuild/images/issues/624
 		rpmOptions.DisableDracut = true
 	}
 	pipeline.AddStage(osbuild.NewRPMStage(rpmOptions, osbuild.NewRpmStageSourceFilesInputs(p.packageSpecs)))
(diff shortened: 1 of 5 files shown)


supakeen (dismissed):
Does what it needs to do; perhaps we could use a (boot) test to ensure that user customized systems that add `dracut-config-rescue` build after this.

@achilleas-k How do you feel having the check here, I could also see it living inside the RPM stage wrapper.

achilleas-k:
The edge AMI builds are boot tested in gitlab CI, so the base case here has already passed.
The `edge-ostree-pull-user` config for example is configured to build an `edge-ami`:
https://github.com/osbuild/images/blob/0a9d89e4a15c98fca524591612fac8d90dc18943/test/config-map.json#L89-L93
using the `edge-container` image type with the `empty` config as its base:
https://github.com/osbuild/images/blob/0a9d89e4a15c98fca524591612fac8d90dc18943/test/configs/edge-ostree-pull-user.json#L16-L19

In the CI run for this PR, the edge-container was built with dracut disabled: https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/7725683968#L1903
and the AMI that was built using it was successfully booted in AWS:
https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/7725901360

So we're covered for the base case in terms of testing, which is nice.  But since this PR was made to address a specific use case, we could add an extra test config that installs `@minimal-environment` (or `dracut-config-rescue` directly) to make sure the intended use case is also working as expected.

achilleas-k:
Added new build configs for this scenario.

supakeen (approved):

achilleas-k:
Messed something up with the configs.  Fixing.

supakeen (dismissed):

supakeen (approved):

PR #896, opened 2024-09-03
cmd/otk-*: do basic input validation to avoid crashing too easily

mvo5:
This commit is a lesson from https://github.com/osbuild/otk/pull/186 as it is right now extremly easy to crash the otk disk externals with bad inputs.

This commit adds basic input validation, it's probably a good idea to also go deeper into the library and make it less trusting about pointers but that is (slightly) orthogonal.

### DIFF
--- a/cmd/otk-make-fstab-stage/main.go
+++ b/cmd/otk-make-fstab-stage/main.go
@@ -20,6 +20,9 @@ func run(r io.Reader, w io.Writer) error {
 	if err := json.NewDecoder(r).Decode(&inp); err != nil {
 		return err
 	}
+	if err := inp.Tree.Validate(); err != nil {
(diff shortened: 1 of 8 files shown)


supakeen (commented):
Do we want this right now or do we want to postpone this to do the whole schema dance (which is probably going to happen in the future)?

mvo5:
> Do we want this right now or do we want to postpone this to do the whole schema dance (which is probably going to happen in the future)?

I don't see a downside in merging it now, we can expand `Validate()` in otkdisk as needed and the "frontends" are calling it in the right place already in this commit. But maybe I'm missing something?

supakeen (approved):

achilleas-k (approved):

PR #897, opened 2024-09-03
anaconda: enable SELinux for live-installer

### LINKED ISSUE #460: SELinux is in permissive mode for `live-installer`
I just hooked up openQA testing of the osbuild-built Fedora Workstation live image (the one from https://fedoraproject.org/wiki/Changes/FedoraWorkstationImageBuilder ), and it [found a bug right away](https://openqa.stg.fedoraproject.org/tests/3591526#step/base_selinux/7). (Finally, after about five years of having it, this test failed on something!)

After running an install from the image, on the installed system, SELinux is in permissive mode. By policy SELinux must be in enforcing mode by default for Fedora installs - this is in [the release criteria](https://fedoraproject.org/wiki/Basic_Release_Criteria#SELinux_configuration).

I wasn't sure where would be best to file this, chose this repo as it's where the existing "known issue" mentioned in the Change is filed.

The image tested was https://kojipkgs.fedoraproject.org/compose/branched/Fedora-40-20240220.n.0/compose/Workstation/x86_64/iso/Fedora-Workstation-Live-osb-40-20240220.n.0.x86_64.iso , the test just runs an install, boots the installed system, runs `getenforce` and checks the output.

supakeen:
The live installer build has been installing non-enforcing systems. This is due to a leftover from the image installer setting the system to permissive.

Make it so we apply the contexts and don't write a permissive config for the live installer.

Closes #460 which has been flying under my radar for entirely too long.

### DIFF
--- a/pkg/manifest/anaconda_installer.go
+++ b/pkg/manifest/anaconda_installer.go
@@ -84,6 +84,11 @@ type AnacondaInstaller struct {
 	// Uses the old, deprecated, Anaconda config option "kickstart-modules".
 	// Only for RHEL 8.
 	UseLegacyAnacondaConfig bool
+
+	// SELinux policy, when set it enables the labeling of the installer
+	// tree with the selected profile and selects the required package
+	// for depsolving
+	SElinux string
 }
 
 func NewAnacondaInstaller(installerType AnacondaInstallerType,
@@ -159,14 +164,24 @@ func (p *AnacondaInstaller) getBuildPackages(Distro) []string {
 		)
 	}
 
+	if p.SElinux != "" {
+		packages = append(packages, "policycoreutils", fmt.Sprintf("selinux-policy-%s", p.SElinux))
+	}
+
 	return packages
 }
 
 func (p *AnacondaInstaller) getPackageSetChain(Distro) []rpmmd.PackageSet {
 	packages := p.anacondaBootPackageSet()
+
 	if p.Biosdevname {
 		packages = append(packages, "biosdevname")
 	}
+
+	if p.SElinux != "" {
+		packages = append(packages, fmt.Sprintf("selinux-policy-%s", p.SElinux))
+	}
+
 	return []rpmmd.PackageSet{
 		{
 			Include:         append(packages, p.ExtraPackages...),
@@ -306,6 +321,13 @@ func (p *AnacondaInstaller) payloadStages() []*osbuild.Stage {
 
 	stages = append(stages, osbuild.NewSELinuxConfigStage(&osbuild.SELinuxConfigStageOptions{State: osbuild.SELinuxStatePermissive}))
 
+	// SElinux is not supported on the non-live-installers (see the previous
+	// stage setting SELinux to permissive. It's an error to set it to anything
+	// that isn't an empty string
+	if p.SElinux != "" {
+		panic("payload installers do not support SELinux policies")
+	}
+
 	if p.InteractiveDefaults != nil {
 		var ksUsers []users.User
 		var ksGroups []users.Group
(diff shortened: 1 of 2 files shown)


pkg/manifest/anaconda_installer.go
```diff
@@ -370,7 +370,9 @@ func (p *AnacondaInstaller) liveStages() []*osbuild.Stage {
 	dracutOptions.AddDrivers = p.AdditionalDrivers
 	stages = append(stages, osbuild.NewDracutStage(dracutOptions))
 
-	stages = append(stages, osbuild.NewSELinuxConfigStage(&osbuild.SELinuxConfigStageOptions{State: osbuild.SELinuxStatePermissive}))
+	stages = append(stages, osbuild.NewSELinuxStage(&osbuild.SELinuxStageOptions{
+		FileContexts: "etc/selinux/targeted/contexts/files/file_contexts",
```
thozza, line 374: I would suggest to rework the AnacondaInstaller pipeline a bit more, to be closer to how the `Os` pipeline handles SELinux context. The actual policy to be used is set in the pipeline options and then used when adding the stage 
https://github.com/osbuild/images/blob/0a9d89e4a15c98fca524591612fac8d90dc18943/pkg/manifest/os.go#L830

Consequently, it would be good to always make sure that the proper `selinux-policy-*` package is installed in the tree, similar to how it is done in https://github.com/osbuild/images/blob/0a9d89e4a15c98fca524591612fac8d90dc18943/pkg/manifest/os.go#L232-L234.

supakeen: Right I considered doing this the same as the OS pipeline but then felt that it didn't make much sense to configure this in the pipeline as it's non-changing and only used by a single image type.

Making sure `selinux-policy-x` is installed is something that should be done so that probably tips it the other way (make it configurable).

supakeen: New commit added that makes this now (similar to) the os pipeline; the value is hardcoded in the image but could be moved up even more later on if we wish (e.g. supporting `NoSELinux`).

It is a programming error to set a policy on the non-live installers.

supakeen:
Note that I could be convinced to move this to the same paradigm with `NoSELinux`, it might make sense for the live installer to be able to toggle it off on a higher level? It'd still be an error for the non-live installers.

The non-live installers could be done as a follow up but we have the 'problem' there that we need to ensure that the toggle makes sense and goes to the right place as the anaconda-tree is *not* the tree being installed to system so we'd have two SElinux policies there.

thozza (approved):
Thanks for the changes. This looks good enough for me :slightly_smiling_face:

achilleas-k (approved):
Thanks.  LGTM!

achilleas-k:
The CentOS Stream 10 unit tests are failing because `dnf-plugins-core` seems to have disappeared.

nullr0ute:
> The CentOS Stream 10 unit tests are failing because `dnf-plugins-core` seems to have disappeared.

there's dnf5-plugins that replace "core/extras" but they've not done proper provides/obsoleted etc. I think core needs to come back, I am going to file a bug on this because it bit me the other day too in a different usecase, will reference the bug

nullr0ute:
https://bugzilla.redhat.com/show_bug.cgi?id=2309697

achilleas-k:
> > The CentOS Stream 10 unit tests are failing because `dnf-plugins-core` seems to have disappeared.
> 
> there's dnf5-plugins that replace "core/extras" but they've not done proper provides/obsoleted etc. I think core needs to come back, I am going to file a bug on this because it bit me the other day too in a different usecase, will reference the bug

Thanks!  Though, doing a quick check with the `centos:stream10-development` container (which we use for testing), I don't see anything for `dnf search plugins`.  I found this though: https://issues.redhat.com/browse/CPE-2396

PR #900, opened 2024-09-04
cmd/build: add new `--manifest-only`, `--manifest-path` and `-arch` options

mvo5:
This PR tweaks the `build` binary to make it easier for `otk` to use it to generate reference images. With the tweaks here the tests in otk can generate reference images via:
```
$ go run github.com/osbuild/images/cmd/build@latest -type -distro centos-9 type qcow2 -arch aarch64 -manifest-only -manifest-path test/data/reference-images/<our-location>
````
that can then be used to validate that the otk generated images are identical to the ones generated by `images`

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -198,11 +198,11 @@ func run() error {
 	var distroName, imgTypeName, configFile string
 	flag.StringVar(&distroName, "distro", "", "distribution (required)")
 	flag.StringVar(&imgTypeName, "type", "", "image type name (required)")
-	flag.StringVar(&configFile, "config", "", "build config file (required)")
+	flag.StringVar(&configFile, "config", "", "build config file")
 
 	flag.Parse()
 
-	if distroName == "" || imgTypeName == "" || configFile == "" {
+	if distroName == "" || imgTypeName == "" {
 		flag.Usage()
 		os.Exit(1)
 	}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by supakeen — COMMENTED** [review]
The idea of having this is good but there's some confusion for me; how is this different from using `cmd/gen-manifests`. Perhaps it'd be better suited to add the `manifest-path` there?

**Comment by mvo5** [comment]
> The idea of having this is good but there's some confusion for me; how is this different from using `cmd/gen-manifests`. Perhaps it'd be better suited to add the `manifest-path` there?

Sure, that could also be done, I have no super strong opinion either way. My reasoning is that using `gen-manifest` seems to be more geared for the mass build task and it will be harder to add a "--manifest-path" as it will be auto-generated which means the other tool that drives it needs to know what the resulting filename is (or use the same pattern) and it also needs post-processing bcause AFAICT the generated files are not manifests but contain a build-request key and some extra data. Can all be done but I would rather move "build" into a more generally useful tool (that we can eventually use for project excosim) - but happy to hear other thoughts (e.g. @achilleas-k )

**Comment by achilleas-k** [comment]
> My reasoning is that using `gen-manifest` seems to be more geared for the mass build task and it will be harder to add a "--manifest-path" as it will be auto-generated which means the other tool that drives it needs to know what the resulting filename is (or use the same pattern)

This got me thinking if it would be useful to have a `--filename-format` flag that can interpret patterns.  For example, our current (default) format would be (in standard Go template form):
```
{{ distro }}-{{ arch }}-{{ type }}-{{ config_name }}.json
```
(though we also do `s/-/_/` on each component).

> it also needs post-processing bcause AFAICT the generated files are not manifests but contain a build-request key and some extra data.

The extra bits can be skipped with `--metadata=false` in `gen-manifests`.  That produces clean manifests.

mvo5:
Closing as this is no longer used/needed for https://github.com/osbuild/otk/pull/188

PR #902, opened 2024-09-04
[RFC] reporegistry: embed the test repo in the go binary for easier use

mvo5:
[draft as this needs splitting into two commits, more tests and is also a bit of a discussion starter if it's a good idea, it will also make the binaries bigger, e.g. the "build" binary increases from 25MB -> 28MB (stripped), otoh only things that "import testdata" will be affected so should not be too bad]

Running `cmd/build/build` currently requires running it from a git checkout so that it can find the required repository files.

This commit embeds them into the binary via "go:embed" to make it easier to use `build` on it's own.

This will make working with `otk` to generate reference images easier.

### DIFF
--- a/pkg/reporegistry/reporegistry.go
+++ b/pkg/reporegistry/reporegistry.go
@@ -2,10 +2,17 @@ package reporegistry
 
 import (
 	"fmt"
+	"io/fs"
 
 	"github.com/osbuild/images/pkg/distroidparser"
 	"github.com/osbuild/images/pkg/rpmmd"
+	"github.com/osbuild/images/test"
 )
+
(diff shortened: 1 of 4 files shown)


achilleas-k:
I think adding a `--repositories` (or `--repos` to save some typing ^_^) would be a nicer approach to this.  There are many cases where I would like to generate manifests or build an image with the same definitions but different repositories.  For example, if a repository snapshot update causes errors, the investigation would be much easier if I could run the same binary (same image definitions) against two separate repository configurations, without needing to recompile.

mvo5:
Closing as something we currently do not want.

PR #903, opened 2024-09-04
osbuild: add openscap profile and compliance policy id to facts 

croissanne:
Allows registered systems to be coupled to existing compliance policies.

---

As it stands this would require composer to fill in these facts in the image options as it does for the api types. For the openscap profile id we could fill it in automatically if an openscap customisation is defined, open to opinions.

### DIFF
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -5,6 +5,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/google/uuid"
+
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/internal/environment"
 	"github.com/osbuild/images/internal/workload"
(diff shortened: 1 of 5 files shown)


pkg/manifest/os.go
```diff
@@ -135,6 +136,11 @@ type OSCustomizations struct {
 	// The final RHSM config to be applied to the image
 	RHSMConfig *subscription.RHSMConfig
 
+	// Facts
+	RHSMFacts *facts.ImageOptions
+	// FactCompliancePolicyID *uuid.UUID
+	// FactOpenSCAPProfileID  *string
```
achilleas-k, line 142: Delete these?

achilleas-k (changes requested):

pkg/manifest/os.go
croissanne: oh yes :flushed:

kingsleyzissou (approved):

achilleas-k (approved):

achilleas-k:
Need to get #904 merged to unblock.

PR #908, opened 2024-09-05
Test configs for RHEL 10

achilleas-k:
The all-customizations config is limited to RHEL 8 and 9 because a lot of the customizations aren't supported on RHEL 7.  Let's also add RHEL 10 to the mix.

### DIFF
--- a/test/config-map.json
+++ b/test/config-map.json
@@ -1,6 +1,7 @@
 {
   "./configs/all-customizations.json": {
     "distros": [
+      "rhel-10*",
       "rhel-9*",
       "rhel-8*",
       "centos*",


thozza:
So OSCAP is not yet supported on RHEL-10 and also the EC2 (previously RHUI) images are not yet supported.

kingsleyzissou:
> So OSCAP is not yet supported on RHEL-10 and also the EC2 (previously RHUI) images are not yet supported.

Some of the profiles haven't been enabled yet afaik. I will have to double-check which are and which arent'

kingsleyzissou (approved):

PR #912, opened 2024-09-06
manifest: add `sync` to kickstart for bootc iso installer

mvo5:
Small followup for #714 - to add an extra level of safety call "sync" before signaling that the install is finished.

[draft as this will get a coresponding bib PR to ensure it works correctly]

### DIFF
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -515,6 +515,7 @@ halt --eject
 bootc switch --mutate-in-place --transport %s %s
 
 # used during automatic image testing as finished marker
+sync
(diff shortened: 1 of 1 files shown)


ondrejbudai:
I just wonder what's the motivation for this PR? Is there a bug report about this? It looks generally useful, but I would expect anaconda to run `sync` on its own, or that it relies on the fact that `umount` will do the same (correct me if I'm wrong).

mvo5:
> I just wonder what's the motivation for this PR? Is there a bug report about this? It looks generally useful, but I would expect anaconda to run `sync` on its own, or that it relies on the fact that `umount` will do the same (correct me if I'm wrong).

Agreed, this was a bit of a test/stab in the dark, I close it.

PR #924, opened 2024-09-12
Add default OpenSCAP datastream for RHEL10 & cs10 images

kingsleyzissou:
We were missing the default OpenSCAP datastream for RHEL10 and centos stream 10 distros.


thozza (approved):
:+1:

PR #928, opened 2024-09-13
blueprint: remove the sshkey customization

achilleas-k:
This is something I've mentioned a few times in the past but never got around to doing.  With the upcoming changes to partitioning (#926), I think this is a good opportunity to start with a proof of concept and the canonical example for customization simplification in images.  The general idea is that blueprints in osbuild-composer are the **canonical, user-facing version of the structure**.  The images counterpart has no obligation to maintain backwards compatibility, or even have the same exact structure as the real blueprints.

Extra considerations:
- We (well, @croissanne) recently (re)introduced the idea of moving blueprints out into its own project, and having a schema for the _actual real_ blueprint format.  I think this is a great idea and takes care of any issues of divergence between blueprint implementations.
- bootc-image-builder uses the blueprints code from this repository directly.  The particular change in this PR does not affect that project, since we never supported the sshkey customization there.
- In the past we discussed renaming the package and objects in this repository to something other than `blueprint`, to avoid any confusion about what they are.  In that case, any common structure would be considered incidental, and there would be much less implication about the structures here being a separate implementation of the same configs.

---

Both the sshkey and users customizations have been around since the beginning of time [1].  The users customization was always a superset of the sshkey, so we have code in blueprints to convert sshkeys into users.

The blueprint code in images is not exposed directly to users.  The code responsible for deserializing blueprint toml files lives in osbuild-composer and is directly translated to the equivalent types in osbuild/images.  We can simplify the code and types here to only contain the customizations that are strictly necessary and we don't need to maintain backwards compatibility.

The blueprint conversion code in osbuild-composer should convert all sshkeys to users before passing them down to osbuild/images functions.

[1] 967b0e8ce0e87c77a7c681b0c9f86d70f84711a5

### DIFF
--- a/pkg/blueprint/customizations.go
+++ b/pkg/blueprint/customizations.go
@@ -12,7 +12,6 @@ import (
 type Customizations struct {
 	Hostname           *string                        `json:"hostname,omitempty" toml:"hostname,omitempty"`
 	Kernel             *KernelCustomization           `json:"kernel,omitempty" toml:"kernel,omitempty"`
-	SSHKey             []SSHKeyCustomization          `json:"sshkey,omitempty" toml:"sshkey,omitempty"`
 	User               []UserCustomization            `json:"user,omitempty" toml:"user,omitempty"`
 	Group              []GroupCustomization           `json:"group,omitempty" toml:"group,omitempty"`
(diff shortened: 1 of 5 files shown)


croissanne (approved):
ty!

PR #930, opened 2024-09-13
Add initial support for generating SPDX SBOM documents (COMPOSER-2274)

thozza:
This PR depends on https://github.com/osbuild/osbuild/pull/1818

- Add a simple `sbom` package for working with SBOM documents.
- Extend the `dnfjson` `Solver.Depsolve()` to support requesting SBOM documents for depsolved transactions.

### DIFF
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -156,6 +156,13 @@ type Solver struct {
 	subscriptions *rhsm.Subscriptions
 }
 
+// DepsolveResult contains the results of a depsolve operation.
+type DepsolveResult struct {
+	Packages []rpmmd.PackageSpec
+	Repos    []rpmmd.RepoConfig
+	SBOM     *sbom.Document
+}
+
 // Create a new Solver with the given configuration. Initialising a Solver also loads system subscription information.
 func NewSolver(modulePlatformID, releaseVer, arch, distro, cacheDir string) *Solver {
 	s := NewBaseSolver(cacheDir)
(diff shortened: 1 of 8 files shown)


thozza:
The PR is ready for review.

thozza:
The osbuild-composer test failures are expected (and not blocking), since the PR needs code changes to be integrated in osbuild-composer.

pkg/dnfjson/dnfjson.go
```diff
@@ -156,6 +156,13 @@ type Solver struct {
 	subscriptions *rhsm.Subscriptions
 }
 
+// DepsolveResult contains the results of a depsolve operation.
+type DepsolveResult struct {
+	Packages []rpmmd.PackageSpec
+	Repos    []rpmmd.RepoConfig
+	Sbom     *sbom.Document
```
achilleas-k, line 163: Thoughts on changing `Sbom` to `SBOM`?
I kinda expected it to be all caps since it's an initialism, like `RPM`, `URL`, `BIOS`...

pkg/dnfjson/dnfjson.go
```diff
@@ -693,6 +725,9 @@ type depsolveResult struct {
 
 	// (optional) contains the solver used, e.g. "dnf5"
 	Solver string `json:"solver"`
+
+	// (optional) contains the SBOM for the depsolved transaction
+	Sbom json.RawMessage `json:"sbom"`
```
schuellerf, line 730: should this also be a pointer and omitempty as it's optional?
```suggestion
	Sbom *json.RawMessage `json:"sbom,omitempty"`
```

pkg/dnfjson/dnfjson.go
thozza: I have no strong preference. I can change it.

pkg/dnfjson/dnfjson.go
thozza: IMHO no need to make this a pointer, because it is a `[]byte`. `Solver` is also optional and it is not omitempty. While I agree that adding `omitempty` should be there, it has no actual impact on the behavior, because we always only unmarshalling from JSON to the struct.

pkg/dnfjson/dnfjson.go
thozza: renamed it to SBOM

pkg/dnfjson/dnfjson.go
thozza: I've added `omitempty` to optional struct members

pkg/sbom/document.go
```diff
@@ -0,0 +1,70 @@
...
+	default:
+		panic("invalid standard type")
+	}
+}
+
+func (t StandardType) MarshalJSON() ([]byte, error) {
+	var s string
+
+	if t == StandardTypeNone {
+		s = ""
```
kingsleyzissou, line 30: Is the string `"none"` being used anywhere? Because maybe we could change the `String()` method to just return the empty string and this `if/else` wouldn't be necessary.

I may be missing something though

thozza: It is used if you would print the value as string. I wanted to keep there something that's descriptive. For JSON, I want the value to be omitted if empty (being "none") and representing it as an empty string make things easier to work with the standard JSON (un)marshaller.

achilleas-k (approved):
TY!  LGTM

schuellerf (approved):
LGTM, thanks!

pkg/sbom/document.go
kingsleyzissou: Cool thanks, that seems sensible

kingsleyzissou (approved):

PR #935, opened 2024-09-16
Add ostree commit resolver for otk: otk-resolve-ostree-commit (COMPOSER-2349)

achilleas-k:
New `cmd` for resolving ostree commits for otk.

The resolver is a simple call to ostree.Resolve() which handles resolving ostree input options to an ostree.CommitSpec.

The resolver returns the input as-is when the ref is a commit ID (there is nothing to resolve).  This can happen when a user wants to pull an ostree commit by its ID rather than specifying a ref to resolve.

### DIFF
--- a/cmd/otk-resolve-ostree-commit/main_test.go
+++ b/cmd/otk-resolve-ostree-commit/main_test.go
@@ -1,0 +1,79 @@
+package main_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	resolver "github.com/osbuild/images/cmd/otk-resolve-ostree-commit"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Create a test server that responds with the commit ID that corresponds to
+// the ref.
+func createTestServer(refIDs map[string]string) *httptest.Server {
+	handler := http.NewServeMux()
+	handler.HandleFunc("/refs/heads/", func(w http.ResponseWriter, r *http.Request) {
+		reqRef := strings.TrimPrefix(r.URL.Path, "/refs/heads/")
+		id, ok := refIDs[reqRef]
+		if !ok {
+			http.NotFound(w, r)
+			return
+		}
+		fmt.Fprint(w, id)
+	})
+
+	return httptest.NewServer(handler)
+}
+
+func TestResolver(t *testing.T) {
+	commitMap := map[string]string{
+		"centos/9/x86_64/edge": "d04105393ca0617856b34f897842833d785522e41617e17dca2063bf97e294ef",
+		"fake/ref/f":           "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+		"fake/ref/9":           "9999999999999999999999999999999999999999999999999999999999999999",
+		"test/ref/alpha":       "9b1ea9a8e10dc27d4ea40545bec028ad8e360dd26d18de64b0f6217833a8443d",
+		"test/ref/one":         "7433e1b49fb136d61dcca49ebe34e713fdbb8e29bf328fe90819628f71b86105",
+	}
+
+	require := require.New(t)
+	assert := assert.New(t)
+
+	repoServer := createTestServer(commitMap)
+	defer repoServer.Close()
+
+	url := repoServer.URL
+	for ref, id := range commitMap {
+		inputReq, err := json.Marshal(map[string]map[string]string{
+			"tree": {
+				"url": url,
+				"ref": ref,
+			},
+		})
+		require.NoError(err)
+
+		inpBuf := bytes.NewBuffer(inputReq)
+		outBuf := &bytes.Buffer{}
+
+		assert.NoError(resolver.Run(inpBuf, outBuf))
+
+		var output map[string]map[string]map[string]string
+		require.NoError(json.Unmarshal(outBuf.Bytes(), &output))
+
+		expOutput := map[string]map[string]map[string]string{
+			"tree": {
+				"const": {
+					"url":      url,
+					"ref":      ref,
+					"checksum": id,
+				},
+			},
+		}
+		assert.Equal(expOutput, output)
+	}
+
+}
--- a/cmd/otk-resolve-ostree-commit/main_test.go
+++ b/cmd/otk-resolve-ostree-commit/main_test.go
@@ -14,6 +14,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var commitMap = map[string]string{
+	"centos/9/x86_64/edge": "d04105393ca0617856b34f897842833d785522e41617e17dca2063bf97e294ef",
+	"fake/ref/f":           "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+	"fake/ref/9":           "9999999999999999999999999999999999999999999999999999999999999999",
+	"test/ref/alpha":       "9b1ea9a8e10dc27d4ea40545bec028ad8e360dd26d18de64b0f6217833a8443d",
+	"test/ref/one":         "7433e1b49fb136d61dcca49ebe34e713fdbb8e29bf328fe90819628f71b86105",
+}
+
 // Create a test server that responds with the commit ID that corresponds to
 // the ref.
 func createTestServer(refIDs map[string]string) *httptest.Server {
@@ -32,14 +40,6 @@ func createTestServer(refIDs map[string]string) *httptest.Server {
 }
 
 func TestResolver(t *testing.T) {
-	commitMap := map[string]string{
-		"centos/9/x86_64/edge": "d04105393ca0617856b34f897842833d785522e41617e17dca2063bf97e294ef",
-		"fake/ref/f":           "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
-		"fake/ref/9":           "9999999999999999999999999999999999999999999999999999999999999999",
-		"test/ref/alpha":       "9b1ea9a8e10dc27d4ea40545bec028ad8e360dd26d18de64b0f6217833a8443d",
-		"test/ref/one":         "7433e1b49fb136d61dcca49ebe34e713fdbb8e29bf328fe90819628f71b86105",
-	}
-
 	require := require.New(t)
 	assert := assert.New(t)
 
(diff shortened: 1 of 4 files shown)


supakeen (commented):
Just one little bit and maybe @mvo5 can chime in here but I think these outputs should live under a `tree.const` similar to the `otk-gen-partition-table` and `osbuild-gen-depsolve-dnf4` externals that generate defines.

achilleas-k:
> Just one little bit and maybe @mvo5 can chime in here but I think these outputs should live under a `tree.const` similar to the `otk-gen-partition-table` and `osbuild-gen-depsolve-dnf4` externals that generate defines.

Ah, yes.  Probably.  Though, I'm not entirely sure what our rules and conventions are exactly.

Given the way we describe the `Const` object
https://github.com/osbuild/images/blob/fa40f2da9f8a2f15a15687e301421be9fa5f6d3b/internal/otkdisk/partition.go#L23-L25
I think it's more appropriate for outputs from externals that will be fed back into another external.

The output from the resolver here is going to be used directly in stage options, so it's probably not necessary to mark them as `const`.  Though maybe it's still a good convention to specify that modifying one or more might lead to inconsistencies.

supakeen:
> > Just one little bit and maybe @mvo5 can chime in here but I think these outputs should live under a `tree.const` similar to the `otk-gen-partition-table` and `osbuild-gen-depsolve-dnf4` externals that generate defines.
> 
> Ah, yes. Probably. Though, I'm not entirely sure what our rules and conventions are exactly.
> 
> Given the way we describe the `Const` object
> 
> https://github.com/osbuild/images/blob/fa40f2da9f8a2f15a15687e301421be9fa5f6d3b/internal/otkdisk/partition.go#L23-L25
> 
> I think it's more appropriate for outputs from externals that will be fed back into another external.
> 
> The output from the resolver here is going to be used directly in stage options, so it's probably not necessary to mark them as `const`. Though maybe it's still a good convention to specify that modifying one or more might lead to inconsistencies.

That is definitely a discussion we could have for now I think we can go with mashing it under const and then have a separate discussion about it when we have this one and the container resolving one in as well (we'll have 4 externals by then).

cmd/otk-resolve-ostree-commit/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+			return
+		}
+		fmt.Fprint(w, id)
+	})
+
+	return httptest.NewServer(handler)
+}
+
+func TestResolver(t *testing.T) {
+	commitMap := map[string]string{
```
mvo5, line 35: We could get away with a simpler test here I think. It's fine as is given that it's all written etc but I think at this level here all we care about and need to test is that we can read the inputs, call `ostree.Resolve()` with the right inputs and then return the expected outputs. So just doing something like:
```go
diff --git a/cmd/otk-resolve-ostree-commit/export_test.go b/cmd/otk-resolve-ostree-commit/export_test.go
index 77c5f1a93..91ab63416 100644
--- a/cmd/otk-resolve-ostree-commit/export_test.go
+++ b/cmd/otk-resolve-ostree-commit/export_test.go
@@ -3,3 +3,12 @@ package main
 var (
        Run = run
 )
+
+func MockOstreeResolve(new func(...)) (restore func()) {
+       saved := ostreeResolve
+       ostreeResolve = new
+       return func() {
+               ostreeResolved = saved
+       }
+}
+
diff --git a/cmd/otk-resolve-ostree-commit/main.go b/cmd/otk-resolve-ostree-commit/main.go
index 3857dfd37..a7ea2dd5e 100644
--- a/cmd/otk-resolve-ostree-commit/main.go
+++ b/cmd/otk-resolve-ostree-commit/main.go
@@ -44,6 +44,8 @@ type Output struct {
        Checksum string `json:"checksum"`
 }
 
+var ostreeResolve = ostree.Resolve
+
 func run(r io.Reader, w io.Writer) error {
        var inputTree Tree
        if err := json.NewDecoder(r).Decode(&inputTree); err != nil {
@@ -51,7 +53,7 @@ func run(r io.Reader, w io.Writer) error {
        }
 
        sourceSpec := ostree.SourceSpec(inputTree.Tree)
-       commitSpec, err := ostree.Resolve(sourceSpec)
+       commitSpec, err := ostreeResolve(sourceSpec)
        if err != nil {
                return fmt.Errorf("failed to resolve ostree commit: %w", err)
        }

```
(the upside is that we don't need to care about implementation details of ostree.Resolve() at this level and we just trust that it's well tested to reach out to the net etc and if implementation details change we only need to tweak the tests in ostree and here we are blisfully ignorant (which again of course here is not very important as this behavior/test will most likely never change))

cmd/otk-resolve-ostree-commit/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+		"centos/9/x86_64/edge": "d04105393ca0617856b34f897842833d785522e41617e17dca2063bf97e294ef",
+		"fake/ref/f":           "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
```
mvo5, line 37: silly question - do we need all of the cases? If so, maybe a short comment what `fake/ref/f` tests that `fake/ref/9` does not etc.

cmd/otk-resolve-ostree-commit/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+		"fake/ref/9":           "9999999999999999999999999999999999999999999999999999999999999999",
+		"test/ref/alpha":       "9b1ea9a8e10dc27d4ea40545bec028ad8e360dd26d18de64b0f6217833a8443d",
+		"test/ref/one":         "7433e1b49fb136d61dcca49ebe34e713fdbb8e29bf328fe90819628f71b86105",
+	}
+
+	require := require.New(t)
```
mvo5, line 43: (super nitpick) I personally would just only use one here but I am totally fine to leave this as is, just noticed during the review :)

cmd/otk-resolve-ostree-commit/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+			"tree": {
+				"url":      url,
+				"ref":      ref,
+				"checksum": id,
+			},
+		}
+		assert.Equal(expOutput, output)
+	}
+
+	// repeat test but use the commit ID in the request directly (no ref resolve)
```
mvo5, line 77: (super nitpick) I personally would put this into it's own test and call it something like `TestResolverByID` or something (but happy to just leave as is).

mvo5 (commented):
Thank you! I added some comments but feel free to ignore, mostly nitpicks/ideas/suggestions.

About the `const` vs `inteneral`. I'm happy to have a discussion here if needed, in my head it was mostly this heuristic:
- `const` means that if a consumer modifies it things may fall apart/become inconsistent but const itself is fine to be used in an otk manifest and things under it are an API that we cannot easily break
- `internal` means to be used only by internal tooling, i.e. matching producer/concsumer externals, anything under it is not an API and can change anytime as the consumer/produce pleases

pkg/ostree/ostree.go
```diff
@@ -152,7 +152,10 @@ func ResolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptio
 	if consumerCerts {
 		if subs == nil {
 			subs, err = rhsm.LoadSystemSubscriptions()
-			if subs.Consumer == nil || err != nil {
+			if err != nil {
+				return "", NewResolveRefError("error adding rhsm certificates when resolving ref: %s", err)
+			}
+			if subs.Consumer == nil {
 				return "", NewResolveRefError("error adding rhsm certificates when resolving ref")
```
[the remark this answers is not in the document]
achilleas-k: > we expect certs but got none and no error

That's not entirely accurate and is where the issue comes from.  The `LoadSystemSubscriptions()` function is "allowed" to return empty consumer certs without an error for systems that have RHSM installed but aren't subscribed.  The depsolver calls this function to get subscription certs to include in depsolve requests.

In the case of ostree, the request for consumer certs is explicit, so we need to check that both (1) the function didn't fail to find RHSM system certs and (2) the consumer certs were also loaded.

It's confusing and strange and I'd like to address it in a separate PR.  We should either make the function fail if it doesn't find consumer certs, because I think there's never a case where we care about RHSM certs but don't care about consumer certs, or, if there really are cases where we care about one and not the other, add an argument or variant of the function (`LoadSystemSubscriptionWithConsumerCerts()`) or `func LoadSystemSubscription(consumerRequired bool)` (ignoring the known issues of bools in fn args).

cmd/otk-resolve-ostree-commit/main_test.go
achilleas-k: There's some logic in the resolver that I'd rather not mock.  The resolver is tested elsewhere, yes, but I'd also like for the external to test its own expectations and assumptions.  The interaction between the arguments for example is something that might change and it'd be good to have those tested in the external.  Mocking the function would make us miss those changes.  On the other hand, mocking the server itself is perfect because it's so simple, it's just a single GET with a very easy to fake response.  I think given how cheap it is to catch regressions and changes in the `ostree.Resolve()` call for the external, mocking the server and not the function is a lot more beneficial.

cmd/otk-resolve-ostree-commit/main_test.go
achilleas-k: Actually nothing.  What we should add is some negative tests, tbh.

cmd/otk-resolve-ostree-commit/main_test.go
achilleas-k: You mean assert vs require?  I know what you mean, it's weird to have both, but some errors are showstoppers (like, if the output from `resolver.Run()` doesn't unmarshal without error, that should kill the whole test) and others aren't (ref mismatch).

cmd/otk-resolve-ostree-commit/main_test.go
achilleas-k: To summarise: With the mock of the server we are in fact testing `ostree.Resolve()`, but only as a means of making sure that the `otk-resolve-ostree-commit` tool is preparing the right inputs.  The validity of a given input can change without notice if we mock `ostree.Resolve()`.

achilleas-k:
The `Output` has been `Const`ed.

cmd/otk-resolve-ostree-commit/main_test.go
achilleas-k: Split!

cmd/otk-resolve-ostree-commit/main_test.go
mvo5: I'm may misunderstand the argument (my apologies if I did) - once we release this tool the inputs for this tool become an API that will be used in many otk manifests. That means that any inputs for `ostree.Resolve()` are now also an API, we cannot restrict what was valid before anymore (also afaik osbuild composer uses `ostree.Resolve()` too?).  Again, totally fine to leave as is (and stop this thread) I'm really just trying to understand (and sorry for the extra work my questions cause).

mvo5 (approved):
Thank you!

supakeen (approved):
Needs the other comments resolved, awesome.

cmd/otk-resolve-ostree-commit/main_test.go
achilleas-k: The arguments to `ostree.Resolve()` become API indirectly, but we might change that function's behaviour.

Here's a concrete example that we might actually do at some point:

Let's say we decide that the `ref` argument to `ostree.Resolve()` **must** be a valid ref (similar to a branch name) and cannot be an ostree commit ID (sha256 checksum).  After this change, passing a checksum to the resolver will result in an error.  This makes it the job of the caller to check what it's passing through to the resolver, basically moving this block outside the function: https://github.com/achilleas-k/images/blob/31d94be4ec71566036c397b43deeffb349c71ba7/pkg/ostree/ostree.go#L238-L242

If that happens and we want to keep the otk external's API consistent, we would run the checksum check in the `run()` function of the otk external before calling `ostree.Resolve()`.  The external's API and behaviour don't change even though the resolver function's did.

Without calling the real `ostree.Resolve()` in tests, no test would fail from this change.

cmd/otk-resolve-ostree-commit/main_test.go
```diff
@@ -0,0 +1,193 @@
...
+func createTestServer(refIDs map[string]string) *httptest.Server {
+	handler := http.NewServeMux()
+	handler.HandleFunc("/refs/heads/", func(w http.ResponseWriter, r *http.Request) {
+		reqRef := strings.TrimPrefix(r.URL.Path, "/refs/heads/")
+		id, ok := refIDs[reqRef]
+		if !ok {
+			http.NotFound(w, r)
+			return
+		}
+		fmt.Fprint(w, id)
```
achilleas-k, line 36: Poked into Snyk to see what the problem is on this line:
```
Unsanitized input from the request URL flows into fmt.Fprint, where it is used to render an HTML page returned to the user. This may result in a Reflected Cross-Site Scripting attack (XSS).
```

supakeen: Can be ignored on the test case :)

PR #936, opened 2024-09-16
Add container resolver for otk: otk-resolve-containers (COMPOSER-2293)

achilleas-k:
New `cmd` for resolving containers for otk.

The resolver is a thin wrapper around the container.Resolver.  It can handle multiple containers for the same architecture in one call.

### DIFF
--- a/cmd/otk-resolve-containers/main.go
+++ b/cmd/otk-resolve-containers/main.go
@@ -1,0 +1,114 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/osbuild/images/pkg/container"
+)
+
+// All otk external inputs are nested under a top-level "tree"
+type Tree struct {
+	Tree Input `json:"tree"`
+}
+
+// Input represents the user-provided inputs that will be used to resolve a
+// container image.
+type Input struct {
+	// The architecture of the container images to resolve.
+	Arch string `json:"arch"`
+
+	// List of container refs to resolve.
+	Containers []blueprint.Container `json:"containers"`
+}
+
+// Output contains everything needed to write a manifest that requires pulling
+// container images.
+type Output struct {
+	Const OutputConst `json:"const"`
+}
+
+type OutputConst struct {
+	Containers []ContainerInfo `json:"containers"`
+}
+
+type ContainerInfo struct {
+	Source string `json:"source"`
+
+	// Digest of the manifest at the source.
+	Digest string `json:"digest"`
+
+	// Container image identifier.
+	ImageID string `json:"imageid"`
+
+	// Name to use inside the image.
+	LocalName string `json:"local-name"`
+
+	// Digest of the list manifest at the source
+	ListDigest string `json:"list-digest,omitempty"`
+
+	// The architecture of the image
+	Arch string `json:"arch"`
+
+	TLSVerify *bool `json:"tls-verify,omitempty"`
+}
+
+func run(r io.Reader, w io.Writer) error {
+	var inputTree Tree
+	if err := json.NewDecoder(r).Decode(&inputTree); err != nil {
+		return err
+	}
+
+	resolver := container.NewResolver(inputTree.Tree.Arch)
+
+	for _, bpSpec := range inputTree.Tree.Containers {
+		srcSpec := container.SourceSpec{
+			Source:    bpSpec.Source,
+			Name:      bpSpec.Name,
+			TLSVerify: bpSpec.TLSVerify,
+		}
+		resolver.Add(srcSpec)
+	}
+
+	containerSpecs, err := resolver.Finish()
+	if err != nil {
+		return err
+	}
+	containerInfos := make([]ContainerInfo, len(containerSpecs))
+	for idx := range containerSpecs {
+		spec := containerSpecs[idx]
+		containerInfos[idx] = ContainerInfo{
+			Source:     spec.Source,
+			Digest:     spec.Digest,
+			ImageID:    spec.ImageID,
+			LocalName:  spec.LocalName,
+			ListDigest: spec.ListDigest,
+			Arch:       spec.Arch.String(),
+			TLSVerify:  spec.TLSVerify,
+		}
+	}
+
+	output := map[string]Output{
+		"tree": {
+			Const: OutputConst{
+				Containers: containerInfos,
+			},
+		},
+	}
+	outputJson, err := json.MarshalIndent(output, "", "  ")
+	if err != nil {
+		return fmt.Errorf("cannot marshal response: %w", err)
+	}
+	fmt.Fprintf(w, "%s\n", outputJson)
+	return nil
+}
+
+func main() {
+	if err := run(os.Stdin, os.Stdout); err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err.Error())
+		os.Exit(1)
+	}
+}
--- a/cmd/otk-resolve-containers/main_test.go
+++ b/cmd/otk-resolve-containers/main_test.go
@@ -18,12 +18,6 @@ import (
 	resolver "github.com/osbuild/images/cmd/otk-resolve-containers"
 )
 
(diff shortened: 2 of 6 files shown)


supakeen (commented):
Going to leave the same comment as on #935: what about the `const`. We can resolve both in one go there. Otherwise very nice.

cmd/otk-resolve-containers/main_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/osbuild/images/internal/testregistry"
+	"github.com/osbuild/images/pkg/arch"
+	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	resolver "github.com/osbuild/images/cmd/otk-resolve-containers"
+)
+
+const rootLayer = `H4sIAAAJbogA/+SWUYqDMBCG53lP4V5g9x8dzRX2Bvtc0VIhEIhKe/wSKxgU6ktjC/O9hMzAQDL8
```
mvo5, line 21: Maybe a small comment how this is generated, what it represents etc?

cmd/otk-resolve-containers/main_test.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+
+	for _, containerArch := range []string{"amd64", "ppc64le"} {
+		t.Run(containerArch, func(t *testing.T) {
+
+			require := require.New(t)
+			assert := assert.New(t)
+
+			amd64input := map[string]interface{}{
```
mvo5, line 74: I may sound silly here, but I'm a fan of writing the json out mostly because to me a test is also documentation and having a look into the test file to see what the json input and expected json output looks like feels rather nice. I can see it's tedious and work so feel free to ignore (both there and in the expected output)

cmd/otk-resolve-containers/main_test.go
```diff
@@ -0,0 +1,194 @@
...
+	defer registry.Close()
+
+	type testCase struct {
+		source    string
+		arch      string
+		tlsverify bool
+		errSubstr string
+	}
+
+	testCases := map[string]testCase{
```
mvo5, line 133: I wonder if we could get away with testing slightly less cases here? It's fine to keep as is as it's already written but at this level we are mostly caring about that when we call `Run()` and it results in an error but the details of the various ways that `run` can fail. Which seems mostly `resolver.Finish()` so testing that is great but at the same time just checking one way (or even mocking) is probably enough.

cmd/otk-resolve-containers/main_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+		tag := fmt.Sprintf("tag-%d", idx)
+		repo.AddTag(checksum, tag)
+		refs[idx] = fmt.Sprintf("%s:%s", ref, tag)
+	}
+	return registry, refs
+}
+
+func TestResolver(t *testing.T) {
+	registry, refs := createTestRegistry()
```
mvo5, line 55: Thanks, this looks like a really nice test abstraction and given that `NewResolver()` is hard to mock it's seems the best approach to test this.

mvo5 (commented):
Thanks! This looks nice, ideally we would apply the outcome of the `const` discusison here and there are some ideas/suggestions inline for your considerations, mostly about spelling out the inputs/outputs slightly more explicitly

cmd/otk-resolve-containers/main_test.go
achilleas-k: In this case it's incredibly tedious but for the output it's also mostly impossible to get right because the order or elements (and sometimes keys) isn't predictable.  It's often stable, and the elements are stable in our container resolver, but modifying the tests in any way would require adjusting the expected output and figuring out the new order of elements.

Not to mention, the error output if it fails is a lot less readable when it's just comparing json in string format.

cmd/otk-resolve-containers/main_test.go
achilleas-k: I felt it was more useful to make sure that the error message provides useful information for a command line tool, vs simply testing that "error exists".
Of course, this all comes straight from `resolver.Finish()` like you mentioned, so maybe here is not the place to test for it.  I do like that we're making sure that _the container resolver command line tool isn't "eating" parts of the underlying error messages_.

cmd/otk-resolve-containers/main_test.go
achilleas-k: Thank whoever wrote that mock container registry :D

achilleas-k:
> Going to leave the same comment as on #935: what about the `const`. We can resolve both in one go there. Otherwise very nice.

Done!  Moved output under `const` and also added an extra level, `containers`, in case we need to add more to the output at some point, because the previous top-level object (the one under `.tree.const`) is an array.

cmd/otk-resolve-containers/main_test.go
mvo5: Sorry, I probably expressed this not very well, it's great to have a test that ensures we do pass the error properly and not "eat" part of the underlying error message. I was mostly wondering if we need to test this for all the cases. But as I wrote, it's fine as is.

cmd/otk-resolve-containers/main_test.go
mvo5: Nitpick (and fine for a followup or fine for me to take if you write a coment here in GH) - but I'm still curious where this comes from.

cmd/otk-resolve-containers/main.go
```diff
@@ -0,0 +1,117 @@
...
+	// Name to use inside the image.
+	LocalName string `json:"local-name"`
+
+	// Digest of the list manifest at the source
+	ListDigest string `json:"list-digest,omitempty"`
+
+	// The architecture of the image
+	Arch string `json:"arch"`
+
+	LocalStorage bool  `json:"local-storage,omitempty"`
```
mvo5, line 56: This means that "LocalStorage: false" will not be part of the json output, not sure that is desirable or now, we could consider not exposing it at all as I'm not sure how we would use it in an otk file or just remove the omitempty(?)

mvo5 (dismissed):
Thank you! This is nice and I really like the detail that pkg/containers/spec.go:Spec is not directly exposed as we have now more control.

There are some testing nitpicks/suggestions inline but as they only affect testing (which is already comprehensive) and are very opinionated (not neccessarily in a good way, sorry for that!)  - the testing is very nice and comprehensive (please don't get me wrong here). *If* you like the ideas I am happy to do a followup, if you don't I will just be silent :)

The only thing that we might consider is the coment/question about `LocalStorage` (but given that most of the times I think it will be false/not visible(?) it should cause no harm even if we merge as is).

supakeen (dismissed):
Needs the other comments resolved, awesome.

cmd/otk-resolve-containers/main_test.go
achilleas-k: Sorry.  Forgot to add this.  I'll include a comment.

cmd/otk-resolve-containers/main_test.go
achilleas-k: I just noticed I accidentally left the test running with only 1 ref when the comment says 10 (it should be 10).  But I'll add this test with raw json and a single ref too.

cmd/otk-resolve-containers/main_test.go
achilleas-k: Not really necessary to cover all cases but I think it's nice to have.

mvo5 (approved):
Thank you

cmd/otk-resolve-containers/main.go
achilleas-k: I think you're right.  We should probably hide it completely and see if we ever need it.

cmd/otk-resolve-containers/main_test.go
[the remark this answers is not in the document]
achilleas-k: Yup.  Renaming.

cmd/otk-resolve-containers/main.go
achilleas-k: Dropped!

cmd/otk-resolve-containers/main_test.go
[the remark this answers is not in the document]
achilleas-k: Done

PR #937, opened 2024-09-16
gen-manifests: create predictable version/release numbers in mockDepsolve

mvo5:
The current code will only generate "0" version numbers for packages and releases. This limits the usefulness of this mock in the tests for otk. Instead generate predictable but (mostly) non-zero numbers based on the modulo of the first and second char of the pkgname.

Useful for e.g. https://github.com/osbuild/otk/pull/196 as it will ensure that the `fragment/grub2.yaml` actually changes depending on the kernel name.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/gobwas/glob"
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
Not a bad idea.

cmd/gen-manifests/main.go
```diff
@@ -375,11 +376,15 @@ func mockDepsolve(packageSets map[string][]rpmmd.PackageSet, repos []rpmmd.RepoC
...
+				// release/version numbers
+				ver := strconv.Itoa(int(pkgName[0]) % 9)
+				rel := strconv.Itoa(int(pkgName[1]) % 9)
 				spec := rpmmd.PackageSpec{
 					Name:           pkgName,
 					Epoch:          0,
-					Version:        "0",
-					Release:        "0",
+					Version:        ver,
+					Release:        "r" + rel,
```
supakeen, line 387: Where does the `r` in the release come from here? All the release fields I've seen are usually a single number :)

mvo5: I may be missing something and happy to tweak as needed, I was simply looking at:
```
$ dnf info coreutils
Name         : coreutils
Version      : 9.4
Release      : 7.fc40
...
```
or am I looking at the wrong "releas" field here?

achilleas-k: It's the right release field, but to @supakeen's point, I don't see an `r`.

mvo5: Sure, sorry, the example release field is `7.fc40` AFAICT, on centos9 it's `36.el9` so maybe `rel + ".fk1" (or the full fake1)? I was mostly trying to make it distinct from pure numbers as it seems to be a combination of number and distro shorthand in general. Ideas welcome of course!

achilleas-k: Agreed.  Doesn't really matter in the end, as long as it's stable an predictable.

achilleas-k (approved):
LGTM

cmd/gen-manifests/main.go
supakeen: Ack.

PR #938, opened 2024-09-18
[RFC] gen-manifests: add excluded packages via `exclude:` prefix in mock

mvo5:
This commit adds excluded packages to the generated manifest when the `gen-manifests` command is run in the `mockDepsolve()` mode.

This allows us to compare the generated otk and images manifests with more confidence.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -386,6 +386,20 @@ func mockDepsolve(packageSets map[string][]rpmmd.PackageSet, repos []rpmmd.RepoC
 				}
 				specSet = append(specSet, spec)
 			}
+			for _, excludeName := range pkgSet.Exclude {
+				pkgName := fmt.Sprintf("exclude:%s", excludeName)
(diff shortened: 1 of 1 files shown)


supakeen (commented):
Does it make sense to return these in the list of package specs? We wouldn't normally get that information from the depsolver as the packages are excluded so I feel like a list of package names is enough here?

mvo5:
> Does it make sense to return these in the list of package specs? We wouldn't normally get that information from the depsolver as the packages are excluded so I feel like a list of package names is enough here?

Maybe I'm misunderstanding what you are saying but if we just add the excluded package names as "normal" packages to the resolver we would not catch differences like "images" includes "pkg1" but "otk" excludes "pkg1". In both cases it would appear in the package list so I think we need some prefix/marker/thing that ensures they are different. Plus this is very explicit. But open for ideas of course.

achilleas-k (approved):
I was going to suggest we also include 1767b440691e8183cc2d4e36eed04e9f297c14df here, but that's not as useful since the version numbers of excluded packages don't matter.

LGTM

PR #945, opened 2024-09-23
cmd/otk-resolve-ostree-commit: support mocking ostree resolve calls

achilleas-k:
Read the OTK_UNDER_TEST env var, which we use in otk to enable test mode for our resolvers, and return a mocked checksum.


mvo5 (approved):
Thank you

supakeen (approved):

PR #946, opened 2024-09-23
reporegistry: add NoReposLoadedError and use it

thozza:
Add `NoReposLoadedError` custom error type for the case when any function that is supposed to load repository configs, does not load any. Use it in all functions that are directly loading repositories from configuration files.

I considered `NoReposError` name, but that could be later added for functions such as `RepoRegistry.ReposBy...()`, in case there are no repositories in the RepoRegistry for the given image type / arch.

This is a follow-up to https://github.com/osbuild/osbuild-composer/pull/4378

### DIFF
--- a/pkg/reporegistry/error.go
+++ b/pkg/reporegistry/error.go
@@ -1,0 +1,13 @@
+package reporegistry
+
+import "fmt"
+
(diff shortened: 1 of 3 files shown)


mvo5 (approved):
Nice!

PR #948, opened 2024-09-24
github: drop `-v` from `go test`

mvo5:
This commit removes `-v` from `go test`. The test output is very long (17k lines according to GH) and most is just:
```
...
--- PASS: TestRPMDeduplication (0.00s)
...
```
but when trying to find a failing test (like in PR#945) scrolling through to find the actual failing one becomes a bit cumbersome.

So this commit drops `-v`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,10 +50,10 @@ jobs:
         run: git config --global --add safe.directory "$(pwd)"
 
       - name: Run unit tests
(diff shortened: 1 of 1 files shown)


bcl (approved):
Ok, I'd swear that without `-v` there wasn't enough context to debug random errors -- but having just tested that here locally I am still seeing decent output.
ACK!

supakeen (approved):

PR #951, opened 2024-09-24
blueprints: fix filesystem customizations to allow toml Undecoded()

mvo5:
The current toml filesytem customizations will decode a `FilesystemCustomization` as a map[string]interface{}. This means that the underlying toml engine will not konw what keys inside the map are decoded and which are not decoded. This lead to the issue https://github.com/osbuild/bootc-image-builder/issues/655 in `bootc-image-builder` where we check if we have unencoded entries and if so error (in this case incorrectly).

This commit fixes the issue by moving the toml decoding from the struct/map[string]interface{} to primitive types. It's a bit ugly as is (partly because of the limited go type system) but with that the tests pass and len(meta.Undecoded()) is the expected zero.

I opened https://github.com/BurntSushi/toml/issues/425 to see if there is another way via a custom unmarshaler.

### DIFF
--- a/pkg/blueprint/customizations.go
+++ b/pkg/blueprint/customizations.go
@@ -301,7 +301,7 @@ func (c *Customizations) GetFilesystemsMinSize() uint64 {
 	}
 	var agg uint64
 	for _, m := range c.Filesystem {
-		agg += m.MinSize
+		agg += uint64(m.MinSize)
 	}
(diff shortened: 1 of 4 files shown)


mvo5:
I submited an alternative idea upstream https://github.com/BurntSushi/toml/pull/426

mvo5:
Closing in favor of https://github.com/osbuild/images/pull/1049

PR #952, opened 2024-09-26
cmd/otk-gen-partition-table: support for boot partition type PPC PReP

schuellerf:
Implements support for the legacy PowerPC PowerPC Reference Platform boot partition type


cmd/otk-gen-partition-table/main.go
```diff
@@ -43,9 +43,12 @@ type InputProperties struct {
 
 // InputCreate contains details what partitions to auto-generate
 type InputCreate struct {
-	BIOSBootPartition bool   `json:"bios_boot_partition"`
-	EspPartition      bool   `json:"esp_partition"`
-	EspPartitionSize  string `json:"esp_partition_size"`
+	BIOSBootPartition            bool   `json:"bios_boot_partition"`
+	BIOSBootPartitionSize        string `json:"bios_boot_partition_size"`
+	BIOSBootPartitionTypePPCPReP bool   `json:"bios_boot_partition_type_ppc_prep"`
```
mvo5, line 48: I would love the feedback of @achilleas-k here too but this feels  a bit too specialized, arguable maybe the "create_bios" is a bit of overoptimizing and we may reconsider but for ppc it seems a bit much. My alternative suggestions would be something relatively simple, we just support "raw" partitions with no fs (and we need bootable too) which means the caller can decide what to build, this flexibility will probably help us in other places too e.g.:
```go
index b009c3270..5dc2d64cb 100644
--- a/cmd/otk-gen-partition-table/main.go
+++ b/cmd/otk-gen-partition-table/main.go
@@ -51,6 +51,7 @@ type InputCreate struct {
 // InputPartition represents a single user provided partition input
 type InputPartition struct {
        Name       string `json:"name"`
+       Bootable   bool   `json:"bootable"`
        Mountpoint string `json:"mountpoint"`
        Label      string `json:"label"`
        Size       string `json:"size"`
@@ -183,12 +184,15 @@ func makePartitionTableFromOtkInput(input *Input) (*disk.PartitionTable, error)
                if err != nil {
                        return nil, err
                }
-               pt.Partitions = append(pt.Partitions, disk.Partition{
-                       Size: uintSize,
-                       UUID: part.PartUUID,
-                       Type: part.PartType,
-                       // XXX: support lvm,luks here
-                       Payload: &disk.Filesystem{
+               // XXX: support lvm,luks here
+               newPart := disk.Partition{
+                       Size:     uintSize,
+                       Bootable: part.Bootable,
+                       UUID:     part.PartUUID,
+                       Type:     part.PartType,
+               }
+               if part.Type != "" {
+                       newPart.Payload = &disk.Filesystem{
                                Label:        part.Label,
                                Type:         part.Type,
                                Mountpoint:   part.Mountpoint,
@@ -196,8 +200,9 @@ func makePartitionTableFromOtkInput(input *Input) (*disk.PartitionTable, error)
                                FSTabOptions: part.FSMntOps,
                                FSTabFreq:    part.FSFreq,
                                FSTabPassNo:  part.FSPassNo,
-                       },
-               })
+                       }
+               }
+               pt.Partitions = append(pt.Partitions, newPart)
        }
 
        return pt, nil
```
(diff works but for a real PR needs tests of course :)

with that we can then write a partition table for ppc64el that looks like:
```yaml
otk.define:
  filesystem:
    modifications:
    # empty
    otk.external.otk-gen-partition-table:
      modifications:
        ${filesystem.modifications}
      properties:
        type: dos
        default_size: "10 GiB"
        uuid: "0x14fc63d2"
      partitions:
        - name: ppc-boot
          bootable: true
          size: "4 MiB"
          part_uuid: ""
          part_type: "41"
        - name: boot
          mountpoint: /boot
          label: boot
          size: "1 GiB"
          type: "xfs"
          fs_mntops: defaults
          part_uuid: ""
        - name: root
          mountpoint: /
          type: "xfs"
          size: "2 GiB"
          fs_mntops: defaults
          part_uuid: ""
``` 

afaict this works (except that the uuid needs manual adjustment as ppc64el uses no labels so the dection what uuid is used for root/boot in gen-images-ref fails)

mvo5 (commented):
Thanks for looking into this. I made a suggestion about a slightly different approach but let's make sure this is discussed with some more people (especially our fs gurus :)

cmd/otk-gen-partition-table/main.go
achilleas-k: I really like @mvo5's suggestion here.  `bios_boot_partition_type_ppc_prep` is way too specific, especially since the caller of the external can already specify the `part_type`.  Adding the `bootable` boolean to the input struct makes the external less smart, which I like.  I see externals as straightforward, non-smart helpers for generating osbuild-relevant structures (stage options and other things) and I don't think they should be encoding much knowledge in them, that should be a job for the otk files.

mvo5: @schuellerf I think it would be nice to move this PR forward, do you want me to have a look and push my idea to this branch?

schuellerf: yes, please - you are definitely more into the details of this implementation and their side-effects!

schuellerf:
will be handled by https://github.com/osbuild/images/pull/963

PR #954, opened 2024-09-27
Makefile: initial version of makefile, replace GH action

schuellerf:
Implementing `make test` to be able to run the tests locally
and implements `make gh-action-test` to run github action "unit-tests-fedora" locally

Also changes the github action to just use the implementation in `Makefile`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,8 +24,6 @@ jobs:
       fail-fast: false  # if one fails, keep the other(s) running
     name: "🛃 Unit tests (Fedora ${{ matrix.fedora_version }})"
     runs-on: ubuntu-latest
-    container:
-      image: registry.fedoraproject.org/fedora:${{ matrix.fedora_version }}
 
     steps:
       - name: Check out code into the Go module directory
@@ -33,27 +31,8 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
 
-      - name: Install python3
-        # The Fedora 41 container doesn't have python3 installed by default
-        run: dnf -y install python3
-
-      - name: Set up repository for pinned osbuild commit
-        run: ./test/scripts/setup-osbuild-repo
-
-        # krb5-devel is needed to test internal/upload/koji package
-        # gcc is needed to build the mock depsolver binary for the unit tests
-        # gpgme-devel is needed for container upload dependencies
-      - name: Install build and test dependencies
-        run: dnf -y install krb5-devel gcc git-core go gpgme-devel osbuild-depsolve-dnf btrfs-progs-devel device-mapper-devel
-
-      - name: Mark the working directory as safe for git
-        run: git config --global --add safe.directory "$(pwd)"
-
       - name: Run unit tests
-        run: go test -race  ./...
-
-      - name: Run depsolver tests with force-dnf to make sure it's not skipped for any reason
-        run: go test -race ./pkg/dnfjson/... -force-dnf
+        run: make BASE_CONTAINER_IMAGE_TAG=${{matrix.fedora_version}} gh-action-test
 
   container-resolver-tests:
     name: "🛃 Container resolver tests"
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,22 @@
+ARG BASE_CONTAINER_IMAGE=registry.fedoraproject.org/fedora:latest
+FROM ${BASE_CONTAINER_IMAGE}
+
+# The Fedora 41 container doesn't have python3 installed by default
+RUN dnf install -y python3
+
+WORKDIR /setup
+
+COPY ./test/scripts ./test/scripts/
+COPY Schutzfile .
+RUN ./test/scripts/setup-osbuild-repo
+
+# NOTE: keep in sync with README.md#Contributing
+RUN dnf -y install krb5-devel go gpgme-devel osbuild-depsolve-dnf btrfs-progs-devel device-mapper-devel
+
+COPY go.mod go.sum .
+RUN go mod download
+
+WORKDIR /app
+
+# Mark the working directory as safe for git
+RUN git config --global --add safe.directory /app
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,35 @@
+.PHONY: help
+help:
+	@echo 'Usage:'
+	@echo '  make <target>'
+	@echo ''
+	@echo 'Targets:'
+	@awk 'match($$0, /^([a-zA-Z_\/-]+):.*?## (.*)$$/, m) {printf "  \033[36m%-30s\033[0m %s\n", m[1], m[2]}' $(MAKEFILE_LIST) | sort
+
+BASE_CONTAINER_IMAGE_NAME?=registry.fedoraproject.org/fedora
+BASE_CONTAINER_IMAGE_TAG?=40
+BASE_CONTAINER_IMAGE?=${BASE_CONTAINER_IMAGE_NAME}:${BASE_CONTAINER_IMAGE_TAG}
+
+CONTAINERFILE=Containerfile
+CONTAINER_IMAGE?=osbuild-images_$(shell echo $(BASE_CONTAINER_IMAGE) | tr '/:.' '_')
+CONTAINER_EXECUTABLE?=podman
+
+container_built_$(CONTAINER_IMAGE).info: $(CONTAINERFILE) Schutzfile test/ go.mod go.sum # internal rule to build the container only if needed
+	$(CONTAINER_EXECUTABLE) build --build-arg BASE_CONTAINER_IMAGE="${BASE_CONTAINER_IMAGE}" \
+	                              --tag $(CONTAINER_IMAGE) \
+	                              -f $(CONTAINERFILE) .
+	echo "Container last built on" > $@
+	date >> $@
+
+.PHONY: gh-action-test
+gh-action-test: container_built_$(CONTAINER_IMAGE).info ## run all tests in a container (see BASE_CONTAINER_IMAGE_* in Makefile)
+	podman run -v .:/app:z --rm -t $(CONTAINER_IMAGE) make test
+
+.PHONY: test
+test: ## run all tests locally
+	go test -race  ./...
+	# Run depsolver tests with force-dnf to make sure it's not skipped for any reason
+	go test -race ./pkg/dnfjson/... -force-dnf
+
+clean: ## remove all build files
+	rm -f container_built*.info
--- a/README.md
+++ b/README.md
@@ -1,6 +1,11 @@
 Images
 ======
 
+This repository is, primarily, a Go library for generating osbuild manifests
+([more details here](./docs/developer/code-manifest-generation.md)).
+It also has some libraries for uploading artifacts to cloud platforms and Koji.
+The binaries implemented in `cmd/` are for development and testing purposes and not part of the library.
+
 ## Project
 
  * **Website**: <https://www.osbuild.org>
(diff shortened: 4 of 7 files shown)


.github/workflows/tests.yml
```diff
@@ -24,36 +24,15 @@ jobs:
       fail-fast: false  # if one fails, keep the other(s) running
     name: "🛃 Unit tests (Fedora ${{ matrix.fedora_version }})"
     runs-on: ubuntu-latest
-    container:
```
mvo5, line 27: ~This change is not explained in the commit message AFAICT, why is this removed?~ Now that I read the entire commit I see what is happening. I personally would expand the commit message a bit and describe that this commit replaces the GH action container/run with a makefile target that also builds a container/runs things insides the container.

Containerfile
```diff
@@ -0,0 +1,24 @@
+ARG BASE_CONTAINER_IMAGE=registry.fedoraproject.org/fedora:latest
```
mvo5, line 1: The commit message talks about adding a Makefile, but there is now also a Containerfile, probably worth expanding the commit message a bit more.

Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+container_built.info: $(CONTAINERFILE) Schutzfile test/ go.mod go.sum # internal rule to build the container only if needed
+	$(CONTAINER_EXECUTABLE) build --build-arg BASE_CONTAINER_IMAGE="${BASE_CONTAINER_IMAGE}" \
+	                              --tag $(CONTAINER_IMAGE) \
+	                              -f $(CONTAINERFILE) .
+	echo "Container last built on" > $@
+	date >> $@
+
+.PHONY: test
+test: container_built.info ## run all tests
```
mvo5, line 25: I would be surprised if `make test` runs a container for me, maybe that is just me but I would expect more something like `go test` here. Maybe just a different makefile target (`make test-in-container`, `gh-testss`, `test-gh` or something)?

And/or two targets: "test" that just runs the test locally (which is I think what most ppl expect) and test-gh or something that puts them into the container?

mvo5 (commented):
Thanks, I like the idea - I was initially a bit confused as the described looked different from the content but after reading it in full I see what is going on. I still would tweak the description a bit and maybe title: `test: initial makefile, container, incoperate GH action` or something. *maybe* split even into two commits, not sure though but the fact that the title has so many things indicates it might be nice (but idk, just a hunch). Some more comments inline, hope they are useful.

Makefile
schuellerf: In general you are right. I think this is more something we can enrich at any point in time, not to enlarge this PR even more.
e.g. rename the implementation of this PR to a `gh-test` target and implement a `test` target without a container.

For me the most important thing is that `make test` just runs the tests and they should behave the same as the github tests as close as possible. (That was the whole plan of this issue)

I'd postpone the implementation of the second "test target" (testing without the container) to the future if we happen to need it.

mvo5: Sure, let's keep this PR targeted. I still feel a `make test` that creates/runs a container is not what I (and I assume most users) would expect. Calling it `make gh-test` would still be my suggestion. I am okay to be overridden by the more experienced images developers like @achilleas-k, @thozza or @ondrejbudai here of course.

schuellerf: for convenience are you ok with renaming it to `make gh-test` but still have a
```
# TBD implement checking/installing prerequisites and running the tests without a container
test: gh-test
```
and I'll check for a real `make test` after this PR (not to enrich this one)

schuellerf: … or even name it `test_in_container` or similar

achilleas-k: +1 for naming it `gh-test` or `gh-action-test` (indicating that it's replicating the action, not actually triggering a test on github or something).  `test_in_container` also works, but is an awkward name for a make target IMO.



> for convenience are you ok with renaming it to `make gh-test` but still have a
> 
> ```
> # TBD implement checking/installing prerequisites and running the tests without a container
> test: gh-test
> ```
> 
> and I'll check for a real `make test` after this PR (not to enrich this one)

No, I'd rather `make test` just did
```
go test -race ./...
go test -race ./pkg/dnfjson/... -force-dnf
```

and have it fail on systems that don't have dependencies.

achilleas-k:
A note on

> and have it fail on systems that don't have dependencies.

and

> # TBD implement checking/installing prerequisites and running the tests without a container

_Apologies if I'm making assumptions about the plan for the `test` Makefile target._

I get the appeal of being able to `git clone` the repo, run `make test` and have all the tests run as if by magic without needing to install dependencies and stuff.  But if this is where this is going, if the plan for a future `make test` is to have it install required dependencies on the developer's system, I have to question the utility of it all.

First of all, there's no scenario where we don't assume some prerequisites for users.  Does the container target install `podman` for the user as well?  Are we going to assume what distro they're running or have package lists for all major distributions?  Are we making it rpm only?  What about Fedora vs CentOS/RHEL (no btrfs driver).

Also, any check for dependencies will either need to check that they're installed, or run a no-op script that tries to install them every time `make test` is called, or maybe write a file in the repo like the container info one.  It seems like a lot of steps to avoid having to tell people "if you're going to compile this thing, you're going to need these packages", which is something that happens once per contributor/machine.

I don't think listing dependencies in the README and asking people who clone the repository to install some packages themselves is too much.  I get that a lot of things in this repository are a little hostile to newcomers, but I doubt build dependency installation is one of them.

schuellerf:
@achilleas-k you are right, although I like `Makefiles` that try to care about prerequisites as much as possible reducing the need to update ( / duplicate code in…) READMEs and keeping "infrastructure" or "install instructions" as code.
Similar to a README where also only rpm package names are mentioned (now), this is up to future contributors to just add the section/implementation for other distros.

But sure, this can get out of hands quickly. I'll just update the `README.md` now (as it's [out of date](https://github.com/osbuild/images/commit/80703211691975074b5b75fdd2a6ed17ae911d6d) already a while ago)
and finish this PR

.github/workflows/tests.yml
schuellerf: … enriched the commit message

Containerfile
schuellerf: … enriched the commit message

Makefile
schuellerf: please check the revised implementation and documentation

thozza: I agree with `make test` to just running tests locally and not in a container. One of the use cases may be to intentionally running them in a git checkout on a different distro, such as CentOS Stream or RHEL.

achilleas-k:
> But sure, this can get out of hands quickly. I'll just update the `README.md` now (as it's [out of date](https://github.com/osbuild/images/commit/80703211691975074b5b75fdd2a6ed17ae911d6d) already a while ago)
> and finish this PR

What's outdated in the README?  The "out of date" link opens an old commit for `tests.yaml`.

Which brings me to #959 and something I realised while writing that.
The github action is actually outdated.  The `krb5-devel` package isn't needed for unit tests.  It's needed by the koji uploader, but we don't test that (which is the actual problem, I guess).

The issue of "where should uploaders live" (images or osbuild-composer) is a long-standing question we haven't entirely figured out yet, so I wont go into it here.  But as it stands, right now, krb5-devel isn't a build nor a test dependency of osbuild/images.  It's a build dependency of any project that imports `pkg/upload/koji`.

thozza:
Besides for the action not working, this PR now looks good to me. :+1:

README.md
```diff
@@ -20,7 +20,17 @@ Images
...
-- `gpgme-devel`, `btrfs-progs-devel`, `device-mapper-devel`
+- `btrfs-progs-devel`
+- `device-mapper-devel`
+- `gcc`
+- `git-core`
+- `go`
+- `gpgme-devel`
+- `gpgme-devel`
+- `krb5-devel`
+- `osbuild-depsolve-dnf`
```
achilleas-k, line 31: We might need to be more clear here, or link to a document explaining how things work, so that users and developers of the library understand what's going on.

This repository is, primarily, a Go library for generating osbuild manifests.  It also has some libraries for uploading stuff to cloud platforms and Koji.  All binaries defined in `cmd/` are for development and testing purposes (at least for now, ignoring otk externals).  Given all that, the only _real_ build and test dependencies are:
- `go`
- `gpgme-devel`

Everything else on that list is either a build or runtime dependency of a public package.

- `btrfs-progs-devel`, `device-mapper-devel`: build dependencies for projects that import `pkg/container`, which even in that case can be skipped using `exclude_graphdriver_btrfs` and `exclude_graphdriver_devicemapper` (see bootc-image-builder).
- `gcc`: outdated.  Used to be we needed it to compile a mock dnf-json binary, but that's long gone.
- `git-core`: not a dependency.  Needed to clone the repository, which is why you might have seen it in some test setup scripts.
- `krb5-devel`: build dependency for projects that import `pkg/upload/koji`.  Some unit tests for the package would be good and, when we add them, `krb5-devel` becomes a test dependency.
- `osbuild-depsolve-dnf`: runtime dependency for projects that import `pkg/dnfjson`.

Now if we also account for `cmd/build` and `cmd/gen-manifests`:
- `osbuild-depsolve-dnf`: runtime dependency for `cmd/gen-manifests` and `cmd/build`.
- `osbuild` (and subpackages): runtime dependencies for `cmd/build`.

README.md
```diff
@@ -20,7 +20,17 @@ Images
...
 
 The build-requirements for Fedora and rpm-based distributions are:
-- `gpgme-devel`, `btrfs-progs-devel`, `device-mapper-devel`
+- `btrfs-progs-devel`
+- `device-mapper-devel`
+- `gcc`
+- `git-core`
+- `go`
+- `gpgme-devel`
+- `gpgme-devel`
```
thozza, line 29: This is here twice.

Makefile
```diff
@@ -0,0 +1,35 @@
...
+container_built_$(CONTAINER_IMAGE).info: $(CONTAINERFILE) Schutzfile test/ go.mod go.sum # internal rule to build the container only if needed
+	$(CONTAINER_EXECUTABLE) build --build-arg BASE_CONTAINER_IMAGE="${BASE_CONTAINER_IMAGE}" \
+	                              --tag $(CONTAINER_IMAGE) \
+	                              -f $(CONTAINERFILE) .
+	echo "Container last built on" > $@
+	date >> $@
+
+.PHONY: gh-action-test
+gh-action-test: container_built_$(CONTAINER_IMAGE).info ## run all tests in a container (see BASE_CONTAINER_IMAGE_* in Makefile)
+	podman run -v .:/app:z --rm -t osbuild-images make test
```
achilleas-k, line 26: This should be 
```suggestion
	podman run -v .:/app:z --rm -t "${CONTAINER_IMAGE}" make test
```
right?

achilleas-k (changes requested):

Makefile
```diff
@@ -0,0 +1,35 @@
+.PHONY: help
+help:
+	@echo 'Usage:'
+	@echo '  make <target>'
+	@echo ''
+	@echo 'Targets:'
+	@awk 'match($$0, /^([a-zA-Z_\/-]+):.*?## (.*)$$/, m) {printf "  \033[36m%-30s\033[0m %s\n", m[1], m[2]}' $(MAKEFILE_LIST) | sort
+
+BASE_CONTAINER_IMAGE_NAME?=registry.fedoraproject.org/fedora
+BASE_CONTAINER_IMAGE_TAG?=41
```
achilleas-k, line 10: Fedora 41 unit tests are failing right now because of tests trying to depsolve using dnf5 with an sbom request, which isn't supported.

Makefile
schuellerf: thanks

Makefile
schuellerf: This is overridden by the GH action anyway - but I can make fedora 40 the default, for now

README.md
schuellerf: Thanks, obviously I was also lacking this knowledge
maybe can can align directly where a good place for this knowledge is

Makefile
achilleas-k: > This is overridden by the GH action anyway

Sure, but when people run `make gh-action-test` locally, it will fail.

README.md
achilleas-k: Yes we should have this information somewhere in the repo.  If not in the README, then in a document linked from the README.

schuellerf: I'll try to document in the README somehow (not yet committed)

schuellerf: @achilleas-k please proof read what I understood.
I left `otk*` binaries aside on purpose and we should document them once settled (and not part of this PR)

achilleas-k (approved):
Thanks!  LGTM now.

Should we use the `make gh-action-test` target for the CentOS Stream tests as well?

schuellerf:
> Should we use the make gh-action-test target for the CentOS Stream tests as well?
I'll check the implications and do a followup-PR

schuellerf:
… rebased to main, lost approvals … :-/

achilleas-k (approved):

PR #955, opened 2024-09-27
go.mod: use go 1.22.0

achilleas-k:
#947 needed a go version update in the go.mod which dependabot didn't change.
After I added the change myself, I remembered it needs changes in more places.

Instead of polluting the dependabot PR with all this, let's do the change separately so it's more visible and I'll have dependabot rebase the PR after it's merged.

### DIFF
--- a/.github/workflows/gitlab-helper.yml
+++ b/.github/workflows/gitlab-helper.yml
@@ -18,10 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-      - name: Set up Go 1.21
+      - name: Set up Go 1.22
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21"
+          go-version: "1.22"
         id: go
 
       - name: Check out code into the Go module directory
--- a/.github/workflows/test-osbuild-composer-intergation.yml
+++ b/.github/workflows/test-osbuild-composer-intergation.yml
@@ -50,10 +50,10 @@ jobs:
     name: "⌨ osbuild-composer Golang Lint"
     runs-on: ubuntu-latest
     steps:
-      - name: Set up Go 1.21
+      - name: Set up Go 1.22
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21"
+          go-version: "1.22"
         id: go
 
       - name: Check out osbuild-composer main branch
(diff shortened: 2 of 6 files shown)


croissanne (approved):
stamp

achilleas-k:
Ok, updating golanglint-ci worked.  Now it's yelling at me because it found a bunch of things to complain about.  Better get fixing.

croissanne (approved):
stamp2

achilleas-k:
Hold off on merging, please.  I'm discussing the rng seed with @mvo5.

internal/cmdutil/rand.go
```diff
@@ -44,5 +44,6 @@ func SeedArgFor(bc *buildconfig.BuildConfig, imgTypeName, distributionName, arch
 	h.Write([]byte(imgTypeName))
 	h.Write([]byte(bc.Name))
 
+	// nolint:gosec
```
mvo5, line 47: Fwiw, I think this is fine, at this level we only need predictable bits for the RNG, the fact that it overflows does not matter, the number itself has no meaning (we could make it uint64 in the future but that is a lot of busywork without reallying giving us much)

mvo5 (approved):

lzap:
This breaks image builder which uses downstream Go Toolbox from UBI which sitting at 1.21.13 [at the moment](https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?architecture=amd64&image=66fd9412a88dd59af4f4836d&container-tabs=gti). I do not think this hard dependency can be updated to 1.22 or even 1.23 for tools yet:

https://github.com/osbuild/image-builder/pull/1363

achilleas-k:
Do you know if the next release of the UBI Go Toolbox will be 1.22?  Afaik, 9.5 will ship with Go 1.22.  And osbuild-composer in 9.5 will still be using 1.21 anyway, since the version for that is already frozen.

lzap:
I do not know. But we are unable to upgrade images in image builder until it happens. From prior history, 1.21.13 was released 2024-08-06 but updated in UBI 2024-10-03. That is 2 months lag, I suggest to hold on these upgrades in the library by pinning those dependencies which require new version of Go. Unfortunately, Go tools themselves does not have any features for that, what we did in the past was simply reinstalling particular versions after full update:

```
# update to latest and greatest
go get -u ./...

# pin specific libs
go get -u library/requiring/new/go@v1.2.3

# and modify the go.mod back to a reasonable version
go mod tidy
```

Something like that.

PR #956, opened 2024-09-27
test: build edge-commit with parent

achilleas-k:
We never had a test configuration where an ostree commit is built as child of another one.  Let's add a simple configuration where an ostree commit is built from an existing one.

The parent commit is the base image with no customizations (empty) and the child commit adds a user (edge-ostree-pull-user).

### DIFF
--- a/test/config-map.json
+++ b/test/config-map.json
@@ -89,7 +89,8 @@
   },
   "./configs/edge-ostree-pull-user.json": {
     "image-types": [
-      "edge-ami"
+      "edge-ami",
+      "edge-commit"
     ]
   },
   "./configs/minimal-environment.json": {


lzap:
Sure, thanks. Tho I am not exactly sure what this change does - is there some kind of generator you must run or something? Tests are failing, is this relevant?

I would like to have full end to end test, osrepo is just a plain directory served over HTTP so I could work on publishing some "example" repo that could be used by a test in the osbuild mirror perhaps?

achilleas-k:
> Sure, thanks. Tho I am not exactly sure what this change does - is there some kind of generator you must run or something?

Test setup and configuration is described here: https://github.com/osbuild/images/tree/main/test

> Tests are failing, is this relevant?
> 

We have some non-required reverse dependency tests with osbuild-composer.  They mostly serve as a reminder that updating osbuild/images in osbuild-composer will need work (it's not a clean update).  It would be great if they were warnings instead of failures, but that's not something GitHub offers.  Perhaps we can have it post a message on a PR when it causes the change (from working to non-working), so we don't have a list of PRs that are constantly red.

> I would like to have full end to end test, osrepo is just a plain directory served over HTTP so I could work on publishing some "example" repo that could be used by a test in the osbuild mirror perhaps?

That would be a good idea.  For every other type of source we have static resources: rpmrepo for RPM repository snapshots, and small toy containers for pulling containers in builds.  An static (or regularly/dynamically updated) ostree commit repo would be great.  The problem has always been that there was never a good way to host an ostree repository and especially one that you can _push_ into until Pulp gained support, which is quite recent.  With traditional rpm-ostree being replaced by bootc-based images, it might be tricky to justify the effort to set something like this up, but if you feel like doing it, by all means!

croissanne (approved):

supakeen:
> t would be great if they were warnings instead of failures, but that's not something GitHub offers. Perhaps we can have it post a message on a PR when it causes the change (from working to non-working), so we don't have a list of PRs that are constantly red.

This would be *really* nice; I always go back into my PRs to figure out why they're red and then see it's rawhide or integration :)

PR #963, opened 2024-10-01
otk-gen-partition-table: add support for ppc64/s390x partition tables

mvo5:
This commit adds support for "raw" partitions without a "payload",
i.e. no filesystem or LVM volume or similar. This is used for the
PPC64/s390x partition table that looks like this:
```yaml
otk.define:
  filesystem:
    modifications:
    # empty
    otk.external.otk-gen-partition-table:
      modifications:
        ${filesystem.modifications}
      properties:
        type: dos
        default_size: "10 GiB"
        uuid: "0x14fc63d2"
      partitions:
        - name: ppc-boot
          bootable: true
          size: "4 MiB"
          part_uuid: ""
          part_type: "41"
        - name: boot
          mountpoint: /boot
          label: boot
          size: "1 GiB"
          type: "xfs"
          fs_mntops: defaults
          part_uuid: ""
        - name: root
          mountpoint: /
          type: "xfs"
          size: "2 GiB"
          fs_mntops: defaults
          part_uuid: ""
```
Thanks to Florian Schüller for the initial implementation/research
here, see also https://github.com/osbuild/images/pull/952 

---

otk-gen-partition-table: add `bootable` flag support

This commit adds support to mark partitions as bootable from
an `otk-gen-partition-table` input. E.g.:
```yaml
{
...
  "partitions": [
    {
...
      "bootable": true,
...
    },
```

### DIFF
--- a/cmd/otk-gen-partition-table/main.go
+++ b/cmd/otk-gen-partition-table/main.go
@@ -51,6 +51,7 @@ type InputCreate struct {
 // InputPartition represents a single user provided partition input
 type InputPartition struct {
 	Name       string `json:"name"`
+	Bootable   bool   `json:"bootable"`
 	Mountpoint string `json:"mountpoint"`
 	Label      string `json:"label"`
 	Size       string `json:"size"`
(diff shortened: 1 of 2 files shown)


cmd/otk-gen-partition-table/main.go
```diff
@@ -183,21 +184,25 @@ func makePartitionTableFromOtkInput(input *Input) (*disk.PartitionTable, error)
...
-			Size: uintSize,
-			UUID: part.PartUUID,
-			Type: part.PartType,
-			// XXX: support lvm,luks here
-			Payload: &disk.Filesystem{
+		// XXX: support lvm,luks here
+		newPart := disk.Partition{
+			Size:     uintSize,
+			UUID:     part.PartUUID,
+			Type:     part.PartType,
```
schuellerf, line 191: Should we add some (inline) documentation what the difference between "Type" and "PartType" is?
This is actually the same, just to distinguish the two options, right?

supakeen: At the very least we'll have to decide when we bring the current documentation in otk up to par again with the state of the repository; I was planning on doing that when we start distributing otk.

supakeen (approved):

schuellerf (approved):

PR #964, opened 2024-10-01
cmd/otk-gen-partition-table: implement returning type in partition_map

schuellerf:
to use the fact "type" in the partition_map we need to return it
used at least in the s390x otk example

### DIFF
--- a/cmd/otk-gen-partition-table/main.go
+++ b/cmd/otk-gen-partition-table/main.go
@@ -86,15 +86,15 @@ func makePartMap(pt *disk.PartitionTable) map[string]otkdisk.Partition {
 	for _, part := range pt.Partitions {
 		switch pl := part.Payload.(type) {
 		case *disk.Filesystem:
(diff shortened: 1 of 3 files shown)


mvo5 (changes requested):
I assume you use this for https://github.com/osbuild/otk/pull/257/files#diff-4d2cf1ac3984b9fe677648f7d6d9c1dd13555ab5edf281c407579bc5239659fdR53 ? Please check the other partition tables, they all use 
```
  fs_options:
    otk.external.otk-make-partition-mounts-devices:
      ${filesystem}
```
which encapsulates these details. Setting to "request changes" to make sure this does not get merged accidentially as this adds API that we (probably) not want (at least yet)

schuellerf:
you are right here too!
I wasn't aware that the parameters `org.osbuild.zipl` expects are exactly the same as the other images. It's hard to see that without digging into the sourcecode before. I'll check how I can improve the visibility/documentation of the structures provided by the various externals or otk behavior

PR #967, opened 2024-10-02
coreos-installer: install only the required dracut modules

mmartinv:
Install only the minimal number of dracut modules required
by the coreos installer to avoid issues during the installation.

Resolves: fedora-iot/iot-distro#5

### DIFF
--- a/pkg/manifest/coreos_installer.go
+++ b/pkg/manifest/coreos_installer.go
@@ -185,11 +185,34 @@ func (p *CoreOSInstaller) serialize() osbuild.Pipeline {
 
 	dracutModules := append(
 		p.AdditionalDracutModules,
+		"systemd",
+		"systemd-initrd",
+		"fips",
+		"modsign",
+		"rescue",
+		"i18n",
+		"kernel-modules",
+		"kernel-modules-extra",
+		"network-manager",
+		"network",
+		"drm",
 		"coreos-installer",
 		"fdo",
+		"lvm",
+		"terminfo",
+		"fs-lib",
+		"dracut-systemd",
+		"debug",
+		"shutdown",
 	)
-
-	dracutStageOptions := dracutStageOptions(p.kernelVer, p.Biosdevname, dracutModules)
+	if p.Biosdevname {
+		dracutModules = append(dracutModules, "biosdevname")
+	}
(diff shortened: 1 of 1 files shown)


pkg/distro/fedora/package_sets.go
```diff
@@ -680,6 +680,7 @@ func iotSimplifiedInstallerPackageSet(t *imageType) rpmmd.PackageSet {
 			"sudo",
 			"traceroute",
 			"util-linux",
+			"util-linux-core",
```
mvo5, line 683: Maybe (probably?) I'm missing something, but isn't `util-linux-core` a hard depedency of `util-linux`, i.e. don't we always install this automatically anyway?

mmartinv: Damn, no, you are not missing anything, `util-linux-core` is a hard dependency of `util-linux` so it must be installed anyway. I think the problem appeared when I was "playing" with dracut modules to see which of them where not required. I guess that in one of my attempts I removed `dracut-systemd` from the list which is the one installing `sulogin`

I will remove the commit from the PR, sorry

achilleas-k (approved):

achilleas-k:
Thanks you!!

achilleas-k:
Looks like RHEL 8 doesn't like the dracut module changes:
```
dracut: Executing: /usr/bin/dracut --no-hostonly --kver 4.18.0-372.9.1.el8.aarch64 --force -v --show-modules --no-early-microcode --reproducible --modules "prefixdevname prefixdevname-tools systemd-initrd fips modsign rescue i18n kernel-modules kernel-modules-extra network-manager network drm coreos-installer fdo lvm terminfo fs-lib dracut-systemd debug shutdown" --install /.buildstamp
dracut: systemd-initrd needs systemd in the initramfs
dracut: dracut-systemd needs systemd-initrd in the initramfs
dracut: dracut module 'systemd-initrd' cannot be found or installed.
```

achilleas-k (approved):
:rocket:

achilleas-k:
Something went wrong on the merge queue.  Looks like a flake on first inspection.  No details on the failed job (it was a dynamic trigger job).
Requeued.  Will keep an eye on it.

PR #974, opened 2024-10-10
workflow: generate PR comments when non-required tests fail and let the action itself succeed (part 1)

achilleas-k:
The purpose of this new workflow is to run when certain actions are allowed to cause errors without failing but PR authors and reviewers should be notified of issues.  The goal is, instead of having actions that are allowed to fail and are marked as not required, the actions themselves will succeed but will communicate an error message through a shared artifact that this action can read and use to post a comment on the PR.

The action is run as a 'workflow_run' to have access to secrets (Schutzbot's access token).

The action is not finalised in this commit.

The plan is to merge this into `main` as is, so that GitHub will pick up the new workflow and we can iterate on the implementation in a follow-up PR, since GitHub will not run actions defined in workflows that are not part of the main branch but will run modified versions of a pre-existing action.

### DIFF
--- a/.github/workflows/comment-integration-failure.yml
+++ b/.github/workflows/comment-integration-failure.yml
@@ -1,0 +1,29 @@
+---
+name: "Schutzbot comments"
+# This workflow runs when the [integration] workflow reports a failure through
+# an artifact. The [integration] workflow itself doesn't fail if the testing or
+# linting fails, instead it reports a failure through a shared artifact that
+# this workflow reads and reports on. The failure is communicated to the author
+# and reviewers of a PR without causing a failed action to appear.
+#
+# This workflow is run on a workflow_run trigger to have access to secrets.
+#
+# TODO: Actually do what the comment says.
+
+on:
+  workflow_run:
+    workflows: ["[integration]"]
+    types: [completed]
+
+jobs:
+  on-failure:
+    runs-on: ubuntu-latest
+
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+      - name: Add comment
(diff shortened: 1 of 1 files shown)


.github/workflows/comment-integration-failure.yml
```diff
@@ -0,0 +1,29 @@
...
+name: "Schutzbot comments"
+# This workflow runs when the [integration] workflow reports a failure through
+# an artifact. The [integration] workflow itself doesn't fail if the testing or
+# linting fails, instead it reports a failure through a shared artifact that
+# this workflow reads and reports on. The failure is communicated to the author
+# and reviewers of a PR without causing a failed action to appear.
+#
+# This workflow is run on a workflow_run trigger to have access to secrets.
+#
+# TODO: Actually do what the comment says.
```
mvo5, line 11: (nitpick) the comment could be a bit clearer, i.e. something like "//TODO: add integration workflows once this is merge to main ..." or something but the commit message explains the plan so it's a minor thing

mvo5 (approved):
The title could be of the form "workflow: generate PR comments..." like we do for most other PRs. I approve anyway as these are minor points.

supakeen (approved):
Very cool.

.github/workflows/comment-integration-failure.yml
achilleas-k: I really hope the TODO comment doesn't stay around long enough for it to matter, but you're right.  I can change it if you really feel it's important.

mvo5: No, nitpick only :) Let's get it merged

PR #541, opened 2024-03-22
support org.osbuild.mtls secret in curl source (HMS-3798)

croissanne:
Support the mtls secret in curl sources. If the worker sets mtls data on non-rhsm repos, make sure we add the mtls secret.

PR #975, opened 2024-10-10
ostree: configurable MTLS config for ostree resolve

lzap:
To complete edge service Pulp migration, we need to be able to resolve and download ostree commits from Pulp running behind `console.redhat.com` proxy which only supports MTLS for authentication. I have found that some MTLS support is already present but it does not make much sense why RHSM is used. I cannot figure out a workflow why ostree repo would be published via RHSM cert. I am not aware of any RH ostree content hosted via RH CDN.

_The patch quite evolved during the review so editing the description so it is more appropriate._

### DIFF
--- a/cmd/otk/osbuild-resolve-ostree-commit/main.go
+++ b/cmd/otk/osbuild-resolve-ostree-commit/main.go
@@ -49,7 +49,11 @@ func run(r io.Reader, w io.Writer) error {
 		return err
 	}
 
-	sourceSpec := ostree.SourceSpec(inputTree.Tree)
+	sourceSpec := ostree.SourceSpec{
+		URL:  inputTree.Tree.URL,
+		Ref:  inputTree.Tree.Ref,
+		RHSM: inputTree.Tree.RHSM,
+	}
 
 	var commitSpec ostree.CommitSpec
 	if !underTest() {
--- a/pkg/ostree/ostree.go
+++ b/pkg/ostree/ostree.go
@@ -141,7 +141,7 @@ func verifyChecksum(commit string) bool {
 // ResolveRef resolves the URL path specified by the location and ref
 // (location+"refs/heads/"+ref) and returns the commit ID for the named ref. If
 // there is an error, it will be of type ResolveRefError.
-func ResolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptions, ca *string) (string, error) {
+func resolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptions, ca *string) (string, error) {
 	u, err := url.Parse(location)
 	if err != nil {
 		return "", NewResolveRefError("error parsing ostree repository location: %v", err)
@@ -252,7 +252,7 @@ func Resolve(source SourceSpec) (CommitSpec, error) {
 	// URL set: Resolve checksum
 	if source.URL != "" {
 		// If a URL is specified, we need to fetch the commit at the URL.
-		checksum, err := ResolveRef(source.URL, source.Ref, source.RHSM, nil, nil)
+		checksum, err := resolveRef(source.URL, source.Ref, source.RHSM, nil, nil)
 		if err != nil {
 			return CommitSpec{}, err // ResolveRefError
 		}
--- a/pkg/ostree/ostree.go
+++ b/pkg/ostree/ostree.go
@@ -25,9 +25,21 @@ var (
 // SourceSpec serves as input for ResolveParams, and contains all necessary
 // variables to resolve a ref, which can then be turned into a CommitSpec.
 type SourceSpec struct {
-	URL  string
-	Ref  string
+	URL string
+	Ref string
+	// RHSM indicates to use RHSM secrets when pulling from the remote. Alternatively, you can use MTLS with plain certs.
 	RHSM bool
+	// MTLS information. Will be ignored if RHSM is set.
+	MTLS *MTLS
+	// Proxy as HTTP proxy to use when fetching the ref.
+	Proxy string
+}
+
+// MTLS contains the options for resolving an ostree source.
+type MTLS struct {
+	CA         string
+	ClientCert string
+	ClientKey  string
 }
 
 // CommitSpec specifies an ostree commit using any combination of Ref (branch), URL (source), and Checksum (commit ID).
@@ -138,59 +150,53 @@ func verifyChecksum(commit string) bool {
 	return len(commit) > 0 && ostreeCommitRE.MatchString(commit)
 }
 
-// ResolveRef resolves the URL path specified by the location and ref
+// resolveRef resolves the URL path specified by the location and ref
 // (location+"refs/heads/"+ref) and returns the commit ID for the named ref. If
 // there is an error, it will be of type ResolveRefError.
-func resolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptions, ca *string) (string, error) {
-	u, err := url.Parse(location)
+func resolveRef(ss SourceSpec) (string, error) {
+	u, err := url.Parse(ss.URL)
 	if err != nil {
 		return "", NewResolveRefError("error parsing ostree repository location: %v", err)
 	}
-	u.Path = path.Join(u.Path, "refs/heads/", ref)
+	u.Path = path.Join(u.Path, "refs/heads/", ss.Ref)
 
-	var client *http.Client
-	if consumerCerts {
-		if subs == nil {
-			subs, err = rhsm.LoadSystemSubscriptions()
-			if err != nil {
-				return "", NewResolveRefError("error adding rhsm certificates when resolving ref: %s", err)
-			}
-			if subs.Consumer == nil {
-				return "", NewResolveRefError("error adding rhsm certificates when resolving ref")
-			}
-		}
-
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	client := &http.Client{
+		Transport: transport,
+		Timeout:   300 * time.Second,
+	}
+	if u.Scheme == "https" {
 		tlsConf := &tls.Config{
 			MinVersion: tls.VersionTLS12,
 		}
 
-		if ca != nil {
-			caCertPEM, err := os.ReadFile(*ca)
+		// If CA is set, load the CA certificate and add it to the TLS configuration. Otherwise, use the system CA.
+		if ss.MTLS.CA != "" {
+			caCertPEM, err := os.ReadFile(ss.MTLS.CA)
 			if err != nil {
-				return "", NewResolveRefError("error adding rhsm certificates when resolving ref: %s", err)
+				return "", NewResolveRefError("error adding ca certificate when resolving ref: %s", err)
 			}
-			roots := x509.NewCertPool()
-			ok := roots.AppendCertsFromPEM(caCertPEM)
-			if !ok {
-				return "", NewResolveRefError("error adding rhsm certificates when resolving ref")
+			tlsConf.RootCAs = x509.NewCertPool()
+			if ok := tlsConf.RootCAs.AppendCertsFromPEM(caCertPEM); !ok {
+				return "", NewResolveRefError("error adding ca certificate when resolving ref")
 			}
-			tlsConf.RootCAs = roots
 		}
 
-		cert, err := tls.LoadX509KeyPair(subs.Consumer.ConsumerCert, subs.Consumer.ConsumerKey)
-		if err != nil {
-			return "", NewResolveRefError("error adding rhsm certificates when resolving ref: %s", err)
+		if ss.MTLS.ClientCert != "" && ss.MTLS.ClientKey != "" {
+			cert, err := tls.LoadX509KeyPair(ss.MTLS.ClientCert, ss.MTLS.ClientKey)
+			if err != nil {
+				return "", NewResolveRefError("error adding client certificate when resolving ref: %s", err)
+			}
+			tlsConf.Certificates = []tls.Certificate{cert}
 		}
-		tlsConf.Certificates = []tls.Certificate{cert}
 
-		client = &http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: tlsConf,
-			},
-			Timeout: 300 * time.Second,
+		transport.TLSClientConfig = tlsConf
+	}
+
+	if ss.Proxy != "" {
+		transport.Proxy = func(request *http.Request) (*url.URL, error) {
+			return url.Parse(ss.Proxy)
 		}
-	} else {
-		client = &http.Client{}
 	}
 
 	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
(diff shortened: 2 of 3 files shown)


cmd/ostree-resolve/main.go
```diff
@@ -0,0 +1,25 @@
...
+	spec := ostree.SourceSpec{
+		URL: "https://builder.home.lan/ccb2194f-9876-4e76-9e64-a338a32df230/",
+		Ref: "fedora/40/x86_64/iot",
+	}
+	cs, err := ostree.Resolve(spec)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Printf("Resolved checksum: %s", cs.Checksum)
+}
```
lzap, line 25: This is a test program I used to test the feature. Ah, I see I left my home `URL` by accident, will remove this this should be configurable.

croissanne (changes requested):
The mtls secrets need to be passed in the same way as https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-worker/jobimpl-depsolve.go.

So they're set in the [worker config](https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-worker/config.go), and then just pass them along to the ostree resolve job.

cmd/ostree-resolve/main.go
```diff
@@ -0,0 +1,36 @@
...
+	proxy, err := url.Parse(os.Getenv("OSBUILD_SOURCES_OSTREE_PROXY"))
+	if err != nil {
+		panic(err)
+	}
+	cs, err := ostree.Resolve(
+		spec,
+		ostree.WithResolveCA(os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CA_CERT")),
+		ostree.WithResolveMTLSClient(os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CLIENT_CERT"), os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CLIENT_KEY")),
+		ostree.WithResolveProxy(proxy),
+	)
```
lzap, line 31: The `Resolve` function is kept compatible with the old codebase in composer, the options are completely optional therefore passing just spec will work and in fact not use any MTLS configuration.

pkg/ostree/ostree.go
```diff
@@ -141,56 +134,46 @@ func verifyChecksum(commit string) bool {
 // ResolveRef resolves the URL path specified by the location and ref
 // (location+"refs/heads/"+ref) and returns the commit ID for the named ref. If
 // there is an error, it will be of type ResolveRefError.
-func ResolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptions, ca *string) (string, error) {
+func ResolveRef(location, ref, caCert, cert, key string, proxy *url.URL) (string, error) {
```
lzap, line 137: Possible API break but `composer` does not appear to use this directly.

pkg/ostree/ostree.go
```diff
@@ -228,16 +211,17 @@ func ResolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptio
 // resolved or checked against the repository.
 //
 // If the ref is malformed, the function returns with a RefError.
-func Resolve(source SourceSpec) (CommitSpec, error) {
+func Resolve(source SourceSpec, opt ...ResolveOptionFn) (CommitSpec, error) {
+	options := ResolveOptions{}
+	for _, o := range opt {
+		o(&options)
+	}
```
lzap, line 218: And as I stated before, this is backwards compatible change.

lzap:
```
CentOS Stream 10 - BaseOS                       647  B/s | 3.9 kB     00:06    
Errors during downloading metadata for repository 'baseos':
  - Downloading successful, but checksum doesn't match. Calculated: 09f9f3040a68ccceb349f04463cad23b3af1dc36ccf61c1fada52b9955a441742d8434dcb66f265924aed2d2b9dee3edaf16e0534fe475c6a5faf45db021d9a1(sha512)  Expected: f3765f12125ff6f09532cd5e74a0ab0bf0b275aa28794c1171b11ed247ed46fe60a7869d2ddc4658e31ac155632b4def908fd473050e6f68c32322212b921f62(sha512) 
Error: Failed to download metadata for repo 'baseos': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

pkg/ostree/ostree.go
```diff
@@ -218,6 +210,33 @@ func ResolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptio
 	return checksum, nil
 }
 
+// ResolveRef is deprecated, use ResolveRefMTLS.
+func ResolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptions, ca *string) (string, error) {
```
achilleas-k, line 214: Just break the API.  The mess of keeping deprecated functions around isn't worth the effort.

pkg/ostree/ostree.go
```diff
@@ -25,8 +25,9 @@ var (
 // SourceSpec serves as input for ResolveParams, and contains all necessary
 // variables to resolve a ref, which can then be turned into a CommitSpec.
 type SourceSpec struct {
-	URL  string
-	Ref  string
+	URL string
+	Ref string
+	// RHSM is deprecated, use Resolve function with MTLS configuration
```
achilleas-k, line 30: Could the mTLS configuration be added to the SourceSpec?  I imagine that would make things simpler, but I haven't written it out to be sure.

pkg/ostree/ostree.go
```diff
@@ -228,16 +247,17 @@ func ResolveRef(location, ref string, consumerCerts bool, subs *rhsm.Subscriptio
 // resolved or checked against the repository.
 //
 // If the ref is malformed, the function returns with a RefError.
-func Resolve(source SourceSpec) (CommitSpec, error) {
+func Resolve(source SourceSpec, opt ...ResolveOptionFn) (CommitSpec, error) {
+	options := ResolveOptions{}
+	for _, o := range opt {
+		o(&options)
+	}
```
achilleas-k, line 254: This is a very odd pattern for something that could just be an extension of the `SourceSpec` struct.

achilleas-k (changes requested):

pkg/ostree/ostree.go
lzap: How about I do a followup PR once this is integrated into composer? Because if I remove this function tests immediately breaks, I feel like I would make my life harder by straight-up breaking it.

achilleas-k: Which tests break?  The integration ones with composer?  Those are non-required for a reason.  They're meant as a warning to the PR creator that they will have to change osbuild-composer when updating.

> How about I do a followup PR once this is integrated into composer?

I'd rather have one clean PR than two messy ones.  The way it is now it's hard to review (though not too hard).  I suspect that a single PR that makes the appropriate change would be easier for me to wrap my head around, rather than try to imagine what it will look like _eventually_.

pkg/ostree/ostree.go
lzap: Yeah well I thought that `SourceSpec` is a type I should not touch. Now that I think about it, I can add fields directly in there that is more simple.

pkg/ostree/ostree.go
lzap: Yeah well I thought that `SourceSpec` is a type I should not touch. Now that I think about it, I can add fields directly in there that is more simple. Let me try.

lzap:
Oh when working on the changes you proposed, I realized I completely missed this input type which maps to the `SourceSpec`:

```go
// Input represents the user-provided inputs that will be used to resolve an
// ostree commit ID.
type Input struct {
	// URL of the repo where the commit can be fetched.
	URL string `json:"url"`

	// Ref to resolve.
	Ref string `json:"ref"`

	// Whether to use RHSM secrets when resolving and fetching the commit.
	RHSM bool `json:"rhsm,omitempty"`
}
```

So it was not complete at all, I fixed it. How about this?

The solution now properly reads [these ENV variables](https://github.com/osbuild/osbuild-composer/pull/4412/files), I hope they are available.

lzap:
Okay I am breaking the API here dropping the RHSM field, secret and creating a new `Connection` field that can be used to pass CA, client cert and proxy. How about this? @achilleas-k

lzap:
So I have rewritten the patch to match the merged `osbuild` PR. I will drop few comments.

cmd/build/main.go
```diff
@@ -119,6 +119,12 @@ func resolvePipelineCommits(commitSources map[string][]ostree.SourceSpec) (map[s
 		commitSpecs := make([]ostree.CommitSpec, len(commitSources))
 		for idx, commitSource := range commitSources {
 			var err error
+			commitSource.MTLS = &ostree.MTLS{
+				CA:         os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CA_CERT"),
+				ClientCert: os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CLIENT_CERT"),
+				ClientKey:  os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CLIENT_KEY"),
+			}
+			commitSource.Proxy = os.Getenv("OSBUILD_SOURCES_OSTREE_PROXY")
```
lzap, line 127: In all three development command line utilities, I am leveraging the same ENV variables which are used to do actual ostree pull.

Note that all the code here is only for the commit resolve HTTP(S) request during manifest build.

cmd/otk/osbuild-resolve-ostree-commit/main.go
```diff
@@ -49,7 +49,16 @@ func run(r io.Reader, w io.Writer) error {
...
+	var sourceSpec ostree.SourceSpec
+	sourceSpec.URL = inputTree.Tree.URL
+	sourceSpec.Ref = inputTree.Tree.Ref
+	sourceSpec.RHSM = inputTree.Tree.RHSM
+	sourceSpec.MTLS = &ostree.MTLS{
+		CA:         os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CA_CERT"),
+		ClientCert: os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CLIENT_CERT"),
+		ClientKey:  os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CLIENT_KEY"),
+	}
+	sourceSpec.Proxy = os.Getenv("OSBUILD_SOURCES_OSTREE_PROXY")
```
lzap, line 61: Here I am not sure if I should actually change the OTK schema and add these as proper fields. Let me know.

pkg/distro/fedora/images.go
```diff
@@ -815,6 +816,12 @@ func makeOSTreeParentCommit(options *ostree.ImageOptions, defaultRef string) (*o
 		URL:  options.URL,
 		Ref:  parentRef,
 		RHSM: options.RHSM,
+		MTLS: &ostree.MTLS{
+			CA:         os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CA_CERT"),
+			ClientCert: os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CLIENT_CERT"),
+			ClientKey:  os.Getenv("OSBUILD_SOURCES_OSTREE_SSL_CLIENT_KEY"),
+		},
+		Proxy: os.Getenv("OSBUILD_SOURCES_OSTREE_PROXY"),
```
lzap, line 824: This is something I am not sure how to do - for the actual fedora/rhel manifest build, where should I read the configuration for MTLS to do ostree commit resolving? For now I am proposing ENV variables, but I am not sure. From what I know, this library is used in composer and bib, the latter is not relevant so only in composer. Let me know how to properly pass the MTLS configuration.

croissanne: You don't set them at this point, they'll be set by the worker when calling osbuild https://github.com/osbuild/osbuild-composer/blob/6ccfc7f81884eedbf3290bb2118d0c6e31c676d5/cmd/osbuild-worker/jobimpl-osbuild.go#L519-L528.

cmd/build/main.go
croissanne: I don't think you need to use any ENV vars in images, take a look at https://github.com/osbuild/images/pull/541/files, something similar should work.

lzap:
So I have vastly simplified the patch after our discussion on Slack. I am no longer touching `cmd/` or `pkg/distro`. The `ostree.SourceSpec` type was extened with `*MTLS` and `Proxy`. MTLS info cannot be set together with RHSM, callers must choose one or the other. This allows to provide arbitrary MTLS certs or RHSM certs. Secrets are set accordingly.

croissanne (approved):
LGTM! very nice

pkg/ostree/ostree.go
```diff
@@ -25,9 +25,21 @@ var (
...
+	// Proxy as HTTP proxy to use when fetching the ref.
+	Proxy string
+}
+
+// MTLS contains the options for resolving an ostree source.
+type MTLS struct {
+	CA         string
+	ClientCert string
+	ClientKey  string
 }
```
lzap, line 43: One thing I am not sure about is to pull Proxy into the MTLS type, it would have been cleaner but Proxy has nothing to do with MTLS technically...

pkg/ostree/ostree_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/osbuild/images/pkg/rhsm"
 )
 
-func TestOstreeResolveRef(t *testing.T) {
+func TestOstreeresolveRef(t *testing.T) {
```
achilleas-k, line 16: Unnecessary rename but I imagine it got caught in the `sed`ing ^_^

pkg/ostree/ostree.go
```diff
@@ -138,59 +150,53 @@ func verifyChecksum(commit string) bool {
 	return len(commit) > 0 && ostreeCommitRE.MatchString(commit)
 }
 
-// ResolveRef resolves the URL path specified by the location and ref
+// resolveRef resolves the URL path specified by the location and ref
```
achilleas-k, line 153: Nitpick: This should be part of the first commit.

pkg/ostree/ostree.go
```diff
@@ -234,8 +240,30 @@ func Resolve(source SourceSpec) (CommitSpec, error) {
...
+			return commit, NewResolveRefError("error adding rhsm certificates when resolving ref: %s", err)
+		}
+
+		if subs.Consumer == nil {
+			return commit, NewResolveRefError("error adding rhsm certificates when resolving ref")
+		}
+
+		source.MTLS = &MTLS{}
+		source.MTLS.ClientCert = subs.Consumer.ConsumerCert
+		source.MTLS.ClientKey = subs.Consumer.ConsumerKey
```
achilleas-k, line 264: This would be more readable if the struct was assigned directly:
```go
source.MTLS = &MTLS{
  ClientCert: subs.Consumer.ConsumerCert
  ClientKey: subs.Consumer.ConsumerKey
}
```

pkg/ostree/ostree.go
```diff
@@ -234,8 +240,30 @@ func Resolve(source SourceSpec) (CommitSpec, error) {
...
+	} else if source.MTLS != nil {
+		commit.Secrets = "org.osbuild.mtls"
```
achilleas-k, line 266: Would it make sense to make it an error when both MTLS and RHSM are set?  It's fine to have a documented priority like we do now, but maybe making it a hard error would be better for troubleshooting unexpected behaviour.

cmd/otk/osbuild-resolve-ostree-commit/main.go
```diff
@@ -49,7 +49,10 @@ func run(r io.Reader, w io.Writer) error {
 		return err
 	}
 
-	sourceSpec := ostree.SourceSpec(inputTree.Tree)
+	sourceSpec := ostree.SourceSpec{}
+	sourceSpec.URL = inputTree.Tree.URL
+	sourceSpec.Ref = inputTree.Tree.Ref
+	sourceSpec.RHSM = inputTree.Tree.RHSM
```
achilleas-k, line 55:
```suggestion
sourceSpec := ostree.SourceSpec{
	URL: inputTree.Tree.URL,
	Ref: inputTree.Tree.Ref,
	RHSM: inputTree.Tree.RHSM,
}
```

(again, nitpick)

achilleas-k (approved):
Thanks!  LGTM.

Couple of nitpicks and comments, but nothing worth blocking the PR over.  Feel free to implement the suggested changes or ignore them if you don't agree.

pkg/ostree/ostree.go
lzap: This is already there see few lines above:

```
	if source.RHSM && source.MTLS != nil {
		return commit, NewResolveRefError("cannot use both RHSM and MTLS when resolving ref")
	}
```

cmd/otk/osbuild-resolve-ostree-commit/main.go
lzap: Is absolutely nicer way of writing that and I would do that, but this is like 5th version of this patch so I kept moving things around and this was the reasult. Fixed both.

lzap:
Amended all your comments.

lzap:
The SBOM error is not related, integration tests are failing that is expected. What is the next step?

achilleas-k (approved):

PR #976, opened 2024-10-11
Pin Go version to 1.21

lzap:
The library is now dependency in image builder and we agreed on Slack that although it is just two functions from the crypt package, it makes sense to share this code (and from otk), therefore we must do the effort of not upgrading the Go version to latest and greatest but keep the version in check with UBI containers (1.21.x currently).

All dependant bot updates which will fail the build must be closed or kept in a queue until Go can be upgraded few months later.


lzap:
_SBOM support for DNF5 is not implemented_

Is this relevant?

lzap:
SBOM failures again.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/osbuild/images
 
-go 1.21.0
+go 1.22.0
```
achilleas-k, line 3: This second commit reverts the revert in the go.mod.

achilleas-k:
> SBOM failures again.

Not required to merge.

go.mod
lzap: Wait what? Uh.

lzap:
So I [wrote a small utility](https://github.com/lzap/gobump) which allows bumping of dependencies while keeping the Go requirement. It allowed me to bump all dependencies which can be updated and keep Go 1.21.

lzap:
@achilleas-k ready for re-review

achilleas-k (approved):

PR #977, opened 2024-10-11
workflow: generate PR comments when non-required tests fail and let the action itself succeed (part 2)

achilleas-k:


supakeen:
:broken:

achilleas-k:
hm... didn't trigger

achilleas-k:
I set up [a test repo and opened a PR](https://github.com/achilleas-k/pr-comment-experiment/pull/1) to check how it would work and well things get a bit tricky.  Since we're running on a `workflow_run`, the issue/PR number isn't immediately available.  But we can pass it over with an artifact, which I was planning to do anyway to communicate the status of the lint and test jobs.  [This page](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#using-data-from-the-triggering-workflow) explains how that's done, but it seems a bit convoluted.  Either way, it's official github docs, so I guess that's how we'll do it.

achilleas-k:
See #980

PR #980, opened 2024-10-14
github: remove workflow "Schutzbot comments"

achilleas-k:
This workflow was with the plan to make it post a comment on a PR when certain non-required checks failed, without failing the workflows themselves.  However, this turns out to require a bit more work than originally anticipated and is probably not worth the effort right now.

It might be worth waiting for github to add support for something like this, or we could make a github action for ourselves that abstracts this idea to make it reusable across projects.

For now, I'd like to remove the workflow so it doesn't stay in the repo unfinished as clutter.

This reverts commit 0fc99d2ebb8e8d2e7f66a0f1f88203ffdc063b4e.


supakeen (approved):
Sad approval, do we have a ticket somewhere to track if we can retry it sometime?

achilleas-k:
> Sad approval, do we have a ticket somewhere to track if we can retry it sometime?

No.  It was just an idea I had that I thought would be quick to solve.  I'll open a GH issue in the hopes we don't forget this down the line.

PR #981, opened 2024-10-14
pkg: extra new `datasizes` package for `DataSizeToUint64`

mvo5:
This commit extracts the `common.DataSizeToUint64()` helper as `datasizes.Parse` and the `common.MiB` to `datasizes.MiB` (etc).

This is required so that we can import the datasizes package in the upcoming blueprints git module.

This will be needed for https://github.com/mvo5/blueprints (well, the "osbuild/blueprints" of course - with this the "blueprints" package that consists of pkg/{blueprints,customizations} is self sufficient. We may still need to look closer as there is a bunch of logic and imports (like pathpolicy ) that we need to get rid of.

### DIFF
--- a/cmd/osbuild-playground/playground.go
+++ b/cmd/osbuild-playground/playground.go
@@ -6,7 +6,7 @@ import (
 	"os"
 	"path"
 
-	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/pkg/datasizes"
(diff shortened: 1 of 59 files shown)


supakeen (approved):

achilleas-k (commented):
Neat!

achilleas-k (approved):

PR #985, opened 2024-10-16
many: extract newGCETarPipelineForImg() helper

mvo5:
This commit extracts a common helper to setup the right tar options for the GCE tar file generation. The options that need to be passed to tar for GCE need to be exactly right so let's make sure we only have a single place where they are set.

### DIFF
--- a/pkg/image/gce_test.go
+++ b/pkg/image/gce_test.go
@@ -1,0 +1,42 @@
+package image
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/images/pkg/osbuild"
+	"github.com/osbuild/images/pkg/rpmmd"
+	"github.com/osbuild/images/pkg/runner"
+)
+
+func TestNewGCETarPipeline(t *testing.T) {
+	for _, tc := range []struct {
+		imgFilename       string
+		expectedTransform string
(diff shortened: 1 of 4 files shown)


pkg/image/gce_test.go
```diff
@@ -0,0 +1,33 @@
...
+func TestNewGCETarPipeline(t *testing.T) {
+	var repos []rpmmd.RepoConfig
+	m := &manifest.Manifest{}
+	runner := &runner.Fedora{}
+
+	buildPipeline := manifest.NewBuild(m, runner, repos, nil)
+	buildPipeline.Checkpoint()
+
+	imgPipeline := manifest.NewRawImage(buildPipeline, nil)
+	imgPipeline.SetFilename("foo.img")
```
bcl, line 23: May as well also have a test for `disk.raw` just to be sure.
Looks good to me though, other than the merge conflict :)

bcl (changes requested):

pkg/image/gce_test.go
mvo5: Thanks, excellent point. I updated this now to test both cases (transform and no-transform).

bcl (commented):
Looks ok to me, but the f41 unit test is failing with odd looking metalink urls, I'm not sure why so I restarted it.

thozza (approved):
LGTM, Thanks!

PR #987, opened 2024-10-17
osbuild: Rename GCE output from .tgz to .tar.gz

pastequo:
Import with `gcloud` cli requires the file to ends with `.tar.gz`, otherwise the command fails

```
gcloud compute images create --help
[...]
       --source-uri=SOURCE_URI
          The full Cloud Storage URI where the disk image is stored. This file
          must be a gzip-compressed tarball whose name ends in .tar.gz. For
          more information about Cloud Storage URIs, see
          https://cloud.google.com/storage/docs/request-endpoints#json-api.
```


pastequo:
Also updated the test in bib: https://github.com/osbuild/bootc-image-builder/pull/679

mvo5 (approved):
Thank you! I And sorry that I missed this.

PR #997, opened 2024-10-28
cmd: add new image-builder binary and initial `list-distros` 

mvo5:
[draft as this is sill WIP and questions like where to get the authorative list of distros/repos and if we should embed them via "embed:" to have a simpler distribution mechanism etc]

This commit adds a new `image-builder` binary that will be capable
to build images standalone (daemonless). There are various use-cases
for this, the most important one is CI/CD pipelines.

This binary should eventually also replace the existing "cmd/list-images" and
"cmd/build" binaries (and potentially more) and it will cherry pick a lot of the
functionality from there. I decided against starting from one of the exiting binaries
because it allows a slightly cleaner start but this work is early so we could still
reconsider if people feel strongly it should be done the other way around.


cmd/image-builder/manifest.go
```diff
@@ -30,29 +30,15 @@ func depsolve(cacheDir string, packageSets map[string][]rpmmd.PackageSet, d dist
...
 
 func outputManifest(out io.Writer, distroName, imgTypeStr, archStr string) error {
-	filterExprs := []string{
-		fmt.Sprintf("name:%s", distroName),
-		fmt.Sprintf("arch:%s", archStr),
-		fmt.Sprintf("type:%s", imgTypeStr),
-	}
-	filteredResults, err := getFilteredImages(filterExprs)
+	// XXX: what/how much do we expose here?
+	var options distro.ImageOptions
```
thozza, line 34: For on-prem, we use `ImageOptions` for the stuff that is passed on the CLI and is not part of the BP. E.g. image size.

achilleas-k: This is something we need to think about.  The image options, like Tomáš said, is (mostly) for things that come from the command line.  This includes:
- image size
- ostree options (url, ref, and parent, when using `composer-cli compose start-ostree`).

https://github.com/osbuild/osbuild-composer/blob/a7cd521325504a43d3708f8f91ab63999862b279/internal/weldr/api.go#L2549-L2556

But it's also used for other things that get set automatically (or semi-automatically):
- Facts: this shouldn't be exposed at all.  It adds a value to the RHSM facts so we can see how the image was built when it registers back with subscription manager (hosted service vs on-prem).  Perhaps we could add a new value for this now for the daemonless builder.
- Subscription: Used only by the service (I think, https://github.com/osbuild/osbuild-composer/blob/a7cd521325504a43d3708f8f91ab63999862b279/internal/cloudapi/v2/compose.go#L1194).  But might be useful to expose for on-prem builders.
- PartitioningMode: This is a funny one.  In osbuild-composer, it's set in the blueprint (https://github.com/osbuild/osbuild-composer/blob/a7cd521325504a43d3708f8f91ab63999862b279/internal/blueprint/customizations.go#L23), but it's passed down to images through the image options (https://github.com/osbuild/osbuild-composer/blob/a7cd521325504a43d3708f8f91ab63999862b279/internal/weldr/api.go#L2552).  Not sure what we want to do there.  Since it's in the BP in composer already, I think we should move it there in images too.

cmd/image-builder/main.go
```diff
@@ -0,0 +1,103 @@
...
+		Short: "Build operating system images from a given blueprint",
+		Long: `Build operating system images from a given blueprint
+
+Image-builder builds operating system images for a range of predefined
+operating sytsems like centos and RHEL with easy customizations support.`,
+	}
+
+	// XXX: this will list 802 images right now, we need a sensible
+	// default here, maybe without --filter just list all available
+	// distro names?
```
achilleas-k, line 65: I think it's fine to list everything if it's unfiltered.  Perhaps in the human-readable version we can have a count on the last line with a suggestion to look into `--filter`, so it's clear that the user is getting too much output.
For the machine readable case, listing all is definitely ok as far as I'm concerned.

achilleas-k (commented):
I really like this PR (what it's starting, in particular).  I realise there are many reasons why it's a draft, but it makes me worry that it will be forgotten if we don't move towards something more concrete and start merging soon.

I think we might get the ball rolling if we start small.  A PR that only implements list-images for example would be very quick to merge, and after that we'd have the `image-builder` command ready to expand on and start using in testing (replacing the existing `cmd/list-images`).

mvo5:
Just FTR - this will be transferred to https://github.com/osbuild/image-builder-cli (unless someone has concerns)

PR #999, opened 2024-10-28
build: support specifying repository file path

achilleas-k:
Add a flag `--repositories` for specifying where to find repositories. The default value for the new flag is `test/data/repositories/` so this change is backwards compatible.

The new option is a bit flexible.  It supports either a full path to a file with repository configurations for a single distribution, or a path to a directory containing at least one repository configuration.  If the path is a directory, the distro name is appended to the path as a filename, e.g. `fedora-40.json`.

This makes it possible to build images with repository configurations other than the ones in the test/data/repositories/ directory.  This change also speeds up execution slightly because it doesn't need to initialise the repository configurations for every file in the directory, just the one that we are building.


mvo5 (approved):
Thank you!

PR #1001, opened 2024-10-29
reporegistry: move to logrus instead of log.Printf()

mvo5:
This commit removes the direct log.Printf() from the repo loading. With the coming daemonless CLI frontend to image building these prints are distracting for our users.

Instead logrus (which is already used in the images library) is used with the appropriate "Debugf()" and "Warnf()" methods.

### DIFF
--- a/pkg/reporegistry/repository.go
+++ b/pkg/reporegistry/repository.go
@@ -1,10 +1,11 @@
 package reporegistry
 
 import (
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/sirupsen/logrus"
 
 	"github.com/osbuild/images/pkg/distroidparser"
 	"github.com/osbuild/images/pkg/rpmmd"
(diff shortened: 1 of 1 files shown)


supakeen (approved):

pkg/reporegistry/repository.go
```diff
@@ -58,7 +59,7 @@ func LoadAllRepositories(confPaths []string) (rpmmd.DistrosRepoConfigs, error) {
 					return nil, err
 				}
 
-				log.Println("Loaded repository configuration file:", configFile)
+				logrus.Debugf("Loaded repository configuration file: %s", configFile)
```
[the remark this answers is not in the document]
mvo5: Hm, I would hate to cause regressions with this change. I will switch to `Info()` and set the logging in the CLI interface accordingly.

pkg/reporegistry/repository.go
```diff
@@ -58,7 +59,7 @@ func LoadAllRepositories(confPaths []string) (rpmmd.DistrosRepoConfigs, error) {
 					return nil, err
 				}
 
-				log.Println("Loaded repository configuration file:", configFile)
+				logrus.Info("Loaded repository configuration file: %s", configFile)
```
thozza, line 62: This should be `Infof()`, not `Info()`.

achilleas-k (commented):
LGTM (minus Tomáš' comment).

pkg/reporegistry/repository.go
mvo5: Silly me, thanks for spotting! I am a bit short on time but ideally I would have added a test for this too /o\

achilleas-k (approved):

thozza (approved):
:+1:

PR #1002, opened 2024-10-31
disk: define FSType enum for all supported types

achilleas-k:

Add a new enum, FSType, for referring to all supported filesystem types.  This is currently not used but will be part of the new custom partition table generator.

### DIFF
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -19,6 +19,7 @@ package disk
 
 import (
 	"encoding/hex"
+	"fmt"
 	"io"
 	"math/rand"
 	"reflect"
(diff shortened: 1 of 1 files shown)


ondrejbudai:
The commit message is wrong, there's no `NewCustomPartitionTable` in this reality.

achilleas-k:
> This is basically dead code without more context, but I guess that's fine, we know that it will be useful in the near future, right?

Yes.

achilleas-k:
> The commit message is wrong, there's no `NewCustomPartitionTable` in this reality.

Fixed.

I also mention in the commit message that it's not used yet but it will bee SOON!

ondrejbudai (approved):
Go!

PR #1003, opened 2024-10-31
Refactor LVM LV name generation and extract unique name generator into helper

achilleas-k:

---

**disk: refactor LVM Logical Volume name generation**

The VolumeGroup.CreateLogicalVolume() function assumed that the given
name was a mountpoint and would automatically escape it, append 'lv',
and append numbers in case of collisions.  This made it impossible to
use the helper function for LV creation with desired names.

Refactor the function and modify all the calls so that now it uses the
provided name without validating, escaping, or checking for collisions,
and generates a name from the payload's mountpoint when one is not
provided.

The CreateMountpoint() function's functionality is not affected.

---

**disk: helper function for generating unique names**

The new genUniqueString() function follows the logic used when
generating logical volume names, which was based on the same
functionality in blivet.
It will be used to generate unique logical volume names, labels, and
btrfs volume names.

---

**disk: replace the lvm name generation with genUniqueString()**

Replace the heart of the lvm name generator with the new function.
Update the vg.CreateMountpoint() test, because the old implementation
was off by one and would fail when '99' was available.

---

### DIFF
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -211,6 +211,8 @@ func NewVolIDFromRand(r *rand.Rand) string {
 // the existing set. If the base itself does not exist, it is returned as is,
 // otherwise a two digit number is added and incremented until a unique string
 // is found.
+// This function is mimicking what blivet does for avoiding name collisions.
+// See blivet/blivet.py#L1060 commit 2eb4bd4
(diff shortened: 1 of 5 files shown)


ondrejbudai (approved):
Thanks! :heart:

PR #1005, opened 2024-10-31
pkg/{distro,platform}: move BootMode to platform

achilleas-k:

---

Move the BootMode enum from pkg/distro to pkg/platform.  We want to import BootMode to the disk package, which would create an import loop with distro.


ondrejbudai (approved):
Go! :)

supakeen (approved):

PR #1009, opened 2024-11-01
Partition table helper functions for adding root filesystem and boot partition

achilleas-k:

---

The PR introduces two helper functions, `EnsureRootFilesystem()` and `EnsureBootPartition()`.  These will add, respectively, a root filesystem and a boot partition to a partition table if they don't already exist.

### DIFF
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -60,6 +60,9 @@ const (
 
 	// DosFat16B used for the ESP-System partition
 	DosFat16B = "06"
+
+	// Partition type ID for any native Linux filesystem on dos
+	DosLinuxTypeID = "83"
 )
 
 // FSType is the filesystem type enum.
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -990,3 +990,47 @@ func EnsureRootFilesystem(pt *PartitionTable, defaultFsType FSType) error {
 	pt.Partitions = append(pt.Partitions, rootpart)
 	return nil
 }
+
+// EnsureBootPartition creates a boot partition if one does not already exist.
+// The function will append the boot partition to the end of the existing
+// partition table therefore it is best to call this function early to put boot
+// near the front (as is conventional).
+func EnsureBootPartition(pt *PartitionTable, bootFsType FSType) error {
+	// collect all labels to avoid conflicts
+	labels := make(map[string]bool)
+	var foundBoot bool
+	_ = pt.ForEachMountable(func(mnt Mountable, path []Entity) error {
+		if mnt.GetMountpoint() == "/boot" {
+			foundBoot = true
+			return nil
+		}
+
+		labels[mnt.GetFSSpec().Label] = true
+		return nil
+	})
+	if foundBoot {
+		// nothing to do
+		return nil
+	}
+
+	if bootFsType == FS_NONE {
+		return fmt.Errorf("error creating boot partition: no filesystem type")
+	}
+
+	bootLabel, err := genUniqueString("boot", labels)
+	if err != nil {
+		return fmt.Errorf("error creating boot partition: %w", err)
+	}
+	bootPart := Partition{
+		Type: XBootLDRPartitionGUID,
+		Size: 512 * datasizes.MiB,
+		Payload: &Filesystem{
+			Type:         bootFsType.String(),
+			Label:        bootLabel,
+			Mountpoint:   "/boot",
+			FSTabOptions: "defaults",
+		},
+	}
+	pt.Partitions = append(pt.Partitions, bootPart)
+	return nil
+}
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -894,3 +894,100 @@ func (pt *PartitionTable) GetMountpointSize(mountpoint string) (uint64, error) {
 
 	panic(fmt.Sprintf("no sizeable of the entity path for mountpoint %s, this is a programming error", mountpoint))
 }
+
+// EnsureRootFilesystem adds a root filesystem if the partition table doesn't
+// already have one.
+//
+// When adding the root filesystem, add it to:
+//
+//   - The first LVM Volume Group if one exists, otherwise
+//   - The first Btrfs volume if one exists, otherwise
+//   - At the end of the plain partitions.
+//
+// For LVM and Plain, the fsType argument must be a valid filesystem type.
+func EnsureRootFilesystem(pt *PartitionTable, defaultFsType FSType) error {
+	// collect all labels and subvolume names to avoid conflicts
+	subvolNames := make(map[string]bool)
+	labels := make(map[string]bool)
+	var foundRoot bool
+	_ = pt.ForEachMountable(func(mnt Mountable, path []Entity) error {
+		if mnt.GetMountpoint() == "/" {
+			foundRoot = true
+			return nil
+		}
+
+		labels[mnt.GetFSSpec().Label] = true
+		switch mountable := mnt.(type) {
+		case *BtrfsSubvolume:
+			subvolNames[mountable.Name] = true
+		}
+		return nil
+	})
+	if foundRoot {
+		// nothing to do
+		return nil
+	}
+
+	for _, part := range pt.Partitions {
+		switch payload := part.Payload.(type) {
+		case *LVMVolumeGroup:
+			if defaultFsType == FS_NONE {
+				return fmt.Errorf("error creating root logical volume: no default filesystem type")
+			}
+
+			rootLabel, err := genUniqueString("root", labels)
+			if err != nil {
+				return fmt.Errorf("error creating root logical volume: %w", err)
+			}
+			rootfs := &Filesystem{
+				Type:         defaultFsType.String(),
+				Label:        rootLabel,
+				Mountpoint:   "/",
+				FSTabOptions: "defaults",
+			}
+			// Let the function autogenerate the name to avoid conflicts
+			// with LV names from customizations.
+			// Set the size to 0 and it will be adjusted by
+			// EnsureDirectorySizes() and relayout().
+			if _, err := payload.CreateLogicalVolume("", 0, rootfs); err != nil {
+				return fmt.Errorf("error creating root logical volume: %w", err)
+			}
+			return nil
+		case *Btrfs:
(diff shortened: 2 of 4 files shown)


pkg/disk/partition_table_test.go
```diff
@@ -105,3 +105,552 @@ func TestPartitionTable_GenerateUUIDs_VFAT(t *testing.T) {
...
+									Name:       "root",
+									Mountpoint: "/",
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		"err-empty": {
```
mvo5, line 396: (Super nitpick) *maybe* spliting into happy and unhappy tests might make things feel a little bit more organized but that might be just me (and slightly less struct/ content for each). I generally like mixing when the test cases are not overly long but when they are multiple pages like here it might be slightly nicer to split (but again could be preferences)

pkg/disk/partition_table.go
```diff
@@ -894,3 +894,145 @@ func (pt *PartitionTable) GetMountpointSize(mountpoint string) (uint64, error) {
...
+// already have one.
+//
+// When adding the root filesystem, add it to:
+//
+//   - The first LVM Volume Group if one exists, otherwise
+//   - The first Btrfs volume if one exists, otherwise
+//   - At the end of the plain partitions.
+//
+// For LVM and Plain, the fsType argument must be a valid filesystem type.
+func EnsureRootFilesystem(pt *PartitionTable, fsType FSType) error {
```
mvo5, line 908: (super nitpick) maybe defaultFsType because this will not "ensure" that the fsType is of the given type just what to use if it's missing

pkg/disk/partition_table_test.go
```diff
@@ -105,3 +105,552 @@ func TestPartitionTable_GenerateUUIDs_VFAT(t *testing.T) {
 
 	assert.Equal(t, "6e4ff95f", pt.Partitions[0].Payload.(*disk.Filesystem).UUID)
 }
+
+func TestEnsureRootFilesystem(t *testing.T) {
+	type testCase struct {
+		pt       disk.PartitionTable
+		expected disk.PartitionTable
+		fsType   disk.FSType
```
mvo5, line 113: (nipitck) *maybe* also here `defaultFstype` because we do not switch fstypes

mvo5 (approved):
Looks good, some ideas/suggestions inline but no blockers just my usual nitpicky self (but some might still be worthwhile).

pkg/disk/partition_table.go
```diff
@@ -894,3 +894,145 @@ func (pt *PartitionTable) GetMountpointSize(mountpoint string) (uint64, error) {
...
+	// collect all labels and subvolume names to avoid conflicts
```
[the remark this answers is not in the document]
achilleas-k: We already have `entityPath()` which we use in a lot of places for this exact purpose, for example:
https://github.com/osbuild/images/blob/d9ffdd686ff3307d84c36b566dc48cacc37ca393/pkg/disk/partition_table.go#L468-L473

In fact, the first version of this function was doing just that before I realised we need to collect labels: https://github.com/osbuild/images/pull/926/commits/462ceca5f17f2e3045e476aee4b7d43086571da9#diff-ffc14a35d85eb6fd9724427b51786e40bc3e910bc41cbd6762ab977ae9689765R1170

Using that when we're going to scan through the PT anyway to collect the labels feels a bit wrong.  That said, one thing I noticed now is that we don't need to use the bool and can just return.

pkg/disk/partition_table.go
```diff
@@ -894,3 +894,145 @@ func (pt *PartitionTable) GetMountpointSize(mountpoint string) (uint64, error) {
...
+	}
+
+	// add a plain root partition at the end of the partition table
+	rootLabel, err := genUniqueString("root", labels)
+	if err != nil {
+		return fmt.Errorf("error creating root partition: %w", err)
+	}
+	rootpart := Partition{
+		Type:     FilesystemDataGUID,
+		Bootable: false,
```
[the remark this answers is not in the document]
achilleas-k: Probably from autofilling the struct.  Dropping.

pkg/disk/partition_table.go
```diff
@@ -894,3 +894,145 @@ func (pt *PartitionTable) GetMountpointSize(mountpoint string) (uint64, error) {
...
+		return fmt.Errorf("error creating boot partition: no filesystem type")
+	}
+
+	bootLabel, err := genUniqueString("boot", labels)
+	if err != nil {
+		return fmt.Errorf("error creating boot partition: %w", err)
+	}
+	bootPart := Partition{
+		Type:     XBootLDRPartitionGUID,
+		Bootable: false,
```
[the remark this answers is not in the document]
achilleas-k: Probably from autofilling the struct.  Dropping.

pkg/disk/partition_table.go
```diff
@@ -894,3 +894,145 @@ func (pt *PartitionTable) GetMountpointSize(mountpoint string) (uint64, error) {
...
+		Size:     512 * datasizes.MiB,
```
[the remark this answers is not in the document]
achilleas-k: I'd rather we keep it consistent with the existing behaviour: https://github.com/achilleas-k/images/blob/d6365bdc7a5980cfa2a5ee1fa3c14a64a688d6fa/pkg/disk/partition_table.go#L692-L702

And maybe we can change it separately for both and unify the declaration to one place.

pkg/disk/partition_table_test.go
achilleas-k: I did at first but then wanted to deduplicate the common bits between the two.  Which isn't a lot, admittedly.

pkg/disk/partition_table.go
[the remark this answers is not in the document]
mvo5: Yeah, it's fine. I thought this was why it was done - to avoid the double traversal. I (personally) would still have made the tradeoff to do it twice and have a helper so that it can return early from a helper (as it feels easier for a human to read) but fine to leave as is, it is probably just personal.

pkg/disk/partition_table.go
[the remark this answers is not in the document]
mvo5: Aha, that makes sense, thank you

mvo5 (approved):
Thank you

pkg/disk/partition_table.go
```diff
@@ -990,3 +990,47 @@ func EnsureRootFilesystem(pt *PartitionTable, defaultFsType FSType) error {
 	pt.Partitions = append(pt.Partitions, rootpart)
 	return nil
 }
+
+// EnsureBootPartition creates a boot partition if one does not already exist.
+// The function will append the boot partition to the end of the existing
+// partition table therefore it is best to call this function early to put boot
+// near the front (as is conventional).
+func EnsureBootPartition(pt *PartitionTable, bootFsType FSType) error {
+	// collect all labels and subvolume names to avoid conflicts
```
ondrejbudai, line 999: nitpick: we don't gathet subvolumes here.

pkg/disk/partition_table.go
```diff
@@ -894,3 +894,100 @@ func (pt *PartitionTable) GetMountpointSize(mountpoint string) (uint64, error) {
...
+		return fmt.Errorf("error creating root partition: no default filesystem type")
+	}
+
+	// add a plain root partition at the end of the partition table
+	rootLabel, err := genUniqueString("root", labels)
+	if err != nil {
+		return fmt.Errorf("error creating root partition: %w", err)
+	}
+	rootpart := Partition{
+		Type:     FilesystemDataGUID,
```
ondrejbudai: Should we adjust the partition type based on the type of the partition table? In other words, if the partition type is MBR, we would need to use the 1 byte value instead of UUID here.

pkg/disk/partition_table.go
```diff
@@ -990,3 +990,47 @@ func EnsureRootFilesystem(pt *PartitionTable, defaultFsType FSType) error {
...
+	if bootFsType == FS_NONE {
+		return fmt.Errorf("error creating boot partition: no filesystem type")
+	}
+
+	bootLabel, err := genUniqueString("boot", labels)
+	if err != nil {
+		return fmt.Errorf("error creating boot partition: %w", err)
+	}
+	bootPart := Partition{
+		Type: XBootLDRPartitionGUID,
```
ondrejbudai, line 1025: Same concerns about UUIDs and MBR here.

ondrejbudai (dismissed):
I think I found a tiny issue regarding MBR partitions, but I believe that we can fix these in a follow-up since this code isn't actually used anywhere in production yet.

pkg/disk/partition_table.go
achilleas-k: This is true.

pkg/disk/disk.go
```diff
@@ -110,6 +113,41 @@ func NewFSType(s string) (FSType, error) {
 	}
 }
 
+// PartitionTableType is the partition table type enum.
+type PartitionTableType uint64
```
mvo5, line 117: I'm confused, why is this part of the displayed diff here? Is GH having a bad day? This should already be in `main`, no?

pkg/disk/partition_table_test.go
```diff
@@ -105,3 +105,922 @@ func TestPartitionTable_GenerateUUIDs_VFAT(t *testing.T) {
 
 	assert.Equal(t, "6e4ff95f", pt.Partitions[0].Payload.(*disk.Filesystem).UUID)
 }
+
+func TestEnsureRootFilesystem(t *testing.T) {
+	type testCase struct {
+		pt            disk.PartitionTable
+		expected      disk.PartitionTable
+		defaultFsType disk.FSType
+		errmsg        string
```
mvo5, line 114: This is now no longer needed AFAICT

mvo5 (approved):
Thank you, looks good, there is lint error that probably needs fixing

pkg/disk/partition_table_test.go
achilleas-k: Yup

pkg/disk/disk.go
achilleas-k: That's my fault.  I rebased on the head of the merge queue while the other PR was queued up to see if I get any issues.  Should be good now.  Sorry for the confusion.

mvo5 (approved):

PR #1012, opened 2024-11-04
distrosort: add new `distrosort.Names` helper

mvo5:
[edit: alternative version of https://github.com/osbuild/images/pull/1000 - does not support `-beta` and similar semver features but reuses more of the existing infrastructure]

distrosort: add new `distrosort.Names` helper

This commit adds a new `distrosort.Names()` helper that can be
used sort sort distro names. It will take version numbers of
the distro into account and sort by semantic version. This
is useful in e.g. user-interfaces where we want to present
the version in a nicely sorted way.

Sorting follows the go-version library which supports semantic
versioning [0] so anything supported there will be sorted
correctly including `-beta` suffixes and similar (see the
spec for details).

Note that this is an extra pkg because the `distro` and `distroid`
would be circular imports otherwise. The alternative might be to
create a distro.ID package.

[0] https://semver.org/

---

distro: add new 'ID.Version()` method

This commit adds a new method `ID.Version()` that returns a
`go-version.Version` object so that we can easily sort the distro
versions.

The `ID.Version()` currently also has to return an error because
the ID struct exposes MajorVersion/MinorVersion which means that
during the lifetime the Version may become invalid even after
`ParseID` was run. We could fix this by making {Major,Minor}Version
getters instead of the current way. Then we could also use the
version.NewVersion() library directy to parse the version string
from the ID and support things like `-beta` or other extras that
semantic versioning allows. But then that may interfere with the
distroid.Parser interface so definitely needs some more thinking.

### DIFF
--- a/pkg/distro/id.go
+++ b/pkg/distro/id.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+
+	"github.com/hashicorp/go-version"
 )
 
 // ID represents a distro name and version
@@ -20,6 +22,16 @@ func (id ID) String() string {
 	}
 
 	return fmt.Sprintf("%s-%d.%d", id.Name, id.MajorVersion, id.MinorVersion)
+}
+
+func (id ID) Version() (*version.Version, error) {
+	var verStr string
+	if id.MinorVersion == -1 {
+		verStr = fmt.Sprintf("%d", id.MajorVersion)
+	} else {
+		verStr = fmt.Sprintf("%d.%d", id.MajorVersion, id.MinorVersion)
+	}
+	return version.NewVersion(verStr)
 }
 
 type ParseError struct {
--- a/pkg/distro/id.go
+++ b/pkg/distro/id.go
@@ -16,22 +16,20 @@ type ID struct {
 	MinorVersion int
 }
 
-func (id ID) String() string {
+func (id ID) versionString() string {
 	if id.MinorVersion == -1 {
-		return fmt.Sprintf("%s-%d", id.Name, id.MajorVersion)
+		return fmt.Sprintf("%d", id.MajorVersion)
+	} else {
+		return fmt.Sprintf("%d.%d", id.MajorVersion, id.MinorVersion)
 	}
+}
 
-	return fmt.Sprintf("%s-%d.%d", id.Name, id.MajorVersion, id.MinorVersion)
+func (id ID) String() string {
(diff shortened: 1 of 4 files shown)


thozza (approved):
Nice, I like this version :+1:

pkg/distro/id.go
```diff
@@ -22,6 +24,16 @@ func (id ID) String() string {
 	return fmt.Sprintf("%s-%d.%d", id.Name, id.MajorVersion, id.MinorVersion)
 }
 
+func (id ID) Version() (*version.Version, error) {
+	var verStr string
+	if id.MinorVersion == -1 {
+		verStr = fmt.Sprintf("%d", id.MajorVersion)
+	} else {
+		verStr = fmt.Sprintf("%d.%d", id.MajorVersion, id.MinorVersion)
+	}
```
supakeen, line 33: Could we split this out into a separate `VersionString()` and then also re-use it in `String()`? They currently duplicated the `%d` vs `%d.%d` logic.

pkg/distrosort/sort_test.go
```diff
@@ -0,0 +1,58 @@
...
+			[]string{"foo-1.x", "foo-2"},
+			"error when parsing distro name (foo-1.x): parsing minor version failed, inner error:\nstrconv.Atoi: parsing \"x\": invalid syntax",
+		}, {
+			// missing "-" is not supported
+			[]string{"foo", "bar-1"},
+			`error when parsing distro name (foo): A dash is expected to separate distro name and version`,
+		}, {
+			// foo-1.4-beta is not supported
+			[]string{"foo-1.4", "foo-1.4-beta", "foo-1.0"},
+			"error when parsing distro name (foo-1.4-beta): parsing major version failed, inner error:\nstrconv.Atoi: parsing \"beta\": invalid syntax",
```
supakeen, line 52: Is it correct that this is failing on the *major* here? It seems to me that if we don't support the `-beta` part that it would be part of the minor version?

mvo5: The parser is currently quite simple, it will assume the major/minor after the first or second "-" (second to support `centos-stream-9`. I will investigate if we can easily fix this without breaking compatbility

thozza: I would also point out that each distro may have its own parser implementation (e.g., an RHEL distro before RHEL-10 to support release numbers without a dot).

supakeen: In that case feel free to resolve this :)

supakeen (approved):

PR #1013, opened 2024-11-04
awscloud: add import role parameter to AMI registration

ondrejbudai:
We got a request for supporting custom role names, let's start by
adding support for it in the images library.

### DIFF
--- a/cmd/boot-aws/main.go
+++ b/cmd/boot-aws/main.go
@@ -154,6 +154,19 @@ func newClientFromArgs(flags *pflag.FlagSet) (*awscloud.AWS, error) {
 	return awscloud.New(region, keyID, secretKey, sessionToken)
 }
 
+// getOptionalStringFlag returns the value of a string flag if it's set, or nil
+// if it's not set.
+func getOptionalStringFlag(flags *pflag.FlagSet, name string) (*string, error) {
+	value, err := flags.GetString(name)
+	if err != nil {
+		return nil, err
+	}
+	if value == "" {
+		return nil, nil
+	}
+	return &value, nil
+}
+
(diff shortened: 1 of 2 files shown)


cmd/boot-aws/main.go
```diff
@@ -192,6 +192,13 @@ func doSetup(a *awscloud.AWS, filename string, flags *pflag.FlagSet, res *resour
 		return err
 	}
 
+	var importRolePtr *string
+	if importRole, err := flags.GetString("import-role"); importRole != "" {
```
mvo5, line 196: (nitpick) Maybe the slightly more canoical:
```go
	var importRolePtr *string
	importRole, err := flags.GetString("import-role")
	if err != nil {
		return err
	}
	if importRole != "" {
		importRolePtr = &importRole
	}
```
ie. error check first? I know it's the same but when reading go code anything that diverges from this pattern automatically stops my flow.

mvo5 (approved):
Looks good, one tiny style suggestion

supakeen (approved):

cmd/boot-aws/main.go
achilleas-k: I agree with errors first but I do like the limited scoping you get from the `if ret, err := func(); err != nil ...`

ondrejbudai: What about extracting a helper function to get both error handling first, and limiting the scope? See my latest force-push.

mvo5 (approved):

cmd/boot-aws/main.go
achilleas-k: Well, Michael approved, so I'm resolving this and merging.

achilleas-k (approved):
LGTM.  Optional string flag getter is nice.

PR #1017, opened 2024-11-06
dnfjson: skip sbom tests with dnf5

achilleas-k:
**dnfjson: add Solver string to DepsolveResult**

The solver string is returned from the depsolver with the response and
identifies the solver library that was used.  Currently, it is either
dnf or dnf5.

Returning it with the DepsolveResult can be useful for reproducibility
and troubleshooting.

---

**dnfjson: skip sbom tests with dnf5**

SBOMs are not yet supported when using the dnf5 solver in
osbuild-depsolve-dnf.  Running the test on Fedora 41, which uses dnf5 by
default, makes the test fail.

Sniff the solver used using a quick, simple depsolve to get the result,
and skip any SBOM tests if the solver is dnf5.

---

### DIFF
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -161,6 +161,7 @@ type DepsolveResult struct {
 	Packages []rpmmd.PackageSpec
 	Repos    []rpmmd.RepoConfig
 	SBOM     *sbom.Document
+	Solver   string
 }
 
 // Create a new Solver with the given configuration. Initialising a Solver also loads system subscription information.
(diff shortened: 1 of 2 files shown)


supakeen (approved):
This is fine to get the tests working on Fedora 41 again however I feel like we might want a more 'official' approach to get the solver-in-use than a fake depsolve in the future :)

thozza (approved):
Thank you for fixing this.

I'm OK with this being a temporary workaround to make tests pass on F41. However, This made me think about a few things:

1. Should the test case not explicitly configure the `solver.json` and test SBOM support only with DNF4 (and later with DNF5)? Alternatively, whether the decision to choose the DNF version should not be left up to the caller if it specifies it in the request.
2. Using DNF5 by default for our depsolving is not a good idea, even on F41. The implementation is still WIP AFAICT and does not support [essential use cases](https://github.com/osbuild/osbuild/blob/798ffa7daccf775b8d68743a669adadbe7a8a828/tools/test/test_depsolve.py#L1262-L1267). So people may experience regressions on F41+ on-prem image builds, depending on what they are doing.

PR #1019, opened 2024-11-07
imagefilter: add `ResultFormatter` type to support flexible  output

mvo5:
This commit adds a new `ResultFormatter` interface that supports
outputing the results as text or JSON.

(this is another splitout from https://github.com/osbuild/images/pull/997)

### DIFF
--- a/pkg/imagefilter/formatter.go
+++ b/pkg/imagefilter/formatter.go
@@ -1,0 +1,96 @@
+package imagefilter
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+)
+
+// OutputFormat contains the valid output formats for formatting results
+type OutputFormat string
+
+const (
+	OutputFormatDefault OutputFormat = ""
+	OutputFormatText    OutputFormat = "text"
+	OutputFormatJSON    OutputFormat = "json"
+)
+
+// ResultFormatter will format the given result list to the given io.Writer
+type ResultsFormatter interface {
+	Output(io.Writer, []Result) error
+}
+
+// NewResultFormatter will create a formatter based on the given format.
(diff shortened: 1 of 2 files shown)


pkg/imagefilter/formatter.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+	"io"
+)
+
+// OutputFormat contains the valid output formats for formatting results
+type OutputFormat string
+
+const (
+	OutputFormatDefault OutputFormat = ""
+	OutputFormatText    OutputFormat = "text"
```
mvo5, line 15: Fwiw, we could easily support `text-fmt` as formatter here to allow comamndline things like `--output=text --format={{.Distro.Name}}` but let's YAGNI this for now (unless we/our users need it)

mvo5: We could also have a "json-experimental" or something formatter that would give us a lot more details without API gurantees (again YAGNI, but we are fleixble enough here)

pkg/imagefilter/formatter.go
```diff
@@ -0,0 +1,93 @@
...
+
+type archResultJSON struct {
+	Name string `json:"name"`
+}
+
+type imgTypeResultJSON struct {
+	Name string `json:"name"`
+}
+
+type filteredResultJSON struct {
```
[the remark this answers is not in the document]
achilleas-k: Add a few config parameters and package lists to this and we have a good way to "dump" image definitions in a portable way ^_^

pkg/imagefilter/formatter.go
achilleas-k: I think it's good to experiment with a few things and see what works, what feels good.  We're not releasing anything (yet), but we should start using it in CI and our own day-to-day.

PR #1020, opened 2024-11-07
rhel: fix crash when PartitionType() is requested

mvo5:
This commit fixes a crash in when the rhel ImageType.PartitionType() is called, here the BasePartitionTables function pointer is sometimes not initialized (e.g. in the `centos9/edge-commit` image type).

### DIFF
--- a/pkg/distro/rhel/imagetype.go
+++ b/pkg/distro/rhel/imagetype.go
@@ -216,6 +216,10 @@ func (t *ImageType) getDefaultInstallerConfig() (*distro.InstallerConfig, error)
 }
 
 func (t *ImageType) PartitionType() string {
+	if t.BasePartitionTables == nil {
+		return ""
+	}
+
 	basePartitionTable, exists := t.BasePartitionTables(t)
 	if !exists {
 		return ""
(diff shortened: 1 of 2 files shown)


achilleas-k (approved):
Ty!!

pkg/distro/rhel/imagetype.go
```diff
@@ -216,6 +216,10 @@ func (t *ImageType) getDefaultInstallerConfig() (*distro.InstallerConfig, error)
 }
 
 func (t *ImageType) PartitionType() string {
+	if t.BasePartitionTables == nil {
+		return ""
```
thozza, line 220: This is just food for thought, but it made me wonder if we should not define `PartitionType` as an enum and have an explicit value for `None` instead of an empty string. Comparing enums is IMHO less error-prone than comparing strings against literal string values.

thozza (approved):
Thanks for fixing this!

pkg/distro/rhel/imagetype.go
mvo5: We have a type like this https://github.com/osbuild/images/pull/1004/files now

mvo5: I opened https://github.com/osbuild/images/pull/1025 ?

thozza: Sweet, thanks!

PR #1024, opened 2024-11-07
disk: AddPartitionsForBootMode() helper function 

achilleas-k:

---

The PR introduces a helper function, `AddPartitionsForBootMode()`.  The function takes a partition table and boot mode and adds partitions to satisfy the boot mode requirements:
- BIOS/legacy: adds a 1 MiB BIOS boot partition.
- UEFI: adds a 200 MiB EFI system partition.
- Hybrid: adds both.

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -1097,7 +1097,7 @@ func AddPartitionsForBootMode(pt *PartitionTable, bootMode platform.BootMode) er
 	case platform.BOOT_NONE:
 		return nil
 	default:
-		return fmt.Errorf("invalid boot mode specified: %s", bootMode)
+		return fmt.Errorf("unknown or unsupported boot mode type with enum value %d", bootMode)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/disk/partition_table.go
```diff
@@ -1054,3 +1055,92 @@ func EnsureBootPartition(pt *PartitionTable, bootFsType FSType) error {
...
+		esp, err := mkESP(200*datasizes.MiB, pt.Type)
+		if err != nil {
+			return err
+		}
+		pt.Partitions = append(pt.Partitions, bios, esp)
+		return nil
+	case platform.BOOT_NONE:
+		return nil
+	default:
+		return fmt.Errorf("invalid boot mode specified: %s", bootMode)
```
mvo5, line 1100: (nitpick) maybe add a test for this too?

mvo5 (approved):
Thank you, this looks very nice! One tiny suggestion about a possible unhappy test but not a blocker, it's not really criticial.

pkg/disk/partition_table.go
achilleas-k: Changed error message and added test.

mvo5 (approved):
Super nice!

PR #1025, opened 2024-11-07
many: use new `disk.PartitionTableType` instead of string

mvo5:
This commit makes use of the new `disk.PartitionTableType instead of just using an untyped string (thanks to Achilleas for adding the new type and to Tomáš for suggesting to use it more).

---

disk: add json marshal/unmarshal to `PartitionTableType`

This commit adds json marshal/unmarshal to `PartitionTableType`
The reason is that the `otk` utils serialize this type
as a JSON. While the json is strictly internal and we make no
gurantees about it is still nicer to serialize as strings because
otherwise the diff in the json looks something like:
```json
@@ -147,7 +147,7 @@ func TestUnmarshalOutput(t *testing.T) {
       "partition-table": {
         "Size": 911,
         "UUID": "",
-        "Type": "gpt",
+        "Type": 2,
         "Partitions": [
           {
             "Start": 0,
```
and it is also nicer for general debug-ability.

### DIFF
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -19,6 +19,7 @@ package disk
 
 import (
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io"
 	"math/rand"
@@ -139,6 +140,24 @@ func (t PartitionTableType) String() string {
 	default:
 		panic(fmt.Sprintf("unknown or unsupported partition table type with enum value %d", t))
 	}
+}
+
+func (t PartitionTableType) MarshalJSON() ([]byte, error) {
+	return json.Marshal(t.String())
+}
+
+func (t *PartitionTableType) UnmarshalJSON(data []byte) error {
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+
+	new, err := NewPartitionTableType(s)
+	if err != nil {
+		return err
+	}
+	*t = new
+	return nil
 }
 
 func NewPartitionTableType(s string) (PartitionTableType, error) {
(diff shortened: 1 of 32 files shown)


pkg/disk/disk.go
```diff
@@ -119,39 +119,33 @@ func NewFSType(s string) (FSType, error) {
 	}
 }
 
-// PartitionTableType is the partition table type enum.
-type PartitionTableType uint64
+// PartitionTableType is the partition table type
+type PartitionTableType string
```
thozza, line 123: I'd keep it being an enum and implement a custom JSON (un)marshaler. Such as in https://github.com/osbuild/images/blob/863bb93228db5e6e289f2d972e7af0e3e07eec69/pkg/sbom/document.go#L8-L48

mvo5: That is a good suggestion - I guess the advantage is that we get "free" error checking on unmarshal this way, something that just moving to a string would not give us. I like that.

mvo5: Updated now.

achilleas-k (approved):
Very nice.  Should we do the same for filesystem types?

achilleas-k:
Also, can you update the top PR comment/description to avoid confusion in the future?

mvo5:
> Very nice. Should we do the same for filesystem types?

Yeah, I guess we want this for `GetFSType()` too (and friends, not sure how many there are)

achilleas-k:
> > Very nice. Should we do the same for filesystem types?
> 
> Yeah, I guess we want this for `GetFSType()` too (and friends, not sure how many there are)

Fine to do in a separate PR.

thozza (approved):
Amazing :heart_eyes:

PR #1031, opened 2024-11-11
fix: install anaconda-install-img-deps package in anaconda installers.

mmartinv:
According to the anaconda team we should be using the `anaconda-install-img-deps`
metapackage to install all the boot.iso image dependencies.

Relates: fedora-iot/iot-distro#66

### DIFF
--- a/pkg/distro/fedora/package_sets.go
+++ b/pkg/distro/fedora/package_sets.go
@@ -354,7 +354,7 @@ func anacondaPackageSet(t *imageType) rpmmd.PackageSet {
 			"alsa-tools-firmware",
 			"anaconda",
 			"anaconda-dracut",
-			"anaconda-install-env-deps",
+			"anaconda-install-img-deps",
 			"anaconda-widgets",
 			"atheros-firmware",
 			"audit",


supakeen (approved):
Going by the descriptions of these packages `anaconda-install-img-deps` is indeed the correct dependency (and it in turn depends on `anaconda-install-env-deps`). Ack.

thozza:
Note that this needs to be fixed also on RHEL-9 and older since I made the change for RHEL-10 when adding the image-installer img type. Unfortunately, I didn't get to verify if the package is present on all older RHEL versions.

Consequently, this change means that a bunch of other packages can be removed from the pkg list, and we do not need to list them explicitly (as these are pulled in as `anaconda-install-img-deps` dependency). Based on my previous work, these are mostly some UI-related dependencies.

mmartinv:
> Note that this needs to be fixed also on RHEL-9 and older since I made the change for RHEL-10 when adding the image-installer img type. Unfortunately, I didn't get to verify if the package is present on all older RHEL versions.
> 

Just checked and it's present in RHEL 8 and RHEL 9

> Consequently, this change means that a bunch of other packages can be removed from the pkg list, and we do not need to list them explicitly (as these are pulled in as `anaconda-install-img-deps` dependency). Based on my previous work, these are mostly some UI-related dependencies.

Yes, we can do a follow up PR to remove the unneeded deps I guess, the dependencies for both metapackages in Fedora 42 are currently:
```
sudo dnf repoquery --requires anaconda-install-img-deps anaconda-install-env-deps
Updating and loading repositories:
Repositories loaded.
anaconda-install-env-deps = 42.13-1.fc42
brltty
btrfs-progs
createrepo_c
device-mapper-multipath
dosfstools
e2fsprogs
f2fs-tools
fcoe-utils >= 1.0.12-3.20100323git
gdb
gnome-kiosk
gnome-remote-desktop
gsettings-desktop-schemas
hfsplus-tools
isomd5sum >= 1.0.10
kdump-anaconda-addon
kexec-tools
libblockdev-lvm-dbus
libblockdev-plugins-all >= 2.1
librsvg2
mesa-dri-drivers
nm-connection-editor
ntfs-3g
ntfsprogs
nvme-cli
openssl
ostree
podman
python3-pam
realmd
rpm-ostree >= 2023.2
rsync
skopeo
tmux
udisks2-iscsi
xfsprogs
zram-generator-defaults
```

PR #1033, opened 2024-11-11
blueprint: add map conversion check to fs customization unmarshaller

achilleas-k:
If an element in the `customizations.filesystem` array is not an object, the existing code would fail the conversion without an error, but would fail on the `d["mountpoint"].(type)` conversion, which produces an error message that isn't very useful:
```
  mountpoint must be string, got <nil> of type <nil>
```
The commit catches the conversion error and instead prints a more useful error message.


bcl (approved):
Look ok

supakeen (approved):

mvo5 (approved):
:heart:

PR #1034, opened 2024-11-11
Validate OSTree options on RHEL10

bcl:
This check seems to have been lost in the initial commit of this file. Discovered while working on weldr-client tests running on RHEL10.

Related: RHEL-67056

### DIFF
--- a/pkg/distro/rhel/rhel10/options.go
+++ b/pkg/distro/rhel/rhel10/options.go
@@ -22,6 +22,12 @@ func checkOptions(t *rhel.ImageType, bp *blueprint.Blueprint, options distro.Ima
 	// holds warnings (e.g. deprecation notices)
 	var warnings []string
 
+	if options.OSTree != nil {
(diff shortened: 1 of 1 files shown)


bcl:
There may be other things missing from this file, it looks quite a bit different from the fedora and rhel9 versions but I can't tell what supposed to be there and what isn't -- other than this, which was caught by my weldr-client tests when I run it on RHEL10.

achilleas-k:
We're not supporting edge for RHEL 10, so it might be more appropriate for this check to return an error when `ostree != nil` right away.

thozza:
I went ahead and implemented the check for all distros in https://github.com/osbuild/images/pull/1071

thozza:
> I went ahead and implemented the check for all distros in #1071

The PR was merged, so any el10 image types should now generate an error if OSTree image options are provided. @bcl, are you okay with closing this PR?

bcl:
Thanks!

PR #1035, opened 2024-11-12
image: add support for FIPS customization to `BootcDiskImage`

mvo5:
[draft as I think we want to double check if this makes sense in the bootc world first]

This commit is a followup for
https://github.com/osbuild/bootc-image-builder/pull/709 and adds support for FIPS to the `BootcDiskImage` image type.

One open question is if this should be done at this level or if the container itself should set the kernel cmdline to FIPS and bib would not bother.

### DIFF
--- a/pkg/image/bootc_disk.go
+++ b/pkg/image/bootc_disk.go
@@ -25,6 +25,7 @@ type BootcDiskImage struct {
 
 	// Customizations
 	KernelOptionsAppend []string
+	FIPS                bool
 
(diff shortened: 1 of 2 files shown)


ondrejbudai:
I don't think this is neccessary, the `fips=1` kernel argument should be set inside the source container image, see https://docs.fedoraproject.org/en-US/bootc/security-and-hardening/

mvo5:
Excellent, less work

PR #1036, opened 2024-11-12
reporegistry: introduce new  `distro.aliases` file

mvo5:
This commit introduces a new `distro.aliases` file that replaces the
role of the existing symlinks to provide an alias name for a distro.
The reason is that when using `go:embed` symlinks are not supported
and we have currently `centos-9.json -> centos-stream-9.json` in
our `osbuild-comopser` distro definitions.

By moving them from symlinks to alias files we can start using
`go:embed` which is nice for the daemonless work, i.e. the
`image-builder` binary will be self-contained and one can do:
```
$ go run github.com/osbuild/image-builder/cmd/image-builder --list-images
```
directly without any further configuration. It also solves how to
keep the default repository configuration between `osbuild-composer` and
`image-builder-cli` in sync.

---

reporegistry: remove duplication in LoadRepositories()

This commit removes some directory traversal duplication from
LoadRepositories() on the expense of more "expensive" repo loading.
The new code will just always load all repos and return the one
matching the distro. The advantage of this approach is that when
aliases are introduced there is only a single code path that needs
changing.

### DIFF
--- a/pkg/reporegistry/repository.go
+++ b/pkg/reporegistry/repository.go
@@ -1,6 +1,7 @@
 package reporegistry
 
 import (
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"strings"
@@ -63,6 +64,28 @@ func LoadAllRepositories(confPaths []string) (rpmmd.DistrosRepoConfigs, error) {
 
 				distrosRepoConfigs[distro] = distroRepos
 			}
+		}
+
+		// resolve aliases after buildng distroRepoConfigs
+		aliasesPath := filepath.Join(reposPath, "distro.aliases")
+		f, err := os.Open(aliasesPath)
+		if os.IsNotExist(err) {
+			continue
+		}
+		if err != nil {
+			return nil, err
+		}
+		defer f.Close()
+		var aliases []struct {
+			Name  string `json:"name"`
+			Alias string `json:"alias"`
+		}
+		err = json.NewDecoder(f).Decode(&aliases)
+		if err != nil {
+			return nil, err
+		}
+		for _, alias := range aliases {
+			distrosRepoConfigs[alias.Alias] = distrosRepoConfigs[alias.Name]
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/reporegistry/repository.go
```diff
@@ -64,6 +65,28 @@ func LoadAllRepositories(confPaths []string) (rpmmd.DistrosRepoConfigs, error) {
...
+		f, err := os.Open(aliasesPath)
+		if os.IsNotExist(err) {
+			continue
+		}
+		if err != nil {
+			return nil, err
+		}
+		defer f.Close()
+		var aliases []struct {
+			Name  string `json:"name"`
```
mvo5, line 80: I'm not too happy with the names, ideas for better naming welcome here  :)

thozza:
Just a quick comment: I am not sure if `distro.aliases` is the best name since we already have distro aliases, and it is a different thing:

- https://github.com/osbuild/images/blob/70fd73a592e867865394e30cdc418ea77b9986b5/pkg/distrofactory/distrofactory.go#L68-L99
- https://github.com/osbuild/osbuild-composer/blob/2a5d25d9c02e29a8898166f1e1cdc9c58313e78c/cmd/osbuild-composer/config.go#L18

From my PoV, the ability to add a new repo config file or a symlink is a feature. I have yet to look at the code, but I would keep the option to redefine the alias for the user (ideally, not compiled into the binary).

mvo5:
> Just a quick comment: I am not sure if `distro.aliases` is the best name since we already have distro aliases, and it is a different thing:
[..]

Oh, this super interesting - the only reason I am adding this is to avoid the symlink `centos-9.json -> centos-stream-9.json` in the *default* repo configuration. If a distro_alias in the composer config fixes this equally nicely and we can drop the symlink none of the `distro.alises` support is needed.

> From my PoV, the ability to add a new repo config file or a symlink is a feature. I have yet to look at the code, but I would keep the option to redefine the alias for the user (ideally, not compiled into the binary).

Both the ability to add new repo config, symlinks and to mask existing repos would be preserved. The goal of what I'm doing is strictly about making it easier to share the default/canonical repos from composer across multiple projects (image-builder-cli in my specific case).

mvo5:
Moving back to "draft" as it seems people do not really like this idea very much so far. So I will think a bit more what else we can do.

mvo5:
Closing in favor of https://github.com/osbuild/images/pull/1038 that hopefully addresses the concerns here.

PR #1037, opened 2024-11-12
reporegistry: support loading repos from fs.FS

mvo5:
[based on https://github.com/osbuild/images/pull/1036 and draft because it need 1036 and a tweak in osbuild-composer to make the repos there embeddable - mostly here so that people can see where this is heading]

reporegistry: support loading repos from fs.FS

This is a first step towards loading repositories via `go:embed`
and fully compatible with the old way of loading repos.


mvo5:
Closing in favor of https://github.com/osbuild/images/pull/1038

PR #1038, opened 2024-11-12
reporegistry: add new `LoadAllRepositoriesFromFS` and use in tests

mvo5:
This commit is a (hopefully) less controversial version of: https://github.com/osbuild/images/pull/1037

Here only a new helper to load repository information from an `fs.FS` is added and only used for loading the test repositories.

This will help to have a single test repository that can be embedded accross osbuild-composer, image-builder and images.

It also removes the public "NewTestedDefault()" in favor of a new `testrepos.New()` helper so that the `reporegistry` only has non-test API left.

[*maybe* this should be split into two commits actually]

This will allow me to use the test repos in the new `image-builder-cli` repo without the need to copy/import files from another git repo.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -28,10 +28,10 @@ import (
 	"github.com/osbuild/images/pkg/dnfjson"
 	"github.com/osbuild/images/pkg/manifest"
 	"github.com/osbuild/images/pkg/ostree"
-	"github.com/osbuild/images/pkg/reporegistry"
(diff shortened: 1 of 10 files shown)


thozza (approved):
This looks good to me.

I admit that it feels a bit weird to use embedded repos in the `gen-manifests` tool (other places are perfectly fine), but I could not come up with a reasonable scenario in which it would be an issue. Since previously, it would default to the repositories from `<project_root>/test/data/repositories`, this looks OK.

mvo5:
Thanks! With this in I can now open the first real `ibuilder` PR :)

PR #1050, opened 2024-11-19
ostree: fix proxy parsing

lzap:
Finally found the ostree resolving problem - it was in proxy parsing. The configuration value is not in URL form, it is in `hostname:port` form. And `url.Parse` silently parses such URI setting a nonsense scheme to `hostname`. TIL that `URL` in Go actually is technically URI: https://pkg.go.dev/net/url

The patch also adds missing fields into the OTK schema, if present, the ostree parse function can be used and tested separately.


pkg/ostree/ostree.go
```diff
@@ -158,7 +159,7 @@ func resolveRef(ss SourceSpec) (string, error) {
 	if err != nil {
 		return "", NewResolveRefError("error parsing ostree repository location: %v", err)
 	}
-	u.Path = path.Join(u.Path, "refs/heads/", ss.Ref)
+	u.Path = path.Join(u.Path, "refs", "heads", ss.Ref)
```
lzap, line 162: Irrelevant change, thought it is nicer looking.

lzap:
I did split the OTK hunks into a separate PR: https://github.com/osbuild/images/pull/1051

croissanne (approved):
wow, nice find :)

ezr-ondrej (approved):
Nice find! Thanks! :tada:

mvo5:
Would it make sense to add a regression test here? Also in the "test-as-documentation" sense this might be useful?

ezr-ondrej:
It would make sense to add test for sure... I'd prefer a follow-up tho, I wouldn't like to block this, unless someone disagrees.

lzap:
> Would it make sense to add a regression test here?

I can do a followup, I actually tried by running a tiny HTTP proxy via library but it was not great test to be honest.

PR #1052, opened 2024-11-19
ostree: extract `httpClientForRef` from resolveRef

mvo5:
This commit extracts a `httpClientForRef` helper from the `resolveRef()`
function. This helps with readability and testability. It also adds
a test for the proxy setting.

Not ideal as the test knows a bit too much about the implementation details but OTOH realtively straightfoward to add.

[build on top of https://github.com/osbuild/images/pull/1050]

### DIFF
--- a/pkg/ostree/ostree_test.go
+++ b/pkg/ostree/ostree_test.go
@@ -1,6 +1,7 @@
 package ostree
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -85,7 +86,7 @@ func TestOstreeResolveRef(t *testing.T) {
 				&MTLS{mTLSSrv.CAPath, mTLSSrv.ClientCrtPath, mTLSSrv.ClientKeyPath},
 				"",
 			})
-			assert.NoError(t, err)
+			require.NoError(t, err)
 			assert.Equal(t, expOut, out)
 		}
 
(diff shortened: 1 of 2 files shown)


lzap:
I think this is great, can you add a test for client cert/key as well as CA certs?

mvo5:
> I think this is great, can you add a test for client cert/key as well as CA certs?

Added

pkg/ostree/ostree_test.go
```diff
@@ -226,3 +227,54 @@ func TestValidate(t *testing.T) {
...
+	}
+	client, err := httpClientForRef("https", ss)
+	assert.NoError(t, err)
+
+	tlsConf := client.Transport.(*http.Transport).TLSClientConfig
+	assert.NoError(t, err)
+	// checking this from the outside is hard, there is no way to
+	// poke at the RootCAs internals so we just ensure one got added
+	assert.Equal(t, 1, len(tlsConf.RootCAs.Subjects()))
+	// no certificates got added
```
mvo5, line 278: Hm, this lint error is unfortunate, AFAICT this is the only way to introspect a golang CertPool, I hope that before they remove it they will also add something to introspect again, it seems strange

bcl (approved):
Looks good to me!

supakeen (approved):

lzap (approved):
Thank you.

PR #1055, opened 2024-11-22
blueprint: fix DiskCustomization.MinSize with existing  pattern

mvo5:
This commit fixes the issue that minsize cannot be a string
by using the existing pattern.

See https://github.com/osbuild/images/pull/1049 for alternative
ideas.

----

The disk customizations min-size setting is not accepting strings
or strings with units. This commit adds a failing test that will
validate the fix.

[this was found as part of the work in bib, https://github.com/osbuild/bootc-image-builder/pull/721]

PR #1057, opened 2024-11-22
datasize, blueprint: add new datasizes.Size type and use in blueprints

mvo5:
This is an alternative implementation of the idea in https://github.com/osbuild/images/pull/1049 based on feedback from @thozza (thanks!). 

---

blueprint: decouple FilesystemCustomization from marshaling

This commit decouples the marshaling representation of the
`FilesystemCustomization` from the actual struct. This means that
we can reuse custom unmarshaling via datasizes.Size and also
have our external represenation "pure".

If we continue using this pattern is also means that we can rename
fields in the marshaling and provide compatbility easily.

Thanks to Thozza, c.f.
https://github.com/osbuild/images/pull/1049#issuecomment-2493261642
and
https://github.com/osbuild/images/pull/983

---

datasizes: add new `Size` type with json/toml decoding
 support

This commit adds a new `datasizes.Size` type that is an alias
for uint64 but has supports direct json/toml decoding. So sizes
can be specified as 123, "123" or "123 GiB" and this is transparently
handled.

### DIFF
--- a/pkg/blueprint/filesystem_customizations.go
+++ b/pkg/blueprint/filesystem_customizations.go
@@ -10,49 +10,43 @@ import (
 )
 
 type FilesystemCustomization struct {
-	Mountpoint string `json:"mountpoint,omitempty" toml:"mountpoint,omitempty"`
-	MinSize    uint64 `json:"minsize,omitempty" toml:"minsize,omitempty"`
+	Mountpoint string
+	MinSize    uint64
 }
 
-func (fsc *FilesystemCustomization) UnmarshalTOML(data interface{}) error {
-	d, ok := data.(map[string]interface{})
-	if !ok {
-		return fmt.Errorf("customizations.filesystem is not an object")
-	}
+type filesystemCustomizationMarshaling struct {
+	Mountpoint string         `json:"mountpoint,omitempty" toml:"mountpoint,omitempty"`
+	MinSize    datasizes.Size `json:"minsize,omitempty" toml:"minsize,omitempty"`
+}
 
-	switch d["mountpoint"].(type) {
-	case string:
-		fsc.Mountpoint = d["mountpoint"].(string)
-	default:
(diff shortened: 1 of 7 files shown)


achilleas-k (dismissed):

pkg/blueprint/filesystem_customizations.go
```diff
@@ -10,49 +10,43 @@ import (
...
-		return fmt.Errorf("TOML unmarshal: mountpoint must be string, got \"%v\" of type %T", d["mountpoint"], d["mountpoint"])
-	}
-	minSize, err := decodeSize(d["minsize"])
+func (fsc *FilesystemCustomization) UnmarshalTOML(data any) error {
+	// This is the most efficient way to reuse code when unmarshaling
+	// structs in toml, it leaks json errors which is a bit sad but
+	// because the toml unmarshaler gives us not "[]byte" but an
+	// already pre-processed "any" we cannot just unmarshal into our
+	// "fooMarshaling" struct and reuse the result so we resort to
+	// this hack^Wworkaround (but toml will go long term anyway).
```
achilleas-k, line 28: Let's not be too silly in comments that might stick around for years or be read by people who might not fully get the meaning of `^W` for corrections and just call it a "workaround" (or a "hack", if you prefer).

Also, the phrase in parentheses should probably be "but toml will go *away* long term anyway", right?

mvo5: Thanks, fixed and force pushed together with a fix for the https://github.com/osbuild/images/pull/1055

pkg/blueprint/toml_json_bridge.go
```diff
@@ -0,0 +1,24 @@
+package blueprint
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// XXX: move to interal/common ?
+func unmarshalTOMLviaJSON(u json.Unmarshaler, data any) error {
```
mvo5, line 9: This should probably grow it's own test.

achilleas-k (approved):

PR #1063, opened 2024-11-25
distro: run `ValidateLayoutConstraints()` when using PartitionCustomizations

mvo5:
This commit will run `ValidateLayoutConstraints()` when a `DiskCustomization` is used.

Fwiw, the fact that we have to implement this in every distro (and in bib) and must not forget about it makes me wonder if we shouldn't flip this around. Call it by default but return a distinct error type (ValidateLayoutConstraintsError) when it is violated that clients (like a potential low-level otk partitioner) could choose to ignore. Wdyt?

### DIFF
--- a/pkg/distro/fedora/distro_test.go
+++ b/pkg/distro/fedora/distro_test.go
@@ -1032,3 +1032,40 @@ func TestDistroFactory(t *testing.T) {
 		})
 	}
 }
+
+func TestDistro_DiskCustomizationRunsValidateLayoutConstraints(t *testing.T) {
+	bp := blueprint.Blueprint{
+		Customizations: &blueprint.Customizations{
+			Disk: &blueprint.DiskCustomization{
+				Partitions: []blueprint.PartitionCustomization{
+					{
+						Type:            "lvm",
+						VGCustomization: blueprint.VGCustomization{},
+					},
+					{
+						Type:            "lvm",
+						VGCustomization: blueprint.VGCustomization{},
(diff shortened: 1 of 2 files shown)


achilleas-k:
> Fwiw, the fact that we have to implement this in every distro (and in bib) and must not forget about it makes me wonder if we shouldn't flip this around. Call it by default but return a distinct error type (ValidateLayoutConstraintsError) when it is violated that clients (like a potential low-level otk partitioner) could choose to ignore. Wdyt?

It does seem cleaner, implementation wise, I just don't know if an error is the right way to handle this.  Another alternative would be an argument, which is a bit ugly (unless we had optional arguments, but sadly we don't).

achilleas-k (approved):
LGTM.  We can reconsider the split later, but for now, let's at least get this in so we have the validation applied.

mvo5:
> > Fwiw, the fact that we have to implement this in every distro (and in bib) and must not forget about it makes me wonder if we shouldn't flip this around. Call it by default but return a distinct error type (ValidateLayoutConstraintsError) when it is violated that clients (like a potential low-level otk partitioner) could choose to ignore. Wdyt?
> 
> It does seem cleaner, implementation wise, I just don't know if an error is the right way to handle this. Another alternative would be an argument, which is a bit ugly (unless we had optional arguments, but sadly we don't).

Yeah, I was thinking argument at first too but then we have no way (except for in the customization itself) to set it :/

PR #1064, opened 2024-11-25
blueprint: add extra comments to `Validate{,LayoutConstraints}`

mvo5:
This commit ensures that the roles of `Validate{,LayoutConstraints}` are clearer from the code comments.

Follwup, c.f.
https://github.com/osbuild/images/pull/1041#discussion_r1848792511

[fwiw, writing it down like this still makes me feel like we currently optimize for the uncommon case here and users can easily forget to call the right API, but adding these comments is useful hopefully in any case]


pkg/blueprint/disk_customizations.go
```diff
@@ -326,6 +326,10 @@ func (v *PartitionCustomization) UnmarshalTOML(data any) error {
 //   - Plain filesystem types are valid for the partition type
 //   - All non-empty properties are valid for the partition type (e.g.
 //     LogicalVolumes is empty when the type is "plain" or "btrfs")
+//
+// Note that in *addition* consumers should also call
+// ValidateLayoutConstraints() to validate that the policy for disk
+// customizations is meet.
```
achilleas-k, line 332:
```suggestion
// customizations is met.
```

pkg/blueprint/disk_customizations.go
```diff
@@ -373,6 +377,11 @@ func validateMountpoint(path string) error {
 // ValidateLayoutConstraints checks that at most one LVM Volume Group or btrfs
 // volume is defined. Returns an error if both LVM and btrfs are set and if
 // either has more than one element.
+//
+// Note that this is a *policy* validations, in theory the "disk" code
+// does support the constraints but we choose not to allow them for now.
+// Each consumer of "DiskCustomization" should call this *unless* it's very
+// low-level and not end-user-facing.
```
achilleas-k, line 384:
```suggestion
// Note that this is a *policy* validation, in theory the "disk" code
// does support the constraints but we choose not to allow them for now.
// Each consumer of "DiskCustomization" should call this *unless* it's very
// low-level and not end-user-facing.
```

achilleas-k (changes requested):
Typos

pkg/blueprint/disk_customizations.go
mvo5: Thank you, nice catch!

achilleas-k (approved):

achilleas-k:
Skipping the tests and the queue.  PR is only docstring changes.

PR #1067, opened 2024-11-26
[RHEL-10] add internal-only image types (COMPOSER-2357)

thozza:
Add the following image types:

- ec2 (x86_64, aarch64)
- ec2-ha (x86_64)
- ec2-sap (x86_64)
- azure-rhui (x86_64, aarch64)
- azure-sap-rhui (x86_64)

These are based on RHEL-9 versions, with one difference for SAP images not including `gtk2` and  `libcanberra-gtk2` packages removed from RHEL-10. The internal team working on SAP is aware of this, and we may need to adjust this eventually, but this is a good starting point.

Related to https://issues.redhat.com/browse/COMPOSER-2357

### DIFF
--- a/pkg/distro/rhel/rhel10/ami.go
+++ b/pkg/distro/rhel/rhel10/ami.go
@@ -46,7 +46,7 @@ func mkAMIImgTypeAarch64() *rhel.ImageType {
 		[]string{"image"},
 	)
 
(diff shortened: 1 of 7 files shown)


thozza:
I did some unification of image definition code (just moving things around so that files in different distros look as similar as possible) as part of this work, but I'll submit it as a separate PR.

achilleas-k (approved):
Thanks for this.  I especially love all the cleanups and dedupes/unifications/syncing between RHEL 9 and 10 <3

PR #1068, opened 2024-11-26
ostree: allow passing any ostree source secrets

lzap:
Need this for the newly added `org.osbuild.mtls` secret that is required for Pulp MTLS feature.


achilleas-k (approved):
Tnanks!!

PR #1069, opened 2024-11-27
linter: exclude vendor/ directory

lzap:
SSIA


achilleas-k (approved):
We don't vendor, so I don't see why this is necessary.  Is it for potential local vendoring?

Regardless, can't hurt.

lzap:
Oh it was a mass PR, haven’t realized that :-)

PR #1073, opened 2024-11-28
disk: tweak ensure{Btrfs,LVM}() to be more linear

mvo5:
Tiny drive-by commit to make `ensure{Btrfs,LVM}()` more linear to read. I.e. avoid nested if/else when returning and do error checks early.

Best viewed side-by-side as the diff makes it hard to follow.

[edit: moved to draft as I have not inspected how much testing this function has, I'm pretty confident in the change but would really like to double check the test situation as well]
[edit2: played around a bit more with the tests and it seems it's well covered so moving out of draft again]

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -731,35 +731,35 @@ func (pt *PartitionTable) ensureLVM() error {
 
 	if _, ok := parent.(*LVMLogicalVolume); ok {
 		return nil
-	} else if part, ok := parent.(*Partition); ok {
-		filesystem := part.Payload
-
-		vg := &LVMVolumeGroup{
-			Name:        "rootvg",
-			Description: "created via lvm2 and osbuild",
-		}
-
-		// create root logical volume on the new volume group with the same
-		// size and filesystem as the previous root partition
-		_, err := vg.CreateLogicalVolume("rootlv", part.Size, filesystem)
-		if err != nil {
-			panic(fmt.Sprintf("Could not create LV: %v", err))
-		}
-
-		// replace the top-level partition payload with the new volume group
-		part.Payload = vg
-
-		// reset the vg partition size - it will be grown later
-		part.Size = 0
-
-		if pt.Type == PT_GPT {
-			part.Type = LVMPartitionGUID
-		} else {
-			part.Type = LVMPartitionDOSID
-		}
-
-	} else {
+	}
+	part, ok := parent.(*Partition)
+	if !ok {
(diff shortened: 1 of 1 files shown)


pkg/disk/partition_table.go
```diff
@@ -705,35 +705,35 @@ func (pt *PartitionTable) ensureLVM() error {
...
-			part.Type = "8e"
-		}
+	// reset the vg partition size - it will be grown later
+	part.Size = 0
 
+	if pt.Type == PT_GPT {
+		part.Type = LVMPartitionGUID
 	} else {
-		return fmt.Errorf("Unsupported parent for LVM")
+		part.Type = "8e"
```
achilleas-k, line 736: Let's make this a package constant alongside the others.

achilleas-k (approved):
:pinching_hand: tiny nitpick

pkg/disk/partition_table.go
mvo5: I can push it here:
```go
diff --git a/pkg/disk/disk.go b/pkg/disk/disk.go
index fdfd027cb..25da6d657 100644
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -70,6 +70,9 @@ const (
 
        // Partition type ID for ESP on dos
        DosESPID = "ef00"
+
+       // Partition type ID for LVM partitions on dos
+       DosLvmID = "8e"
 )
 
 // pt type -> type -> ID mapping for convenience
diff --git a/pkg/disk/partition_table.go b/pkg/disk/partition_table.go
index d91df09bd..947631b37 100644
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -733,7 +733,7 @@ func (pt *PartitionTable) ensureLVM() error {
        if pt.Type == PT_GPT {
                part.Type = LVMPartitionGUID
        } else {
-               part.Type = "8e"
+               part.Type = DosLvmID
        }
 
        return nil

```
but what is curious is that https://github.com/osbuild/images/blob/main/pkg/disk/disk.go#L82 uses a different DOS id for lvm it seems?

achilleas-k: I think `83` is valid for generic Linux but according to https://tldp.org/HOWTO/Partition-Mass-Storage-Definitions-Naming-HOWTO/x190.html you're right, `8e` is specific for LVM, so let's use that!

mvo5:
Had to rebase to fix conflicts, sorry

bcl (approved):
Looks good!

PR #1072, opened 2024-11-27
Support swap partitions and logical volumes in DiskCustomization (COMPOSER-2400)

achilleas-k:
This PR adds support for swap partitions and logical volumes when using the new `disk` blueprint customizations.

Internally, swap space is represented by a new `disk.Swap` type (a `disk.Element` type) which implements the `disk.PayloadEntity` interface.  I also added a new interface: `disk.FSTabEntity`.  This new entity variant is for representing items that can appear in the fstab file.  It differs from `disk.Mountable` in that it doesn't have a mountpoint, which makes it useful for distinguishing between filesystems and swap areas.  It will also be useful for defining swap files as part of the partitioning layout (even though it is not, strictly speaking, part of the partition table).

In the customizations, swap is represented by setting the filesystem type to "swap".  This is valid for both plain partitions and lvm logical volumes.  When the fs type is "swap", the mountpoint must always be empty.

PR #1074, opened 2024-11-28
disk: cover all entities in PartitionTable.features()

achilleas-k:
Follow-up from #1072 (https://github.com/osbuild/images/pull/1072#discussion_r1861786532)

---

Add a case for every entity type in the PartitionTable.features() callback.  More importantly, panic if any entity type is not covered by a case.  This ensures that when a new entity type is added, we need to consider whether it needs to be represented as a feature, added to an existing one, or ignored.

Co-authored-by: Michael Vogt <michael.vogt@gmail.com>

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -841,16 +841,20 @@ type partitionTableFeatures struct {
 	Swap  bool
 }
 
-// features examines all of the PartitionTable entities
-// and returns a struct with flags set for each feature used
+// features examines all of the PartitionTable entities and returns a struct
+// with flags set for each feature used. The meaning of "feature" here is quite
+// broad. Most disk Entity types are represented by a feature and the existence
+// of at least one type in the partition table means the feature is
+// represented. For Filesystem entities, there is a separate feature for each
+// filesystem type
 func (pt *PartitionTable) features() partitionTableFeatures {
 	var ptFeatures partitionTableFeatures
 
 	introspectPT := func(e Entity, path []Entity) error {
 		switch ent := e.(type) {
-		case *LVMLogicalVolume:
+		case *LVMVolumeGroup, *LVMLogicalVolume:
 			ptFeatures.LVM = true
-		case *Btrfs:
+		case *Btrfs, *BtrfsSubvolume:
 			ptFeatures.Btrfs = true
 		case *Filesystem:
 			switch ent.GetFSType() {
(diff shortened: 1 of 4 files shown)


ondrejbudai (dismissed):
Awesome! :)

pkg/disk/partition_table.go
```diff
@@ -867,6 +871,9 @@ func (pt *PartitionTable) features() partitionTableFeatures {
 			ptFeatures.Swap = true
 		case *LUKSContainer:
 			ptFeatures.LUKS = true
+		case *PartitionTable, *Partition:
```
mvo5, line 874: (super nitpick) - I personally like a small comment like `// nothing`, `// pass`, `// nothing to do` or similar for empty cases do clarify that this is intentionally left blank and for those faimilar with C it remind them that this is not an auto-fallthrough

mvo5 (approved):
I like it, tiny suggestion inline.

Fwiw, I was debating in my head if we should panic() or error, here it seems panic is okay, I was wondering if we could make it easier to ensure that we catch when new types are added, *maybe* something like this:
```go
diff --git a/pkg/disk/partition_table_test.go b/pkg/disk/partition_table_test.go
index 2924a5f00..258648e72 100644
--- a/pkg/disk/partition_table_test.go
+++ b/pkg/disk/partition_table_test.go
@@ -2291,10 +2291,13 @@ func TestPartitionTableFeatures(t *testing.T) {
        testCases := []testCase{
                {"plain", disk.PartitionTableFeatures{XFS: true, FAT: true}},
                {"plain-swap", disk.PartitionTableFeatures{XFS: true, FAT: true, Swap: true}},
+               {"plain-noboot", disk.PartitionTableFeatures{XFS: true, FAT: true}},
                {"luks", disk.PartitionTableFeatures{XFS: true, FAT: true, LUKS: true}},
                {"luks+lvm", disk.PartitionTableFeatures{XFS: true, FAT: true, LUKS: true, LVM: true}},
                {"btrfs", disk.PartitionTableFeatures{XFS: true, FAT: true, Btrfs: true}},
        }
+       assert.Equal(t, len(testCases), len(testdisk.TestPartitionTables),
+               "new partition table test cases got added but features not updated")
 
        for _, tc := range testCases {
                pt := testdisk.TestPartitionTables[tc.partitionType]

```
would ensure this - but I'm also probably overthinking it as it will be a bit before we add new Entities I suppose (but who knows what we need if bcachefs becomes a thing ;)

thozza (dismissed):
LGTM. FWIW, I like @mvo5 's nitpick :innocent:

pkg/disk/partition_table.go
achilleas-k: Done!

achilleas-k:
> Fwiw, I was debating in my head if we should panic() or error, here it seems panic is okay

I think panicking is appropriate for these kinds of programming level errors.

> I was wondering if we could make it easier to ensure that we catch when new types are added

I did a thing.  PTAL.

thozza (approved):
LGTM

PR #907, opened 2024-09-05
osbuild: new stage 'cacert' (HMS-4839)

lzap:
This pull request adds a new stage called 'cacert' to the osbuild package. It also includes file changes to support the CACustomization feature, which allows users to specify a list of certificates for the CA (Certificate Authority). The changes include updates to the Customizations struct, the osCustomizations function, and the OSCustomizations struct. Additionally, a new function called parseCerts is added to parse the certificate strings. The changes also include updates to the serialize and prependKernelCmdlineStage functions in the OS struct. Finally, a new file called ca_stage.go is added to the osbuild package, which contains the implementation of the NewCAStageStage function.

Needs: https://github.com/osbuild/osbuild/pull/1854

PR #1076, opened 2024-11-29
blueprint: fix cacerts name for TOML

lzap:
@avitova found some issues, I am going to address them.

To test the CA generation run this:

```
go run ./cmd/gen-manifests --arches x86_64 --distros fedora-41 --types qcow2 -config ./test/configs/all-customizations.json
```

It creates the following file in `test/data/manifests`:

```
...
         {
            "type": "org.osbuild.copy",
            "inputs": {
              "file-4c4e8c734e4ee3a117ca8f9363ba7b706d02bfc8f297c00b02ce02babddef51c": {
                "type": "org.osbuild.files",
                "origin": "org.osbuild.source",
                "references": [
                  {
                    "id": "sha256:4c4e8c734e4ee3a117ca8f9363ba7b706d02bfc8f297c00b02ce02babddef51c"
                  }
                ]
              }
            },
            "options": {
              "paths": [
                {
                  "from": "input://file-4c4e8c734e4ee3a117ca8f9363ba7b706d02bfc8f297c00b02ce02babddef51c/sha256:4c4e8c734e4ee3a117ca8f9363ba7b706d02bfc8f297c00b02ce02babddef51c",
                  "to": "tree:///etc/pki/ca-trust/source/anchors/27894af897dd2423607045716438a725f28a6d0b.pem",
                  "remove_destination": true
                }
              ]
            }
          },
          {
            "type": "org.osbuild.chown",
            "options": {
              "items": {
                "/etc/pki/ca-trust/source/anchors/27894af897dd2423607045716438a725f28a6d0b.pem": {
                  "user": "root",
                  "group": "root"
                }
              }
            }
          },
          {
            "type": "org.osbuild.pki.update-ca-trust"
          },
...
```

I think it works.

### DIFF
--- a/test/scripts/base-host-check.sh
+++ b/test/scripts/base-host-check.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# vim: sw=4:et
 set -euo pipefail
 
 running_wait() {
@@ -78,6 +79,22 @@ get_oscap_score() {
     fi
 }
 
+check_ca_cert() {
+    serial=$(jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -serial | cut -d= -f 2-)
+
+    echo "📗 Checking CA cert anchor file"
+    if ! [ -e "/etc/pki/ca-trust/source/anchors/${serial}.pem" ]; then
+        echo "Anchor CA file does not exist"
+        exit 1
+    fi
+
+    echo "📗 Checking extracted CA cert file"
+    if ! [ -e "/etc/pki/ca-trust/source/extracted/pem/directory-hash/Test_CA_for_osbuild.pem.pem" ]; then
+        echo "Extracted CA file does not exist"
+        exit 1
+    fi
+}
+
 echo "❓ Checking system status"
 if ! running_wait; then
 
@@ -113,5 +130,9 @@ if (( $# > 0 )); then
     config="$1"
     if jq -e .blueprint.customizations.openscap "${config}"; then
         get_oscap_score "${config}"
+    fi
+
+    if jq -e '.blueprint.customizations.cacerts.pem_certs[0]' "${config}"; then
+        check_ca_cert "${config}"
     fi
 fi
--- a/cmd/otk/osbuild-resolve-ostree-commit/main_test.go
+++ b/cmd/otk/osbuild-resolve-ostree-commit/main_test.go
@@ -202,7 +202,13 @@ func TestMockResolve(t *testing.T) {
 {
   "tree": {
     "ref": "otk/ostree/test",
-    "url": "https://ostree.example.org/repo"
+    "url": "https://ostree.example.org/repo",
+    "mtls": {
+        "ca": "ca.crt",
+        "client_cert": "client.crt",
+        "client_key": "client.key"
+    },
+    "proxy": "proxy.example.com:8080"
   }
 }
 `
--- a/pkg/blueprint/ca_customizations.go
+++ b/pkg/blueprint/ca_customizations.go
@@ -1,5 +1,0 @@
-package blueprint
-
-type CACustomization struct {
-	PEMCerts []string `json:"pem_certs,omitempty" toml:"pem_certs,omitempty"`
-}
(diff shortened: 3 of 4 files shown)


bcl:
I don't think this can be changed, any existing blueprint using this field will not be parsed correctly.

pkg/manifest/os.go
```diff
@@ -793,7 +794,9 @@ func (p *OS) serialize() osbuild.Pipeline {
 		}))
 	}
 
+	logger, _ := syslog.NewLogger(syslog.LOG_DEBUG, 0)
```
thozza, line 797: This looks like a debugging leftover, which should be removed.

thozza (changes requested):
Except for the leftover debug prints, this looks OK to me. :+1:

thozza:
> I don't think this can be changed, any existing blueprint using this field will not be parsed correctly.

This has not yet been exposed to users in any form AFAICT, and the TOML parsing here is not the same as the one used on-prem in `osbuild-composer`. So, all in all, this should be fine.

pkg/manifest/os.go
lzap: Busted! :-) Sure that was not intentional.

lzap:
So removed the logging line and added a test, but I am not sure where exactly this shell test is being executed.

thozza (approved):

lzap:
Tests look good, why this isn’t getting merged? Not sure if I should rebase.

thozza:
> Tests look good, why this isn’t getting merged? Not sure if I should rebase.

You dismissed my review by force-pushing. If you need just to rebase, using GH UI does not dismiss approvals.

thozza (approved):

lzap:
Ah now I remember, I was on the wrong branch and accidentally pushed here instead, sorry! Cheers.

test/scripts/base-host-check.sh
```diff
@@ -78,6 +79,22 @@ get_oscap_score() {
...
+    serial=$(jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -serial | cut -d= -f 2-)
+
+    echo "📗 Checking CA cert anchor file"
+    if ! [ -e "/etc/pki/ca-trust/source/anchors/${serial}.pem" ]; then
+        echo "Anchor CA file does not exist"
+        exit 1
+    fi
+
+    echo "📗 Checking extracted CA cert file"
+    if ! [ -e "/etc/pki/ca-trust/source/extracted/pem/directory-hash/Test_CA_for_osbuild.pem.pem" ]; then
```
thozza, line 92: @lzap I just noticed a potential typo in the filename - note the double `.pem.pem` :thinking:

test/scripts/base-host-check.sh
```diff
@@ -78,6 +79,22 @@ get_oscap_score() {
...
+        echo "Extracted CA file does not exist"
+        exit 1
+    fi
+}
+
```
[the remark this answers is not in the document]
thozza: If the manifest were not rebuilt during the CI run (because it didn't change), you'd find a link to the run in which it was tested in the job that generated it (e.g. https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/8565731980). Since the manifest didn't change and you added the test, it was not run at all. A complete manifest regeneration would be required. That can be achieved by bumping the `rngseed` in https://github.com/osbuild/images/blob/048c8c6745068f7a91c8fe48422b64d37d34c5ff/Schutzfile#L3

I did it in https://github.com/osbuild/images/pull/1093 and if my comment about the potential typo from above is correct, it should fail :thinking:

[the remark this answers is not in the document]
thozza: The story continuous:

- https://github.com/osbuild/images/pull/907 added the `cacert` customization only to `all-customizations.json`,
- `all-customizations.json` is applied only when building `qcow2` https://github.com/osbuild/images/blob/048c8c6745068f7a91c8fe48422b64d37d34c5ff/test/config-map.json#L2-L13
- We do not boot-test `qcow2` in this repository CI, so the change in the `base-host-check.sh` won't run in the CI.

test/scripts/base-host-check.sh
mvo5: Yeah, this is curious, do we generate funky filenames here or is the test not running correctly?

thozza: @mvo5 see https://github.com/osbuild/images/pull/1076#discussion_r1874828981. The modification is not applied to any image we boot-test, so this code is never executed.

test/scripts/base-host-check.sh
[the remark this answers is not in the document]
lzap: Yeah, apologies: https://github.com/osbuild/images/pull/1096

Issue #1065, opened 2024-11-25
Disk customization validation


achilleas-k:
With the new `DiskCustomization` we have two levels of validation:
- [`Validate()`](https://github.com/osbuild/images/blob/42f6babfae8118caea9e60a7024a2c127f5044a1/pkg/blueprint/disk_customizations.go#L315-L333) checks the structure for inconsistencies, conflicts, and missing values that make the customization invalid.
- [`ValidateLayoutConstraints()`](https://github.com/osbuild/images/blob/42f6babfae8118caea9e60a7024a2c127f5044a1/pkg/blueprint/disk_customizations.go#L377-L385) checks for things that are, strictly speaking, possible but not supported by the library due to our own internal policy (for example, multiple LVM volume groups or btrfs volumes).

This separation is useful for users of the library that may want to validate one but not the other.  Currently, `Validate()` is run automatically when calling `customizations.GetPartitioning()`, but the layout constraints aren't checked.  Also, the `Validate()` function is called in the [`NewCustomPartitionTable()`](https://github.com/osbuild/images/blob/42f6babfae8118caea9e60a7024a2c127f5044a1/pkg/disk/partition_table.go#L1191-L1194) function to check for the inconsistencies but doesn't disallow creating a partition table that violates the layout constraints.

[There is a concern](https://github.com/osbuild/images/pull/1063#issuecomment-2498364840) that this separation might cause issues with validations being dropped because callers might forget or not realise that there are multiple validations to run.

I should also add that there is a **third** validation, [`CheckMountpointPolicy()`](https://github.com/osbuild/images/blob/42f6babfae8118caea9e60a7024a2c127f5044a1/pkg/blueprint/filesystem_customizations.go#L40-L41), for checking if custom mountpoints are allowed for a given image type.

We should make it easier to validate customizations, harder to skip a specific validation, but also make it possible when needed.

Proposals:
1. Add options/arguments to the getter or the validator.
2. Validate everything in the getter function but also keep validator functions separate, so any caller can, if needed, get the customizations without calling the getter and call each validation separately.
3. Return typed errors and let the caller handle each validation failure type.
4. Validate everything always and separate it down the line if the need arises.

bcl:
I'd add disable options to the validator. This way there's one things to call, and by default it validates everything. And when someone copy and pastes an example they don't have to know about optional other validators :) If users need it to be less strict they can disable the layout and or constraint validation.

PR #1077, opened 2024-11-29
Add support for the new partitioning customizations to RHEL and CentOS

achilleas-k:
Enabling the new partitioning customizations for RHEL and CentOS.

I also found some duplication in Fedora that I cleaned up.

### DIFF
--- a/pkg/distro/rhel/imagetype.go
+++ b/pkg/distro/rhel/imagetype.go
@@ -39,6 +39,12 @@ const (
 	BlueprintPkgsKey = "blueprint"
 )
 
+// Default directory size minimums for all image types.
+var requiredDirectorySizes = map[string]uint64{
+	"/":    1 * datasizes.GiB,
+	"/usr": 2 * datasizes.GiB,
+}
+
 type ImageFunc func(workload workload.Workload, t *ImageType, customizations *blueprint.Customizations, options distro.ImageOptions, packageSets map[string]rpmmd.PackageSet, containers []container.SourceSpec, rng *rand.Rand) (image.ImageKind, error)
 
 type PackageSetFunc func(t *ImageType) rpmmd.PackageSet
@@ -193,6 +199,27 @@ func (t *ImageType) GetPartitionTable(
 	}
 
 	imageSize := t.Size(options.Size)
+	partitioning, err := customizations.GetPartitioning()
+	if err != nil {
+		return nil, err
+	}
+	if partitioning != nil {
+		// Use the new custom partition table to create a PT fully based on the user's customizations.
+		// This overrides FilesystemCustomizations, but we should never have both defined.
+		if options.Size > 0 {
+			// user specified a size on the command line, so let's override the
+			// customization with the calculated/rounded imageSize
+			partitioning.MinSize = imageSize
+		}
+
+		partOptions := &disk.CustomPartitionTableOptions{
+			PartitionTableType: basePartitionTable.Type, // PT type is not customizable, it is determined by the base PT for an image type or architecture
+			BootMode:           t.BootMode(),
+			DefaultFSType:      disk.FS_XFS, // default fs type for RHEL
+			RequiredMinSizes:   requiredDirectorySizes,
+		}
+		return disk.NewCustomPartitionTable(partitioning, partOptions, rng)
+	}
 
 	return disk.NewPartitionTable(&basePartitionTable, customizations.GetFilesystems(), imageSize, options.PartitioningMode, nil, rng)
 }
--- a/pkg/distro/rhel/imagetype.go
+++ b/pkg/distro/rhel/imagetype.go
@@ -180,7 +180,7 @@ func (t *ImageType) BootMode() platform.BootMode {
 }
 
 func (t *ImageType) GetPartitionTable(
-	mountpoints []blueprint.FilesystemCustomization,
+	customizations *blueprint.Customizations,
 	options distro.ImageOptions,
 	rng *rand.Rand,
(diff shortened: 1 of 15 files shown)


mvo5 (approved):
Thank you for adding the new customizations to all image types.

The duplication makes me slightly sad but I guess there is no way with the current design to do it differently, I would still love to explore ideas here. I was thinking about at least suggesting a comment like "// keep in sync with the other distros" but I guess that is implied already so redundant.

achilleas-k:
> The duplication makes me slightly sad but I guess there is no way with the current design to do it differently

If you're referring to the `checkOptions()` parts, I agree, there's a tonne of duplication there and it's also quite hard to trace.  There have been several attempts to make it nicer, but none of them reached completion.
(Curious what you think about one of my more recent attempts from earlier this year: https://github.com/osbuild/images/compare/main...achilleas-k:images:config-validation).

The way it's structured currently, it's a bit of a necessary evil because for some (but perhaps not most) customizations, the requirements are different between distributions and major versions.  Perhaps extracting the common ones would be enough to get the ball rolling on some of the simplification and to avoid diverging where possible.

achilleas-k:
RHEL 8 aarch64 builds are failing.  :thinking: 

Digging.

pkg/distro/rhel/rhel9/options.go
```diff
@@ -124,19 +124,33 @@ func checkOptions(t *rhel.ImageType, bp *blueprint.Blueprint, options distro.Ima
...
+
+	if err := blueprint.CheckMountpointsPolicy(mountpoints, policies.MountpointPolicies); err != nil {
+		return warnings, err
+	}
+
+	if err := blueprint.CheckDiskMountpointsPolicy(partitioning, policies.MountpointPolicies); err != nil {
+		return warnings, err
+	}
+
+	if err := partitioning.ValidateLayoutConstraints(); err != nil {
```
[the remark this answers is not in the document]
achilleas-k: xref #1065

achilleas-k:
A comment in @ondrejbudai #1080 PR got me thinking:
> This test is only performed on image types with a partition table, thus it doesn't run on e.g. installers and ostree commits.

Instead of listing all supported customizations for each image type like I tried to do in that linked branch, it might be simpler (and more readable) to have properties and capabilities for each image type, and then customizations would be valid or invalid for a given property.  For example "has partition table" is a boolean image type property.  Defining disk or filesystem customizations for any image type that doesn't have a PT is invalid.

This could probably be cleanly extended to all blueprint options: "has package payload" (if false, don't allow packages), "has ostree payload" (requires ostree commit), "can oscap"...

Those payload ones in particular will simplify things a lot.  Right now, when we're dealing with ostree types, we first check for `imageType.RPMOstree` (bool), which is true for all ostree image types.  But that's bad and wrong, because `edge-commit` has a package payload (and therefore installing packages is supported) but `edge-ami` and edge-installer` have an ostree commit payload, so they require an ostree commit and disallow packages.

achilleas-k:
> RHEL 8 aarch64 builds are failing. 🤔
> 
> Digging.

I managed to reproduce the issue.  When building a RHEL 8 image on a Fedora host, it seems the swap partition doesn't get created properly, or at least there's some incompatibility with the RHEL 8 kernel, resulting in:
```
Unable to find swap-space signature
```
when trying to enable the swap space.

Building the same image on RHEL 8 works fine.

It's strange though that this only happens on aarch64.

achilleas-k:
Might be time to switch to running each build on the same host as the target.  Not a huge change to set up, but might need some changes in the test scripts that were written to run on newer python versions.

OTOH, maybe running stuff on RHEL 8 doesn't make sense.  We're not targeting RHEL 8 any more (no more releases), so it might make more sense to test RHEL 8 builds on RHEL 9.

mvo5:
> > RHEL 8 aarch64 builds are failing. 🤔
> > Digging.
> 
> I managed to reproduce the issue. When building a RHEL 8 image on a Fedora host, it seems the swap partition doesn't get created properly, or at least there's some incompatibility with the RHEL 8 kernel, resulting in:
> 
> ```
> Unable to find swap-space signature
> ```
> 
> when trying to enable the swap space.
> 
> Building the same image on RHEL 8 works fine.
> 
> It's strange though that this only happens on aarch64.

I suspect the issue is the pagesize on aarch64. Apparently this got changed recently(ish).c.f. https://bugzilla.redhat.com/show_bug.cgi?id=1978730 and https://bugzilla.redhat.com/show_bug.cgi?id=2001569 - I guess we could try to be smart about it in images too and add a "mkswap -p 64k" option if we target arm and rhel < 9.2(?)

pkg/distro/rhel/imagetype.go
```diff
@@ -193,8 +199,29 @@ func (t *ImageType) GetPartitionTable(
...
+		if options.Size > 0 {
+			// user specified a size on the command line, so let's override the
+			// customization with the calculated/rounded imageSize
+			partitioning.MinSize = imageSize
+		}
+
+		partOptions := &disk.CustomPartitionTableOptions{
+			PartitionTableType: basePartitionTable.Type, // PT type is not customizable, it is determined by the base PT for an image type or architecture
+			BootMode:           t.BootMode(),
+			DefaultFSType:      disk.FS_EXT4, // default fs type for Fedora
```
mvo5, line 218: Hm, this is RHEL here, right? so should this be updated to XFS and comment twekaed?

achilleas-k: Copy pastaaaaaa!!

mvo5 (approved):
Thank you

PR #1080, opened 2024-12-02
Fix boot modes and ESP partitions for RHEL 8 and 9 (RHEL-69628)

ondrejbudai:
RHEL got support for UEFI on AWS in 8.9 and 9.3. However, we introduced some bugs in how we treat old distributions:

- the `ami` image type for older RHEL 8 versions was incorrectly marked as hybrid bootable, but the base partition table didn't have `/boot/efi` rendering the image unbootable on UEFI.
- The same is true for RHEL 9. This situation is even trickier here: the `ec2*` images were marked as legacy-only for older releases, but they actually contained an ESP partition, which is not needed.

I fixed all bugs, and added 2 tests to all distribution:
1) TestESP checks whether all images marked as UEFI or hybrid bootable have an ESP partition in their base partition table. The opposite is checked as well: legacy-only images shall not have an ESP partition.
2) TestAMIHybridBoot checks the boot mode of AMIs for RHEL 9 and RHEL 10.

There's a quite a lot of code duplication here, I'm happy to hear you ideas about how to deduplicate it. The use of internal structures/methods is making this a bit tricky.


ondrejbudai:
Agreed, let's put it on the plan when we are finished with all the required stuff?

achilleas-k (dismissed):
LGTM!

It looks like the `TestESP` test could be defined in only one place if the `ImageType` interface had a `GetPartitionTable()` method or similar, so that the tests could run from anywhere.

Something to consider for a follow-up I think.

thozza (commented):
I just looked at https://github.com/osbuild/images/commit/3eb5779e8bc948ba26bef611d27003458cfaa4e3 and the intention (and it looks that the implementation matched) was for AMIs build in the service or on-prem after RHEL-8.9 to be all hybrid-boot for any RHEL-8 version. And the boot mode was supposed to stay as it was only for EC2 images built in Brew. 

Then https://github.com/osbuild/images/commit/7acf235f8349552b705232773ed429336e42e066 removed `/boot/efi` for all RHEL-8 AMI/EC2 images and made the same partition table to be used for RHEL EC2 and AMI.

Then I fixed the missing `/boot/efi` by https://github.com/osbuild/images/commit/fa5aec481f7c62b7afddeb9fee8e172d4f4b0d6e.

And later reverted the `/boot/efi` for EC2 images from before 8.9 by https://github.com/osbuild/images/commit/35215e6215454a21cb48d35b02932b12b4301533, but due to previous changes, the same partitioning table is used for RHEL AMIs. The Platform for AMIs and EC2 images stayed different and this is from where the inconsistency is coming from.

I'll ask if we don't want to get to the initial version and use hybrid boot mode for RHEL-8 and RHEL-9 AMIs in the SaaS version and on-prem, regardless of the minor version? The initial idea was that the SaaS version builds the latest "version" of RHEL images. But we didn't want to change old images built in Brew when rebuilt.

Lastly, I like the tests that you've added. I'd suggest replacing the four copies by a 
single implementation in `pkg/distro/distro_test_common/distro_test_common.go` and then calling it from each distro.

thozza (changes requested):
Blocking to have the discussion first.

ondrejbudai:
Yeah, I vaguely remember that we discussed this before. I wonder if we should just keep it simple, and treat the ec2 and ami images as close to each other as posslble. It seems like there was quite a lot of confusion about this in the past, so making it dead simple might just be better.

No hard feelings, though. :)

thozza:
I also don't insist on moving back to the original idea and don't have a strong preference for using hybrid boot mode for older RHEL minor versions (shameless plug: which we should not be building them anyway in SaaS and on-prem without EUS repos :innocent: ).

It would be good to call this out explicitly in the code. Maybe something like the following comment could replace the existing comments and/or be added where appropriate:
```
// RHEL-internal EC2 images and RHEL AMI images are kept intentionally in sync
// with regard to not supporting hybrid boot mode before RHEL version X.Y.
// The partitioning table for these reflects that and is also intentionally in sync.
```

Would you have the time to factor out the new tests so that you have just a single copy? I'm happy to approve the PR afterward.

ondrejbudai:
Working on it, I'm fighting with import cycles! 🔪➿

ondrejbudai:
okay, I deduped it as much as possible. If we made `GetPartitionTable` a part of the `distro.ImageType` interface, we can even make it fully generic, but I don't like making methods public just for the sake of tests. But I can be convinced pretty easily. Maybe we can do this after #1077 lands.

I also added the comment suggested by @thozza, I couldn't figure out a better place, suggestions are welcome.

pkg/distro/distro_test_common/distro_test_common.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/osbuild/images/pkg/blueprint"
 	"github.com/osbuild/images/pkg/distro"
 	"github.com/osbuild/images/pkg/ostree"
-	testrepos "github.com/osbuild/images/test/data/repositories"
+	"github.com/osbuild/images/pkg/platform"
```
thozza, line 15: This looks suspicious in the context of the commit in which it is changed.

pkg/distro/rhel/rhel8/distro.go
```diff
@@ -115,6 +115,15 @@ func newDistro(name string, minor int) *rhel.Distribution {
 			ImageFormat: platform.FORMAT_RAW,
 		},
 	}
+
+	// Keep the RHEL EC2 x86_64 images before 8.9 BIOS-only for backward compatibility.
+	// RHEL-internal EC2 images and RHEL AMI images are kept intentionally in sync
+	// with regard to not supporting hybrid boot mode before RHEL version X.Y.
```
thozza, line 121: Nitpick: I'd replace `X.Y` with `8.9`, but I can live with it in the context of the first line.

pkg/distro/rhel/rhel9/distro.go
```diff
@@ -202,6 +202,15 @@ func newDistro(name string, major, minor int) *rhel.Distribution {
 			ImageFormat: platform.FORMAT_RAW,
 		},
 	}
+
+	// Keep the RHEL EC2 x86_64 images before 9.3 BIOS-only for backward compatibility.
+	// RHEL-internal EC2 images and RHEL AMI images are kept intentionally in sync
+	// with regard to not supporting hybrid boot mode before RHEL version X.Y.
```
thozza, line 208: ditto

thozza (commented):
Thanks for the changes. This looks nice.

The addition of `"github.com/osbuild/images/pkg/platform"` to `pkg/distro/distro_test_common/distro_test_common.go` is in the wrong commit.

pkg/distro/distro_test_common/distro_test_common.go
```diff
@@ -476,12 +477,43 @@ func TestDistro_OSTreeOptions(t *testing.T, d distro.Distro) {
...
-	require.Nil(t, err)
-	require.NotEmpty(t, testRepoRegistry)
-	distros := testRepoRegistry.ListDistros()
-	require.NotEmpty(t, distros)
-	return distros
+// TestESP checks whether all UEFI and hybrid images with a partition table have an ESP partition.
+// It also checks the opposite, i.e. that legacy images don't have an ESP. This test is only
+// performed on image types with a partition table, thus it doesn't run on e.g. installers and
+// ostree commits.
+// distros is a list of distros to test
```
achilleas-k, line 484:
```suggestion
// distros is a list of distros to test.
```

pkg/distro/fedora/distro_internal_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/osbuild/images/pkg/distro/distro_test_common"
+)
+
+// math/rand is good enough in this case
+/* #nosec G404 */
+var rng = rand.New(rand.NewSource(0))
+
+func TestESP(t *testing.T) {
+	var distros []distro.Distro
+	for _, distroName := range []string{"fedora-40", "fedora-41", "fedora-42"} {
```
achilleas-k, line 19: We could extend this to `fedora-99` and not have to worry about keeping the list up-to-date.  Or maybe we can add a follow-up item to iterate over all known distro versions based on `test/data/repositories/`.

pkg/distro/rhel/distro_test.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// math/rand is good enough in this case
+/* #nosec G404 */
+var rng = rand.New(rand.NewSource(0))
+
+func TestESP(t *testing.T) {
+	var distros []distro.Distro
+	distroFactory := distrofactory.NewDefault()
+	for _, distroName := range []string{"rhel-7.9", "rhel-8.8", "rhel-8.9", "rhel-8.10", "centos-8", "rhel-9.0", "rhel-9.2", "rhel-9.4", "centos-9", "rhel-10.0", "centos-10"} {
```
achilleas-k, line 26: Same as for Fedora.

pkg/distro/rhel/rhel8/distro.go
ondrejbudai: Thanks!

ondrejbudai:
I think I managed to fix the rebase fails, and I fixed the X.Y placeholders. :)

achilleas-k (commented):
LGTM.  Couple of suggestions/nitpicks, nothing to block on.  Except Tomáš comment about the import change being in the wrong commit.  That one should get fixed :)

thozza (approved):

pkg/distro/fedora/distro_internal_test.go
ondrejbudai: I tried this, and it creates a hard-to-break import loop. Let's figure this in a separate PR. :)

PR #1083, opened 2024-12-02
github: prevent script injections via PR branch names

ondrejbudai:
Prior this commit, ${{ github.event.workflow_run.head_branch }} got expanded in the bash script. A malicious actor could inject an arbitrary shell script. Since this action has access to a token with write rights the malicious actor can easily steal this token.

This commit moves the expansion into an env block where such an injection cannot happen. This is the preferred way according to the github docs:
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

Note that we don't allow external contributors to run CI without our manual approval, so this could not actually be abused.

### DIFF
--- a/.github/workflows/trigger-gitlab.yml
+++ b/.github/workflows/trigger-gitlab.yml
@@ -45,6 +45,8 @@ jobs:
 
       - name: Checkout branch
         id: pr_data
+        env:
+          BRANCH: ${{ github.event.workflow_run.head_branch }}
         run: |
           PR_DATA=$(mktemp)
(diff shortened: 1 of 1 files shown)


supakeen (approved):

achilleas-k (approved):
I see!

achilleas-k:
> Note that we don't allow external contributors to run CI without our manual approval, so this could not actually be abused.

True, but I'll admit I never pay too much attention to the BRANCH NAME before approving a run.

thozza (approved):

PR #1085, opened 2024-12-03
Add support for selecting a partition table type

ondrejbudai:
Users need to have control over the partition table type in certain cases. Most prominently, they need a dos-style partition table for single-board computers.
    
Let's add a simple customizations for it.

### DIFF
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -67,11 +67,17 @@ const (
 	// Partition type ID for any native Linux filesystem on dos
 	DosLinuxTypeID = "83"
 
-	// Partition type ID for BIOS boot partition on dos
-	DosBIOSBootID = "ef02"
+	// Partition type ID for LVM on dos
+	DosLVMTypeID = "8e"
+
+	// Partition type ID for BIOS boot partition on dos.
+	// Type ID is for 'empty'.
+	// TODO: drop this completely when we convert the bios BOOT space to a
+	// partitionless gap/offset.
+	DosBIOSBootID = "00"
 
 	// Partition type ID for ESP on dos
-	DosESPID = "ef00"
+	DosESPID = "ef"
 
 	// Partition type ID for swap
 	DosSwapID = "82"
@@ -87,7 +93,7 @@ var idMap = map[PartitionTableType]map[string]string{
 		"boot": DosLinuxTypeID,
 		"data": DosLinuxTypeID,
 		"esp":  DosESPID,
-		"lvm":  DosLinuxTypeID,
+		"lvm":  DosLVMTypeID,
 		"swap": DosSwapID,
 	},
 	PT_GPT: {
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -75,6 +75,9 @@ const (
 
 	// Partition type ID for swap
 	DosSwapID = "82"
+
+	// Partition type ID for PRep on dos
+	DosPRePID = "41"
 )
 
 // pt type -> type -> ID mapping for convenience
--- a/pkg/disk/partition.go
+++ b/pkg/disk/partition.go
@@ -87,7 +87,7 @@ func (p *Partition) IsBIOSBoot() bool {
 		return false
 	}
 
-	return p.Type == BIOSBootPartitionGUID
+	return p.Type == BIOSBootPartitionGUID || p.Type == DosBIOSBootID
 }
 
 func (p *Partition) IsPReP() bool {
--- a/pkg/disk/partition.go
+++ b/pkg/disk/partition.go
@@ -95,7 +95,7 @@ func (p *Partition) IsPReP() bool {
 		return false
 	}
 
-	return p.Type == "41" || p.Type == PRePartitionGUID
+	return p.Type == DosPRePID || p.Type == PRePartitionGUID
 }
 
 func (p *Partition) MarshalJSON() ([]byte, error) {
--- a/pkg/osbuild/sfdisk_stage.go
+++ b/pkg/osbuild/sfdisk_stage.go
@@ -1,5 +1,11 @@
 package osbuild
 
+import (
+	"fmt"
+
+	"github.com/osbuild/images/pkg/disk"
+)
+
 // Partition a target using sfdisk(8)
 
 type SfdiskStageOptions struct {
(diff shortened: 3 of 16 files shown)


achilleas-k (approved):
Thank you!

supakeen (approved):
Love the 🐦.

ondrejbudai:
Pushed a commit that fixes DOS partition tables with legacy boot. Our code for generation grub stage options require having a protective bios boot partition, but this partition was not recognized by the code responsible for finding it, leading to a panic.

I think that this fix will unblock us, but I wonder whether creating a protective bios boot partition on dos partition tables is the correct behaviour. IIRC, the BIOS core image was not protected, but I'm not sure why (maybe because of the partition count limits on dos?)

supakeen (approved):

pkg/disk/partition.go
```diff
@@ -87,7 +87,7 @@ func (p *Partition) IsBIOSBoot() bool {
 		return false
 	}
 
-	return p.Type == BIOSBootPartitionGUID
+	return p.Type == BIOSBootPartitionGUID || p.Type == DosBIOSBootID
```
achilleas-k, line 90: Let's add this for now but maybe add a note that we should probably drop BIOS boot from DOS PTs.  From what I understand, it's not needed and the `DosBIOSBootID` I added was wrong.

mvo5: Yeah, after a bit of reading up on this I came to the same conclusion, AIUI BIOS boot is not needed on dot PTs, there the MBR gap is used to store the second stage of grub.

mvo5 (approved):
Thank you, this seems fine and we should probably merge to unblock the Pi users but I would also like to see a followup just like @achilleas-k suggested that cleans up the DosBIOSBootID/DosESPID

achilleas-k (approved):

achilleas-k:
LVM test configs are failing to boot.

achilleas-k:
> LVM test configs are failing to boot.

Not sure what I saw when I said this.  It's not a boot failure, it's a build failure.  One without a good error message too:
```
org.osbuild.sfdisk: 2b4d8878a950bce7632c82f685b3f146c4b3fb60d69706f3438047c5a057330a {
  "label": "dos",
  "uuid": "4e2e4024-f881-4025-94c1-b248eb7542a6",
  "partitions": [
    {
      "bootable": true,
      "size": 2048,
      "start": 2048,
      "type": "ef02",
      "uuid": "FAC7F1FB-3E8D-4137-A512-961DE09A5549"
    },
    {
      "size": 409600,
      "start": 4096,
      "type": "ef00",
      "uuid": "68B[2905](https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/8560780982#L2905)B-DF3E-4FB3-80FA-49D1E773AA33"
    },
    {
      "size": 1048576,
      "start": 413696,
      "type": "83"
    },
    {
      "size": 2097152,
      "start": 1462272,
      "type": "83"
    },
    {
      "size": 37756928,
      "start": 3559424,
      "type": "E6D6D379-F507-44C2-A23C-238F2A3DF928"
    }
  ]
}
device/device (org.osbuild.loopback): loop0 acquired (locked: True)
label: dos
label-id: 4e2e4024-f881-4025-94c1-b248eb7542a6
start="2048", size="2048", type="ef02", uuid="FAC7F1FB-3E8D-4137-A512-961DE09A5549", bootable
start="4096", size="409600", type="ef00", uuid="68B2905B-DF3E-4FB3-80FA-49D1E773AA33"
start="413696", size="1048576", type="83"
start="1462272", size="2097152", type="83"
start="3559424", size="37756928", type="E6D6D379-F507-44C2-A23C-238F2A3DF928"
  File "/run/osbuild/bin/org.osbuild.sfdisk", line 164, in <module>
    ret = main(args["devices"], args["options"])
  File "/run/osbuild/bin/org.osbuild.sfdisk", line 155, in main
    pt.write_to(device)
    ~~~~~~~~~~~^^^^^^^^
  File "/run/osbuild/bin/org.osbuild.sfdisk", line 111, in write_to
    self.update_from(target)
    ~~~~~~~~~~~~~~~~^^^^^^^^
  File "/run/osbuild/bin/org.osbuild.sfdisk", line 122, in update_from
    assert len(disk_parts) == len(self.partitions)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
⏱  Duration: 0s
manifest - failed
```

Fully reproducible locally.

achilleas-k:
@ondrejbudai correctly identified this is because we're trying to write a dos partition table with more than 4 partitions.
`sfdisk` silently stops writing at 4 and then the read-back / sync fails to validate against the original list of commands.

pkg/osbuild/sfdisk_stage.go
```diff
@@ -36,7 +42,18 @@ type SfdiskPartition struct {
...
+func (o SfdiskStageOptions) validate() error {
+	if o.Label == disk.PT_DOS.String() && len(o.Partitions) > 4 {
+		return fmt.Errorf("sfdisk stage creation failed: \"dos\" partition table only supports up to 4 partitions: got %d", len(o.Partitions))
+	}
+	return nil
+}
+
 func NewSfdiskStage(options *SfdiskStageOptions, device *Device) *Stage {
+	if err := options.validate(); err != nil {
+		panic(err)
```
mvo5, line 54: (nitpick) slightly sad

pkg/osbuild/sfdisk_stage_test.go
```diff
@@ -41,3 +41,22 @@ func TestNewSfdiskStage(t *testing.T) {
 	actualStageDOS := NewSfdiskStage(&options, device)
 	assert.Equal(t, expectedStage, actualStageDOS)
 }
+
+func TestNewSfdiskStageInvalid(t *testing.T) {
```
mvo5, line 45: (super nitpick) I personally would combine this commit and the previous one (i.e. have code and test in the same commit)

mvo5 (dismissed):
Thank you! Some minor nitpicks inline but feel free to ignore

pkg/osbuild/sfdisk_stage_test.go
achilleas-k: Squashed.

achilleas-k:
This should be green now, but I noticed a couple of issues with partition types being set to GUIDs for DOS.  It doesn't break anything in light testing, but the partitions are marked as `empty` (`00`), which might be a problem.

I can fix it here or in a quick follow-up.

pkg/disk/disk.go
```diff
@@ -67,14 +67,23 @@ const (
 	// Partition type ID for any native Linux filesystem on dos
 	DosLinuxTypeID = "83"
 
-	// Partition type ID for BIOS boot partition on dos
-	DosBIOSBootID = "ef02"
+	// Partition type ID for LVM on dos
+	DosLVMTypeID = "8e"
```
mvo5, line 71: Maybe for a followup, let's use this new constant everywhere, right now I still see:
```console
$ git grep '"8e"'
cmd/otk/osbuild-gen-partition-table/main_test.go:                                                       Type:  "8e",
pkg/disk/partition_table.go:                    part.Type = "8e"
```

mvo5 (approved):
Thank you!

PR #1086, opened 2024-12-04
Minor fixes and changes to warnings handling in option checkers

achilleas-k:
**distro/fedora: don't return early from FIPS check in checkOptions()**

When the check was first added [1] it was the last in the function and
was written to immediately return.  New checks were added after the
check, without changing the early return, which means invalid
configurations were not caught.

Append to warnings instead of returning.

[1] 665a128f049c62b14285498f753a8c5fdbaa646d

---

**distro/fedora: always return warnings from checkOptions()**

Always return the warnings list from checkOptions() so that future
modifications or reshuffling of the checks doesn't drop any warnings
collected before the return.

---

**distro/rhel: always return warnings from checkOptions()**

Always return the warnings list from checkOptions() so that future
modifications or reshuffling of the checks doesn't drop any warnings
collected before the return.

---

**distro/rhel: don't print warnings in checkOptions()**

When collecting warnings in checkOptions() functions, don't
`log.Print()` the warning.  Let the caller decide whether to print them,
suppress them, make them errors, format them, etc.

---

### DIFF
--- a/pkg/distro/fedora/imagetype.go
+++ b/pkg/distro/fedora/imagetype.go
@@ -290,6 +290,8 @@ func (t *imageType) Manifest(bp *blueprint.Blueprint,
 func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOptions) ([]string, error) {
 	customizations := bp.Customizations
 
+	var warnings []string
+
 	if !t.rpmOstree && options.OSTree != nil {
 		return nil, fmt.Errorf("OSTree is not supported for %q", t.Name())
 	}
@@ -448,8 +450,7 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
 	}
 
 	if customizations.GetFIPS() && !common.IsBuildHostFIPSEnabled() {
-		w := fmt.Sprintln(common.FIPSEnabledImageWarning)
-		return []string{w}, nil
+		warnings = append(warnings, fmt.Sprintln(common.FIPSEnabledImageWarning))
 	}
 
 	instCust, err := customizations.GetInstaller()
@@ -474,5 +475,5 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
 		}
 	}
 
-	return nil, nil
+	return warnings, nil
 }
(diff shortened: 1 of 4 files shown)


thozza (approved):
Nice catch with the Fedora FIPS check :+1: The rest is also great, thanks :heart_eyes:

achilleas-k:

pkg/distro/fedora/imagetype.go
```diff
@@ -293,31 +293,31 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
 	var warnings []string
 
 	if !t.rpmOstree && options.OSTree != nil {
-		return nil, fmt.Errorf("OSTree is not supported for %q", t.Name())
+		return warnings, fmt.Errorf("OSTree is not supported for %q", t.Name())
```
mvo5, line 296: I'm not sure I follow here (for this commit and for commit 811bc5fa344a10334dae206d4e94436b5591bcdc). The way I see it is that the go convention is that for cases like this `return nil, err` is the norm, adding the warnings here seems to be unnecessary. I read in the commit message that it is to guard against code-shuffling but looking at this case here the only way that warnings would get dropped is if we chnage the code from a) erroring in this case and b) still returning early. That seems quite unlikely (and would hopefully get caught in code reviews).

I guess there is a argument that it might be nice to show the user both the error and the warnings so that they can fix both but that interpretation does not matches the commit message. Maybe I'm missing something? 


what exactly would this protect against?

mvo5 (commented):
I love the first and last commit and I think we should land them asap as they fix real issues.I'm less sure about the middle two (see inline comment), I am probably missing something though.

pkg/distro/fedora/imagetype.go
achilleas-k: > I guess there is a argument that it might be nice to show the user both the error and the warnings so that they can fix both

That's the idea, yeah.

> but that interpretation does not matches the commit message. Maybe I'm missing something?

The commit message mentioning reshuffling is because I was expecting (or assuming) an argument that it doesn't make sense to return `warnings` before there's any reason for it to be set.  Warnings only really get set pretty late in the function.  I suppose a more complete message would say something like:
> Always return warnings so the caller gets both warnings and error messages for the customizations and options.  Return from the top even if no warnings could possibly be generated to cover potential future additions or reordering.

Which makes me think, maybe we should be collecting all errors too and only returning them at the end, either as a list or `Join()`ed.

PR #1092, opened 2024-12-06
disk: align logical volumes up in EnsureSize()

mvo5:
When creating an LV they need to be aligned up boundaries or lvcreate will fail. When a lv is create via CreateLogicalVolume() this hapens because the size is calculcated with `v.AlignUp()` but when a resize is required via `lv.EnsureSize()` this is not taken into account.

This fixes the failure in bib PR#750


mvo5:
And indeed https://github.com/osbuild/bootc-image-builder/pull/750 is green now

ondrejbudai (approved):
Thanks!

PR #1096, opened 2024-12-09
blueprint: fix CA cert testing

lzap:
SSIA @thozza @mvo5

### DIFF
--- a/test/scripts/base-host-check.sh
+++ b/test/scripts/base-host-check.sh
@@ -80,17 +80,20 @@ get_oscap_score() {
 }
 
 check_ca_cert() {
-    serial=$(jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -serial | cut -d= -f 2-)
+    serial=$(jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -serial | cut -d= -f 2- | tr '[:upper:]' '[:lower:]')
+    cn=$(jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -subject | sed -E 's/.*CN ?= ?//')
 
-    echo "📗 Checking CA cert anchor file"
+    echo "📗 Checking CA cert anchor file serial '${serial}'"
     if ! [ -e "/etc/pki/ca-trust/source/anchors/${serial}.pem" ]; then
-        echo "Anchor CA file does not exist"
+        echo "Anchor CA file does not exist, directory contents:"
+        find /etc/pki/ca-trust/source/anchors
         exit 1
     fi
 
-    echo "📗 Checking extracted CA cert file"
-    if ! [ -e "/etc/pki/ca-trust/source/extracted/pem/directory-hash/Test_CA_for_osbuild.pem.pem" ]; then
-        echo "Extracted CA file does not exist"
+    echo "📗 Checking extracted CA cert file named '${cn}'"
+    if ! grep -q "${cn}" /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem; then
+        echo "Extracted CA cert not found in the bundle, tls-ca-bundle.pem contents:"
+        grep '^#' /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
         exit 1
     fi
 }
--- a/test/config-map.json
+++ b/test/config-map.json
@@ -8,7 +8,7 @@
       "fedora*"
     ],
     "image-types": [
-      "qcow2"
+      "ami"
     ]
   },
   "./configs/all-with-fips.json": {


test/scripts/base-host-check.sh
```diff
@@ -80,16 +80,21 @@ get_oscap_score() {
...
 check_ca_cert() {
+    ls /etc/pki/ca-trust/source/anchors
+
+    jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}"
+    jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -serial
     serial=$(jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -serial | cut -d= -f 2-)
 
+
     echo "📗 Checking CA cert anchor file"
     if ! [ -e "/etc/pki/ca-trust/source/anchors/${serial}.pem" ]; then
```
thozza, line 91: Looking at https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/8586354060, the `${serial}` is uppercase, while the actual filename is lowercase.

test/scripts/base-host-check.sh
```diff
@@ -80,16 +80,16 @@ get_oscap_score() {
 }
 
 check_ca_cert() {
-    serial=$(jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -serial | cut -d= -f 2-)
+    serial=$(jq -r '.blueprint.customizations.cacerts.pem_certs[0]' "${config}" | openssl x509 -noout -serial | cut -d= -f 2- | tr '[:upper:]' '[:lower:]')
 
-    echo "📗 Checking CA cert anchor file"
-    if ! [ -e "/etc/pki/ca-trust/source/anchors/${serial}.pem" ]; then
+    echo "📗 Checking CA cert anchor file serial $serial"
+    if ! [ -e "/etc/pki/ca-trust/source/anchors/27894af897dd2423607045716438a725f28a6d0b.pem" ]; then
```
thozza, line 86: Please use the variable `${serial}` instead.

lzap: Done, note the subject is hardcoded few lines below, I could not figure out an easy way of finding the filename, it is generated by `c_rehash` OpenSSL command. A bit more challenging.

lzap: I have found a dumb way to generalize this a bit replacing spaces with underscores. To give you some context, the `update-ca-certificates` command copies the certs into the `extracted` directory under `Common_name_with_underscores.pem` and then creates some symlinks. These are not serials, it is something different. The stage does copy the file into `anchors` directory with serial in the name.

thozza (dismissed):
Thanks!

lzap:
Finally it is passing, I figured out how to generalize the test. Ready for review.

test/scripts/base-host-check.sh
```diff
@@ -80,17 +80,20 @@ get_oscap_score() {
...
-    echo "📗 Checking extracted CA cert file"
-    if ! [ -e "/etc/pki/ca-trust/source/extracted/pem/directory-hash/Test_CA_for_osbuild.pem.pem" ]; then
-        echo "Extracted CA file does not exist"
+    echo "📗 Checking extracted CA cert file named '${cn}'"
+    if ! [ -e "/etc/pki/ca-trust/extracted/pem/directory-hash/${cn}.pem" ]; then
+        echo "Extracted CA file does not exist, directory contents:"
+        find /etc/pki/ca-trust/extracted/pem/directory-hash
         exit 1
     fi
 }
```
lzap, line 99: @achilleas-k fyi this test was introduced last week but it was broken (a typo in `directory-hash` path) and due to the config-map it was not executed properly so it sneaked into `main`. This fixes the map and the test.

achilleas-k: Thank you!

achilleas-k (approved):

lzap:
Also I have found out that on some (older) systems the directory-hash directory does not exist, so I came with a better test case: it now searches for CN in the bundle PEM file which is concatenated list of all certs with CNs as comments:

```
grep "Test CA for osbuild" /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
```

This should work also on older systems.

lzap:
Is green, let's put it on the queue @achilleas-k @thozza if you like the grep approach.

test/scripts/base-host-check.sh
```diff
@@ -80,17 +80,20 @@ get_oscap_score() {
...
         exit 1
     fi
 
-    echo "📗 Checking extracted CA cert file"
-    if ! [ -e "/etc/pki/ca-trust/source/extracted/pem/directory-hash/Test_CA_for_osbuild.pem.pem" ]; then
-        echo "Extracted CA file does not exist"
+    echo "📗 Checking extracted CA cert file named '${cn}'"
+    if ! grep -q "${cn}" /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem; then
+        echo "Extracted CA file does not exist, directory contents:"
+        find /etc/pki/ca-trust/extracted/pem
```
thozza, line 96: The grep-ing of `tls-ca-bundle.pem` makes me wonder if listing the content of `/etc/pki/ca-trust/extracted/pem` still makes sense? :thinking:

thozza: So, judging from https://github.com/osbuild/osbuild-composer/pull/4487/files#diff-a0a3d509a4e8903c0bede204d82662223fa270c20360d5bf2dcd2f1d2b68bbcdR264, it should be probably
```shell
grep '^#' /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
```

thozza: Thank you for the change. Please also amend the comment above it.

lzap:
Right, done.

thozza (approved):
Thanks!

lzap:
I do not see any comments by @achilleas-k the tests were fixed, adding to the merge queue. Thanks again for help with this, now I understand how tests do work in here a little bit better :-D

PR #1099, opened 2024-12-10
disk: rename all DOS type ID consts to match GUIDs

achilleas-k:
**disk,otk: use disk.DosLVMTypeID constant**

---

**disk: rename all DOS type ID consts to match GUIDs**

Rename the global constants to follow the same convention as the GPT
GUIDs and UUIDs.

---


bcl (approved):
Looks good!

supakeen (approved):

PR #1101, opened 2024-12-10
Main rootfs type

bcl:


bcl:
I've run into a problem with this, the tests seem to expect a static list of pipelines so they fail when it builds a manifest that skips the rootfs-image. Any ideas on how to deal with this?

PR #1105, opened 2024-12-12
Add ability to use plain squashfs as the iso rootfs

bcl:


bcl:
It seems to be making f39 and f40 squashfs as well, which it isn't supposed to. Changing to a draft for now.

pkg/image/image.go
```diff
@@ -27,3 +27,10 @@ func NewBase(name string) Base {
 		name: name,
 	}
 }
+
+type RootfsType uint64
+
+const ( // Rootfs type enum
+	SquashfsExt4Rootfs RootfsType = iota
```
ondrejbudai, line 34: Can you add docstrings here explaining the values?

ondrejbudai (commented):
Looks good to me, thanks! 😍

ondrejbudai (approved):
Thanks! :)

pkg/distro/fedora/images.go
```diff
@@ -675,6 +680,11 @@ func iotInstallerImage(workload workload.Workload,
 
 	img.Filename = t.Filename()
 
+	img.SquashfsCompression = "lz4"
+	if common.VersionGreaterThanOrEqual(img.OSVersion, VERSION_ROOTFS_SQUASHFS) {
```
cgwalters, line 684: This is the kind of thing though that I think should be owned by the OS, not by this repo. Especially because a lot of the code to support this at runtime had to have landed in the OS and is versioned with it already.

Maybe something like just having a config file in the dracut-live package - or even having this repo inspect a systemd `.mount` unit as configured in that root for the Live OS, etc.

I'm mainly familiar with the FCOS LiveOS code, see e.g.
https://github.com/coreos/fedora-coreos-config/blob/e8073f8d1933fa22b79b7ccc43069264fe22a823/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator#L73
And there probably what would be better is to move that to a static unit instead of being generated, and then code like this can parse it.

cgwalters (approved):
Looks sane to me but I will do my usual thing that I think the OS content should drive partitioning.

pkg/distro/fedora/images.go
bcl: Maybe? But such speculation is outside the scope of this PR. The way things are working now is this is defined in the distro specific code in images.

bcl:
Moving this back to draft for the moment, I need to move RootfsType into manifest instead of image to make it easier to add Erofs support later (it will need to switch out the Squashfs stage in anaconda_installer_iso_tree.go, I'll do that Monday.

bcl:
Ok, this should be ready now. There followup to this later that will add erofs support instead of squashfs, but that's for later.

pkg/distro/distro_test.go
```diff
@@ -171,6 +172,15 @@ func TestImageTypePipelineNames(t *testing.T) {
 					// sure they match.
 					assert.Equal(imageType.Exports()[0], pm.Pipelines[len(pm.Pipelines)-1].Name)
 
+					// The pipelines named in allPipelines must exist in the manifest, and in the
+					// order specified (eg. 'build' first) but it does not need to be an exact
+					// match. Only the pipelines with rpm or ostree metadata are required.
```
achilleas-k, line 177: I think technically the order doesn't matter, but might as well check for it.

Thanks for this!

achilleas-k (approved):
Cheers!  This all LGTM.

pkg/distro/distro_test.go
bcl: Order was inherent in the previous behavior, but I also swear I saw a comment about it someplace. Which I cannot find now :)

PR #1107, opened 2024-12-13
pkg: make `seed` in `ImageType.Manifest()` random by default

mvo5:
This commit changes the API of `ImageType.Manifest()` to take a pointer to the seed value. If no seed value is given it will automatically generate a random one.

This avoid the pitfall that the consumer needs to know about properly seeding the manifests. The default use-case is to generate non-fixed seed manifests so this commit makes it easier.

[edit: this is one possible way to avoid https://github.com/osbuild/image-builder-cli/pull/6#discussion_r1882807289]


ondrejbudai (commented):
I like that it's safe by default.

pkg/distro/distro.go
```diff
@@ -120,7 +122,9 @@ type ImageType interface {
 	// specified in the given blueprint; it also returns any warnings (e.g.
 	// deprecation notices) generated by the manifest.
 	// The packageSpecSets must be labelled in the same way as the originating PackageSets.
-	Manifest(bp *blueprint.Blueprint, options ImageOptions, repos []rpmmd.RepoConfig, seed int64) (*manifest.Manifest, []string, error)
+	// A custom seed for the rng can be specified, if nil the seed will
+	// be random.
```
achilleas-k, line 126: Maybe we can be more accurate and informative here and say that it will be derived from the current time.

mvo5: Thanks! I'm happy to do that, I was thinking of using crypto.Rand to seed the generator but maybe I'm overthinking it. Happy to either document or change.

thozza (approved):
Thanks, this looks OK to me.

achilleas-k (approved):

pkg/distro/distro.go
achilleas-k: Not important.  Can do later or not at all.

mvo5:
I prepared the followups to fix the consumers

PR #1115, opened 2024-12-20
Add support for custom partition types/GUIDs in blueprints

ondrejbudai:
Introduce the ability to specify custom partition types/GUIDs for both DOS and GPT partition types.

For DOS partitions, the field is technically called type, but this is already a taken field, so I went with GUID from the GPT vocabulary for the new field name.

PR #1116, opened 2024-12-20
disk: enable customizing ESP partitions

ondrejbudai:
This commits allow people to customize their ESP partition. Previously,
we always added an ESP partition ourselves, regardless of the given
customizations, so the customization errored out if there was an ESP
because of a conflict. This commit alters the behaviour, so ESP is
only added automatically if it's not in the customizations.

Based on #1115.

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -1086,7 +1086,7 @@ func addBootPartition(pt *PartitionTable, bootFsType FSType) error {
 // The function will append the new partitions to the end of the existing
 // partition table therefore it is best to call this function early to put them
 // near the front (as is conventional).
-func addPartitionsForBootMode(pt *PartitionTable, bootMode platform.BootMode) error {
+func addPartitionsForBootMode(pt *PartitionTable, disk *blueprint.DiskCustomization, bootMode platform.BootMode) error {
 	switch bootMode {
 	case platform.BOOT_LEGACY:
 		// add BIOS boot partition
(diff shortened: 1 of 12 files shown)


pkg/blueprint/disk_customizations.go
```diff
@@ -64,6 +66,15 @@ type PartitionCustomization struct {
...
 	MinSize uint64 `json:"minsize" toml:"minsize"`
 
+	// The partition type GUID for GPT partitions. For DOS partitions, this
+	// field can be used to set the partition type (e.g. "swap").
+	// If not set, the code will make the best guess based on the mountpoint or
+	// the payload type.
+	// Examples:
+	// 3B8F8425-20E0-4F3B-907F-1A25A76F98E8 (/srv on GPT)
+	// 06 (FAT16 on DOS)
+	GUID string `json:"guid,omitempty" toml:"guid,omitempty"`
```
bcl, line 76: I don't like calling this a GUID when it isn't always a GUID. In other places this is called Type, but that's obviously already been used. Maybe `PartType`?

mvo5: This got renamed to PartType in https://github.com/osbuild/images/pull/1115/files and is no longer needed in this PR.

achilleas-k:
We need to relax the customization policy for the mountpoints as well.

thozza (dismissed):
LGTM, Thanks.

FWIW, you could also unit test that one can only see `vfat` for `/boot/efi` if we don't already (I didn't check :innocent: ).

mvo5:
> We need to relax the customization policy for the mountpoints as well.

Yeah, I noticed when adding the test, I added this now, maybe a bit too relaxed :) See the rational in the commit message but if its considered risky I can add code to have different policies for disk and filesystem customizations.

Ideally this should be better tested, either in base-host-check.sh so that it validates that /boot/efi got actually changed in size or (preferable IMHO) have a manifest test that just generates a manifest with the changed size and validates that the osbuild stage options are set correctly. But that is probably followup work and ties into the wider discussion how to unify the bib/ibcli/images testing more.

achilleas-k:
Tests are failing to boot in AWS.

supakeen (approved):

mvo5 (approved):

supakeen (approved):

thozza (approved):

supakeen:
What's the status on this; do we still want it or has it been superseded by recent partition table changes (cc @achilleas-k?)? I can rebase it if necessary.

PR #1117, opened 2024-12-20
Add erofs support 

bcl:
Things that are needed:
* allow zstd compression in org.osbuild.erofs - https://github.com/osbuild/osbuild/pull/1969
* Add block size support to org.osbuild.erofs - also in https://github.com/osbuild/osbuild/pull/1969
* ~~RHEL 10 kernel needs to support erofs and zstd compression~~

### DIFF
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -50,6 +50,7 @@ type AnacondaInstallerISOTree struct {
 	isoLabel string
 
 	RootfsCompression string
+	RootfsType        RootfsType
 
 	OSPipeline         *OS
 	OSTreeCommitSource *ostree.SourceSpec
@@ -289,7 +290,7 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 	// The iso's rootfs can either be an ext4 filesystem compressed with squashfs, or
 	// a squashfs of the plain directory tree
 	var squashfsStage *osbuild.Stage
-	if p.rootfsPipeline != nil {
+	if p.RootfsType == SquashfsExt4Rootfs {
 		squashfsStage = osbuild.NewSquashfsStage(&squashfsOptions, p.rootfsPipeline.Name())
 	} else {
 		squashfsStage = osbuild.NewSquashfsStage(&squashfsOptions, p.anacondaPipeline.Name())
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -49,7 +49,7 @@ type AnacondaInstallerISOTree struct {
 
 	isoLabel string
 
-	SquashfsCompression string
+	RootfsCompression string
 
 	OSPipeline         *OS
 	OSTreeCommitSource *ostree.SourceSpec
@@ -273,8 +273,8 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 		}
 	}
 
-	if p.SquashfsCompression != "" {
-		squashfsOptions.Compression.Method = p.SquashfsCompression
+	if p.RootfsCompression != "" {
+		squashfsOptions.Compression.Method = p.RootfsCompression
 	} else {
 		// default to xz if not specified
 		squashfsOptions.Compression.Method = "xz"
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -155,6 +155,42 @@ func (p *AnacondaInstallerISOTree) getBuildPackages(_ Distro) []string {
 	return packages
 }
 
+// NewSquashfsStage returns an osbuild stage configured to build
+// the squashfs root filesystem for the ISO.
+func (p *AnacondaInstallerISOTree) NewSquashfsStage() *osbuild.Stage {
+	var squashfsOptions osbuild.SquashfsStageOptions
+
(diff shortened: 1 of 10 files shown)


bcl:
Currently works to build an erofs `image-installer` image on RHEL10 but will not boot due to the kernel not supporting zstd on erofs.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -23,6 +23,7 @@ type RootfsType uint64
 const ( // Rootfs type enum
 	SquashfsExt4Rootfs RootfsType = iota // Create an EXT4 rootfs compressed by Squashfs
 	SquashfsRootfs                       // Create a plain squashfs rootfs
+	ErofsRootfs                          // Create a plain erofs rootfs
```
mvo5, line 26: Nice how this fits together with the previous work!

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -154,6 +161,72 @@ func (p *AnacondaInstallerISOTree) getBuildPackages(_ Distro) []string {
...
+			BCJ: osbuild.BCJOption(p.anacondaPipeline.platform.GetArch().String()),
+		}
+	}
+
+	// The iso's rootfs can either be an ext4 filesystem compressed with squashfs, or
+	// a squashfs of the plain directory tree
+	if p.RootfsType == SquashfsExt4Rootfs && p.rootfsPipeline != nil {
+		return osbuild.NewSquashfsStage(&squashfsOptions, p.rootfsPipeline.Name())
+	}
+	return osbuild.NewSquashfsStage(&squashfsOptions, p.anacondaPipeline.Name())
```
mvo5, line 197: We have have tests in `anaconda_installer_iso_tree_test.go`  that instances the manifest, I wonder if we should add a test there that generates a squashfs and an erofs manifest and check the json for the right stages? But could also be a followup of course :)

mvo5 (commented):
This looks very nice, one question about a possible test. Also, what is keeping this in draft? It seems quite complete (code-wise) or am I missing something?

thozza:
Rebased due to image build cache layout change (https://github.com/osbuild/images/pull/1130)

supakeen (dismissed):
Very nice. I would only wonder if we want to make `RootfsCompression` also some kinda-enum; or if we want to move them together into a single property just called `rootfs` that contains both the rootfstype and the rootfscompression, just so we can (more easily) validate them together in case some compressions do not fit some rootfstypes?

pkg/osbuild/erofs_stage.go
```diff
@@ -0,0 +1,24 @@
...
+	Filename string `json:"filename"`
+
+	Compression     *ErofsCompression `json:"compression,omitempty"`
+	ExtendedOptions []string          `json:"options,omitempty"`
+	ClusterSize     *int              `json:"cluster-size,omitempty"`
+}
+
+func (ErofsStageOptions) isStageOptions() {}
+
+func NewErofsStage(options *ErofsStageOptions, inputPipeline string) *Stage {
```
mvo5, line 18: (nitpick) I personally like accompanying test for the stages, I put a tiny one in https://github.com/osbuild/images/commit/a368a6dd7c76183b1dad2f8148c863f6f28fa0a7 and I'm happy to do it as a followup. It may seem silly and its definitely busywork but I see (small) benefits because a) it ensures the json tags are right initially and next time something is added at least the omitempty will not forgoten :) and b) that there is a ready-made example how this will look like in json in a real manifest. But totally optional of course, I can propose it via followup or we ignore it :)

mvo5 (dismissed):
Thank you! This looks very nice (small idea inline for your consideration but no blockers from my pov).

pkg/manifest/anaconda_installer_iso_tree_test.go
```diff
@@ -404,6 +423,31 @@ func TestAnacondaISOTreeSerializeWithOS(t *testing.T) {
...
+
+	t.Run("plain+erofs-rootfs", func(t *testing.T) {
+		pipeline := newTestAnacondaISOTree()
+		pipeline.OSPipeline = osPayload
+		pipeline.RootfsType = ErofsRootfs
+		pipeline.serializeStart(nil, nil, nil, nil)
+		sp := pipeline.serialize()
+		pipeline.serializeEnd()
+		assert.NoError(t, checkISOTreeStages(sp.Stages,
+			append(payloadStages, []string{"org.osbuild.erofs"}...),
```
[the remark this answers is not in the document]
thozza: I'd add that IMHO `[]string{"org.osbuild.kickstart", "org.osbuild.isolinux"}...` above also seems like an unnecessary complication.

thozza (dismissed):
LGTM

achilleas-k (commented):
LGTM but I'd like to see the comments resolved.  They're all minor (the `append()`ing and the test that @mvo5 already wrote), but good to have it clean from the start.

I don't mind it being a follow-up PR if merging this unblocks other work.

pkg/manifest/anaconda_installer_iso_tree_test.go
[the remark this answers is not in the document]
bcl: Ha, whoops, no need to expand a single entry there.
As for the others, I left them for consistency with the other tests.

pkg/osbuild/erofs_stage.go
bcl: Ugh, I totally forgot to add that, thanks!

thozza (approved):

PR #1124, opened 2025-01-03
Test/imgtestlib/get_osbuild_commit(): consider only gitlab-ci-runner distro (HMS-5243)

thozza:
Previously, the osbuild commit stored in image build `info.json` was based on the host distro. However, we always use the same distro for all image build jobs, which is defined in the Schutzfile.

Modify the `get_osbuild_commit()` function to not take the distro as an argument, but instead read the `gitlab-ci-runner` from the Schutzfile and determine the osbuild commit based on its distro.

This will make it much easier to determine the actual osbuild commit that was (supposed to be) used for the image builds on the checked out branch. Consistency of data in `image.info` and the reality will be ensured.

Make `get_osbuild_commit()` raise a KeyError in case there's no osbuild commit defined in Shutzfile for the gitlab-ci-runner distro, so that we do not accidentally miss it and use unknown osbuild version from the distro repos.

/jira-epic COMPOSER-2318

JIRA: [HMS-5243](https://issues.redhat.com/browse/HMS-5243)

### DIFF
--- a/test/scripts/build-image
+++ b/test/scripts/build-image
@@ -46,9 +46,7 @@ def main():
 
     osbuild_ver, _ = testlib.runcmd(["osbuild", "--version"])
 
-    osrelease = testlib.read_osrelease()
-    distro_version = osrelease["ID"] + "-" + osrelease["VERSION_ID"]
-    osbuild_commit = testlib.get_osbuild_commit(distro_version)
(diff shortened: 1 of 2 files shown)


thozza:
I've missed the other dependencies. I'll rethink the whole change from a different angle...

PR #1127, opened 2025-01-06
GH/test-osbuild-composer-integration: don't set env.pr_head

thozza:
The `pr_head` env variable set by the step was always empty, because there was a typo in the source variable "hase" vs. "head" / "base". I was not able to find any actual use of the `pr_head` variable, so let's delete it altogether from the action step.


achilleas-k (approved):
Oops :)

PR #1130, opened 2025-01-07
Rework S3 CI cache prefix (HMS-5274)

thozza:
To replace the manifest-db in osbuild CI with the images repo image build cache, we will need to be able to easily construct the S3 cache prefix corresponding to a specific commit checkout of this repository.

The current S3 prefix for the image build cache looks like this:
```
<s3_bucket>/images/builds/<img_distro>/<img_arch>/<osbuild_nevra>/<img_manifest_id>/
```
e.g.
```
<s3_bucket>/images/builds/centos-9/x86_64/osbuild-136-1.20190819git9071cd0.fc40.noarch/ede3be49f4a6748fb65948aea9eccdf1a6114c2d270a05178f92c3494b3854e1/
```

The most problematic part of constructing the correct prefix without running scripts on the CI runner is the osbuild NEVRA. The specific NEVRA is an RPM build of a particular osbuild commit for a specific distribution (the CI runner distro). The osbuild commit and the default CI runner distro can be determined from a local copy of the `Schutzfile`. However, getting the NEVRA requires querying a DNF repository for the CI runner distro.

After brainstorming with @achilleas-k, we agreed to rework the image build cache prefix by:

- Replacing the osbuild NEVRA with a combination of osbuild commit ref and the CI runner distro. The reason is that this mimics the use of the current use of NEVRA, resulting in a full image build cache rebuild in case the osbuild version or the CI runner distro changes.
- Moving the osbuild ref and CI runner distro to the beginning of the prefix. This will result in fewer files downloaded when checking if any image needs to be rebuilt. Previously, we would need to download cache for any old osbuild NEVRA if we wanted to download cache files for all distros or architectures.
- Prefixing some values in the path prefix to fully specify their meaning (e.g. runner distro vs. image distro; osbuild commit SHA vs. manifest ID SHA).

The newly proposed S3 prefix for the image build cache looks like this:
```
<s3_bucket>/images/builds/osbuild-ref-<commit_ref>/runner-<runner_distro>/<img_distro>/<img_arch>/manifest-id-<img_manifest_id>/
```
e.g.
```
<s3_bucket>/images/builds/osbuild-ref-fcb93bde01d4a027c67b5747c8f976ceb4fc8d80/runner-fedora-41/rhel-9.6/x86_64/manifest-id-1ef93e5548a00c1ffe6a33d543da30c19924f5ff0ad1b06440030f63dd4ce4ee/
```

/jira-epic COMPOSER-2318

JIRA: [HMS-5274](https://issues.redhat.com/browse/HMS-5274)


achilleas-k (approved):
Very nice.  All LGTM.

Since I was part of the planning, I wont merge yet and wait for others to review as well.

schuellerf (approved):
LGTM, although I'm not sure if the `ref-` is really needed as it's pretty obvious and just clutters the name? But works either way, I guess.

PR #1132, opened 2025-01-08
many: enable support for librepo based rpm downloading

mvo5:
[draft as it is light on tests and also needs osbuild pr#1974 first]

This PR adds support for (optional) librepo based rpm downloading. It can be controlled when generating the manifest via the new `RpmDownloader` iota. This needs osbuild PR#1974.

Note that this PR implements manifest.Manifest.SerializeFull() which I think we do not want in the final PR, I did that just to keep the size of the PR low so that validating the approach is easier.


mvo5:
This is all split out/improved and open as real PRs. closing this draft.

PR #1137, opened 2025-01-13
anaconda: Use erroronfail

cgwalters:
Not doing this can silently mask errors.

xref https://github.com/containers/bootc/issues/931


mvo5 (dismissed):
Thank you, this looks very reasonable. There is a risk of existing things to break though (when the error in the script was inconsequential) but those are arguably are already broken :)

cgwalters:
There's only two places that can go wrong here:

- The switch
- Writing the completion to `/dev/ttyS0`

If the switch fails then we definitely want to fatally error. The write to `/dev/ttyS0` is more arguable...I pushed a commit which continues to ignore errors there.

supakeen (approved):

supakeen:
Since this only catches errors in `bootc switch` (debatable if the write-to-terminal can fail?) I'm ACK.

PR #1138, opened 2025-01-13
Remove dnsmasq from iot package sets

djach7:
Removes the `dnsmasq` package from iot package sets to bring package list in line with current iot package list here: https://pagure.io/fedora-iot/ostree/blob/main/f/fedora-iot-base.yaml

Also see the original PR that removed `dnsmasq` here: https://pagure.io/fedora-iot/ostree/c/176b0add790ece0b16b41337569b9170b0514117?branch=main

### DIFF
--- a/pkg/distro/fedora/package_sets.go
+++ b/pkg/distro/fedora/package_sets.go
@@ -97,7 +97,6 @@ func iotCommitPackageSet(t *imageType) rpmmd.PackageSet {
 			"cryptsetup",
 			"curl",
 			"dbus-parsec",
-			"dnsmasq",
 			"dosfstools",
 			"dracut-config-generic",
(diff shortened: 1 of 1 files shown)


supakeen (changes requested):
I think should be conditionally added back when the version is <= 40 as it is still supported for a few more months and the package was only dropped for fedora 41 onwards right?

See: https://pagure.io/fedora-iot/ostree/blob/f40/f/fedora-iot-base.yaml#_98 (note the branch).

djach7:
@supakeen added some conditionals to keep `dnsmasq` present for versions before 41, thanks for the reminder because I totally blanked on it

achilleas-k (approved):
Thanks!

PR #1095, opened 2024-12-09
modularity: depsolver and manifests (COMPOSER-2366)

supakeen:
TBD, draft, do not look. Depends on https://github.com/osbuild/osbuild/pull/1933.

PR #1139, opened 2025-01-14
manifest: change the serializeStart() to take a new `Inputs` arg

mvo5:
The signature of serializeStart() is a bit unwieldy and when adding librepo based sources this prompted for this proposed cleanup. The idea is to just pass a single "Inputs" struct to the serialization so that if we add field we don't need to change the signature all over the place. It also makes the calls easier to read (IMHO).

This contains only mechanical changes so the review shouldbe straightforward.

Unfortunately there will be conflicts with https://github.com/osbuild/images/pull/1095/files but after this is merged the other PR will be a lot smaller

We should merge this relatively soon as pretty much any PR will conflict with this one.

[edit: happy to discuss naming, I think `manifest.Inputs` makes sense in the context and `serializeStart(inputs)` too but open for ideas, some other possibilities:
- manifest.Sources
- manifest.SerializationInputs, SerialkzationSources
- manifest.Resources
- manifest.InputData
- ...
- ?

### DIFF
--- a/pkg/manifest/anaconda_installer.go
+++ b/pkg/manifest/anaconda_installer.go
@@ -6,12 +6,10 @@ import (
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/arch"
-	"github.com/osbuild/images/pkg/container"
 	"github.com/osbuild/images/pkg/customizations/fsnode"
 	"github.com/osbuild/images/pkg/customizations/kickstart"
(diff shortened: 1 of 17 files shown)


thozza (approved):
Makes sense, thanks!

supakeen (approved):
Quite like this; could argue the name but seems unnecessary.

ondrejbudai (approved):
Love it!

achilleas-k (approved):

PR #1140, opened 2025-01-14
rpmmd,dnfjson: add `Path`, `RepoID` to PackageSpec for librepo

mvo5:
This commit adds the two new `Path` and `RepoID` fields to the rpmmd.PackageSpec so that librepo can use them as it needs to correlate a package to a repo to generate the right mirror id.

Ideally we would also make sure the inputs to manifest.Serialize() only accepts depsolved packageSets/repoConfigs via static type checking. There is a TODO for this now but it's not done in this commit as it requires some more thinking/discussion. But because of the strict requirement for matching packageSpec.RepoID and repoConfig.Id we should try to make this type safe. The repoConfig.Id is really only guranteed after the depsolve() run.

[split out from https://github.com/osbuild/images/pull/1132 with improved docs/tests]


thozza (approved):
LGTM

achilleas-k (approved):

PR #1141, opened 2025-01-14
osbuild: add org.osbuild.librepo source wrapper

mvo5:
This commit adds a wrapper for the org.osbuild.librepo source
to download RPMs.

Note that this does not allow customizing the librepo options
MaxParallels, FastestMirror yet. We need to brainstorm this
a bit. The reason is that GenSources() does not currently allow
much control so our (library) users cannot customize it easily.

Split out/cleaned up version of https://github.com/osbuild/images/pull/1132

### DIFF
--- a/pkg/osbuild/librepo_source.go
+++ b/pkg/osbuild/librepo_source.go
@@ -1,0 +1,126 @@
+package osbuild
+
+import (
+	"fmt"
+
+	"github.com/osbuild/images/pkg/rpmmd"
+)
+
+// LibrepoSource wraps the org.osbuild.librepo osbuild source
+type LibrepoSource struct {
+	Items   map[string]*LibrepoSourceItem `json:"items"`
+	Options *LibrepoSourceOptions         `json:"options"`
+}
+
+func NewLibrepoSource() *LibrepoSource {
+	return &LibrepoSource{
+		Items: make(map[string]*LibrepoSourceItem),
+		Options: &LibrepoSourceOptions{
+			Mirrors: make(map[string]*LibrepoSourceMirror),
+		},
+	}
+}
+
+// AddPackage adds the given *depsolved* pkg to the downloading. It
+// needs the *depsovled* repoConfig so that the repoID of the two can
+// be matched up
+func (source *LibrepoSource) AddPackage(pkg rpmmd.PackageSpec, repos []rpmmd.RepoConfig) error {
+	pkgRepo, err := findRepoById(repos, pkg.RepoID)
+	if err != nil {
+		return fmt.Errorf("cannot find repo-id for pkg %v: %v", pkg.Name, err)
+	}
+	if _, ok := source.Options.Mirrors[pkgRepo.Id]; !ok {
+		mirror, err := mirrorFromRepo(pkgRepo)
+		if err != nil {
+			return err
+		}
+		source.Options.Mirrors[pkgRepo.Id] = mirror
+	}
+	mirror := source.Options.Mirrors[pkgRepo.Id]
+	if pkg.IgnoreSSL {
+		mirror.Insecure = true
+	}
+	// this should never happen but we should still check to avoid
+	// potential security issues
+	if mirror.Insecure && !pkg.IgnoreSSL {
+		return fmt.Errorf("inconsistent SSL configuration: package %v requires SSL but mirror %v is configured to ignore SSL", pkg.Name, mirror.URL)
+	}
+	if pkg.Secrets == "org.osbuild.rhsm" {
+		mirror.Secrets = &URLSecrets{
+			Name: "org.osbuild.rhsm",
+		}
+	} else if pkg.Secrets == "org.osbuild.mtls" {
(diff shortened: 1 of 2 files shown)


pkg/osbuild/librepo_source.go
```diff
@@ -0,0 +1,120 @@
...
+	Type string `json:"type"`
+
+	Insecure bool        `json:"insecure,omitempty"`
+	Secrets  *URLSecrets `json:"secrets,omitempty"`
+
+	// XXX: should we expose those? if so we need a way to set them,
+	// current this is done in manifest.GenSources which cannot take
+	// options.
+	// MaxParallels  *int `json:"max-parallels,omitempty"`
+	// FastestMirror bool `json:"fastest-mirror,omitempty"`
```
achilleas-k, line 119: Expose them anyway.  I think it's better if the bindings/wrappers in this package reflect the osbuild stages even when we don't use specific options.  It's always possible to import the `osbuild` package and create stages with their options manually, without using a helper function or even the `NewXStage()` initialiser.

I'd also make `librepoSourceMirror` an exported type.

pkg/osbuild/librepo_source.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+// AddPackage adds the given *depsolved* pkg to the downloading. It
+// needs the *depsovled* repoConfig so that the repoID of the two can
+// be matched up
+func (source *LibrepoSource) AddPackage(pkg rpmmd.PackageSpec, repos []rpmmd.RepoConfig) error {
+	pkgRepo := findRepoById(repos, pkg.RepoID)
+	if pkgRepo == nil {
+		return fmt.Errorf("cannot find repo-id %v for %v in %+v", pkg.RepoID, pkg.Name, repos)
```
achilleas-k, line 30: The `repos` array can contain a lot of text because of gpg keys and long URLs.  Maybe we should make this error message easier to read by printing only the identifiable information from the repos array?  Maybe just the ID and mirror URLs.

Perhaps not worth the effort though.

pkg/osbuild/librepo_source.go
```diff
@@ -0,0 +1,120 @@
...
+	if _, ok := source.Options.Mirrors[pkgRepo.Id]; !ok {
+		mirror, err := mirrorFromRepo(pkgRepo)
+		if err != nil {
+			return err
+		}
+		source.Options.Mirrors[pkgRepo.Id] = mirror
+	}
+	mirror := source.Options.Mirrors[pkgRepo.Id]
+	// XXX: should we error here if one package requests IgnoreSSL
+	// and one does not for the same mirror?
```
achilleas-k, line 41: This shouldn't be possible if the response from `osbuild-depsolve-dnf` hasn't been modified, right?  The ignore ssl option is per-repo so it shouldn't be inconsistent unless something was modified.  If that's the case, maybe an error is a good idea.

pkg/osbuild/librepo_source.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"fmt"
+
+	"github.com/osbuild/images/pkg/rpmmd"
+)
+
+// LibrepoSource wraps the org.osbuild.librepo osbuild source
+type LibrepoSource struct {
+	Items   map[string]*librepoSourceItem `json:"items"`
+	Options *librepoSourceOptions         `json:"options"`
```
achilleas-k, line 12: I'd prefer we make the types exported/public so that one can, if they want, create a `LibrepoSource{}` without functions or helpers.  It's not the preferred way to do it, but I think `pkg/osbuild/` should support that workflow generally.

achilleas-k (changes requested):

pkg/osbuild/librepo_source.go
mvo5: I can change them back to exported (there were in my first version), I changed them because it easy to non-working osbuild.LibrepoSources when doing this by hand, the repos and package IDs must match etc so this is meant as a hint "use-the-helper". But fine to change them as the rest of the package exports all types (I guess consistency here is more important than my whims)

pkg/osbuild/librepo_source.go
mvo5: Yeah, this should never happen in practise but it would be a potential security issue so I agree, lets check it.

pkg/osbuild/librepo_source.go
mvo5: I changed it back to export everything. I would love to reduce our API surface though longer-term but not a discussion we need for this PR :)

pkg/osbuild/librepo_source.go
mvo5: Done

pkg/osbuild/librepo_source.go
mvo5: Done

ondrejbudai (approved):
Thanks! ❤️

achilleas-k (approved):

PR #1150, opened 2025-01-17
[RFC] move "progress" from bib to images for ibcli

mvo5:
This is a bit of a RFC - it would be nice to have ibcli use progress as bib does. This is pretty trivial and done in https://github.com/osbuild/image-builder-cli/pull/21/files - *but* it needs the "progress" library that is currently living in bib.

This PR is the most simple approach - just copy it over with its warts, including that there is a `progress.RunOSBuild()` that is slightly different from `osbuild.RunOSBuild`. I guss one coulld justify this by saying one is about progress and the other is not. Then we can have ibcli use it and get it with more and nicer features into fedora.

Alternatively we can do the cleanup of this first, the most important part to cleanup is the unification of the to way to run osbuild.


achilleas-k:
I'm not sure about this.  I think there should definitely be some changes in images to accommodate progress reporting, but it feels strange to have everything here.  I'm thinking it'd be cleaner if the osbuild bindings in images implemented a convenient way to start osbuild with progress output and stream the info to the caller, but then we'd be duplicating the output handling and actual progress printing in BIB and IB-cli.  The cleanest would be to have an importable progress reporting package separately that BIB and IB-cli could import, but I don't think I want yet another project repository just for this.

In the end I think this makes sense.  What does everyone else think?

mvo5:
> I'm not sure about this. I think there should definitely be some changes in images to accommodate progress reporting, but it feels strange to have everything here. I'm thinking it'd be cleaner if the osbuild bindings in images implemented a convenient way to start osbuild with progress output and stream the info to the caller, but then we'd be duplicating the output handling and actual progress printing in BIB and IB-cli. The cleanest would be to have an importable progress reporting package separately that BIB and IB-cli could import, but I don't think I want yet another project repository just for this.
[..]

Thanks for the feedback. I guess one (slightly odd) alternative would be to make the package "progress" public in image-builder-cli and then import it from bib. With the long term goal to unify bib and ibcli into the same codebase it would eventually converge. However the downside is that I see that eventually we want this progress parsing in the osbuild-composer osbuild-executor as well and then it become strange again to import for from image-builder-cli - that is mostly why I put it in "images" but I agree it's not a perfect fit, but I also agree that yet another project repo is probably also a bit overkill.

achilleas-k:
> yet another project repo is probably also a bit overkill

Yeah... let's put it here then.  We can think about splitting it in the future if it it feels out of place still.

supakeen:
I'd like to duplicate the code between `bootc-image-builder` and `image-builder` for now; it doesn't need to be moved into `images`. The long term plan is to merge `bootc-image-builder` into `image-builder`; at which point it would not be duplicated anymore *and* lives in the right spot again.

mvo5:
> > yet another project repo is probably also a bit overkill
> 
> Yeah... let's put it here then. We can think about splitting it in the future if it it feels out of place still.

I understand your concerns and don't want to push things here that you feel uncomfortable about. I opened https://github.com/osbuild/bootc-image-builder/pull/804 which means we can postpone this discussion for a while (until we want to use the progress module inside osbuild-composer, see the commit message there).

achilleas-k:
> Alternative: I'd be OK to duplicate the code between `bootc-image-builder` and `image-builder` for **now**; it doesn't need to be moved into `images`. The long term plan is to merge `bootc-image-builder` into `image-builder`; at which point it would not be duplicated anymore _and_ lives in the right spot again.
> 
> Note; this is only about the current situation. If as @mvo5 says we start using this code in other packages again then at that point we can split it out/have this discussion with all the places identified.

FTR, I'm fine with this too.
(I thought I had already posted this message yesterday and came back to find that I hadn't)

mvo5:
Closing for now in favor of https://github.com/osbuild/bootc-image-builder/pull/804

PR #1154, opened 2025-01-20
Fix user and group creation order in ostree deployments

achilleas-k:
**manifest: create groups before users in ostree deployments**

In regular, package-based builds and in base ostree commits, we always
create groups before creating users, i.e., we always run
org.osbuild.groups before org.osbuild.users.  This is the correct way to
do it because a user in a blueprint can have a gid that is defined in
the group section of the blueprint.  In ostree deployments (e.g.
edge-qcow2/iot-qcow2 images) this scenario would make the build fail,
because we were creating users before groups.

Flip the order to fix the bug.

---

**test: add a group to the iot-ostree-pull-user config**

Add a group to the iot-ostree-pull-user config with name osbuild and gid
1002.  The group is assigned to the user in the same config.  This
covers the scenario that causes the bug that was fixed in the previous
commit.

---


supakeen (approved):
Nice find; thank you.

mvo5 (approved):
Thank you!

PR #1155, opened 2025-01-20
Test/scripts: add scripts for convenient download of image build CI cache (HMS-5356)

thozza:
Add scripts for downloading image build CI cache to test osbuild behavior when rebuilding manifests in the osbuild repository (so-called manifest tests).

As I worked on using these scripts in the osbuild repo for manifest tests, I realized that there is a need to be able to split the test cases (corresponding to specific manifest) into batches to allow parallelization. For this reason, it is desirable to determine the current CI cache to download (which includes generating relevant manifests to determine their checksum), but don't download any image artifacts. The reason is that these are huge, and if traffic is not an issue, it still takes time to download files, which would not be used at all. Image artifacts are downloaded once test cases for the given batch are determined.

Instead of adding an overly complex script to handle both scenarios described above, I decided to split them and add two scripts (naming ideas are welcomed :innocent:):

- The `dl-image-build-cache` script will determine the image build cache S3 paths based on the current branch of the images repository and download the image build cache files. Determining the S3 paths involves generating manifests for the relevant distro/arch/image_type/config combinations to determine manifest checksums. The script, by default, downloads only image build metadata (`*.json` files, but mainly `info.json`) and the manifest. It can optionally download also image artifacts. While the reworked manifest tests in osbuild do not use this functionality, I kept it there in the hope that it may be helpful for manual downloads.
- The `dl-one-image-build-cache` script reads a single image build metadata (`info.json`) and downloads all corresponding files.

So the idea of the manifest tests (as implemented in https://github.com/osbuild/osbuild/pull/1983) is that it will:

1. Run `dl-image-build-cache` with specific arguments without downloading all image artifacts.
2. Filter the test cases to a subset based on the test chunk configuration.
3. Run `dl-one-image-build-cache` to download image artifacts for the selected test cases.
4. Run manifest tests for the selected test cases.

/jira-epic COMPOSER-2318

JIRA: [HMS-5356](https://issues.redhat.com/browse/HMS-5356)

### DIFF
--- a/test/scripts/dl-image-build-cache
+++ b/test/scripts/dl-image-build-cache
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+
+"""
+Download the image build CI cache files for the current branch from AWS S3.
+
+This script downloads the image build cache files for the current branch from AWS S3.
+The script generates the current manifests to determine the image build cache files to download.
+"""
+
+import argparse
+import os
+import sys
+import tempfile
+from fnmatch import fnmatch
+from typing import Dict, List, Optional
+
+import imgtestlib as testlib
+
+
+def get_argparser():
+
+    class ExtendAction(argparse.Action):
+        """
+        Custom argparse action to append multiple values to a list option
+        to prevent overwriting the list with each new value.
+
+        This may be removed when Python 3.8 is the minimum supported version (in osbuild).
+        """
+        def __call__(self, parser, namespace, values, option_string=None):
+            items = getattr(namespace, self.dest) or []
+            items.extend(values)
+            setattr(namespace, self.dest, items)
+
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.register('action', 'extend', ExtendAction)
+
+    parser.add_argument(
+        "--distro", action="extend", metavar="DISTRO", nargs="+",
+        help="Distro for which the image build cache is downloaded. Can be specified multiple times. " +
+             "If not provided, all distros are downloaded.",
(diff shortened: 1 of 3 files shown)


thozza:
I'm opening this PR for review, because the cache downloading part for osbuild manifest-tests is IMO sufficient, e.g. see https://gitlab.com/redhat/services/products/image-builder/ci/osbuild/-/jobs/8911649202

schuellerf:
two minor questions:
- Is there a special reason why this isn't one script with a command line argument to select behavior?
(just seems much duplicated code)
- why don't they have a `.py` ending (I know it's technically not needed :-| )

thozza:
> two minor questions:

First of all, thanks for having a look at the code.
 
> * Is there a special reason why this isn't one script with a command line argument to select behavior?
>   (just seems much duplicated code)

Each script does fundamentally a different thing before downloading anything from the S3 cache, as described in the PR description.

Personally, I disagree with the statement that there is a lot of duplicate code. The shared parts are in the `imgtestlib.py`. Merging the two scripts and adding a command to select the behavior, where each takes a different set of arguments, would not be for free, so the lines saved are IMHO questionable. In addition, it would complicate the script. I prefer simpler and single-purpose scripts to ones that do everything.

> * why don't they have a `.py` ending (I know it's technically not needed :-| )

The reason is consistency with other Python scripts in the directory.

achilleas-k (approved):
Good stuff.  TY.  LGTM!

PR #1156, opened 2025-01-20
Validate config map and fix bad Fedora IoT test config

achilleas-k:
1. Follow-up from #1154 with the same change but for the edge-ami image type.
2. Add a new test script that validates the build configurations in the config map¹.
3. Change the `iot-ami` image type in the config map to `iot-qcow2-image`.  `iot-ami` is not a valid image type name.

¹ Note that the validation script does not catch all issues with the config map.  For
example, the following configuration is considered valid:
```json
{
  "./configs/some-config.json": {
    "distros": [
      "rhel-10*",
    ],
    "image-types": [
      "ami",
      "invalid-type-name"
    ]
  }
}
```

because it still produces (at least) one valid configuration:

    rhel-10.0, ami, any architecture

This is (somewhat) necessary because we also want the following to be
valid:
```json
{
  "./configs/some-config.json": {
    "distros": [
      "rhel-9*",
      "fedora*"
    ],
    "image-types": [
      "iot-commit",
      "edge-commit"
    ]
  }
}
```

where iot-commit is only valid for fedora* and edge-commit is only valid
for rhel-9*.

### DIFF
--- a/.github/workflows/test-osbuild-composer-integration.yml
+++ b/.github/workflows/test-osbuild-composer-integration.yml
@@ -53,6 +53,10 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: registry.fedoraproject.org/fedora:latest
+    env:
+      # workaround for expired cert at source of indirect dependency
+      # (go.opencensus.io/trace)
+      GOPROXY: "https://proxy.golang.org|direct"
     outputs:
       # Define job outputs
       # (see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs#example-defining-outputs-for-a-job)
--- a/.github/workflows/gitlab-helper.yml
+++ b/.github/workflows/gitlab-helper.yml
@@ -16,6 +16,10 @@ jobs:
   prepare:
     name: "🔍 Check source preparation and test configs"
     runs-on: ubuntu-latest
+    env:
+      # workaround for expired cert at source of indirect dependency
+      # (go.opencensus.io/trace)
+      GOPROXY: "https://proxy.golang.org|direct"
     steps:
 
       - name: Set up Go 1.21
--- a/.github/workflows/gitlab-helper.yml
+++ b/.github/workflows/gitlab-helper.yml
@@ -14,7 +14,7 @@ on:  # yamllint disable-line rule:truthy
 
 jobs:
   prepare:
-    name: "🔍 Check source preparation"
+    name: "🔍 Check source preparation and test configs"
     runs-on: ubuntu-latest
     steps:
 
@@ -45,6 +45,11 @@ jobs:
           else
             exit "0"
           fi
+
+      - name: Check that the config-map is valid
+        run: |
+          ./test/scripts/validate-config-map
+
 
   gitlab-ci-helper:
     name: "Gitlab CI trigger helper"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,7 +48,9 @@ jobs:
         run: git config --global --add safe.directory "$(pwd)"
 
       - name: Run unit tests
-        run: go test -race ./...
+        # skip the TestResolverLocalManifest test. It is tested separately
+        # (see below: requires root)
+        run: go test -race ./... -test.skip TestResolverLocalManifest
 
       - name: Run depsolver tests with force-dnf to make sure it's not skipped
         run: go test -race ./pkg/dnfjson/... -force-dnf
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -160,21 +160,21 @@ jobs:
           SHELLCHECK_OPTS: -e SC1091 -e SC2002 -e SC2317
 
   python-test:
-    name: "🐍 pytest (imgtestlib)"
+    name: "🐍 pytest (imgtestlib and test scripts)"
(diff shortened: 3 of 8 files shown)


achilleas-k:
I changed some test names in the github actions which means I had to change the required test list for PRs.  We'll need to rebase all other open PRs when this is merged.

test/scripts/imgtestlib.py
```diff
@@ -63,7 +63,7 @@
 """
 
 
-def runcmd(cmd, stdin=None, extra_env=None, capture_output=True):
+def runcmd(cmd, stdin=None, extra_env=None, capture_output=True, *, cmd_root="."):
```
thozza, line 66: Nitpick: `subprocess.run()`'s `cwd` defaults to `None`. I'd suggest defaulting to `None` here for `cmd_root`...

In addition, is there any particular reason not to allow specifying `cmd_root` as a positional argument? :thinking:

test/scripts/imgtestlib.py
```diff
@@ -213,7 +214,7 @@ def check_config_names():
 
 
 def gen_manifests(outputdir, config_map=None, distros=None, arches=None, images=None,
-                  commits=False, skip_no_config=False):
+                  commits=False, skip_no_config=False, *, cmd_root="."):
```
thozza, line 217: Does setting `cmd_root="."` make any real difference?

thozza (approved):
I've added a few nitpicks and a question, but in general, this looks good to me :+1: 
Extra points for adding a test for the new script :sweat_smile:

test/scripts/imgtestlib.py
achilleas-k: > Nitpick: `subprocess.run()`'s `cwd` defaults to `None`. I'd suggest defaulting to `None` here for `cmd_root`...
> 

You're right.  Good call.

> In addition, is there any particular reason not to allow specifying `cmd_root` as a positional argument? 🤔

I wanted to make it forced named for `gen_manifests()` and `list_images()` and I ran with it for `runcmd()`.  I'll make it a regular arg here too, but I'll keep it named for the other two because it doesn't really _fit_ with the rest of the positional arguments, so I'd like it to stand out.

achilleas-k:
Dropped the `cmd_root`.  If we need it, we can re-add it.

test/scripts/imgtestlib.py
thozza: > > In addition, is there any particular reason not to allow specifying `cmd_root` as a positional argument? 🤔
> 
> I wanted to make it forced named for `gen_manifests()` and `list_images()` and I ran with it for `runcmd()`. I'll make it a regular arg here too if I don't drop it entirely, but I'll keep it named for the other two because it doesn't really _fit_ with the rest of the positional arguments, so I wanted it to stand out.

That's fair, thanks for the explanation. I was mostly curious...

thozza (dismissed):
LGTM

achilleas-k:
Dismissed again.  Had to set the GOPROXY on all the github jobs that need it because of an expired cert at the source of one of the dependencies.

achilleas-k:
GitLab didn't seem to need it.  Do we set it on our gitlab runners by default?

.github/workflows/test-osbuild-composer-integration.yml
```diff
@@ -53,6 +53,10 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: registry.fedoraproject.org/fedora:latest
+    env:
+      # workaround for expired cert at source of indirect dependency
+      # (go.opencensus.io/trace)
+      GOPROXY: "https://proxy.golang.org|direct"
```
schuellerf, line 59: I'd document as making the downloading more robust - as opencensus was only the trigger but this should stay…
Also I'd move it to the Makefile to implement this only once and also benefit from it locally (when running on any fedora spin)
I'll suggest a  PR for this

schuellerf (changes requested):

.github/workflows/test-osbuild-composer-integration.yml
schuellerf: I can also do the PR after this is merged 🤷

achilleas-k: The Makefile setup is a bit annoying.  The whole container build mean to replicate the GH setup obfuscates more than it helps IMO.  I much prefer the old setup where things are visible in the action and there are fewer layers between the action config, the logs, and what's actually being run.

Also, if the Makefile is meant to be for people that clone the repo and want to run the tests without understanding all the details, I think it should also not make many assumptions about those people's setup.  Overriding their (potentially system-wide, sometimes necessary) GOPROXY seems a bit invasive and can cause more problems than it solves.

schuellerf: I think setting GOPROXY with `?=` leaving people to just override if needed could be ok.
Also I'm not sure if duplicating `GOPROXY` definition 7 times is better than one indirection to make 🤔

thozza (approved):
FWIW, LGTM

.github/workflows/test-osbuild-composer-integration.yml
schuellerf: For build stability, I'd definitely go for keeping it. It says in the documentation the `|` pipe can be used to also cover timeouts…
but we can do this in a followup PR for sure…

schuellerf (approved):

PR #1157, opened 2025-01-21
many: add "uri" in file customizations to support host  resources

mvo5:
[edit:
- this got switched from "ref"->"uri" now as suggested by the reviews
- this will need adding to the new osbuild/blueprint repo and the converter there
- it will also need support in the new BPv2 from @lzap (but iirc it was speced there already)
]

We have a feature request that asks to support including local
files in blueprints. This also came up recently in a dicussion
about the blueprints v2 format (that is where the name "ref" comes
from). This is tricky of course for the service and would probably
not be supported there initially (maybe later via a extra upload
option). But for `image-builder-cli` it does make sense and is
a useful feature that is support by other image-building tools
too.

This initial commit show how we can use blueprints and a new "ref"
field to give access to host files [0]. It works internally by
generating an osbuild manifest with `org.osbuild.files` inputs
that get populated via a `curl` stage [1] that pulls the files
from the host into the osbuild store. Because they are addressed
by hash (just like inline files) the manifest is still reproducible
and if anything changes in the file(s) from the host the manifest
will fail to build. Note that all of this is an implementaion detail
and we could change "osbuild" to do something else, the blueprint
API would not change.

Note also that this means that an attacker could potentially use
this in combination with custom rpm packages and maintainer scripts
to gain information about files outside the buildroot. As our isolation
is not a security feature this shouldn't be an issue (and arguably
running maintainer scripts in the rpm stage as root is already
opening a lot of possibilities to break out of the buildroot).
This is only an issue when the inputs are untrusted, e.g. in the
service in the future.

Blueprints generated like this are no longer portable as they
require the local files and when those change the manifest breaks.

They will become portable when we implement "offline" bundles,
i.e. a way to generate a manifest and their dependencies in
a single file that can then be rebuild by image-builder/osbuild.
This implementation works nicely with that idea because in the
most simple case a "offline-bundle" is just a manifest+store
directory. Because everything is looked up in the store first
by hash this will "just work". Offline bundles will also
fix the other issue with manifests that if sources like packages
go away (because they are outdated) the manifest also stops
building. Note that we can even have "half-offline" bundles
where we would put only the "file" refs into the .store we
generate but keep the package urls out.

The downside of using the store of course is that it means
duplication of data and may seem wasteful for big files.
However without that it is very hard to keep the gurantee
that a manifest is reproducible because if we access the
file directly we have a race-condition that the hash may
change anytime. Any trick like bind-mounting or hardlinking
will have the same issue. So unless we drop the reproducibility
for local file refs (which if of course an option) I see
no other way than to have a copy in the store.

[0] Initial commit because only regular files are reported
for now. The mechanism is generic enough so that we could
support dirs (via simple recursion) and http{,s} URLs as
well (actually anything that curl supports but we should
currate it carefully).
[1] IIRC coreos uses this trick as well in their manifests

Jira: RHEL-21334


supakeen:
I absolutely love the idea and flexibility this gives; it will cover quite a few usecases that I have personally, especially when we can include entire trees. It will also severely simplify blueprints when you want to, for example, write 15 quadlet files.

However. I do feel that we should only introduce this together with manifest bundles, I prefer being able to rebuild images on several (different) hosts.

mvo5:
> I absolutely love the idea and flexibility this gives; it will cover quite a few usecases that I have personally, especially when we can include entire trees. It will also severely simplify blueprints when you want to, for example, write 15 quadlet files.
> 
> However. I do feel that we should only introduce this together with manifest bundles, I prefer being able to rebuild images on several (different) hosts.

Offline/manifest-bundles seem pretty straightforward, we can do them first, I have no strong opinion here, I have a PoC for image-builder-cli in https://github.com/osbuild/image-builder-cli/compare/main...mvo5:offline-bundles?expand=1#diff-dd961e42853d4345e2cf6a29f188f80386ed6b583056d7078a38c0c099c97c26R63 - the "backend" code should land in images but the concept is pretty straightforward (the linked test works fwiw).

supakeen:
> > I absolutely love the idea and flexibility this gives; it will cover quite a few usecases that I have personally, especially when we can include entire trees. It will also severely simplify blueprints when you want to, for example, write 15 quadlet files.
> > However. I do feel that we should only introduce this together with manifest bundles, I prefer being able to rebuild images on several (different) hosts.
> 
> Offline/manifest-bundles seem pretty straightforward, we can do them first, I have no strong opinion here, I have a PoC for image-builder-cli in https://github.com/osbuild/image-builder-cli/compare/main...mvo5:offline-bundles?expand=1#diff-dd961e42853d4345e2cf6a29f188f80386ed6b583056d7078a38c0c099c97c26R63 - the "backend" code should land in images but the concept is pretty straightforward (the linked test works fwiw).

I'll read through that, awesome, thanks.

pkg/blueprint/fsnode_customizations.go
```diff
@@ -190,6 +190,15 @@ type FileCustomization struct {
...
 	Data string `json:"data,omitempty" toml:"data,omitempty"`
+
+	// Ref references the given files from the host, this makes
+	// the manifest alone no-longer portable (but future offline
+	// manifest bundles will fix that). It will still be
+	// reproducible as the manifest will include all the hashes of
+	// the content so any change will make the build fail.
+	//
+	// XXX: In this PoC only single files are supported.
+	Ref string `json:"ref" toml:"ref,omitempty"`
```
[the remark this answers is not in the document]
mvo5: Good catch, thank you! I will fix that (and before this becomes non-draft add a test as well, pinky-swear!)

ondrejbudai:
I very much support this, we indeed have quite a lot of users that want to have the files stored outside of blueprints.

I appreciate the simplicity of this solution as it just uses the pre-existing source in osbuild. @mvo5 mentioned to me that it isnt great that we firstly need to copy the file inside the osbuild's store which effectively creates a double-copy. I think that's a valid concern, but I don't consider this as a blocker. We can revisit this later when we know how this feature is used by users. The nice thing is that we can just change the implementation without changing the user interface.

thozza (commented):
I think that this approach is very reasonable.

Even without offline bundles, it would be possible to optionally embed the file in the manifest as inline source ,if needed. But letting it be `curl` source makes sense, especially once the `ref` would be an actual remote URL and we would have a resolver that would generate the ref.

I think that extracting the resolver logic (at the moment only for local files, but without making the assumption that they must be local) into a common resolver and passing the resolved spec as an input to manifest generation would make sense.

Lastly, the `ref` name is a but overloaded, in my opinion, and initially, I though that it is just the SHA256 of the file. While in reality it is a path or a URL (we could standardize on URLs even for local files).

supakeen:
Before looking at the code, I like the `url` name more than `ref` it hints at what the value should be :)

test/configs/all-customizations.json
```diff
@@ -155,6 +155,10 @@
           "path": "/etc/empty_file.txt",
           "user": 0,
           "group": 0
+        },
+        {
+          "path": "/etc/host-os-release",
+	  "ref": "file:///etc/os-release"
```
achilleas-k, line 161: Inconsistent indentation.

test/configs/file-customizations.json
```diff
@@ -33,6 +33,10 @@
           "path": "/etc/empty_file.txt",
           "user": 0,
           "group": 0
+        },
+        {
+          "path": "/etc/host-os-release",
+	  "ref": "file:///etc/os-release"
```
achilleas-k, line 39: Inconsistent indentation.

achilleas-k (commented):
This is great.  One thing I'd like to do is bikeshed / brainstorm a bit on the term `ref`.  I'm not sure it's the best key for this, but I'm also not sure I have a better idea.  `source_path`?  `content_path`?  I like that `ref` is short though.

thozza:
> This is great. One thing I'd like to do is bikeshed / brainstorm a bit on the term `ref`. I'm not sure it's the best key for this, but I'm also not sure I have a better idea. `source_path`? `content_path`? I like that `ref` is short though.

What about 'uri' or 'url' if we end up using `file:///...`? This would eventually allow us to use `https://..`, etc.

pkg/customizations/fsnode/file.go
```diff
@@ -20,9 +27,53 @@ func (f *File) Data() []byte {
...
+	}
+	if mode == nil {
+		mode = common.ToPtr(st.Mode())
+	}
+	if user == nil {
+		user = int64(st.Sys().(*syscall.Stat_t).Uid)
+	}
+	if group == nil {
+		group = int64(st.Sys().(*syscall.Stat_t).Gid)
+	}
```
thozza, line 71: Nitpick: I'm not sure if it makes sense to take the local host UID and GID, which may not exist in the image. Using the default (`root`) if not specified may IMO make more sense.

thozza (commented):
LGTM in general.

I'm not sure about inheriting UIDs and GIDs of local files and using them for the image content. Inheriting the mode is a different story.

I know that the eventual goal is to use the "offline bundle" for manifest+source distribution, but I can't help myself not think about the low hanging fruit to optionally embed the referenced local file as an inline source, to optionally make the generated manifest rebuildable on its own.

achilleas-k:
> What about 'uri' or 'url' if we end up using `file:///...`? This would eventually allow us to use `https://..`, etc.

yeah, `uri` makes more sense to me.

supakeen:
I think everyone has suggested `uri` (or `url`) so far and brought up no other concerns :)

pkg/customizations/fsnode/file.go
mvo5: Yeah, makes sense, removed.

supakeen (dismissed):
Awesome, thank you. I had read the code way back then and it seems it hasn't changed aside from the name.

thozza (dismissed):

pkg/manifest/os.go
```diff
@@ -1160,3 +1161,25 @@ func (p *OS) addInlineDataAndStages(pipeline *osbuild.Pipeline, files []*fsnode.
 		p.inlineData = append(p.inlineData, string(file.Data()))
 	}
 }
+
+func (p *OS) fileRefs() []string {
```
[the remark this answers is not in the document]
mvo5: This is a bit confusing and maybe we should make it less confusing, I added a comment and also a small integration test to demonstrate that this actually works (I had one for ibcli pending but it really belongs here).

The stages are generated via the `OS.addInlineDataAndStages()` which will do the stage generation for *all* files and then also add the inline data. Maybe its a bit too confusing this way, idk.

test/configs/file-customizations.json
```diff
@@ -33,6 +33,10 @@
           "path": "/etc/empty_file.txt",
           "user": 0,
           "group": 0
+        },
+        {
+          "path": "/etc/file-from-host",
+          "uri": "file:///proc/1/cpu_resctrl_groups"
```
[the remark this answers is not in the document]
achilleas-k: You're right, it's not used.  It probably got folded into all-customizations at some point and then wasn't deleted (just guessing; didn't check).

As for a stable host file, what about `/etc/os-release`?
- In the build tests in gitlab-ci, it wont change unless the runner OS changes and in those cases we trigger rebuilds anyway.
- For the mock manifest generation, we mock this along with all other sources by hashing the `uri` (same thing we do for ostree commits).

pkg/blueprint/fsnode_customizations.go
```diff
@@ -190,6 +190,17 @@ type FileCustomization struct {
 	Mode string `json:"mode,omitempty" toml:"mode,omitempty"`
 	// Data is the file content in plain text
 	Data string `json:"data,omitempty" toml:"data,omitempty"`
+
+	// URI references the given URI, this makes the manifest alone
+	// no-longer portable (but future offline manifest bundles
```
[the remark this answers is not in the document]
mvo5: Yeah, I did the naive version in https://github.com/osbuild/images/pull/1610 as a draft for preview but I agree, its not ideal because of the double downloading. I'm not sure how to avoid it though.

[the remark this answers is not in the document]
mvo5: I guess we could consider using the osbuld objectstore and cache the downloaded file there but that would be a big depature what we are currently doing

test/configs/file-customizations.json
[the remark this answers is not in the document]
mvo5: I added this now, I hope I captured what you have in mind here.

cmd/gen-manifests/main.go
```diff
@@ -229,6 +230,24 @@ func makeManifestJob(
...
+				if err := os.MkdirAll(filepath.Dir(newBpFileUrl), 0755); err != nil {
+					panic(err)
+				}
+				if err := os.WriteFile(newBpFileUrl, []byte(fc.URI), 0600); err != nil {
+					panic(err)
+				}
+				bp.Customizations.Files[i].URI = "file://" + newBpFileUrl
+			}
+		}
+	}
```
achilleas-k, line 250: I guess the plan is to make it a command line option after it's merged?

.github/workflows/validate-checksums.yml
```diff
@@ -56,5 +56,6 @@ jobs:
               echo
               echo "Please refer to the developer docs at docs/developer for more information."
               echo "-----------------------------------------------------------------------------"
+              ./tools/gen-manifests-diff || true
```
achilleas-k, line 59: I worry this might make the log difficult to read, essentially hiding the instructions, if the diff is large enough.  The message above already guides people towards generating the diff and reviewing it.  I get that it might be useful to have the diff in the log, especially if it's small, but if it's not, it becomes annoying.

Let's keep it for now and see how it goes.

test/integration/fscust_test.go
```diff
@@ -0,0 +1,70 @@
...
+	l, err := filepath.Glob(tmpdir + "/*/archive/*.tar.xz")
+	assert.NoError(t, err)
+	require.Equal(t, 1, len(l))
+
+	// ensure we get the expected content from the locally added file
+	// #nosec G204
+	output, err := exec.Command("tar", "xOf", l[0], "./etc/file-from-host").CombinedOutput()
+	assert.NoError(t, err, "tar output: %s", output)
+	assert.Equal(t, canaryString, string(output))
+}
```
achilleas-k, line 70: This is a bit strange to me, compiling and running the build command from inside a go test.  Let's keep it for now, but I'd prefer to have something a bit cleaner.  It's probably not a lot of work to generate a manifest in code and inspect it for the correct stages, like we do in a lot of cases.

I'm also not sure why we can't have a customization like this in one of the build tests.  We could use a known file (like `os-release`) or set up a mechanism where we create a file on the host before building the test image.
A discussion for another time though.

achilleas-k (approved):

.github/workflows/validate-checksums.yml
mvo5: I can remove it again, its mostly here for debugging why it currently gives me different results in github than locally, its quite a mystery

test/integration/fscust_test.go
mvo5: Thanks, I guess we need to sit down and converge on the wider discussion around testing in images long-term. We probably have slightly different views on what "clean" means and how to approach the tests and maybe I should have not written this before we had this discussion.

I would love to hear your thoughts how this would look concretely with the existing framework (I want an end-to-end test, having just a manifest inspection is good but I think its important to test this end-to-end), I'm happy to rework this. In parallel I will try to help kickstart this wider discussion.

pkg/manifest/os.go
```diff
@@ -1150,13 +1151,45 @@ func (p *OS) getInline() []string {
 	return p.inlineData
 }
 
-// addInlineDataAndStages generates stages for creating files and adds them to
+// addStagesForAllFilesAndInlineData generates stages for creating files and adds them to
 // the pipeline. It also adds their data to the inlineData for the pipeline so
 // that the appropriate sources are created.
-func (p *OS) addInlineDataAndStages(pipeline *osbuild.Pipeline, files []*fsnode.File) {
+//
+// Note that this also creates stages for non-inline files that crom via "fileRefs()"
```
thozza, line 1158: Nitpick: typo `crom` -> `come` ?

test/integration/fscust_test.go
thozza: My 2 cents: This end-to-end testing should be done as part of the boot test of test manifests (see https://github.com/osbuild/images/blob/main/test/scripts/base-host-check.sh). We do this now for other types of customizations. This adds an entirely new approach, which is not ideal from my PoV because it creates friction. I'd suggest using our current approach until we have a consensus on the future form of integration testing of images (vanilla and with customizations).

thozza (commented):
The changes LGTM, but I'm not on board with adding a new way of integration testing of images (and their customizations) at this point.

pkg/manifest/os.go
mvo5: Thanks, nice catch. Fixed.

achilleas-k (approved):
Thank you!  LGTM

I agree that the last commit (end-to-end test) isn't ideal, but I'm looking forward to improving the host check script and functionality.

supakeen (approved):

PR #1160, opened 2025-01-22
rhel: On RHEL10 use erofs to compress the root filesystem of the Anaconda ISO

bcl:
This will use lzma compression with 128k blocks and the dedupe and all-fragments erofs options (see the manpage for mkfs.erofs for details).

Resolves: RHEL-75932

### DIFF
--- a/pkg/distro/rhel/images.go
+++ b/pkg/distro/rhel/images.go
@@ -495,9 +495,12 @@ func EdgeInstallerImage(workload workload.Workload,
 	// kickstart though kickstart does support setting them
 	img.Kickstart.Timezone, _ = customizations.GetTimezoneSettings()
 
-	img.RootfsCompression = "xz"
 	if t.Arch().Distro().Releasever() == "10" {
-		img.RootfsType = manifest.SquashfsRootfs
+		// On RHEL10 use erofs and lzma compression for the iso root filesystem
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
Thanks!

achilleas-k:
One of the RHEL 10 ISO build jobs timed out twice (2h timeout).  One of them was right at the end of the build.  I restarted it, but I'll also increase the timeout a bit.

Is this build time increase expected?

supakeen:
Fedora/FCOS has also noticed time and size increases, the maintainer is working on them; see: https://github.com/coreos/fedora-coreos-tracker/issues/1852

bcl:
Yeah, erofs lzma compression is slow, depending on how you run it. On bare metal with my laptop (12 core 2.6GGz) it is reasonable (6 minutes or so) but in a VM it can take 20 minutes or more.

Also, this should not be merged until we're really ready for it in RHEL-10 (see RHEL-2702 issue) so I'll switch it to draft for now.

bcl:
Almost there, just waiting. Things have improved in the meantime and the size is no longer an issue, erofs can now produce compressed images that are slightly smaller than squashfs has been making.

bcl:
Postponed until RHEL11.

PR #1162, opened 2025-01-23
rpmmd: add NEVRA

supakeen:
The depsolver has started returning a pre-formatted NEVRA in https://github.com/osbuild/osbuild/pull/1933. This breaks deserialization, this PR adds the field to the `PackageSpec` type.

I'd prefer to do this here instead of dropping the NEVRA from the depsolve response; the EVR part from the response comes directly from DNF.

### DIFF
--- a/pkg/rpmmd/repository.go
+++ b/pkg/rpmmd/repository.go
@@ -160,6 +160,7 @@ type PackageSpec struct {
 	Version        string `json:"version,omitempty"`
 	Release        string `json:"release,omitempty"`
 	Arch           string `json:"arch,omitempty"`
+	NEVRA          string `json:"nevra,omitempty"`
 	RemoteLocation string `json:"remote_location,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/rpmmd/repository.go
```diff
@@ -160,6 +160,7 @@ type PackageSpec struct {
 	Version        string `json:"version,omitempty"`
 	Release        string `json:"release,omitempty"`
 	Arch           string `json:"arch,omitempty"`
+	NEVRA          string `json:"nevra,omitempty"`
```
mvo5, line 163: I wasn't (much) involved with osbuild https://github.com/osbuild/osbuild/pull/1933 - why was this new string added? It seems it contains the concatenated information that is already provided by "name,epoch,version,arch,release", why do we need it again?

supakeen:
I've spoken to some people and it is preferable not to bump the API version; there will be a new PR which reverts that in osbuild (and removes the nevra field).

PR #1163, opened 2025-01-23
depsolve: add `modules` support

supakeen:
Since the modularity PR https://github.com/osbuild/osbuild/pull/1933 landed in `osbuild` the depsolve result now contains a `modules` key with data if modules were resolved.

Add the appropriate structures to the depsolve result, without propagating them into rpmmd for now.

Also adds a quick test for the internal `depsolveResult` type deserialization to ensure it understands some JSON serialized data that contains a `modules` key.

This PR will serve as the basis for #1095 where the types will be propagated to rpmmd and the stage will be implemented; but this one is important to unblock our CI as everything now fails on the unexpected field.

---

Note that, as opposed to the nevra field on packages which was a big red herring today; this field is actually *expected* to be added. Hence we can't just remove it from the depsolve response. This also means that the DNF JSON version *will* need to be bumped; even though we didn't want to do that (cc: @ondrejbudai). See https://github.com/osbuild/osbuild/pull/1992 for that bump.

---

When this commit is to be depended on by `osbuild-composer` and `image-builder-cli` we will likely need to depend on the new API version as well if I understand correctly (cc @thozza).

### DIFF
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -722,10 +722,12 @@ type transactionArgs struct {
 }
 
 type packageSpecs []PackageSpec
+type moduleSpecs map[string]ModuleSpec
 
 type depsolveResult struct {
 	Packages packageSpecs          `json:"packages"`
 	Repos    map[string]repoConfig `json:"repos"`
+	Modules  moduleSpecs           `json:"modules"`
 
 	// (optional) contains the solver used, e.g. "dnf5"
 	Solver string `json:"solver,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/dnfjson/dnfjson.go
```diff
@@ -722,10 +722,12 @@ type transactionArgs struct {
 }
 
 type packageSpecs []PackageSpec
+type moduleSpecs map[string]ModuleSpec
 
 type depsolveResult struct {
 	Packages packageSpecs          `json:"packages"`
 	Repos    map[string]repoConfig `json:"repos"`
+	Modules  moduleSpecs           `json:"modules"`
```
thozza, line 730: Nitpick: could probably be marked as `omitempty`.

thozza (approved):
LGTM.

I guess that adding the ability to enable specific module stream will be added in a follow-up? The same for bumping the pinned osbuild ref in Schutzfile?

thozza:
> Should the pinned osbuild ref be part of this PR?

I don't think it will make any difference, so not a blocker.

pkg/dnfjson/dnfjson.go
supakeen: I thought `omitempty` only applies when encoding, when decoding `moduleSpecs` is already initialized to its zero value if not present?

thozza: Correct. Although functionally irrelevant, it also expresses that it is optional and may not be present. I'm OK with keeping it as is.

pkg/dnfjson/dnfjson_test.go
```diff
@@ -887,3 +888,17 @@ echo '{"solver": "zypper"}'
 	assert.Equal(t, 0, len(res.Packages))
 	assert.Equal(t, 0, len(res.Repos))
 }
+
+func TestDepsolveResultWithModulesKey(t *testing.T) {
+	// quick test that verifies that `depsolveResult` understands JSON that contains
+	// a `modules` key
+	data := []byte(`{"modules": {}}`)
```
mvo5, line 895: I wonder how much value this test adds as it is.I personally would either omit it or expand it to cover some real world module example (in the sense of tests as documentation). Fine in a followup of course if we should get this in ASAP

supakeen: I'm going to prefer to do that as a followup. I mostly used this test case to have a very minor smoke test that the depsolver no longer breaks when a `modules` key is present.

I'll be updating to an idiom much like what you implemented in `osbuild` in https://github.com/osbuild/osbuild/pull/1993

PR #1164, opened 2025-01-24
osbuild: add consts for more "sources" in sources.go

mvo5:
This is a followup for the suggestion in PR#1142 to not use strings for defining sources but constants. This helps avoiding typos.

Note that this does not change the tests to ensure that we keep the API promises here. I.e. an accidental change of e.g. the `SourceNameSkopeoIndex` const should trigger a test failure.


thozza (approved):
:heart:

PR #1165, opened 2025-01-24
cmd/build: use the new manifestgen package

mvo5:
Note that I could also create a PR for the first two commits that extend
manifestgen and then do the build.go changes separately (or not
at all if that is preferred). The first two commits will help ibcli as well,
once those are in we can add `--ignore-warnings` (or something)
like this to "image-builder-cli" and also a "--override-repository"
option that works differently from "datadir" (just like we do here).
There is a question in the PR about how to best deal with repos
on the API level (which will affect how this will be done in ibcli).

Individual commits below:

---

This commit moves the `cmd/build` command to the new manifestgen
package to the full extend. Some of the output will slightly
change now but this commit tries to keep the errors/messages
as similar as possible.

---

manifestgen: add new `manifestgen.Options.WarningsOutput`

This commit adds a new `manifestgen.Options.WarningsOutput`
io.Writer that can be used to make warnings visible. Without
the WarningsOutput option any warnings will generate an errror.

---

manifestgen: add support for custom repos override

This commit adds support to override the manifest generator with
custom repositories. This is useful for e.g. the `cmd/build`
command that has the ability to override the repostiory information
via a custom file.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -2,7 +2,7 @@
 package main
 
 import (
-	"encoding/json"
+	"bytes"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


cmd/build/main.go
```diff
@@ -204,32 +82,65 @@ func run() error {
...
+		repoConfig, err := rpmmd.LoadRepositoriesFromFile(repositories)
+		if err != nil {
+			return fmt.Errorf("failed to load repositories from %q: %w", repositories, err)
+		}
+		customRepos = repoConfig[archName]
+	} else {
+		// HACK: reporegistry hardcodes adding the "repositories"
+		// dir to the "repoConfigPaths" but the interface of
+		// "images" does not expect this.
+		// XXX: should we fix this in reporegistry?
```
thozza, line 97: Yes, we could remove it and adjust all callers, specifically osbuild-composer.

cmd/build/main.go
```diff
@@ -204,32 +82,65 @@ func run() error {
...
+		repositories = filepath.Join(repositories, "..")
+		reporeg, err = reporegistry.New([]string{repositories})
+		if err != nil {
+			return fmt.Errorf("failed to load repositories from %q: %w", repositories, err)
+		}
 	}
```
thozza, line 103: I'm not sure I like that depending on the `repositories` being a regular file (BTW, it also could be a symlink, no?) the code either directly loads the repo config or constructs the RepoRegistry.

I'd suggest unifying how we call `manifestgen.New()` in both cases. IOW either always load the repo file (the distro is known, so this can also be done in case `repositories` is a directory) or always construct `RepoRegistry` (if `repositories` is a file, then load the parent directory. However, the behavior may be unexpected if the provided filename does not conform to `<distro>.json` naming scheme).

But I can live with this version as well...

pkg/manifestgen/manifestgen.go
```diff
@@ -126,9 +144,12 @@ func (mg *Generator) Generate(bp *blueprint.Blueprint, dist distro.Distro, imgTy
 		return err
 	}
 	if len(warnings) > 0 {
-		// XXX: what can we do here? for things like json output?
-		// what are these warnings?
-		return fmt.Errorf("warnings during manifest creation: %v", strings.Join(warnings, "\n"))
+		warn := strings.Join(warnings, "\n")
+		if mg.warningsOutput != nil {
+			fmt.Fprintf(mg.warningsOutput, warn)
```
thozza, line 149: It should have been probably just `fmt.Fprint()` :thinking:

thozza (commented):
Thanks for the PR. I've added a few nitpicks. Also, linter is sad...

pkg/manifestgen/manifestgen.go
```diff
@@ -48,6 +48,10 @@ type Options struct {
 	// useful for testing
 	CustomSeed *int64
 
+	// CustomRepos overrides the default repository selection.
+	// This is mostly useful for testing
+	CustomRepos []rpmmd.RepoConfig
```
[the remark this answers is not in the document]
mvo5: I renamed the field now, I agree that there is a risk of confusion here.

I personally would not mind to drop this, it is needed if we want to port "build" to fully use manifestgen but as visible in "build" it is not easy. I think @achilleas-k mentioned at some point he like the ability to point to a single repo file for testing.

I guess we can:
a) drop it and leave images as is (using parts of manifestgen only)
b) include it 

I don't strongly mind either way.

pkg/manifestgen/manifestgen.go
```diff
@@ -44,10 +44,19 @@ type Options struct {
 	// content can be read
 	SBOMWriter SBOMWriterFunc
 
+	// WarningsOutput will recieve any warnings that are part
```
achilleas-k, line 47:
```suggestion
	// WarningsOutput will receive any warnings that are part
```

pkg/manifestgen/manifestgen.go
```diff
@@ -44,10 +44,19 @@ type Options struct {
...
+	// of the manifest generation. If that options is empty
```
achilleas-k, line 48:
```suggestion
	// of the manifest generation. If it is unset
```
???

cmd/build/main.go
mvo5: Yeah, excellent point. I will think about this, part of the issue is exactly what you point out, "build" currently supports pointing to arbitrarily named files that most likely not conform to <distro>.json - but I can rework this or (see above) we just keep "build" as is and only do a minimal port to manifestgen.

I'm not sure we have good options here though, moving single-file loading into reporegistry itself leaves the open question to what distro the repo should be attached (and it goes against the spirit of the package). Chaning OverrideReos to a path feels also awkward (and limits how this can be used, one could imagine an override via e.g. a network request). Ideas very welcome, I would love to see this improved.

cmd/build/main.go
```diff
@@ -204,32 +82,67 @@ func run() error {
...
+	}
+	if err := mg.Generate(config.Blueprint, distribution, imgType, arch, &config.Options); err != nil {
+		return fmt.Errorf("[ERROR] manifest generation failed: %w", err)
 	}
 	fmt.Print("DONE\n")
 
 	manifestPath := filepath.Join(buildDir, "manifest.json")
-	if err := save(mf, manifestPath); err != nil {
-		return err
+	if err := os.WriteFile(manifestPath, mf.Bytes(), 0644); err != nil {
```
achilleas-k, line 138: gosec doesn't like this any more it seems.  Needs 600 (or less).  That's fine here but maybe we can consider disabling that rule globally if we need to write more files with more open permissions.

achilleas-k (commented):
Typos.
LGTM otherwise (minus Tomas' comments).

cmd/build/main.go
achilleas-k: I think dropping the forced `repositories` path from `LoadAllRepositories()` in reporegistry is fine, but let's make a big deal out of it and update osbuild-composer immediately so we don't forget and end up scratching our heads next time we update the dependency.

pkg/manifestgen/manifestgen.go
mvo5: I like this suggestion, feels less clunky than my original

cmd/build/main.go
mvo5: Thanks, I added https://github.com/osbuild/images/pull/1179 which does this and also bundles a change that breaks the build so that it is impossible to build unchanged code with this api break (see the comment in there too).

pkg/manifestgen/manifestgen.go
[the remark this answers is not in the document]
thozza: Well, this depends on the answer to my question in the original comment - will this be the way to explicitly define repos to use for build / depsolving / manifest-gen when using  tools built on this functionality (ibcli), when being used as part of a CI/CD pipeline or distro build system? I suspect that "probably yes". So, from that point of view, it makes sense to have a way to define repos without using a repo registry at all.

pkg/manifestgen/manifestgen.go
```diff
@@ -89,6 +94,7 @@ func New(reporegistry *reporegistry.RepoRegistry, opts *Options) (*Generator, er
 		rpmDownloader:     opts.RpmDownloader,
 		sbomWriter:        opts.SBOMWriter,
 		customSeed:        opts.CustomSeed,
+		overrideRepos:     opts.OverrideRepos,
```
thozza, line 104: Nitpick: would it make sense to return an `error` in case `opts.OverrideRepos` and `reporegistry` are both provided?, Because in such case, the `reporegistry` will always be ignored. Similarly, would an error make sense if none is provided?

thozza (commented):
LGTM. My comment can be a follow-up.

thozza (approved):

cmd/build/main.go
thozza: Let's keep it as is for now...

achilleas-k (approved):

PR #1166, opened 2025-01-24
pkg/imagefilter/formatter: add `shell` output

schuellerf:
This should implement the initial idea of a
copy'n'paste ready list of possibilities.

### DIFF
--- a/pkg/imagefilter/formatter.go
+++ b/pkg/imagefilter/formatter.go
@@ -11,9 +11,10 @@ import (
 type OutputFormat string
 
 const (
-	OutputFormatDefault OutputFormat = ""
-	OutputFormatText    OutputFormat = "text"
-	OutputFormatJSON    OutputFormat = "json"
+	OutputFormatDefault   OutputFormat = ""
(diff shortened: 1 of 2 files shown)


schuellerf:
of course I missed the test 😭 sorry

pkg/imagefilter/formatter.go
```diff
@@ -54,6 +57,26 @@ func (*textResultsFormatter) Output(w io.Writer, all []Result) error {
...
+			res.Distro.Name(),
+			res.Arch.Name()); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+
+	return nil
```
supakeen, line 77: I believe you can do `return errors.Join(errs...)` directly here. `errors.Join` discards `nil-values` in errors and returns `nil` when it only has `nil`s in its arguments.

Since `errs` contains a bunch of `nil`s anyhow its `len` is likely always over zero.

mvo5 (approved):
Thank you for adding the test. This looks fine to me now.

PR #1168, opened 2025-01-24
imagefilter: add SupportedOutputFormats

schuellerf:
Convenience function for concise help output

(This is based on #1166 and #1167)

The last commit could be split off, it's not directly related to the implementation of `SupportedOutputFormats` just extends the tests

### DIFF
--- a/pkg/imagefilter/export_test.go
+++ b/pkg/imagefilter/export_test.go
@@ -1,0 +1,3 @@
+package imagefilter
+
+var SupportedFormatters = supportedFormatters
--- a/pkg/imagefilter/formatter.go
+++ b/pkg/imagefilter/formatter.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 
 	"github.com/osbuild/images/pkg/distrosort"
+	// we cannot use "maps" yet, as it needs go1.23
+	"golang.org/x/exp/maps"
 )
 
 // OutputFormat contains the valid output formats for formatting results
@@ -27,20 +29,28 @@ type ResultsFormatter interface {
 	Output(io.Writer, []Result) error
 }
 
-// NewResultFormatter will create a formatter based on the given format.
+var supportedFormatters = map[string]ResultsFormatter{
+	string(OutputFormatDefault):   &textResultsFormatter{},
+	string(OutputFormatText):      &textResultsFormatter{},
+	string(OutputFormatJSON):      &jsonResultsFormatter{},
+	string(OutputFormatTextShell): &shellResultsFormatter{},
+	string(OutputFormatTextShort): &textShortResultsFormatter{},
+}
+
+// SupportedOutputFormats returns a list of supported output formats
+func SupportedOutputFormats() []string {
+	keys := maps.Keys(supportedFormatters)
+	sort.Strings(keys)
+	return keys
+}
+
+// NewResultsFormatter will create a formatter based on the given format.
 func NewResultsFormatter(format OutputFormat) (ResultsFormatter, error) {
-	switch format {
-	case OutputFormatDefault, OutputFormatText:
-		return &textResultsFormatter{}, nil
-	case OutputFormatJSON:
-		return &jsonResultsFormatter{}, nil
-	case OutputFormatTextShell:
-		return &shellResultsFormatter{}, nil
-	case OutputFormatTextShort:
-		return &textShortResultsFormatter{}, nil
-	default:
+	rs, ok := supportedFormatters[string(format)]
+	if !ok {
 		return nil, fmt.Errorf("unsupported formatter %q", format)
 	}
+	return rs, nil
 }
 
 type textResultsFormatter struct{}
--- a/pkg/imagefilter/formatter_test.go
+++ b/pkg/imagefilter/formatter_test.go
@@ -116,3 +116,9 @@ func TestSupportedOutputFormats(t *testing.T) {
 	assert.Contains(t, formatters, "short")
 	assert.True(t, sort.StringsAreSorted(formatters))
 }
+
+func TestResultsFormatterError(t *testing.T) {
+	fmter, err := imagefilter.NewResultsFormatter(imagefilter.OutputFormat("unsupported"))
+	assert.Error(t, err)
+	assert.Nil(t, fmter)
+}
--- a/pkg/imagefilter/formatter_test.go
+++ b/pkg/imagefilter/formatter_test.go
@@ -2,6 +2,7 @@ package imagefilter_test
 
 import (
 	"bytes"
+	"sort"
 	"strings"
 	"testing"
 
@@ -104,4 +105,14 @@ func TestResultsFormatter(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Equal(t, tc.expectsOutput, buf.String(), tc)
 	}
+}
+
+func TestSupportedOutputFormats(t *testing.T) {
+	formatters := imagefilter.SupportedOutputFormats()
+	assert.Len(t, formatters, len(imagefilter.SupportedFormatters))
+	assert.Contains(t, formatters, "text")
+	assert.Contains(t, formatters, "json")
+	assert.Contains(t, formatters, "shell")
+	assert.Contains(t, formatters, "short")
+	assert.True(t, sort.StringsAreSorted(formatters))
 }
--- a/go.mod
+++ b/go.mod
@@ -32,6 +32,7 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/ubccr/kerby v0.0.0-20230802201021-412be7bfaee5
 	github.com/vmware/govmomi v0.48.0
+	golang.org/x/exp v0.0.0-20241009180824-f66d83c29e7c
 	golang.org/x/oauth2 v0.25.0
 	golang.org/x/sys v0.29.0
 	golang.org/x/tools v0.26.0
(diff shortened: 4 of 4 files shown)


pkg/imagefilter/formatter.go
```diff
@@ -5,17 +5,30 @@ import (
...
-	OutputFormatText    OutputFormat = "text"
-	OutputFormatJSON    OutputFormat = "json"
+	OutputFormatDefault   OutputFormat = ""
+	OutputFormatText      OutputFormat = "text"
+	OutputFormatJSON      OutputFormat = "json"
+	OutputFormatTextShell OutputFormat = "shell"
+	OutputFormatTextShort OutputFormat = "short"
 )
 
+func GetSupportedOutputFormats() []string {
```
mvo5, line 23: The convention in go is to not use the `Get` prefix (c.f. https://go.dev/doc/effective_go#Getters) so maybe just `SupportedOuptutFormats()` ?

pkg/imagefilter/formatter.go
```diff
@@ -5,17 +5,30 @@ import (
...
+	return []string{
+		string(OutputFormatText),
```
mvo5, line 25: We have a small risk here that NewResultsFormatter and SupportedOutputFormats get out of sync when a new formatter is added. How about a small variation that avoid this by having a common source of truth? Something like:
```go
import (
	// we cannot use "maps" yet as it needs go1.23
	"golang.org/x/exp/maps"
)
...
// XXX: this could also be supportedFormattersMap or some other name of course
var formatterMap = map[string]ResultsFormatter{
	string(OutputFormatDefault):   &textResultsFormatter{},
	string(OutputFormatText):      &textResultsFormatter{},
	string(OutputFormatJSON):      &jsonResultsFormatter{},
	string(OutputFormatTextShell): &shellResultsFormatter{},
	string(OutputFormatTextShort): &textShortResultsFormatter{},
}

func SupportedOutputFormats() []string {
	keys := maps.Keys(formatterMap)
	sort.Strings(keys)
	return keys
}

// NewResultFormatter will create a formatter based on the given format.
func NewResultsFormatter(format OutputFormat) (ResultsFormatter, error) {
	rs, ok := formatterMap[string(format)]
	if !ok {
		return nil, fmt.Errorf("unsupported formatter %q", format)
	}
	return rs, nil
}
```
and then ideally a test. I think we have two options here:
a) just ensure imagefilter.SupportedOuptutFormats() has exactly the strings we expect
b)  or we just ensures at least one of the expected values is present (e.g. text) and then check that"len(formatterMap) == len(imagefilter.SupportedOutputFormats())" as a cheap/simple validation that the two are not drifting (mabye overkill the above tweaks should ensure this)

Both have pros/cons, a) is simple and more precise but needs an update every time we add a formatter (which is arguably a good thing). b) is good enough but less precise but does not require us to touch the tests again if we add one.

mvo5 (commented):
Thank you! I like the idea of having the SupportedOutputFormats() helper. Some suggestions for the implementation inline.

mvo5 (commented):
Quick drive-by:
1. the PR description should follow the convention: `module: change` (e.g. imagefilter: add SupportedOutputFormats)
2. Rebasing this on top of main and dropping the diff for the "short" output would most likely get this merged much quicker as I feel "short" output will need some more design/discussion but this here is very nice and straightforward.

pkg/imagefilter/formatter.go
schuellerf: Thanks for you patience while I'm trying to learn go :-D

pkg/imagefilter/formatter_test.go
```diff
@@ -1,4 +1,4 @@
-package imagefilter_test
+package imagefilter
```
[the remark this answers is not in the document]
mvo5: Sorry, I realize this was a rather terse comment. Please just drop the first commit (and I personally would just squash 2+3) and then with this diff:
```go
diff --git a/pkg/imagefilter/formatter_test.go b/pkg/imagefilter/formatter_test.go
index 2a4f2ab8c..5bc7e105f 100644
--- a/pkg/imagefilter/formatter_test.go
+++ b/pkg/imagefilter/formatter_test.go
@@ -107,8 +107,8 @@ func TestResultsFormatter(t *testing.T) {
 }
 
 func TestSupportedOutputFormats(t *testing.T) {
-       formatters := SupportedOutputFormats()
-       assert.Len(t, formatters, len(supportedFormatterMap))
+       formatters := imagefilter.SupportedOutputFormats()
+       assert.Len(t, formatters, len(imagefilter.SupportedFormatterMap))
        assert.Contains(t, formatters, "text")
        assert.Contains(t, formatters, "json")
        assert.Contains(t, formatters, "shell")
@@ -116,7 +116,7 @@ func TestSupportedOutputFormats(t *testing.T) {
 }
 
 func TestResultsFormatterError(t *testing.T) {
-       fmter, err := NewResultsFormatter(OutputFormat("unsupported"))
+       fmter, err := imagefilter.NewResultsFormatter(imagefilter.OutputFormat("unsupported"))
        assert.Error(t, err)
        assert.Nil(t, fmter)
 }

```
things should be fine.

pkg/imagefilter/formatter.go
```diff
@@ -27,20 +29,28 @@ type ResultsFormatter interface {
 	Output(io.Writer, []Result) error
 }
 
-// NewResultFormatter will create a formatter based on the given format.
+var supportedFormatterMap = map[string]ResultsFormatter{
```
supakeen, line 32: Two notes:

1. We normally don't use Hungarian notation; e.g. writing the type (`Map`) in the variable name.
2. The behavior here is slightly different because all key lookups now return the same instance of a formatter. Since formatters don't keep state (right now) this might be OK, just wanted to write it down as it's a behavioral change I think?

pkg/imagefilter/formatter.go
```diff
@@ -27,20 +29,28 @@ type ResultsFormatter interface {
...
+	string(OutputFormatText):      &textResultsFormatter{},
+	string(OutputFormatJSON):      &jsonResultsFormatter{},
+	string(OutputFormatTextShell): &shellResultsFormatter{},
+	string(OutputFormatTextShort): &textShortResultsFormatter{},
+}
+
+// SupportedOutputFormats returns a list of supported output formats
+func SupportedOutputFormats() []string {
+	keys := maps.Keys(supportedFormatterMap)
+	sort.Strings(keys)
```
supakeen, line 43: Just a question; but `maps.Keys` returns an iterator. Which part of this function turns that into a slice, is it the `return` or the `sort.Strings`?

pkg/imagefilter/formatter_test.go
```diff
@@ -92,16 +91,31 @@ func TestResultsFormatter(t *testing.T) {
...
-		fmter, err := imagefilter.NewResultsFormatter(imagefilter.OutputFormat(tc.formatter))
+		fmter, err := NewResultsFormatter(OutputFormat(tc.formatter))
 		require.NoError(t, err)
 		err = fmter.Output(&buf, res)
 		assert.NoError(t, err)
 		assert.Equal(t, tc.expectsOutput, buf.String(), tc)
 	}
 }
+
+func TestSupportedOutputFormats(t *testing.T) {
```
supakeen, line 108: I think we should test the sorted-ness here as well as it seems like an important feature of the `SupportedOutputFormats` function :)

supakeen (commented):
Some minor questions and bits.

supakeen:
> the PR description should follow the convention: `module: change` (e.g. imagefilter: add SupportedOutputFormats)

I noticed the commits are now prefixed with the full package path instead of the module name. I don't know how I feel about this :)

pkg/imagefilter/formatter.go
mvo5: This is a bit of a funny one, the new "maps" package in 1.23 does return an iterator but the compat package golang.org/x/exp/maps (that we need to use as we are not on 1.23 yet) just returns a []string slice. We can keep the compat maps until it gets removed and then we either port to iterators or just implement our own .Keys() (its trivial enough).

pkg/imagefilter/formatter_test.go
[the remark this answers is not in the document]
schuellerf: thanks - I had that before but it failed…

now I found, that the reason was not the package name but rather a typo - thanks 🙇

schuellerf:
oh - that shouldn't happen? 
https://github.com/osbuild/images/actions/runs/13071656708/job/36474442726?pr=1168
we are supposed to _support_ 1.22, right?

```
go: github.com/vmware/govmomi/govc@v0.0.0-20250127215336-040a6330365b requires go >= 1.23; switching to go1.23.5
go: downloading go1.23.5 (linux/amd64)
```
should I downgrade that lib so it works with 1.22?

pkg/imagefilter/export_test.go
```diff
@@ -0,0 +1,3 @@
+package imagefilter
+
+var SupportedFormatterMap = supportedFormatters
```
supakeen, line 3: `SupportedFormatters` (the type is already a `map`) is the only bit.

schuellerf: grr - overlooked that one, thanks

pkg/imagefilter/formatter.go
mvo5: Thanks for the careful review. I think this is fine because currently the formatters are an empty map, once we add state the map would become a `map[string]func() Foramtter `most likely to generate the new things (we could do it today but see below).

A fun fact of go is that empty structs get optimized away by the go-compiler, so at least on my system:
```go
func TestNewDifferentObj(t *testing.T) {
	fmt1, err := imagefilter.NewResultsFormatter(imagefilter.OutputFormatText)
	assert.NoError(t, err)
	fmt2, err := imagefilter.NewResultsFormatter(imagefilter.OutputFormatText)
	assert.NoError(t, err)
	assert.NotEqual(t, fmt1, fmt2)
}
```
is false even with the current implementation that creates a "new" formatter with every new() call.

This is a long winded way of saying, I'm okay with as it is but open for ideas.

mvo5 (approved):
Thank you

supakeen (approved):

PR #1170, opened 2025-01-27
distro: use `rand.Int63` intead of `time.UnixNano()`

mvo5:
This is a slightly silly commit. Instead of using `time.UnixNano()` its much more sensible to use `rand.Int63` which is automatically seeded since go1.20 [0] and also gives us a wider range of possible seeds than UnixNano (which will always be of a preditable length).

None of this really matters in practise as we just need something "good enough" but using rand.Int63() is the correct thing to do.

[0] https://tip.golang.org/doc/go1.20#mathrandpkgmathrand

### DIFF
--- a/pkg/distro/distro.go
+++ b/pkg/distro/distro.go
@@ -1,7 +1,7 @@
 package distro
 
 import (
-	"time"
+	"math/rand"
 
 	"github.com/osbuild/images/pkg/blueprint"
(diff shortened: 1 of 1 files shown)


supakeen (approved):

achilleas-k (commented):
Heed to disable the linter check for the weak rng on the line: G404

achilleas-k (approved):

PR #1173, opened 2025-01-28
imagefilter: make `short` output format as unstable

mvo5:
This commit changes the "short" output format so that it is marked as "unstable". There are some open questions about this format and how it relates to e.g.
https://github.com/osbuild/osbuild-composer/pull/4336 which adds very similar API is a slightly different format.

So to prevent users from relying on the API before that is resolved we need to at least make them aware that this may still change.


schuellerf (dismissed):
perfect remediation, THANKS!

mvo5:
Sorry, I'm a muppet, I pushed the wrong commit here accidentally (was meant for #1175), please re-approve.

schuellerf (approved):

PR #1179, opened 2025-01-29
reporegistry: do not prepend `repositories` in confPaths

mvo5:
This PR changes the API for reporegistry in two incompatible
ways. The important change is that we want to drop the
prepending of `repositories` when loading from confPaths.

Doing this on its own has the disadvantage of breaking
the API without breaking the builds. So this includes a
small tweak to allow adding repos via fs.FS directly
which will mean the code in ibcli will become simpler:
```go
diff --git a/cmd/image-builder/repos.go b/cmd/image-builder/repos.go
index bc238dc..5d16a95 100644
--- a/cmd/image-builder/repos.go
+++ b/cmd/image-builder/repos.go
@@ -2,8 +2,6 @@ package main
 
 import (
        "io/fs"
-       "os"
-       "path/filepath"
 
        "github.com/osbuild/images/data/repositories"
        "github.com/osbuild/images/pkg/reporegistry"
@@ -27,17 +25,6 @@ var newRepoRegistry = func(dataDir string) (*reporegistry.RepoRegistry, error) {
                dataDirs = defaultDataDirs
        }
 
-       // XXX: think about sharing this with reporegistry?
-       var fses []fs.FS
-       for _, d := range dataDirs {
-               fses = append(fses, os.DirFS(filepath.Join(d, "repositories")))
-       }
-       fses = append(fses, repos.FS)
-
        // XXX: should we support disabling the build-ins somehow?
-       conf, err := reporegistry.LoadAllRepositoriesFromFS(fses)
-       if err != nil {
-               return nil, err
-       }
-       return reporegistry.NewFromDistrosRepoConfigs(conf), nil
+       return reporegistry.New(dataDirs, []fs.FS{repos.FS})
 }
```
And for composer it will just be adding a `nil` to the New()
call.

Note that this also allows us to unexport the LoadAllRepositories
repository loading as this is only used by composer or ibcli.

---

reporegistry: do not prepend `repositories` in confPaths

This commit breaks the API by changing the confPaths loading
in reporegistry to stop prepending `repositories/`. This makes
it easier to use this interface in e.g. `cmd/build`.

Note that this requires changes in the upstream projecs, notably
osbuild-composer and image-builder-cli.

---

This commit changes `reporegistry.New()` to take a new parameter
`repoConfigFS []fs.FS`. With that we can unexport the public
`LoadAllRepositories{,FromFS}` as the only user for those is
`composer` and that can simply use `.New()`.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -93,12 +93,7 @@ func run() error {
 		}
 		overrideRepos = repoConfig[archName]
 	} else {
-		// HACK: reporegistry hardcodes adding the "repositories"
-		// dir to the "repoConfigPaths" but the interface of
-		// "images" does not expect this.
-		// XXX: should we fix this in reporegistry?
-		repositories = filepath.Join(repositories, "..")
-		reporeg, err = reporegistry.New([]string{repositories})
+		reporeg, err = reporegistry.New([]string{repositories}, nil)
 		if err != nil {
 			return fmt.Errorf("failed to load repositories from %q: %w", repositories, err)
 		}
--- a/pkg/reporegistry/error.go
+++ b/pkg/reporegistry/error.go
@@ -1,13 +1,17 @@
 package reporegistry
 
-import "fmt"
+import (
+	"fmt"
+	"io/fs"
+)
 
 // NoReposLoadedError is an error type that is returned when no repositories
 // are loaded from the given paths.
 type NoReposLoadedError struct {
 	Paths []string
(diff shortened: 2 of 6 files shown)


thozza (commented):
The idea of the change looks good to me, but the code needs some fixes for tests to pass...

thozza (approved):
This looks OK to me, thanks.

achilleas-k (approved):

PR #1183, opened 2025-01-30
data: add fedora-42 repository

mvo5:
This commit adds the fedora-42.json repository. The gpg key is imported from
https://raw.githubusercontent.com/rpm-software-management/distribution-gpg-keys/refs/heads/main/keys/fedora/RPM-GPG-KEY-fedora-42-primary and then manually converted to json (do we have a tool for this already?)

Closes: https://github.com/osbuild/image-builder-cli/issues/107

I did a quick manual build wth ibcli and it does build and I can login but I assume as it developes we will need to tweak more things in the fedora distro pkg.


achilleas-k (approved):

ondrejbudai (approved):
Thanks! ❤️

PR #1184, opened 2025-01-30
dnfjson: support redirecting dnfjsons `stderr`

mvo5:
This commit tweaks the dnfjson.Solver to also have a `Stderr` attribute. This is a minimal change to prevent `image-builder-cli` from showing:
```console
image-builder build qcow2 --distro centos-9 --progress=verbose
Manifest generation step
Building manifest for qcow2-centos-9
No match for group package "iwl5150-firmware"
No match for group package "iwl6050-firmware"
No match for group package "firewalld"
No match for group package "iwl7260-firmware"
No match for group package "dracut-config-rescue"
No match for group package "iwl105-firmware"
No match for group package "iwl2000-firmware"
No match for group package "iwl100-firmware"
No match for group package "iwl6000g2a-firmware"
No match for group package "iwl1000-firmware"
No match for group package "iwl3160-firmware"
No match for group package "iwl135-firmware"
No match for group package "iwl2030-firmware"
No match for group package "iwl5000-firmware"
...
```
which seems to be resulting from the "exclude" directive of the transaction. When qcow2 in centos-9 is build it contains the following:
```go
	ps := rpmmd.PackageSet{
		Include: []string{
			"@core",
			...
		},
		Exclude: []string{
			...
			"firewalld",
			"iwl7260-firmware",
			...
		},
```
and each package that is both in @core and in the exclude seems to trigger a message like the above, e.g.:
```
No match for group package "firewalld"
```
This is just confusing for our users so image-builder-cli will just discard this stderr output from dnfjson.

I'm not sure if there is a better way to handle this though, ideas (very) welcome.

### DIFF
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -18,6 +18,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


achilleas-k:
> I'm not sure if there is a better way to handle this though, ideas (very) welcome.

A way to disable or catch warnings in libdnf would have been nice, which is why we never did anything about it.  Also because they were never visible outside the journal (and our own testing / development tooling).

achilleas-k (approved):
Small, clean, and unobtrusive.  LGTM.

PR #1185, opened 2025-01-30
cloud: add new Uploader interface and implement for AWS

mvo5:
This commit adds a new `cloud.Uploader` interface that combines the upload and register into a single operation. The rational is that with that we avoid leaking resource if e.g. the upload works but the registration fails
(c.f. https://github.com/osbuild/images/pull/1145). This would replace most of the existing bib uplodaer code so that it is shareable.

This would looks like this:
* https://github.com/osbuild/bootc-image-builder/pull/815
* https://github.com/osbuild/image-builder-cli/compare/main...mvo5:add-upload-cmd?expand=1

After that I would like to look into the following:
- I would like remake the cloud interface for images around something like this
- ~can we remove (most?) of the existing awscloud in images. because osbuild-compser impleents it all too but has now moved to aws-v2 and it seems we have a lot of unused code in images right now(?)~
- ~how important is cmd/boot-aws ? it uses a lot of the existing API but the composer does not have it~

### DIFF
--- a/pkg/cloud/awscloud/awscloud.go
+++ b/pkg/cloud/awscloud/awscloud.go
@@ -243,6 +243,8 @@ func WaitUntilImportSnapshotTaskCompletedWithContext(c *ec2.EC2, ctx aws.Context
 // The caller can also specify the name of the role used to do the import.
 // If nil is given, the default one from the SDK is used (vmimport).
 // Returns the image ID and the snapshot ID.
+//
+// XXX: make this return (string, string, error) instead of pointers
 func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch string, bootMode, importRole *string) (*string, *string, error) {
 	rpmArchToEC2Arch := map[string]string{
 		"x86_64":  "x86_64",
@@ -294,11 +296,7 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 
 	// we no longer need the object in s3, let's just delete it
 	logrus.Infof("[AWS] 🧹 Deleting image from S3: %s/%s", bucket, key)
-	_, err = a.s3.DeleteObject(&s3.DeleteObjectInput{
-		Bucket: aws.String(bucket),
-		Key:    aws.String(key),
-	})
-	if err != nil {
+	if err = a.DeleteObject(bucket, key); err != nil {
 		return nil, nil, err
 	}
 
@@ -386,6 +384,14 @@ func (a *AWS) Register(name, bucket, key string, shareWith []string, rpmArch str
 	}
 
 	return registerOutput.ImageId, snapshotID, nil
+}
+
+func (a *AWS) DeleteObject(bucket, key string) error {
+	_, err := a.s3.DeleteObject(&s3.DeleteObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key),
+	})
+	return err
 }
 
 // target region is determined by the region configured in the aws session
--- a/pkg/cloud/awscloud/export_test.go
+++ b/pkg/cloud/awscloud/export_test.go
@@ -1,0 +1,11 @@
+package awscloud
+
+type AwsClient = awsClient
+
+func MockNewAwsClient(f func(string) (awsClient, error)) (restore func()) {
+	saved := newAwsClient
+	newAwsClient = f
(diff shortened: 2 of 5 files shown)


achilleas-k:
> * how important is cmd/boot-aws ? it uses a lot of the existing API but the composer does not have it

It's used in our testing and was meant to be the first of a handful of similar commands for different clouds.  We should still do that.

achilleas-k:
> * can we remove (most?) of the existing awscloud in images. because osbuild-compser impleents it all too but has now moved to aws-v2 and it seems we have a lot of unused code in images right now(?)

I would prefer the opposite.  Having cloud uploaders as part of image definitions might seem strange, but we discussed it a few times in the past and came to the agreement that it makes sense because integrating image definitions with a client for their target platform's API makes for a more complete story.  One argument that often comes up goes "an AMI isn't an AMI until it's uploaded and registered on AWS EC2".

achilleas-k (commented):
What's the end goal here and what does this PR do?

The answer I'm assuming is that it defines an interface for uploading stuff to cloud platforms and begins with one implementation based on the existing AWS code.
But I think it doesn't go far enough and isn't providing much value right now (though, the testing is nice).  I'd like to see this common interface happen, but I think it makes more sense to look at the other cloud providers and consider what it might look like when we add those to the mix.  `UploadAndRegister()` sounds generic enough but I don't know if it makes sense for Azure, or GCP, or Oracle (it probably does).  It's not set in stone, we can always change later, so maybe we should go with this and start iterating with the other platforms, but if we're going to start working on cloud clients seriously, I think we should review what we already have and potentially tear it down and build it back up in a way that works for what we're trying to do.

The `boot-aws` command was written specifically for our testing, so the `run`, `setup`, and `teardown` subcommands were written to serve development and testing and those requirements in turn informed some of the choices in the client library (though parts of it were already there from the split from osbuild-composer).  For example, I never considered adding a `Check()` function like we have now because it was never targeted at any user that might need that function; the keys were set in the CI setup and never looked at again and when I use it myself for troubleshooting, I have my own set that works (and have probably way more permissions than they need but let's not think about that right now).  I'm all in favour of making a `cloud.Uploader` interface for all our uploaders, but I also want to do it right, especially if it's going to serve osbuild-composer, ib-cli, and bib.

Now all that said, I'm fine to merge this as is.

achilleas-k (approved):

mvo5:
> What's the end goal here and what does this PR do?

Sorry, I really should have explained this better in the description. The short term (and driver) for the PR is simply
that I wanted to not duplicate the uploading code for aws in bib and ibcli and also to solve the resource leak issue
that lzap described his cleanup PR. 

> The answer I'm assuming is that it defines an interface for uploading stuff to cloud platforms and begins with one implementation based on the existing AWS code. But I think it doesn't go far enough and isn't providing much value right now (though, the testing is nice). I'd like to see this common interface happen, but I think it makes more sense to look at the other cloud providers and consider what it might look like when we add those to the mix. `UploadAndRegister()` sounds generic enough but I don't know if it makes sense for Azure, or GCP, or Oracle (it probably does). It's not set in stone, we can always change later, so maybe we should go with this and start iterating with the other platforms, but if we're going to start working on cloud clients seriously, I think we should review what we already have and potentially tear it down and build it back up in a way that works for what we're trying to do.

Yeah, I think it would be very nice if a generalized interface would work, we may need to tweak things (as you said) but I hope it will be similar enough. The hope is that the amount of changes to add a new cloud on the consumers is minimal this way.

> The `boot-aws` command was written specifically for our testing, so the `run`, `setup`, and `teardown` subcommands were written to serve development and testing and those requirements in turn informed some of the choices in the client library (though parts of it were already there from the split from osbuild-composer). For example, I never considered adding a `Check()` function like we have now because it was never targeted at any user that might need that function; the keys were set in the CI setup and never looked at again and when I use it myself for troubleshooting, I have my own set that works (and have probably way more permissions than they need but let's not think about that right now). I'm all in favour of making a `cloud.Uploader` interface for all our uploaders, but I also want to do it right, especially if it's going to serve osbuild-composer, ib-cli, and bib.

Thanks! This of course means that my original idea of only keeping small parts of the cloud code in images is silly. It seems we would move all(?)/most(?) of the cloud code from composer in here and make composer use it(?). Obviously all followup material.

PR #1190, opened 2025-02-03
policies: allow custom directories and files under /usr/local

ondrejbudai:
According to FHS, /usr/local is meant for installing software locally.
Thus, this directory feels like the ideal place for extra software
installed by Image Builder's users via the custom files and directories.

This commit relaxes the custom files and directory policies so users
can add arbitrary files and directories under /usr/local.


achilleas-k:
We should also add some tests for the file and directory policies to the package tests here.

achilleas-k:
Not sure what kind of weird glitch made everything fail earlier but it seems to be okay now.

thozza (approved):
LGTM

achilleas-k:
Spoke too soon

achilleas-k:
ah wait.  The merge base of this PR is very old (from Dec).  Updating.

PR #1191, opened 2025-02-04
osbuild: update org.osbuild.systemd.unit.create stage

achilleas-k:
This PR updates the org.osbuild.systemd.unit.create stage to match the options in osbuild, especially the new sections for mount and socket unit types.
Tests added to cover all validation failures.


supakeen (approved):
Random thought but it'd be quite nice if we could partially codegen validate functions based on schemas.

thozza (approved):
Thanks.

We should really somehow mark the minimum `osbuild` version for `images` version, to not generate manifests that will fail with old `osbuild` and also auto-detect (by some kind of test) in all places that use `images`, that we need to bump the dependency in the SPEC.

PR #1192, opened 2025-02-04
manifest: add insights-client.service enable

loadtheaccumulator:
Add insights-client.service enable to kickstart file to fix issue with rpm-ostree updates not being collected and reported via insights client at reboot.

Fixes: HMS: 4827


pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -702,7 +702,9 @@ func (p *AnacondaInstallerISOTree) makeKickstartStages(stageOptions *osbuild.Kic
 
 		}
 		hardcodedKickstartBits += makeKickstartSubscriptionPost(subscriptionPath, systemPath)
-
+		if p.SubscriptionPipeline.Subscription.Insights {
+			hardcodedKickstartBits += makeKickstartInsightsClientBootPost()
+		}
```
achilleas-k, line 707: The `SubscriptionPipeline` is [always created with the subscription options set](https://github.com/osbuild/images/blob/e8310ecccc1db95ee322375eea9cbb5d24ad40fa/pkg/image/anaconda_ostree_installer.go#L138-L141) but let's be cautious here ans also make sure that `p.SubscriptionPipeline.Subscription != nil` before dereferencing.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -780,3 +782,15 @@ systemctl enable osbuild-subscription-register.service
...
+func makeKickstartInsightsClientBootPost() string {
+	// Enabling the insights-client.service is necessary for the insights client to
+	// collect when a system is rebooted after an upgrade.
+	// It is facilitated via insights-client.service.d/override.conf
+	kickstartInsightsClientBootPost := `
+%%post
+systemctl enable insights-client.service
+%%end
+`
+	return fmt.Sprintf(kickstartInsightsClientBootPost)
```
achilleas-k, line 795:
```suggestion
	return fmt.Sprint(kickstartInsightsClientBootPost)
```

achilleas-k (commented):
Thanks!  This LGTM but there's a couple of small issues below.

pkg/manifest/anaconda_installer_iso_tree.go
ezr-ondrej: Why are we even printing it? What change to the string we want to hapen through the print? :thinking:

achilleas-k: Well... now I feel stupid :sweat_smile: 

Yeah, just return the string.

loadtheaccumulator: Yeah. That was a wee hours cut and paste. Can just return the string since it's one var.

loadtheaccumulator:
@achilleas-k It looks like the service override file is written on condition of serviceOptions.InsightsOnBoot being true in subscriptions.go. I'm not sure the preferred way to pull that check in here so it only writes to the kickstart. Could also just write a bash check in the kickstart to check for the service, but the check is cleaner.

https://github.com/osbuild/images/blob/e8310ecccc1db95ee322375eea9cbb5d24ad40fa/pkg/manifest/subscription.go#L125-L131

thozza:
Wouldn't it make more sense to conditionally add `Wants=insights-client.service` `if insightsOnBoot` with the proper ordering, to the `osbuild-subscription-register.service` unit https://github.com/osbuild/images/blob/e8310ecccc1db95ee322375eea9cbb5d24ad40fa/pkg/manifest/subscription.go#L163 ?

loadtheaccumulator:
> Wouldn't it make more sense to conditionally add `Wants=insights-client.service` `if insightsOnBoot` with the proper ordering, to the `osbuild-subscription-register.service` unit
> 

I tried this out with an override and the insights client does run at boot even when the ConditionPathExists condition is false, but execution at first boot errors. I haven't chased the root cause down yet, but assuming it's because it's running before registration and needs to be added to the After in the insights client service. I haven't tested it, but with the After, I would expect collection to happen twice on first boot. Once with the Wants and once with the registration.

The insights-client.service status does reflect it is disabled, so wondering if it running could cause confusion. Disabling will require overriding the Wants in the register unit or removing the insights client override file.

loadtheaccumulator:
The registration service needs to be added to the insights-client After directive anyway. I added an if insights-client override file exists conditional to the kickstart post and everything worked as expected.

thozza:
@loadtheaccumulator `Wants=` just ensures the unit is pulled into the transaction, and `After=` ensures the proper order. Even if multiple units `Wants` the same unit, it is still started only once at boot.

loadtheaccumulator:
Hey @thozza .For Edge updates to reflect in Insights, they have to collect at boot. Insights normally collects periodically, so insights-client.service handles the collection at boot. The kickstart file was originally injected into the ISO by Edge API three years or so ago, and we both configured and enabled the service there because we didn't have access to the ostree image at that time. When we moved that functionality to Image Builder last year, the service was configured, but wasn't enabled. 

For me, the problem with the Wants is that it doesn't reflect the actual status of the service, and the registration service is designed for first boot. Maybe we just need to document it in the override.conf file, but if collection errors on boot at some point, my first step is to check the service. It will reflect disabled even though it runs via the Wants in another service.

The first boot registration service is also enabled via the kickstart.

loadtheaccumulator:
I just ran across an insights-client-boot.service that is enabled by default. Testing overriding it now instead of insights-client.service.

loadtheaccumulator:
Overriding insights-client-boot.service works, and I updated this PR to reflect the change.
- service is already enabled by default (so no need for additional enable step in ks file or other)
- runs and collects at every boot

loadtheaccumulator:
I updated the title and description accordingly.

pkg/manifest/subscription.go
```diff
@@ -211,13 +211,21 @@ func runInsightsClientOnBoot() (*fsnode.Directory, *fsnode.File, error) {
...
+//
+// The insights-client-boot.service is defined as first boot with a conditional file
+// that gets removed when it runs (similar to the registration service).
+// We are overriding that here to run at every boot to collect changes after upgrades, etc.
 func insightsClientDropin() (string, string) {
-	return "/etc/systemd/system/insights-client.service.d/override.conf", `[Unit]
+	return "/etc/systemd/system/insights-client-boot.service.d/override.conf", `[Unit]
 Requisite=greenboot-healthcheck.service
-After=network-online.target greenboot-healthcheck.service osbuild-first-boot.service
+After=network-online.target greenboot-healthcheck.service osbuild-first-boot.service osbuild-subscription-register.service
```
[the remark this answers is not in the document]
thozza: `After=` and `Before=` affect only the ordering when starting or stopping units, but they won't make them being started or stopped if not done by other means (e.g. using `Requires=` or `Wants=`).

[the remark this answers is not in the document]
loadtheaccumulator: @ezr-ondrej systems can also be registered manually after install. osbuild-first-boot "might" be there if the image is built without credentials provided, but I don't think it's required for manual registration.

thozza (approved):
AFAICT, the change looks reasonable from what I was able to sanity-check. However, I have no way to verify that it has the intended effect.

loadtheaccumulator:
I've tested by manually adding to a kickstart file used at build. I'll also test after merge.

achilleas-k (approved):
Thanks!  LGTM

PR #1194, opened 2025-02-05
[RFC] osbuild: define the minimum osbuild version and enforce it

thozza:
Define a new variable storing the minimum osbuild version supported
by the module and an exported function to get it. The actual version is
stored in a text file under `data/dependencies/` directory and embedded
at compile time. Enforce the minimum version when executing osbuild.

This has the potential to prevent surprises when using older osbuild
with `images` implementation that relies on changes in a newer version
of osbuild.

This will allow dependent projects to extract the minimum version and
check if their packaging requirements are up to date with the version of
`images` that they depend on.

This will enable adding a composite GH action that can be used by other
projects depending on `osbuild/images` to check that their SPEC file
depends at least on the specified osbuild version.


achilleas-k (commented):
Ok, this makes sense, I like it.

One thing I would change though is to probably put the minimum version in a text file and compile it in with `go:embed`, just for visibility and to make it easier to work with in some automated way.  For example, if we want to read it during some testing, or if someone is browsing the repo to find it, it would be a lot easier to be able to point to something like `res/osbuild-min-version`.  We could even link to it from the README or some docs without needing to point to specific lines in code.

thozza:
> One thing I would change though is to probably put the minimum version in a text file and compile it in with `go:embed`, just for visibility and to make it easier to work with in some automated way. For example, if we want to read it during some testing, or if someone is browsing the repo to find it, it would be a lot easier to be able to point to something like `res/osbuild-min-version`. We could even link to it from the README or some docs without needing to point to specific lines in code.

Great idea!

supakeen (approved):
I'd probably call the file `osbuild-minimum-version` in case we want to also check other stuff later on but we can do that when we want to check other stuff later on.

PR #1195, opened 2025-02-05
distro: rhel9: edge: exclude bootupd

runcom:
From Colin:

If the Edge-9.6 setup isn't ready for bootupd then at the current time the package needs to be excluded.

we do this already in fedora indeed...


thozza (approved):

runcom:
@thozza can we get a tag for this PR to bump it in osbuild-composer once it's merged? :angel:

PR #1197, opened 2025-02-06
modularity support (HMS-5533)

supakeen:
Adds modularity support to `images`. Needs tests.

JIRA: [HMS-5533](https://issues.redhat.com/browse/HMS-5533)

### DIFF
--- a/internal/workload/custom.go
+++ b/internal/workload/custom.go
@@ -3,12 +3,17 @@ package workload
 type Custom struct {
 	BaseWorkload
 	Packages         []string
+	EnabledModules   []string
 	Services         []string
 	DisabledServices []string
 }
 
 func (p *Custom) GetPackages() []string {
 	return p.Packages
+}
+
+func (p *Custom) GetEnabledModules() []string {
+	return p.EnabledModules
 }
 
 func (p *Custom) GetServices() []string {
--- a/pkg/blueprint/blueprint.go
+++ b/pkg/blueprint/blueprint.go
@@ -3,11 +3,16 @@ package blueprint
 
 // A Blueprint is a high-level description of an image.
 type Blueprint struct {
-	Name           string          `json:"name" toml:"name"`
-	Description    string          `json:"description" toml:"description"`
-	Version        string          `json:"version,omitempty" toml:"version,omitempty"`
-	Packages       []Package       `json:"packages" toml:"packages"`
-	Modules        []Package       `json:"modules" toml:"modules"`
+	Name        string    `json:"name" toml:"name"`
+	Description string    `json:"description" toml:"description"`
+	Version     string    `json:"version,omitempty" toml:"version,omitempty"`
+	Packages    []Package `json:"packages" toml:"packages"`
+	Modules     []Package `json:"modules" toml:"modules"`
+
+	// Note, this is called "enabled modules" because we already have "modules" except
+	// the "modules" refers to packages and "enabled modules" refers to modularity modules.
+	EnabledModules []EnabledModule `json:"enabled_modules" toml:"enabled_modules"`
+
 	Groups         []Group         `json:"groups" toml:"groups"`
 	Containers     []Container     `json:"containers,omitempty" toml:"containers,omitempty"`
 	Customizations *Customizations `json:"customizations,omitempty" toml:"customizations"`
@@ -21,6 +26,12 @@ type Blueprint struct {
 type Package struct {
 	Name    string `json:"name" toml:"name"`
 	Version string `json:"version,omitempty" toml:"version,omitempty"`
+}
+
+// A module specifies a modularity stream.
+type EnabledModule struct {
+	Name   string `json:"name" toml:"name"`
+	Stream string `json:"stream,omitempty" toml:"stream,omitempty"`
 }
 
 // A group specifies an package group.
@@ -63,6 +74,16 @@ func (b *Blueprint) GetPackagesEx(bootable bool) []string {
 	return packages
 }
 
+func (b *Blueprint) GetEnabledModules() []string {
+	modules := []string{}
+
+	for _, mod := range b.EnabledModules {
+		modules = append(modules, mod.ToNameStream())
+	}
+
+	return modules
+}
+
 func (p Package) ToNameVersion() string {
 	// Omit version to prevent all packages with prefix of name to be installed
 	if p.Version == "*" || p.Version == "" {
@@ -70,4 +91,8 @@ func (p Package) ToNameVersion() string {
 	}
 
 	return p.Name + "-" + p.Version
+}
+
+func (p EnabledModule) ToNameStream() string {
+	return p.Name + ":" + p.Stream
 }
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -779,7 +779,7 @@ type ModuleConfigData struct {
 
 type ModuleFailsafeFile struct {
 	Path string `json:"path"`
-	Data string `json:"string"`
+	Data string `json:"data"`
 }
 
 // dnf-json error structure
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -168,6 +168,7 @@ type Solver struct {
 // DepsolveResult contains the results of a depsolve operation.
 type DepsolveResult struct {
 	Packages []rpmmd.PackageSpec
+	Modules  []rpmmd.ModuleSpec
 	Repos    []rpmmd.RepoConfig
 	SBOM     *sbom.Document
 	Solver   string
@@ -240,7 +241,7 @@ func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet, sbomType sbom.StandardType
 		return nil, fmt.Errorf("decoding depsolve result failed: %w", err)
 	}
 
-	packages, repos := result.toRPMMD(rhsmMap)
+	packages, modules, repos := result.toRPMMD(rhsmMap)
 
 	var sbomDoc *sbom.Document
 	if sbomType != sbom.StandardTypeNone {
@@ -252,6 +253,7 @@ func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet, sbomType sbom.StandardType
 
 	return &DepsolveResult{
 		Packages: packages,
+		Modules:  modules,
 		Repos:    repos,
 		SBOM:     sbomDoc,
 		Solver:   result.Solver,
@@ -577,7 +579,7 @@ func (s *Solver) makeSearchRequest(repos []rpmmd.RepoConfig, packages []string) 
 // convert internal a list of PackageSpecs and map of repoConfig to the rpmmd
 // equivalents and attach key and subscription information based on the
 // repository configs.
-func (result depsolveResult) toRPMMD(rhsm map[string]bool) ([]rpmmd.PackageSpec, []rpmmd.RepoConfig) {
+func (result depsolveResult) toRPMMD(rhsm map[string]bool) ([]rpmmd.PackageSpec, []rpmmd.ModuleSpec, []rpmmd.RepoConfig) {
 	pkgs := result.Packages
 	repos := result.Repos
 	rpmDependencies := make([]rpmmd.PackageSpec, len(pkgs))
@@ -610,6 +612,22 @@ func (result depsolveResult) toRPMMD(rhsm map[string]bool) ([]rpmmd.PackageSpec,
 		}
 	}
 
+	mods := result.Modules
+	moduleSpecs := make([]rpmmd.ModuleSpec, len(mods))
+
+	i := 0
+	for _, mod := range mods {
+		moduleSpecs[i].ModuleConfigFile.Data.Name = mod.ModuleConfigFile.Data.Name
+		moduleSpecs[i].ModuleConfigFile.Data.Stream = mod.ModuleConfigFile.Data.Stream
+		moduleSpecs[i].ModuleConfigFile.Data.State = mod.ModuleConfigFile.Data.State
+		moduleSpecs[i].ModuleConfigFile.Data.Profiles = mod.ModuleConfigFile.Data.Profiles
+
+		moduleSpecs[i].FailsafeFile.Path = mod.FailsafeFile.Path
+		moduleSpecs[i].FailsafeFile.Data = mod.FailsafeFile.Data
+
+		i++
+	}
+
 	repoConfigs := make([]rpmmd.RepoConfig, 0, len(repos))
 	for repoID := range repos {
 		repo := repos[repoID]
@@ -635,7 +653,7 @@ func (result depsolveResult) toRPMMD(rhsm map[string]bool) ([]rpmmd.PackageSpec,
 			SSLClientCert:  repo.SSLClientCert,
 		})
 	}
-	return rpmDependencies, repoConfigs
+	return rpmDependencies, moduleSpecs, repoConfigs
 }
 
 // Request command and arguments for dnf-json
--- a/pkg/dnfjson/dnfjson.go
+++ b/pkg/dnfjson/dnfjson.go
@@ -465,9 +465,10 @@ func (s *Solver) makeDepsolveRequest(pkgSets []rpmmd.PackageSet, sbomType sbom.S
 	transactions := make([]transactionArgs, len(pkgSets))
 	for dsIdx, pkgSet := range pkgSets {
 		transactions[dsIdx] = transactionArgs{
-			PackageSpecs:    pkgSet.Include,
-			ExcludeSpecs:    pkgSet.Exclude,
-			InstallWeakDeps: pkgSet.InstallWeakDeps,
+			PackageSpecs:      pkgSet.Include,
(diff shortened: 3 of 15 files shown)


supakeen:
Currently not working as `module_hotfixes` is false?

supakeen:
Now has a test config; has a (quick) test if stages get injected (thanks to @mvo5) and I'll be adding some stage tests.

pkg/osbuild/dnf_module_config_stage.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	return &Stage{
+		Type:    "org.osbuild.dnf.module-config",
+		Options: options,
+	}
+}
+
+func GenDNFModuleConfigStages(modules []rpmmd.ModuleSpec) []*Stage {
+	stages := make([]*Stage, 0)
```
thozza, line 44: Nitpick: you could also `make([]*Stage, 0, len(modules))`, since the size is known in advance.

pkg/manifest/os.go
```diff
@@ -737,6 +739,36 @@ func (p *OS) serialize() osbuild.Pipeline {
 		pipeline.AddStages(osbuild.GenFileNodesStages(p.Files)...)
 	}
 
+	// write modularity related configuration files
+	if len(p.moduleSpecs) > 0 {
+		pipeline.AddStages(osbuild.GenDNFModuleConfigStages(p.moduleSpecs)...)
```
thozza, line 744: Looking at the code below, would it make sense to extend the `GenDNFModuleConfigStages()` (maybe a renamed version), to generate all of the relevant osbuild stages for a given `moduleSpecs`, including any `fsnode.File` that were created? The current version is, however, OK from my PoV.

pkg/manifest/os.go
```diff
@@ -737,6 +739,36 @@ func (p *OS) serialize() osbuild.Pipeline {
...
+			failsafeFiles = append(failsafeFiles, moduleFailsafeFile)
+		}
+
+		failsafeDir, err := fsnode.NewDirectory("/var/lib/dnf/modulefailsafe", nil, nil, nil, true)
+
+		if err != nil {
+			panic("failed to create module failsafe directory")
+		}
+
+		pipeline.AddStages(osbuild.GenDirectoryNodesStages([]*fsnode.Directory{failsafeDir})...)
```
thozza, line 766: Would it make sense to make this conditional `if len(failsafeFiles) > 0` ?

test/configs/modularity.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "modularity",
+  "blueprint": {
+    "packages": [
+      {
+        "name": "@nodejs:18"
```
thozza, line 6: We will need a way to enable a module in the BP and install a specific package from it in the `packages` section, right? You could then extend https://github.com/osbuild/images/blob/main/test/scripts/base-host-check.sh to check for the created module files on the OS. I mean, you could do that even in this scenario.

pkg/manifest/os.go
supakeen: It could be; but it'd be an error to not have any `failsafeFiles` when there *were* modules in the depsolve result.

test/configs/modularity.json
supakeen: I thought we were skipping that part of it in favour of doing everything through module names in the packages; if we also want to support the former (probably up to the frontend?) in blueprints then yes it needs to be expanded.

cc @croissanne / @lucasgarfield?

pkg/dnfjson/dnfjson.go
```diff
@@ -610,6 +612,22 @@ func (result depsolveResult) toRPMMD(rhsm map[string]bool) ([]rpmmd.PackageSpec,
 		}
 	}
 
+	mods := result.Modules
+	moduleSpecs := make([]rpmmd.ModuleSpec, len(mods))
+
+	i := 0
+	for _, mod := range mods {
```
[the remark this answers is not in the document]
supakeen: `mods` is a `map[string]ModuleSpec` here so the first value is the key, not an index.

test/configs/modularity.json
thozza: My understanding was that some modules do not have any default profile or any profile at all. If you want to install a specific package from that module and a specific stream, you'd need to enable it explicitly. I thought that was the whole point for adding `module-enable-specs` to the depsolve request :thinking: 

What this manifest gen does was possible before the modularity work landed, just the resulting OS image would not have the proper module enabled and the package marked as coming from it. However, this does not change anything from the point of view of what the user can do with modules.

While we can probably hide module enabling when using via CloudAPI, it seems to me that we should somehow test it anyway, even if we end up not exposing it in the BP.

supakeen: Right, the important part I believe was that the built images know that, and which, modules have content installed which we do through the module configuration and fallback files.

---

It is possible to enable a module yet to not install any packages from it; however when not installing any packages it also won't be returned by the depsolver as no modular packages would have been part of the solved set and we can't get 'enabled modules' from a transaction (we spoke about this in the depsolve PR in `osbuild`, I believe).

Which bits are enabled here is entirely up to the workflow the frontend wants to use, in my opinion. If it wants to show available modules and install them directly (the usecase currently satisfied in this PR) or if it'd show available modules and selecting them then allows you to select packages from those modules (the usecase that would need `enabled-modules`).

pkg/dnfjson/dnfjson.go
```diff
@@ -779,7 +797,7 @@ type ModuleConfigData struct {
 
 type ModuleFailsafeFile struct {
 	Path string `json:"path"`
-	Data string `json:"string"`
+	Data string `json:"data"`
```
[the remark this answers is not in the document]
supakeen: Changed and force pushed.

pkg/osbuild/dnf_module_config_stage.go
supakeen: Changed and force pushed.

pkg/manifest/os.go
supakeen: Let's keep it as is for now then and I'll note a followup to do that :)

test/configs/modularity.json
croissanne: So I think in the UI we'll actually avoid installing module packages 'by modules'. We'll just continue with package search as usual, except for module packages we'll select the latest supported version by default or they can select another version and mark them as supported or not. So that means from composer and images pov we'll be enabling modules by hand and then installing individual packages from them.

lucasgarfield: :+1:  @croissanne

supakeen: Then I'll be adding/exposing the enabled modules as well.

pkg/manifest/os.go
achilleas-k: Reading above, there's no way for `len(failsafeFiles)` to have be different from `len(moduleSpecs)`, even if one or more `module.FailsafeFile.Data` are empty (which it should never be).

pkg/manifest/os.go
```diff
@@ -737,6 +739,36 @@ func (p *OS) serialize() osbuild.Pipeline {
...
+		pipeline.AddStages(osbuild.GenFileNodesStages(failsafeFiles)...)
+
+		p.Files = append(p.Files, failsafeFiles...)
```
achilleas-k, line 769: Note for another time (future refactoring): We could pull the failsafe file handling out into a separate function, either here or in osbuild as a helper function in `osbuild/dnf_module_config_stage.go`.  The work to generate the files is localised here so it's not a big issue, there's no duplication, but it would make the OS pipeline generation easier to read.

achilleas-k (commented):
LGTM so far.  Waiting for a resolution on the blueprint discussion.

pkg/manifest/os.go
thozza: +1 I had a very similar comment, which has already been resolved.

supakeen:
Pushed a bunch more code to route enabled modules through customizations into workloads into the OS pipeline where it should end up in the depsolve. Needs end-to-end tests still so keeping as draft.

test/configs/modularity.json
croissanne: I marked this as resolved, i think with `enabled_modules` there we can continue / merge.

test/configs/modularity.json
```diff
@@ -1,5 +1,8 @@
 {
   "name": "modularity",
+  "expect_modularity": {
+    "modules": ["nodejs:18"]
+  },
```
thozza, line 5: Maybe instead explicitly enable the module in the blueprint and base your check script on that? :thinking:

test/scripts/base-host-check.sh
```diff
@@ -98,6 +98,18 @@ check_ca_cert() {
     fi
 }
 
+check_modularity() {
+    # Verify modules that are enabled on a system, if any.
+    modules_expected=$(jq -rc '.expect_modularity.modules[]' "${config}")
```
thozza, line 103: You could either check for explicitly enabled modules in the BP or filter any packages in `.blueprint.pakages[]` for any that start with `@` and contain `:` ?

```jq
jq -rc '.blueprint.packages[] | select(.name | startswith("@") and contains(":"))' "${config}"
```

AFAIK, we never added a custom field into the manifest gen config, which is then used by a test case. Instead, we always inspect the BP and base the check on that.

test/config-map.json
```diff
@@ -374,5 +374,13 @@
       "centos-10",
       "centos-9"
     ]
+  },
+  "./configs/modularity.json": {
+    "image-types": [
+      "qcow2"
```
thozza, line 380: Please use `ami`, because only AWS images are boot-tested, so the `base-host-check.sh` script does not get run for `qcow2` :facepalm:

thozza (commented):
I did a full review and added a few comments. I think that once the test case for modularity manifest gets run on a booted image, the PR should be good to be merged from my PoV.

test/configs/modularity.json
supakeen: I think this testcase is from before the enabled modules in blueprints and was based on the package-name-is-module-spec bits. I'll add a second testcase that uses the enabled modules.

test/config-map.json
supakeen: Wow, thanks :)

pkg/rpmmd/repository.go
```diff
@@ -141,6 +141,7 @@ func (pkg Package) ToPackageInfo() PackageInfo {
 type PackageSet struct {
 	Include         []string
 	Exclude         []string
+	Modules         []string
```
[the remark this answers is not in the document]
supakeen: Ah. I had made an explicit choice to transition from `EnabledModules` in customizations to `Modules` in the "more internal" bits but it makes sense to call them `EnabledModules` everywhere too.

internal/workload/custom.go
```diff
@@ -3,6 +3,7 @@ package workload
 type Custom struct {
 	BaseWorkload
 	Packages         []string
+	Modules          []string
```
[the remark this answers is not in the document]
supakeen: Done.

pkg/rpmmd/repository.go
```diff
@@ -199,6 +199,28 @@ type PackageInfo struct {
...
+type ModuleConfigData struct {
+	Name     string   `json:"name"`
+	Stream   string   `json:"stream"`
+	Profiles []string `json:"profiles"`
+	State    string   `json:"state"`
+}
+
+type ModuleFailsafeFile struct {
+	Path string `json:"path"`
+	Data string `json:"string"`
```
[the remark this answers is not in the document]
supakeen: Done.

test/configs/modularity.json
supakeen: I've added a second test config for this.

test/scripts/base-host-check.sh
supakeen: I've adjusted base-host checks to use the blueprint, instead of a new attribute.

test/config-map.json
supakeen: Done.

supakeen:
Working on the last bits for the base-host check but I've addressed @thozza's other comments.

supakeen:
Redrafting as there is an error when modules are enabled, and passed along:

```
$ go run ./cmd/gen-manifests -config test/configs/modularity-enabled-modules.json -distros centos-9 -arches x86_64 -types ami
# ...
  0: [centos_9-x86_64-ami-modularity_enabled_modules.json] depsolve failed: error depsolving: makeDepsolveRequest failed: chained packageSet 2 does not use all of the repos used by its predecessor
```

supakeen:
> Redrafting as there is an error when modules are enabled, and passed along:
> 
> ```
> $ go run ./cmd/gen-manifests -config test/configs/modularity-enabled-modules.json -distros centos-9 -arches x86_64 -types ami
> # ...
>   0: [centos_9-x86_64-ami-modularity_enabled_modules.json] depsolve failed: error depsolving: makeDepsolveRequest failed: chained packageSet 2 does not use all of the repos used by its predecessor
> ```

This was due to adding an *extra* transaction including the modules, which didn't have the repositories set on it. I've changed the code to now include any enabled modules in the workload transaction directly.

Note that this is only active when also installing a package, which is "ok" because that's how this will be used (enable module + install packages from it).

supakeen:
The base host tests are now passing (and saying they're finding the modules as enabled on the system!).

pkg/osbuild/dnf_module_config_stage.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"github.com/osbuild/images/pkg/rpmmd"
+)
+
+type DNFModuleConfig struct {
+	Name     string   `json:"name,omitempty"`
+	Stream   string   `json:"stream,omitempty"`
+	State    string   `json:"state,omitempty"`
+	Profiles []string `json:"profiles"`
+}
```
achilleas-k, line 12: _This is not a comment for this PR directly_

The lack of `omitempty` on `Profiles` lead me to look at the stage schema and I noticed that there are no fields marked `required` there.  The stage itself however accesses properties from the stage options dictionary without using `.get()` but directly with `[]`.
We should fix the stage to use `.get()` for the non-required options, or make them required.  The latter is a backwards-incompatible change, but given the recency of the stage itself, I think that might be okay in this case.

pkg/blueprint/blueprint.go
```diff
@@ -3,11 +3,16 @@ package blueprint
...
-	Modules        []Package       `json:"modules" toml:"modules"`
+	Name        string    `json:"name" toml:"name"`
+	Description string    `json:"description" toml:"description"`
+	Version     string    `json:"version,omitempty" toml:"version,omitempty"`
+	Packages    []Package `json:"packages" toml:"packages"`
+	Modules     []Package `json:"modules" toml:"modules"`
+
+	// Note, this is called "enabled modules" because we already have "modules" except
+	// the "modules" refers to packages and "enabled modules" refers to modularity modules.
+	EnabledModules []EnabledModule `json:"enabled_modules" toml:"enabled_modules"`
```
achilleas-k, line 14: As I understand it, the top-level `modules` was there since the beginning for compatibility with the lorax/weldr blueprint schema.  But since osbuild never had module support, the modules were merged with the package list and treated as plain old packages.

If we weren't doing the big blueprint redesign and unification, I would argue that the current PR is a "bug fix" or perhaps an enhancement for a feature that's had a frontend option wired into a fake backend workaround for the last 5+ years and that we should instead change the top-level `modules` to do what they were originally intended to do.  However, this might be considered a breaking change at this point, so I'm fine with leaving it as it is in the PR, with the expectation that this will all be nice and clean after the BP unification is done.

test/scripts/base-host-check.sh
```diff
@@ -98,6 +98,33 @@ check_ca_cert() {
...
+    # Verify modules that are enabled on a system, if any. Modules can either be enabled separately
+    # or they can be installed through packages directly. We test both cases here.
+    #
+    # Caveat is that when a module is enabled yet _no_ packages are installed from it this breaks.
+    # Let's not do that in the test?
+
+    modules_expected_0=$(jq -rc '.blueprint.enabled_modules[]? | .name + ":" + .stream' "${config}")
+    modules_expected_1=$(jq -rc '.blueprint.packages[]? | select(.name | startswith("@") and contains(":")) | .name' "${config}" | cut -c2-)
+
+    modules_expected="${modules_expected_0}\n${modules_expected_1}"
```
achilleas-k, line 113: I think this isn't working exactly as intended.  I'm seeing the following in the test logs:
- `📗 Module expected: nodejs:18\n` in the [enabled modules](https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/9335396048) test.
- `📗 Module expected: \nnodejs:18` in the [modularity in packages](https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/9335396051) test.

It also makes me wonder if the test will succeed when we have more than one module defined in the packages, or in the enabled_modules, or when we have combinations.

I suppose it's fine for now, the actual feature works, so I wont block, but let's check and fix up the test script soon.

achilleas-k (approved):
I have several comments in-line, but I wont block on any of them.  The blueprint comment is worth discussing here if anyone has thoughts on it.

croissanne (approved):

PR #1198, opened 2025-02-06
imagefilter: improve supported filters discoverability

mvo5:
imagefilter: add `SupportedFilter()` that returns known
 prefixes

This commit adds a new `SupportedFilter()` function that returns
the available filter prefixes.

See https://github.com/osbuild/image-builder-cli/issues/104

---

imagefilter: show supported filters when giving one

This commit tweaks the error message for unsupported filter prefixes
to include what is actually supported.

See https://github.com/osbuild/image-builder-cli/issues/104

### DIFF
--- a/pkg/imagefilter/filter.go
+++ b/pkg/imagefilter/filter.go
@@ -10,6 +10,11 @@ import (
 	"github.com/osbuild/images/pkg/distro"
 )
 
+// SupportedFilters returns what filter prefixes are supported
+func SupportedFilters() []string {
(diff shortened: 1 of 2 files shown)


supakeen (approved):
I'd probably prefer enums maybe but not really a blocker either.

mvo5:
> I'd probably prefer enums maybe but not really a blocker either.

Sure thing, I updated the code with that (and some musings about possible other approaches).

PR #1200, opened 2025-02-07
monitor: increase buffer to deal with very big messages

mvo5:
This commit incrases the buffer os the osbuild message scanner to deal with very big messages. The default size of 64kb is too small for the `dracut` stage which will generate 5447 messages that /dev/log is not available and they come so rapidly that osbuild will pick them up in one big chunk.

This is a short term fix, the deeper fix should be to tweak the osbuild monitor to have a smaller buffer size when sending stdout/stderr output from stages to the monitor. This will fix a crash in the bootc-image-builder anaconda builds that is currently breaking the integration test in konflux.

See also https://github.com/osbuild/osbuild/issues/1976


supakeen (approved):

achilleas-k (approved):

cgwalters:
Wow, that must have been painful to debug...

PR #1201, opened 2025-02-07
osbuild: do not assume osbuild stage messages have a final \n

mvo5:
[draft as this needs coresponding image-builder-cli/bib changs]

This commit tweaks the osbuild monitor code to deal correctly with messages from osbuild stage that may not contain a final newline. So far the code assumed this and stripped the final newline when storing the message in the status. But sometimes there are messages without a final \n (e.g. when messages get very long). So just put the message in verbatim and let the caller do a "print" instead of a "println" when processing the messages (arguably the trim for traces was always strong).


achilleas-k:
What's the status here?

PR #1202, opened 2025-02-07
distro: conditionally include basesystem/filesystem

miabbott:
The `basesystem` package is being obsoleted by `filesystem` in Rawhide (F43). Change the definition of the `iot-commit` to conditionally include the correct package based on the Fedora version.

### DIFF
--- a/pkg/distro/fedora/package_sets.go
+++ b/pkg/distro/fedora/package_sets.go
@@ -80,7 +80,6 @@ func iotCommitPackageSet(t *imageType) rpmmd.PackageSet {
 			"atheros-firmware",
 			"attr",
 			"authselect",
-			"basesystem",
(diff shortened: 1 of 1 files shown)


supakeen (approved):
Yea, I saw the discussion ongoing; let's fix it here. Personally I'd structure this with an else instead of two separate conditionals but not a blocker.

miabbott:
> Yea, I saw the discussion ongoing; let's fix it here. Personally I'd structure this with an else instead of two separate conditionals but not a blocker.

That was my first thought, too, but I didn't see any examples of using an else, so I thought there was some prior style that should be maintained.

Thanks for the review!

achilleas-k (approved):

thozza (approved):

PR #1205, opened 2025-02-10
Extract Fedora package sets into a single YAML file

achilleas-k:
This is a simpler version of #1104 but only does the absolute minimum to pull out package sets into embeddable yaml files.

# Why?

I think the conversation on #1104 stalled because we all (rightly) have too many opinions about every design decision made in that PR.  I hope that with a simpler PR that only makes one or two decisions, we might manage to get an initial version through and start working with it and iterating on it.

# What?

This is not the final version of what the full extraction of image definitions will look like.  This PR only pulls out the package sets for Fedora into a single yaml file.  The file contains the package sets for all image types, keyed by the image type name (with `s/-/_/g`).

# Discuss

With this PR, I think there are two design decisions that we can discuss:
1. the way we handle conditions,
2. the way we handle special version names (like rawhide).

I decided to handle conditions in code without any special syntax in the yaml file, just a special `condition:` block with subkeys representing the three types of conditions we have now: architecture, version upper limit (`version_less_than`), and lower limit (`version_greater_or_equal`).  I wanted to make something that is the smallest change from what we have now, something that maps directly from the existing code.  It's probably not the best solution, but it is the smallest step towards something new and I hope we can all agree that it's a "good first step".

The PR does make some other decisions, like the use of a single file for the whole distribution and the decision to avoid any mechanism for including or merging package sets.
Splitting the package sets into multiple files goes more in the direction of the original PR, which requires making extra decisions about how one could or should split the files.  The same goes for any form of inheritance or inclusion.  These are things we should discuss and design eventually, but I'd like to avoid bringing them up here, so we can focus on one thing at a time.

# Notes

This PR affects only one image type's package sets: the `iot-bootable-container` was not properly appending the arch-specific packages to its set, and that's now fixed.  The extraction of package sets itself, so anything after the second commit, has no effect on any image type's package sets.


mvo5 (approved):
Fwiw, I think this is a nice start, its all internal only so we can tweak the yaml as needed and I think this already makes it more approachable for cases like https://github.com/osbuild/images/pull/1202  where external contributors can just edit yaml now

achilleas-k:
Missed a go.mod update for the yaml import.  Fixed and squashed.

pkg/distro/packagesets/loader.go
```diff
@@ -0,0 +1,95 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+//go:embed */*.yaml
+var Data embed.FS
+
+type packageSet struct {
+	Include   []string   `yaml:"include"`
+	Exclude   []string   `yaml:"exclude"`
+	Condition conditions `yaml:"condition,omitempty"`
```
ondrejbudai, line 21: Shouldn't this be a pointer? I think it's quite common to have optional properties as pointers.

mvo5 (approved):

pkg/distro/packagesets/loader.go
achilleas-k: Done!

mvo5 (dismissed):

supakeen (dismissed):
Nice and manageable, great to start with and expand later. Thanks for pushing this forward.

bcl (approved):
I really like this. I think you've boiled down the previous attempts and discussions to their essential core here.

pkg/distro/packagesets/fedora/package_sets.yaml
```diff
@@ -0,0 +1,902 @@
...
+    - "gobject-introspection"
+    - "plymouth"
+    - "zram-generator-defaults"
+    - "grubby-deprecated"
+    - "extlinux-bootloader"
+
+ova:
+  <<: *vmdk
+
+iot_commit: &iot_commit
```
miabbott, line 79: Could you add a comment here about keeping this list in sync with the package list in pagure over here - https://pagure.io/fedora-iot/ostree/blob/main/f/fedora-iot-base.yaml

We have a comment over there about keeping osbuild/images up-to-date (https://pagure.io/fedora-iot/ostree/blob/main/f/fedora-iot-base.yaml#_76) but never made a similar comment in osbuild/images

mvo5: I wonder how much work it would be to have an integration test that compares the two and fail if they start to diverge. But obviously followup material!

ondrejbudai (approved):
LGTM! :)

@thozza when you have a minute, can you take a look? I would love to know your opinion. :)

pkg/distro/fedora/version.go
```diff
@@ -6,3 +6,11 @@ const VERSION_RAWHIDE = "42"
 // Fedora version 41 and later use a plain squashfs rootfs on the iso instead of
 // compressing an ext4 filesystem.
 const VERSION_ROOTFS_SQUASHFS = "41"
+
+func VersionReplacements() map[string]string {
```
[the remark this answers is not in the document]
achilleas-k: I tend to avoid mutable package globals which could be modified at runtime and cause issues that are difficult to troubleshoot.  Most times it's okay, I don't foresee anyone modifying this map accidentally or otherwise, but I did recently have an issue where a package global struct used for testing was being modified by one test, so the test run order would affect outcomes.  I tend not to think about it too much now and default to making them functions.  Maybe that's a bad pattern, I'm not sure yet.

pkg/distro/packagesets/loader.go
```diff
@@ -16,9 +16,9 @@ import (
 var Data embed.FS
 
 type packageSet struct {
-	Include   []string   `yaml:"include"`
-	Exclude   []string   `yaml:"exclude"`
-	Condition conditions `yaml:"condition,omitempty"`
+	Include   []string    `yaml:"include"`
+	Exclude   []string    `yaml:"exclude"`
+	Condition *conditions `yaml:"condition,omitempty"`
```
thozza, line 21: Nitpick: this change is in commit `172e42b984047a8f8ec2ad40560bc7887c0e1079`, but based on the commit message, it does not belong here.

thozza (approved):
I like it.

I think that using the image type as the key in YAML will and is getting tricky for image types that install packages in multiple pipelines (e.g. some installers). We could consider eventually adding the pipeline names to the YAML pkg set, so that an image type can have pkg sets for all pipelines defined under single key (the image type name). YAML anchors make this quite nice. Anyway, something for the future... :rocket:

ondrejbudai (approved):

mvo5 (approved):

thozza (approved):

PR #1206, opened 2025-02-10
imagefilter: adjust `short` output format

schuellerf:
This is now a short text-only output focused only on a condensed and readable output.

```
centos-10
  ami: aarch64, x86_64
  gce: x86_64
  image-installer: aarch64, x86_64
  oci: x86_64
  ova: x86_64
  qcow2: aarch64, ppc64le, s390x, x86_64
  tar: aarch64, ppc64le, s390x, x86_64
  vhd: aarch64, x86_64
  vmdk: x86_64
  wsl: aarch64, x86_64
fedora-40
  ami: aarch64, x86_64
  container: aarch64, ppc64le, riscv64, s390x, x86_64
  image-installer: aarch64, x86_64
  iot-bootable-container: aarch64, ppc64le, s390x, x86_64
  iot-commit: aarch64, x86_64
  iot-container: aarch64, x86_64
…
```

There is an additional suggestion to implement a proper `yaml` output: #1268

### DIFF
--- a/pkg/imagefilter/formatter.go
+++ b/pkg/imagefilter/formatter.go
@@ -99,12 +99,6 @@ type textShortResultsFormatter struct{}
 func (*textShortResultsFormatter) Output(w io.Writer, all []Result) error {
 	var errs []error
 
-	// deliberately break the yaml until the feature is stable, there
-	// are open questions, e.g. how this relates to:
-	// https://github.com/osbuild/osbuild-composer/pull/4336
-	// which adds a similar but slightly different API
-	fmt.Fprint(w, "@WARNING - the output format is not stable yet and may change\n")
-
 	outputMap := make(map[string]map[string][]string)
 	for _, res := range all {
 		if _, ok := outputMap[res.Distro.Name()]; !ok {
@@ -133,10 +127,10 @@ func (*textShortResultsFormatter) Output(w io.Writer, all []Result) error {
 		for _, t := range types {
 			arches := outputMap[distro][t]
 			sort.Strings(arches)
-			typeArchPairs = append(typeArchPairs, fmt.Sprintf("%s: [ %s ]", t, strings.Join(arches, ", ")))
+			typeArchPairs = append(typeArchPairs, fmt.Sprintf("%s: %s", t, strings.Join(arches, ", ")))
 		}
 
-		if _, err := fmt.Fprintf(w, "%s:\n  %s\n", distro, strings.Join(typeArchPairs, "\n  ")); err != nil {
+		if _, err := fmt.Fprintf(w, "%s\n  %s\n", distro, strings.Join(typeArchPairs, "\n  ")); err != nil {
 			errs = append(errs, err)
 		}
 	}
--- a/pkg/imagefilter/formatter_test.go
+++ b/pkg/imagefilter/formatter_test.go
@@ -74,7 +74,7 @@ func TestResultsFormatter(t *testing.T) {
 		{
 			"short",
 			[]string{"test-distro-1:qcow2:test_arch3"},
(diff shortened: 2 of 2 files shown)


supakeen (commented):
I'm starting to feel very uncomfortable about string formatting all this YAML.

supakeen:
As a side question, is the stability goal here to be equal to the structures as output by `osbuild-composer`'s CloudAPI?

supakeen:
@schuellerf The YAML dependency will likely be there soon due to: https://github.com/osbuild/images/pull/1205 :)

pkg/imagefilter/formatter.go
```diff
@@ -99,18 +99,12 @@ type textShortResultsFormatter struct{}
 func (*textShortResultsFormatter) Output(w io.Writer, all []Result) error {
 	var errs []error
 
-	// deliberately break the yaml until the feature is stable, there
-	// are open questions, e.g. how this relates to:
-	// https://github.com/osbuild/osbuild-composer/pull/4336
-	// which adds a similar but slightly different API
-	fmt.Fprint(w, "@WARNING - the output format is not stable yet and may change\n")
```
mvo5, line 106: I would like to keep this for now, I think we need to consider what API we provide and grantee

mvo5 (changes requested):

pkg/imagefilter/formatter.go
schuellerf: so I'll think about another code-comment as it's aligned with the 4336-PR…

schuellerf: …and it doesn't state it's yaml, just happens to be compatible for now ;)

schuellerf: It doesn't say a warning on the cloud-api though (and it's essentially the same output)
Should I transition to yaml library and remove the warning or make it really short and not making it yaml at all?
@supakeen / @mvo5

supakeen: So, with #1205 landing there's no reason anymore not to pull in a YAML dependency as it's already here.

My preference is that if the output format is outputting YAML that the format should be called `yaml` as well, same goes for the JSON format we have.

That means we can either, remove the YAML-ness in this PR and name it `short` or we can make it go full-YAML and name it `yaml`. No real preference on either.

schuellerf: just pushes me towards making both :-D 
I'm fine with that ;)

mvo5: I am sorry that I missed this earlier but I think APis should be simple and orthogonal, I don't see the use-case for adding yaml in addition to json (happy to reconsider if there is one). 

As I said in https://github.com/osbuild/images/pull/1268#pullrequestreview-2645542052 I think "short" is very interesting (and making it yaml is nice as it is still very readable) because it can be a nice entry point (the current default is not very user friendly with  the wall of text that is the current default of list-images). But I think "short" should optimize for "niceness" (and IMHO we could sacrifice yaml for that even).

supakeen: There isn't a reason to have both. What I said in the previous comment:

- if the output is supposed to be YAML, use a YAML library
- if the output is not supposed to be YAML then let's make sure it can't be misconstrued as such

Not to introduce both.

Previously we were pretending that the `short` output format was YAML hence the warning up top to make sure YAML parsers don't understand it. That's also why I made my previous remarks about using a YAML library.

supakeen (approved):
This is fine now that it isn't acting like it's wanting to be compatible with YAML.

thozza (approved):
FWIW, this looks OK to me. I don't have any strong preference, but not making this look like YAML if it is not makes sense :+1:

mvo5 (approved):
Thank you! I think this is the right compromise!

PR #1209, opened 2025-02-11
go.mod: fix envoyproxy/go-control-plane/envoy version

schuellerf:
`github.com/envoyproxy/go-control-plane/envoy@v1.32.3`
has a changed checksum, so we'll just bump to `v1.32.4` to avoid problems.


thozza (approved):
Thanks.

supakeen (approved):

PR #1210, opened 2025-02-11
cloud/awscloud: extend UploaderOptions with boot mode

thozza:
Allow setting the boot mode to be used for the uploaded image, instead of defaulting to uefi-preferred. The boot mode is not a required field, so if not explicitly set in the UploaderOptions, don't set it at all.

Extend test coverage for the newly added code.

### DIFF
--- a/pkg/cloud/awscloud/uploader.go
+++ b/pkg/cloud/awscloud/uploader.go
@@ -11,8 +11,10 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/google/uuid"
 
+	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/cloud"
+	"github.com/osbuild/images/pkg/platform"
 )
 
 type awsUploader struct {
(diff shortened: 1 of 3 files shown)


pkg/cloud/awscloud/uploader.go
```diff
@@ -22,10 +24,30 @@ type awsUploader struct {
 	bucketName string
 	imageName  string
 	targetArch string
+	bootMode   *string
```
mvo5, line 27: Is seems that bootMode in platform is never an empty string so I wonder if we could simply use `string` here and `""` instead of a pointer (not a fan of pointers). But I guess it would require quite a few more change(?). I could look into it for a followup maybe? (if desired)

mvo5 (approved):
Thank you! This is very nice. One tiny question inline

pkg/cloud/awscloud/uploader.go
achilleas-k: You're right, it should be fine.
Also, the 0 value for the internal enum `BOOT_NONE` is `"none"`, though that's probably not entirely relevant here.

achilleas-k (approved):

pkg/cloud/awscloud/uploader.go
thozza: Yeah, I'm considering either using `BOOT_NONE` or introducing `BOOT_UNSET` as the `0` value. I still need to check the potential impact.

PR #1216, opened 2025-02-13
Generalise and clean up blueprint customization validation for image types (HMS-6871)

achilleas-k:
Year-old branch that I never finished.  I remembered it today during a discussion with @supakeen, @mvo5, and @schuellerf.  Opening as draft because I don't want it to disappear or forget about it again.


pkg/blueprint/customizations.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+
+	"golang.org/x/exp/slices"
```
lzap, line 8: Part of Go 1.21, at least the used `Reverse` [function](https://pkg.go.dev/slices#Reverse).

achilleas-k: As the title mentions, this is a year-old branch :laughing: 
But thanks, I'll try to get back to this soon and rebase.  It's going to need a lot of work.

mvo5:
We will need to keep this open for reference as it will become relevant for the YAMLification and move to a "generic" distro everywhere.

lzap:
For the record, I have a ticket: https://issues.redhat.com/browse/HMS-8676

achilleas-k:
The validation functionality is ready now.  I need to finish this now by adding the customization lists to the image types.

lzap:
It looks good, tho, I am wondering why this patch was mentioned as a blocker for UBP. Or have I [commented](https://github.com/osbuild/images/pull/1216#issuecomment-3000024677) on a wrong PR?

achilleas-k:
> I am wondering why this patch was mentioned as a blocker for UBP

Is it?  I think we talked about it being relevant, but not a blocker.

bcl (changes requested):
I think this is pretty nice. customizations.partitioning_mode and customizations.sshkey are missing -- but they are not supported by the images blueprint so that's a separate problem.

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -1884,46 +1852,69 @@ image_types:
...
+            - "shadow-utils"  # includes passwd
             - "sudo"
+            - "systemd"
+            - "tar"
             - "traceroute"
             - "util-linux"
-            - "shadow-utils"  # includes passwd
+            - "xfsprogs"
+            - "xz"
           conditions:
```
lzap, line 2027: Not sure how this hunk is related to the PR title, but okay.

lzap (dismissed):

pkg/distro/validator.go
```diff
@@ -0,0 +1,183 @@
...
+				cerr := err.(*blueprint.CustomizationError)
+				cerr.RevPath = append(cerr.RevPath, key)
+				return cerr
+			}
+		}
+	}
+	return nil
+}
+
+func ValidateConfig(t ImageType, bp blueprint.Blueprint, _ ImageOptions) error {
```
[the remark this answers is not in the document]
achilleas-k: I completely forgot about this comment.  Perhaps we can simplify in a followup.  It'd be nice to do a wider breakdown of our larger interfaces in general, right?

pkg/distro/defs/fedora/distro.yaml
lzap: No hard feelings and I usually do not care as long as commits are separate, I was just curious. Feel free to keep it in!

pkg/blueprint/blueprint.go
```diff
@@ -96,3 +102,15 @@ func (p Package) ToNameVersion() string {
 func (p EnabledModule) ToNameStream() string {
 	return p.Name + ":" + p.Stream
 }
+
+type BlueprintError struct {
+	// Reverse path to the customization that caused the error.
+	RevPath []string
```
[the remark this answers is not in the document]
achilleas-k: A dedicated type for the RevPath sounds a bit too "overengineery" to me.  But perhaps making it a private field and adding an `Append()` (or `AddPath()`?) method to `BlueprintError` might be nice.

achilleas-k:
Responding to [this comment thread](https://github.com/osbuild/images/pull/1216#discussion_r2255439925) here for better visibility.

> All of these need to also have `name`, `version`, and `description` added, otherwise the most basic blueprint is rejected.

I added the support in a separate commit (I can squash it with the others if people feel it's cleaner) but, more importantly, I added descriptions to all test blueprints under `test/configs/`.  The description explains the purpose of each test configuration, which I think is a useful thing to have.

data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -532,6 +532,91 @@
       aarch64:
         <<: *iot_simplified_installer_partition_tables_x86
 
+  supported_options_lists:
+    # common options supported by all disk image types this includes everything
+    # that is not specific to installers or ostree-based images
+    supported_options_disk: &supported_options_disk
+      - "name"
```
mvo5, line 539: (wondering) Should we make name/version/description implicit? They are not really concepts that are relevant for the image types, i.e. it would be weird to have an image type that refuses a blueprint with a "name"?

Similar wondering about "distro", it goes away in UBP and its a bit of a odd one, so maybe the image types should not list it?

pkg/distro/validator_test.go
```diff
@@ -0,0 +1,1196 @@
...
+	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/osbuild/images/pkg/disk"
+	"github.com/osbuild/images/pkg/distro"
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/images/pkg/platform"
+	"github.com/osbuild/images/pkg/rpmmd"
+	"github.com/stretchr/testify/assert"
+)
+
+type TestImageType struct {
```
mvo5, line 17: (nitpick) I would love if we had only one TestImageType, ideally from `test_distro` and with exported fields so that each test that needs an ImageType can just import and tweak it. But I understand its probably a bit cumbersome right now with the existing `test_distro`

test/configs/all-customizations.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "all-customizations",
   "blueprint": {
+    "name": "all-customizations",
+    "description": "Blueprint that uses (almost) all available customizations for the disk image types",
```
mvo5, line 5: ❤️

mvo5 (commented):
I really like how clean this is and how straightfoward the yaml is. I did not yet do a full in-depth review but have a few questions/ideas inline from my first read.

data/distrodefs/fedora/imagetypes.yaml
bcl: I think that hiding it makes it slightly more difficult to get up to speed with it. It's supposed to be a list of blueprint fields, just make it 1:1 without any magic.

bcl (approved):
Looks good!

lzap (dismissed):

thozza:
Out of curiosity, why didn't you consider defining the supported customizations the other way around? IOW for the image type to define the customizations that it does not support? Do you think that it would make sense to also provide this approach in the future?

pkg/distro/validator_test.go
achilleas-k: Can we move this to a follow-up?  We could also solve this by breaking down the image type interface, which I would like to do at some point anyway.

bcl:
> @bcl What would the supported options list look like for the net installer?

Right now the only things supported are FIPS and language.

pkg/distro/validator.go
```diff
@@ -106,12 +106,139 @@ func validateSupportedConfig(supported []string, conf reflect.Value) *validation
...
+	//    that the requiredMap also lists keys that have required subparts,
+	//    whether they are wholly required or not.
+	//
+	// 2. subMap contains the keys that have sub-parts that are required. Each
+	//    substructure will have to be checked recursively until we reach
+	//    required leaf nodes.
+
+	requiredMap := make(map[string]bool)
+	subMap := make(map[string][]string)
+	for _, key := range required {
```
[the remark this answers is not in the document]
mvo5: Yeah, there is no obvious way (I was thinking an iterator and callback and the callback would be specific to supported/required) - but then I'm absolutely overthinking this (sorry for that)

mvo5 (dismissed):
Thank you!

pkg/distro/validator.go
```diff
@@ -106,12 +106,139 @@ func validateSupportedConfig(supported []string, conf reflect.Value) *validation
...
+			return &validationError{message: err.Error(), revPath: []string{key}}
+		}
+		switch value.Kind() {
+		case reflect.Ptr, reflect.Struct, reflect.String, reflect.Slice:
+			// Required should only be used for Pointer, String, and Slice types.
+			// For other types, the zero value can be valid and not indicate a
+			// missing value.
+			if value.IsZero() {
+				return &validationError{message: "required", revPath: []string{key}}
+			}
```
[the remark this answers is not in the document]
mvo5: (note for other reviewers, this got also added)

supakeen (dismissed):

mvo5 (approved):

data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -752,6 +752,91 @@
...
+      - "customizations.firewall"
+      - "customizations.user"
+      - "customizations.group"
+      - "customizations.hostname"
+      - "customizations.kernel"
+      - "customizations.locale"
+      - "customizations.openscap"
+      - "customizations.repositories"
+      - "customizations.rpm"
+      - "customizations.services"
```
bcl, line 783: This is missing `customizations.sshkey` and `customizations.partition_mode` which I stumbled into by accident when trying a blueprint I had laying around from something else, not through any thorough examination so there may be others :/

bcl (changes requested):
I wish there was an automated way to be sure all the correct options are supported. If we miss one then people's existing blueprints will break.

data/distrodefs/fedora/imagetypes.yaml
supakeen: The former only exists in `osbuild-composer` and (AFAIK) gets translated to a `customizations.user` we mention it as [deprecated](https://osbuild.org/docs/user-guide/blueprint-reference/#ssh-keys) in the documentation though I've gotten bitten by it before as well.

The latter is indeed missing and [documented to exist](https://osbuild.org/docs/user-guide/blueprint-reference/#partitioning-mode).

achilleas-k:
Updated:
- Rebased.
- Fixed conflict.
- Added `customizations.sshkey` to all supported lists that also include `customizations.user`.
- Added `customizations.partitioning_mode` to all supported lists that also include `customizations.filesystem`.

mvo5 (approved):

supakeen (approved):
Thank you.

mvo5:
Its nice, having this will help with https://github.com/osbuild/images/pull/1786 - i.e. we can selectively allow kernel options for the installer types that currently do not support them even if the underlying code is all unified.

PR #1217, opened 2025-02-14
osbuild: set the cache-max-size to 20 G when checkpoints are defined

achilleas-k:
Adding --checkpoints to the osbuild call has no effect right now unless a previous run of osbuild set the max size already. Since we support passing checkpoint args to the osbuild call, let's also set the cache size to a baseline default value to have those checkpoints actually checkpoint.

### DIFF
--- a/pkg/osbuild/osbuild-exec.go
+++ b/pkg/osbuild/osbuild-exec.go
@@ -8,6 +8,8 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+
+	"github.com/osbuild/images/pkg/datasizes"
 )
 
 // Run an instance of osbuild, returning a parsed osbuild.Result.
@@ -32,6 +34,11 @@ func RunOSBuild(manifest []byte, store, outputDirectory string, exports, checkpo
 
 	for _, checkpoint := range checkpoints {
 		cmd.Args = append(cmd.Args, "--checkpoint", checkpoint)
+	}
+
+	if len(checkpoints) > 0 {
+		// set the cache-max-size to a reasonable size that the checkpoints actually get stored
+		cmd.Args = append(cmd.Args, "--cache-max-size", fmt.Sprint(20*datasizes.GiB))
 	}
 
 	if result {


thozza (dismissed):
Makes sense :+1:

supakeen (commented):
I think we use `10 GiB` in `image-builder`, should we do the same here? Should we also default to the same cache location?

achilleas-k:
> I think we use `10 GiB` in `image-builder`, should we do the same here?

How do we set it in osbuild?  Afaict, we use this function to call `osbuild` and it doesn't expose the cache size option.
I have [a follow-up PR ready that exposes it](https://github.com/achilleas-k/images/tree/osbuild/cloptions), but I wanted to do it separately because it's going to change public function signatures and we can hold that one back until the change is also ready in osbuild-composer, ib-cli, and bib.

> Should we also default to the same cache location?

The cache location is configurable (`store`) so callers can set whatever they want.  Omitting it uses osbuild's default `.store`.

achilleas-k:
> > Should we also default to the same cache location?
> 
> The cache location is configurable (`store`) so callers can set whatever they want. Omitting it uses osbuild's default `.store`.

Correction: omitting it by setting it to "" puts everything in the working directory.  We could probably make that an error.

supakeen:
> > I think we use `10 GiB` in `image-builder`, should we do the same here?
> 
> How do we set it in osbuild? Afaict, we use this function (`osbuild.RunOSBuild()`) to call `osbuild` and it doesn't expose the cache size option. I have [a follow-up PR ready that exposes it](https://github.com/achilleas-k/images/tree/osbuild/cloptions), but I wanted to do it separately because it's going to change public function signatures and we can hold that one back until the change is also ready in osbuild-composer, ib-cli, and bib.

My bad, I thought https://github.com/osbuild/bootc-image-builder/pull/812 had landed (and it was 20 GiB instead).

> > Should we also default to the same cache location?
> 
> The cache location is configurable (`store`) so callers can set whatever they want. Omitting it uses osbuild's default `.store`.

Right, the question is do we want to make that default location a global cache (in the same place as where `image-builder` puts it), perhaps as a followup so the cache gives benefit by default even when building in different directories (at least for sources).

achilleas-k:
yeah, a system-wide cache would be much nicer

supakeen:
Ok, let's discuss system-wide separately; could we enlarge the cache to either 10, or 20 GiB max? The cache for a single `minimal-raw` with `--checkpoint="*"` is already 21 GiB.

achilleas-k:
Bumped it to 20 G.

supakeen (approved):
Needs linter appeasement. We might also want to set defaults in `osbuild` as well?

achilleas-k:
> Needs linter appeasement. We might also want to set defaults in `osbuild` as well?

You mean, have a default size in osbuild itself?

thozza (approved):

PR #1218, opened 2025-02-14
[RFC] blueprints: (optinally) include blueprint repos in 2nd depsolve

mvo5:
This commit allows to (optionally) include the custom repositories from the blueprints as part of the depsolving when installing the user packages.

Its a quick proof of concept as this question came up. Thanks to Ondrej for raising it.


achilleas-k:
> The amount of duplication in between rhel/fedora in the pkg/distro is also something I'm not happy about but it seems that is the current design but I would love to get back to this and see we can do something about it.

Indeed, the `ImageType.Manifest()` function is probably identical for all implementations of the interface (in each distro).  In cases where it's not, it's probably by accident, or because a specific distro doesn't support something (like RHEL 7), but that latter issue could be handled differently.

achilleas-k:
> I'm a bit unsure about the best way to go about a full end-to-end test

Indeed, it's a bit tricky.  The obvious thing to do is to add a configuration with a repository in the blueprint to `test/configs/`, but it would have to be a repository that's not used already.  The build tests all use the repositories from `test/data/repositories/` and there's no straightforward way to drop one.

We could do something a bit "silly".  If you look at the test repository configurations, some of them have these [image_type_tags](https://github.com/osbuild/images/blob/f80b2c8023b4d799295de03420cefe129e5005ed/test/data/repositories/rhel-9.6.json#L207-L210), which limits the use of a repository to just those image types that are listed.  This is done so that we don't use, for example, `google-cloud-sdk` repositories when building an image for any other platform.  So one thing we could do is add a build config (blueprint) that adds the `google-cloud-sdk` repository to a qcow2 build (or any other build, really) and also selects one of the `google-cloud-sdk` packages for installation.  One issue with this though is that the configuration wont be updated automatically when our repository configurations are updated after new snapshots are created.

thozza:
> > The amount of duplication in between rhel/fedora in the pkg/distro is also something I'm not happy about but it seems that is the current design but I would love to get back to this and see we can do something about it.
> 
> Indeed, the `ImageType.Manifest()` function is probably identical for all implementations of the interface (in each distro). In cases where it's not, it's probably by accident, or because a specific distro doesn't support something (like RHEL 7), but that latter issue could be handled differently.

+1

It used to be much worse - duplication for every RHEL major version :sweat_smile: 

We can unify Fedora and RHEL. The code can be written to support the super-set of all features. It does not matter if there are bits specific to RHEL-7, that Fedora does not use.

thozza:
> > I'm a bit unsure about the best way to go about a full end-to-end test
> 
> Indeed, it's a bit tricky. The obvious thing to do is to add a configuration with a repository in the blueprint to `test/configs/`, but it would have to be a repository that's not used already. The build tests all use the repositories from `test/data/repositories/` and there's no straightforward way to drop one.

I propose snapshotting and using the EPEL repository for this. We never install packages from EPEL. An alternative would be our own COPR repository with a dummy package. Both are pretty straightforward. EPEL repo would be usable only for a specific version of RHEL major version, so COPR is probably the most universal approach.

pkg/blueprint/repository_customizations.go
```diff
@@ -76,6 +80,17 @@ func (rc *RepositoryCustomization) getFilename() string {
 	return rc.Filename
 }
 
+func RepoCustomizationsFilterEnabledDuringBuild(repos []RepositoryCustomization) []RepositoryCustomization {
```
thozza, line 83: Would it make sense to return `([]rpmmd.RepoConfig, error)` from the function (basically just calling `.customRepoToRepoConfig()` on the repo)? Calling `RepoCustomizationsToRepoConfigAndGPGKeyFiles()` seems to do much more than just conversion and the resulting `[]*fsnode.File` is always discarded for these repos.

thozza (commented):
Nice and simple. I've added two (or really just one) suggestions...

pkg/blueprint/repository_customizations.go
mvo5: Thanks, that is a nice one - much simpler this way

pkg/blueprint/repository_customizations.go
```diff
@@ -25,6 +25,10 @@ type RepositoryCustomization struct {
 	SSLVerify      *bool    `json:"sslverify,omitempty" toml:"sslverify,omitempty"`
 	ModuleHotfixes *bool    `json:"module_hotfixes,omitempty" toml:"module_hotfixes,omitempty"`
 	Filename       string   `json:"filename,omitempty" toml:"filename,omitempty"`
+
+	// When set the repository will be used during the depsolve of
+	// payload repositories
+	EnabledDuringBuild bool `json:"enabled_during_build" toml:"enabled_during_build"`
```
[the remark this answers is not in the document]
mvo5: I have no strong opinion here, I like your suggested and used `install_from` now as it is shorter but I'm open for more ideas (I had a hard time picking a good name)

thozza (approved):
LGTM, thanks :partying_face:

achilleas-k (approved):

Issue #1222, opened 2025-02-16
Add a `rawhide` version for Fedora


supakeen:
In Fedora we explicitly list versions in both the repository files, and code. Let's have a `fedora-rawhide` repository and have our version compare functions treat `rawhide` as `inf`?

This would allow to keep functionality on `rawhide` no matter the branching status of Fedora (currently we often have to bump version numbers if changes didn't land) and to build `rawhide` without using a custom repository.

supakeen:
I've decided to add Fedora 43 first in #1232, #1224 needs a lot more work to actually work with the "rawhide" name as well.

PR #1224, opened 2025-02-17
data: fedora rawhide repositories

supakeen:
DRAFT for #1222.


supakeen:
This is very hard to do quickly; I'll revisit all of this some other time because what I wrote on the weekend is not enough.

PR #1227, opened 2025-02-17
manifest: add Files,Directories to bp customizations

mvo5:
[draft: becaue missing tests]


ondrejbudai:
We should probably do this before selinux labelling, right? This is how the os pipeline does it IIRC.

achilleas-k:
> We should probably do this before selinux labelling, right? This is how the os pipeline does it IIRC.

of course.  good catch!

pkg/manifest/raw_bootc.go
```diff
@@ -199,7 +204,18 @@ func (p *RawBootcImage) serialize() osbuild.Pipeline {
...
+	// First create custom directories, because some of the custom files may depend on them
+	if len(p.Directories) > 0 {
+		pipeline.AddStages(osbuild.GenDirectoryNodesStages(p.Directories)...)
+	}
+
+	if len(p.Files) > 0 {
+		pipeline.AddStages(osbuild.GenFileNodesStages(p.Files)...)
+	}
 
+	if len(p.Users) > 0 || len(p.Files) > 0 || len(p.Directories) > 0 {
```
achilleas-k, line 218: Why did this become conditional?  We're always writing `/etc/fstab`, so that should probably be properly labelled every time.

achilleas-k (changes requested):

pkg/manifest/raw_bootc.go
mvo5: Thanks, good catch! I updated the PR.

achilleas-k (approved):

PR #1229, opened 2025-02-17
many: add basic riscv architecture support, start with container image

mvo5:
This PR starts adding support for the riscv architecture. Initially just the container images.

Huge thanks to Ondrej for many super helpful hints/links/code examples.

I feel uneasy to mark this ready without an integration test but its tricky to do, I tested this via https://github.com/osbuild/image-builder-cli/pull/136 but that is not nearly ready. Ondrej suggested just a simple "generate manifest for riscv64" - that is probably a good compormise.

I think we want something like pr#136 (just a lot nicer) anyway soon.


achilleas-k:
Looks good.  Tests need updating though.

I'd like to change the title of the PR though.  I realise it's saying we're adding a new build target, a RISC-V container, but I think the fact that we're adding a new architecture for the first time in forever is a bit more significant and it would help to have a PR title that makes that clear.

I was thinking something like: "Add new architecture riscv64 with a container image type for Fedora 42"

ondrejbudai:
This is awesome! 😍 Can you also add fedora-41? Did you actually manage to test this?

ondrejbudai (approved):
I built an image using a manifest generated by this PR, fed it into the latest osbuild on a riscv64 machine, and I was able to successfully test the container using `podman load`, and `podman run`. 🥳 

It took about 20 minutes (emulated qemu).

PR #1233, opened 2025-02-18
fedora: image type names follow variants

supakeen:
Fedora consists of many variants. Historically we haven't really based our own definitions on anything. Let's introduce variant name prefixes to all Fedora image types to indicate what variant they are based on.

The next PR will then make these variants similar to (or equal to) the Fedora variants as they are supposed to be.

We keep the old image type names around as aliases since changing them is a breaking change. We should however consider deprecating them and emitting a warning for some time. Perhaps a followup.


supakeen:
I'll assume this needs discussion and followups :)

supakeen:
The config map commit needs more changes as the image types are now different between various distros they need to be split out; which means duplicating the files since those are the paths.

achilleas-k:
> which means duplicating the files since those are the paths.

Not necessary.  The image type names don't need to all be valid for each distro, so for example you can have
```diff
diff --git a/test/config-map.json b/test/config-map.json
index e8add2b20..4d548d7b0 100644
--- a/test/config-map.json
+++ b/test/config-map.json
@@ -135,6 +135,7 @@
       "gce",
       "gce-rhui",
       "image-installer",
+      "minimal-installer",
       "iot-bootable-container",
       "iot-container",
       "live-installer",
@@ -264,7 +265,8 @@
       "x86_64"
     ],
     "image-types": [
-      "image-installer"
+      "image-installer",
+      "minimal-installer"
     ]
   },
   "./configs/unattended-iso.json": {
```

and it will work.  Just ass the new names in the same lists where the old names were.

supakeen:
> and it will work. Just add the new names in the same lists where the old names were.

Ah that's good, I'll do that :)

supakeen:
> 
> This is simply a notice. It will not block this PR from being merged.

Names will need to be updated in `osbuild-composer` after this merges as well. It can *probably* request the old names but it reports back the *new* names.

supakeen:
@achilleas-k I might be missing something but the Schutzbot CI fails on:

```
❌ 2 distro/arch/image combinations have no builds in the cache
   1: fedora-41/x86_64/server-wsl
   2: fedora-40/x86_64/server-wsl
```

I think I have `server-wsl` listed under `./configs/empty.json`?

supakeen:
Still open because of:

> @achilleas-k I might be missing something but the Schutzbot CI fails on:
> 
> ```
> ❌ 2 distro/arch/image combinations have no builds in the cache
>    1: fedora-41/x86_64/server-wsl
>    2: fedora-40/x86_64/server-wsl
> ```
> 
> I think I have `server-wsl` listed under `./configs/empty.json`?

supakeen:
Let's start with just the name changes; package sets and other parity work can be done as followups.

supakeen:
This now fails only on `minimal-installer` (the previous `image-installer`) with:

```
⏱  Duration: 0s
org.osbuild.lorax-script: 8791972f76103c310d2e946e16dfe07472e54536dac1e550ba423c8f52a8a6e5 {
  "path": "99-generic/runtime-postinstall.tmpl",
  "basearch": "x86_64",
  "product": {
    "name": "",
    "version": ""
  }
}
running script: 99-generic
move '/run/osbuild/tree/usr/share/anaconda/list-harddrives-stub' -> '/run/osbuild/tree/usr/bin/list-harddrives'
Error on line: 0 ['move', 'usr/share/anaconda/list-harddrives-stub', 'usr/bin/list-harddrives']
  File "/run/osbuild/bin/org.osbuild.lorax-script", line 45, in <module>
    r = main(stage_args["tree"],
             stage_args["options"])
  File "/run/osbuild/bin/org.osbuild.lorax-script", line 40, in main
    script()
    ~~~~~~^^
  File "/run/osbuild/lib/osbuild/util/lorax.py", line 80, in __call__
    method(self, *args)
    ~~~~~~^^^^^^^^^^^^^
  File "/run/osbuild/lib/osbuild/util/lorax.py", line 116, in move
    os.rename(src, dst)
    ~~~~~~~~~^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/run/osbuild/tree/usr/share/anaconda/list-harddrives-stub' -> '/run/osbuild/tree/usr/bin/list-harddrives'
⏱  Duration: 0s
```

Seems like a file doesn't exist, are we missing a package?

supakeen:
> This now fails only on `minimal-installer` (the previous `image-installer`) with:
> 
> ```
> ⏱  Duration: 0s
> org.osbuild.lorax-script: 8791972f76103c310d2e946e16dfe07472e54536dac1e550ba423c8f52a8a6e5 {
>   "path": "99-generic/runtime-postinstall.tmpl",
>   "basearch": "x86_64",
>   "product": {
>     "name": "",
>     "version": ""
>   }
> }
> Failed to open file "/sys/fs/selinux/checkreqprot": Read-only file system
> running script: 99-generic
> move '/run/osbuild/tree/usr/share/anaconda/list-harddrives-stub' -> '/run/osbuild/tree/usr/bin/list-harddrives'
> Error on line: 0 ['move', 'usr/share/anaconda/list-harddrives-stub', 'usr/bin/list-harddrives']
> Traceback (most recent call last):
>   File "/run/osbuild/bin/org.osbuild.lorax-script", line 45, in <module>
>     r = main(stage_args["tree"],
>              stage_args["options"])
>   File "/run/osbuild/bin/org.osbuild.lorax-script", line 40, in main
>     script()
>     ~~~~~~^^
>   File "/run/osbuild/lib/osbuild/util/lorax.py", line 80, in __call__
>     method(self, *args)
>     ~~~~~~^^^^^^^^^^^^^
>   File "/run/osbuild/lib/osbuild/util/lorax.py", line 116, in move
>     os.rename(src, dst)
>     ~~~~~~~~~^^^^^^^^^^
> FileNotFoundError: [Errno 2] No such file or directory: '/run/osbuild/tree/usr/share/anaconda/list-harddrives-stub' -> '/run/osbuild/tree/usr/bin/list-harddrives'
> ⏱  Duration: 0s
> ```
> 
> Seems like a file doesn't exist, are we missing a package? The file is owned by `anaconda-core` so somewhere in this PR it seems to have dropped (or perhaps a package dependency or weakdep changed?).

Figured this out; it was my mistake :)

supakeen:
Seems like a lot of checksums change that I didn't expect, I'll take a look.

achilleas-k:
Rebased to run builds.  If no other build is running, this should rebuild the Fedora WSL images and pass the coverage check.

achilleas-k (dismissed):

supakeen:
> Rebased to run builds. If no other build is running, this should rebuild the Fedora WSL images and pass the coverage check.

The manual removal of `wsl` images still lead to a no-build `server-wsl`. Since this is causing such a ruckus in this PR we've chatted and will drop the rename of `wsl` to `server-wsl` instead so it can be submitted as a separate PR where we can also tackle the coverage testing.

The reason is that there is already a manifest with the same checksum, hence the `server-wsl` image isn't built, which then causes the coverage test to fail as it sees no manifest built for it.

achilleas-k (approved):
Thanks for this and sorry for the messiness.  I hope we can figure out a nicer way to deal with aliases and renames soon.

PR #1239, opened 2025-02-20
fedora: switch to erofs for fedora >= 43

supakeen:
See the change proposal [1]. This switches our the Fedora IoT Anaconda installer, the Fedora image-installer, and our current live installer implementation over to use EroFS as per the accepted change proposal [2].

[1]: https://fedoraproject.org/wiki/Changes/EROFSforLiveMedia
[2]: https://pagure.io/fesco/issue/3356

---

Note: this change applies to Fedora 43 since I do not want to land this change too late in the cycle in case it causes problems as it's not strictly necessary, just nice.


supakeen:
Needs @pcdubs's approval.

bcl (dismissed):
Looks good!

achilleas-k (commented):
LGTM.

The new code is repeated 3 times and could be a local function in the file.

I'm not approving since @supakeen wants @pcdubs to sign off on this first.

supakeen:
> LGTM.
> 
> The new code is repeated 3 times and could be a local function in the file.
> 
> I'm not approving since @supakeen wants @pcdubs to sign off on this first.

Thanks, I was prefering to keep locality here ("each image type is its own self-contained function") though I'm now second guessing that.

bcl:
> Thanks, I was prefering to keep locality here ("each image type is its own self-contained function") though I'm now second guessing that.

Don't second guess yourself, it makes it easier to read :)

supakeen:
Re-targeted to Fedora 43 as beta freeze has started and it seems like a bad plan to do this between beta and final :)

bcl (approved):
Ack

ondrejbudai (approved):

pkg/distro/fedora/images.go
```diff
@@ -406,8 +406,14 @@ func liveInstallerImage(workload workload.Workload,
...
 
-	if common.VersionGreaterThanOrEqual(img.OSVersion, VERSION_ROOTFS_SQUASHFS) {
+	if common.VersionGreaterThanOrEqual(img.OSVersion, VERSION_ROOTFS_EROFS) {
+		img.RootfsCompression = "lzma"
+		img.RootfsType = manifest.ErofsRootfs
+	} else if common.VersionGreaterThanOrEqual(img.OSVersion, VERSION_ROOTFS_SQUASHFS) {
+		img.RootfsCompression = "lz4"
 		img.RootfsType = manifest.SquashfsRootfs
+	} else {
+		img.RootfsCompression = "lz4"
```
thozza, line 416: For the future: it may be good to extract this to the `InstallerConfig` eventually - https://github.com/osbuild/images/blob/245e0ed49174657eff739b52c35723d9010f8a7b/pkg/distro/installer_config.go#L5-L9

thozza (approved):

supakeen:
I can't really tell on what the CI is failing, we end up with:

> Job's log exceeded limit of 4194304 bytes.
> Job execution will continue but no more output will be collected.

Because it seems `erofs` outputs a line of text for each file that's being compressed?

supakeen:
Closing this PR since it's totally out of line now with all the yaml-ification. I might revisit it later.

PR #1241, opened 2025-02-20
many: add minimal raw for riscv64 

mvo5:
This PR adds support for the minimal-raw image type for riscv64.

I tested thia via a test image is generated with https://github.com/osbuild/images/compare/main...mvo5:images:riscv-minimal-raw-with-hacks which contains cross-build support, i.e.
```
IMAGES_EXPERIMENTAL=buildroot=ghcr.io/mvo5/fedora-buildroot:41 image-builder manifest --arch=riscv64 minimal-raw --distro=fedora-41
```
The buildroot container comes from https://github.com/mvo5/buildroot-containers/


pkg/distro/fedora/imagetype.go
```diff
@@ -145,7 +145,7 @@ func (t *imageType) getPartitionTable(
 ) (*disk.PartitionTable, error) {
 	basePartitionTable, exists := t.basePartitionTables[t.arch.Name()]
 	if !exists {
-		return nil, fmt.Errorf("unknown arch: %s", t.arch.Name())
+		return nil, fmt.Errorf("unknown arch for partiion table: %s", t.arch.Name())
```
supakeen, line 148: parti*t*ion.

pkg/platform/riscv64.go
```diff
@@ -15,6 +16,18 @@ func (p *RISCV64) GetArch() arch.Arch {
 func (p *RISCV64) GetPackages() []string {
 	packages := p.BasePlatform.FirmwarePackages
 
+	if p.UEFIVendor != "" {
+		packages = append(packages,
+			// XXX: this is needed to get a generic bootkernel,
+			// this should probably be part of any bootable img
+			// packagelist
+			"dracut-config-generic",
```
supakeen, line 24: Are we sure about this one?

pkg/manifest/os.go
```diff
@@ -418,6 +418,14 @@ func (p *OS) serialize() osbuild.Pipeline {
 		pipeline.AddStage(osbuild.NewOSTreePasswdStage("org.osbuild.source", p.ostreeParentSpec.Checksum))
 	}
 
+	// XXX: this is added 4 times in the manifest for unknown reasons
+	// XXX2: must be *before* the rpm stage as dracut runs as part of
+	// the rpm stage (and osbuild.dract.conf needs to be able to deal
+	// with that
+	for _, dracutConfConfig := range p.DracutConf {
+		pipeline.AddStage(osbuild.NewDracutConfStage(dracutConfConfig))
+	}
```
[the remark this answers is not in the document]
mvo5: I removed this one again, it turned out I was holding it wrong (well, sort of at least). It turned out that without `dracut-config-generic` (see below) dracut would only add the virtio_blk module when running inside a qemu env (as it conditionally inspect the host without this package). Eventually we might want to get back to this though for the reasons you outline.

pkg/platform/riscv64.go
mvo5: Sort of sure, here is my reasoning:
1. I got the hint from @ondrejbudai (thanks for this!). 
2. We include it in all other platforms too so I think this part is okay. I also think this is not the best place AFAICT this is not arch specific, we have it on all arches.
3. The description of `dnf info dracut-config-generic` says: `This package provides the configuration to turn off the host specific initramfs generation with dracut and generates a generic image by default`

But maybe I'm missing something or misunderstanding the question? (or is this about the comment that is not clear enough?)

supakeen: You're not missing anything the misunderstanding was on my side!

supakeen (dismissed):

ondrejbudai (approved):
Love this!

I manually tested this using qemu, and it works like a charm.

mvo5:
> Oh no, there's a conflict now. :(

Resolved now.

ondrejbudai (approved):
Thanks for the quick resolution!

supakeen (approved):

Issue #1244, opened 2025-02-23
Duplicate rescue kernels


supakeen:
Due to the way we handle `/etc/machine-id` in our images any image that has `dracut-config-rescue` will generate rescue kernel by the name of `vmlinuz-0-rescue-ffffffff[repeat]` (and `initramfs`) in `XBOOTLDR`.

When the system boots a new machine id is provisioned, the next time `kernel-install` runs and triggers `dracut` a new rescue kernel will be generated with `vmlinuz-0-rescue-anactualidhere` (and `initramfs`).

Since these together fill up about 150-200 MiB and 300-400 MiB when duplicated this can and sometimes does cause trouble for systems that try to keep 3 kernels and have a 1 GiB `/boot` partition.

supakeen:
We have turned this off for Fedora Minimal now in #1245. However the issue remains for other images that *if* `dracut-config-rescue` is in their package set the above applies.

Looking at other build tools shows that they generally exclude the package on anything that's a disk image (installers tend to handle this case).

It's probably worthwhile to actually chase this rabbit down into Fedora by dropping the machine id from the BLS config(s) and kernel name(s). This can be done by fiddling with `entry-token` so let's test that and then *maybe* we can suggest it to distros.

PR #1245, opened 2025-02-23
fedora: remove rescue kernels from minimal

supakeen:
After some discussion [1] Fedora Minimal would like to disable the generation of rescue kernels. This is because historically they weren't included in images (before switching to `image-builder`) and also because the current `/etc/machine-id` implementation of `image-builder` leads to duplicate rescue kernels being generated [2].

---

This also slightly improves build time (by about 10-20 seconds on my system).

[1]: https://github.com/fedora-minimal/distribution-minimal/issues/10
[2]: https://github.com/osbuild/images/issues/1244

### DIFF
--- a/pkg/distro/packagesets/fedora/package_sets.yaml
+++ b/pkg/distro/packagesets/fedora/package_sets.yaml
@@ -828,7 +828,11 @@ minimal_raw:
     - "brcmfmac-firmware"
     - "realtek-firmware"
     - "iwlwifi-mvm-firmware"
-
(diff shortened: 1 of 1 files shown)


supakeen:
Note that the handling of rescue kernels is likely a larger problem for images we create that have 1 GiB boot partitions; see: https://github.com/osbuild/images/issues/1244

mvo5 (approved):

schuellerf (approved):

PR #1247, opened 2025-02-24
manifest: `zstd`

supakeen:
Implements the existing `zstd` stage from [osbuild](https://github.com/osbuild/osbuild) and creates the appropriate scaffolding to use it.

Also enables a new image type for Fedora Minimal that uses it [1].

[1]: https://github.com/fedora-minimal/distribution-minimal/issues/12


pkg/distro/fedora/distro.go
```diff
@@ -454,6 +454,36 @@ var (
 		basePartitionTables:    minimalrawPartitionTables,
 		requiredPartitionSizes: requiredDirectorySizes,
 	}
+
+	minimalrawZstdImgType = imageType{
+		name:        "minimal-raw-zst",
```
mvo5, line 459: (wondering) should this be `minimal-raw-zstd` (notice the leading `d`) ? I know the extension if `zst` (which I personally find very confusing) so maybe fine? idk

achilleas-k:
Conflict :(

pkg/distro/fedora/distro.go
```diff
@@ -454,6 +454,36 @@ var (
...
+		filename:    "disk.raw.zst",
+		compression: "zstd",
+		mimeType:    "application/zstd",
+		packageSets: map[string]packageSetFunc{
+			osPkgsKey: packageSetLoader,
+		},
+		defaultImageConfig: &distro.ImageConfig{
+			EnabledServices: minimalRawServices,
+			// NOTE: temporary workaround for a bug in initial-setup that
```
mvo5, line 468: This duplicates a lot of the existing minimal-raw with only minimal changes - including this workaround which means we will (if we ever remove it) keep the two in sync. How about just making a "maker" function to keep them in sync automatically? Something like:
```go
diff --git a/pkg/distro/fedora/distro.go b/pkg/distro/fedora/distro.go
index 5959db425..9921a24e4 100644
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -424,42 +424,16 @@ var (
                exports:                []string{"container"},
                requiredPartitionSizes: requiredDirectorySizes,
        }
+       minimalrawImgType     = makeMinimalRaw("minimal-raw", "xz", "xz")
+       minimalrawZstdImgType = makeMinimalRaw("minimal-raw-zst", "zstd", "zst")
+)
 
-       minimalrawImgType = imageType{
-               name:        "minimal-raw",
-               filename:    "disk.raw.xz",
-               compression: "xz",
-               mimeType:    "application/xz",
-               packageSets: map[string]packageSetFunc{
-                       osPkgsKey: packageSetLoader,
-               },
-               defaultImageConfig: &distro.ImageConfig{
-                       EnabledServices: minimalRawServices,
-                       // NOTE: temporary workaround for a bug in initial-setup that
-                       // requires a kickstart file in the root directory.
-                       Files: []*fsnode.File{initialSetupKickstart()},
-                       Grub2Config: &osbuild.GRUB2Config{
-                               // Overwrite the default Grub2 timeout value.
-                               Timeout: 5,
-                       },
-               },
-               rpmOstree:              false,
-               kernelOptions:          defaultKernelOptions,
-               bootable:               true,
-               defaultSize:            2 * datasizes.GibiByte,
-               image:                  diskImage,
-               buildPipelines:         []string{"build"},
-               payloadPipelines:       []string{"os", "image", "xz"},
-               exports:                []string{"xz"},
-               basePartitionTables:    minimalrawPartitionTables,
-               requiredPartitionSizes: requiredDirectorySizes,
-       }
-
-       minimalrawZstdImgType = imageType{
-               name:        "minimal-raw-zst",
-               filename:    "disk.raw.zst",
-               compression: "zstd",
-               mimeType:    "application/zstd",
+func makeMinimalRaw(name, compression, ext string) imageType {
+       return imageType{
+               name:        name,
+               filename:    "disk.raw." + ext,
+               compression: compression,
+               mimeType:    "application/" + compression,
                packageSets: map[string]packageSetFunc{
                        osPkgsKey: packageSetLoader,
                },
@@ -479,12 +453,12 @@ var (
                defaultSize:            2 * datasizes.GibiByte,
                image:                  diskImage,
                buildPipelines:         []string{"build"},
-               payloadPipelines:       []string{"os", "image", "zstd"},
-               exports:                []string{"zstd"},
+               payloadPipelines:       []string{"os", "image", compression},
+               exports:                []string{compression},
                basePartitionTables:    minimalrawPartitionTables,
                requiredPartitionSizes: requiredDirectorySizes,
        }
-)
+}
 
 type distribution struct {
        name               string

```

mvo5 (approved):
This looks fine, some suggestions inline (could be followups but the duplication is something that might be worth doing now).

pkg/distro/fedora/distro.go
supakeen: I indeed picked it based on the extension so the image type is related to the output artifact, I did things in the same style in #1233.

pkg/osbuild/zstd_stage.go
```diff
@@ -0,0 +1,40 @@
...
+
+func NewZstdStageInputs(references FilesInputRef) *ZstdStageInputs {
+	return &ZstdStageInputs{
+		File: NewFilesInput(references),
+	}
+}
+
+// Compresses a file into a zstd archive.
+func NewZstdStage(options *ZstdStageOptions, inputs *ZstdStageInputs) *Stage {
+	var stageInputs Inputs
```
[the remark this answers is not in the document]
supakeen: I'll skip it for now and do this as a potential followup that includes the `xz` stage.

pkg/manifest/zstd.go
```diff
@@ -0,0 +1,55 @@
...
+	return p.filename
+}
+
+func (p *Zstd) SetFilename(filename string) {
+	p.filename = filename
+}
+
+// NewZstd creates a new Zstd pipeline. imgPipeline is the pipeline producing the
+// raw image that will be zstd compressed.
+func NewZstd(buildPipeline Build, imgPipeline FilePipeline) *Zstd {
```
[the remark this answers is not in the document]
supakeen: I'll apply this.

pkg/distro/fedora/distro.go
mvo5: Good point, yeah, the `newDistro()` way seems fine then.

supakeen: Added a commit to do this.

pkg/manifest/zstd.go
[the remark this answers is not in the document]
supakeen: Added a commit with this test.

supakeen:
Rebased, resolved conflicts, addressed comments.

pkg/distro/fedora/distro.go
```diff
@@ -726,6 +696,14 @@ func newDistro(version int) distro.Distro {
 	}
 	vhdImgType.defaultImageConfig = vhdConfig.InheritFrom(qcow2ImgType.defaultImageConfig)
 
+	minimalrawZstdImgType := minimalrawImgType
```
ondrejbudai, line 699: It's interesting that the image type name is "minimal-raw-zst", but the variable name is "minimalRawZst**d**ImgType". My gut feeling is that the d shouldn't be there.

ondrejbudai (approved):
Awesome, thanks! :) I still think that the final compression should be just configurable per build (with a sane default per image type), but this is a step in the correct direction.

How would you feel about adding a `minimal-raw-zstd` alias for `minimal-raw-zst`? I think it may remove some friction...

PR #1248, opened 2025-02-24
lib: add new `experimentalflags` package

mvo5:
This commit adds a new `experimentalflags` package that can be used to enable experimental features in the image-builder ecosystem.

The initial implementation provides a `Bool()` function that allows to do e.g.:
```
IMAGE_BUILDER_EXPERIMENTAL=skip-arch-checks
IMAGE_BUILDER_EXPERIMENTAL=skip-arch-checks=true
```
to disable architecture mismatch checks. This will be used as part of the `image-builder-cli` but the mechanism is generally useful.


supakeen (approved):
💟

achilleas-k (changes requested):
Blocking for discussion:

How is this going to be used exactly?  I'm asking because currently this is a convenience function for reading env vars from a predefined key.  That's fine, no issues there, but I wonder if this is the right place to do it.

I'm thinking of this from the perspective of the library itself and what it enables.  The way I see it, there are two scenarios:
1. A client of the library wants to implement an experimental feature that the library itself already supports.
2. A client wants to implement an experimental feature that the library *doesn't* support.

The first one, I think, should be up to the client itself.  It just needs to pass the right options to the right function calls and it can do whatever it needs to guard against that use (or not).  But it makes me question whether the library itself needs to implement functionality for deciding whether experimental features should be enabled or not.  The downside of not providing one is code duplication, though reading an env var isn't really that much.  Also, there's the common env key, but this PR doesn't really expose the env key itself, just the reading of a `Bool()` from the environment.

The second needs more consideration.  If, for example, we want to enable experimental features in manifest generation, then we'd need to change the `Manifest()` signature, or one of its configuration arguments to take experimental options.  This would enable any caller to use the experimental feature, but it doesn't tie them to the env var in any way.

I think my main issue here is that, the way I see it, experimental features that don't depend on changing anything in osbuild/images are features of a frontend client of the library and should be handled as such, possibly even with different clients implementing different keys / var names (e.g. `BOOTC_IMAGE_BUILDER_EXPERIMENTAL="someoption=somevalue"`).  Experimental features that do require changing things in the library will need to be implemented in the library under appropriate options marked experimental (with the implication that they can be removed or changed without notice).  But nothing about that change would require the client to make an `experimentalflags.Bool()` call first.  It could be a command line `--experimental="someoption=somevalue"` flag, a build flag, or even a completely separate binary.

Put another way, how would you document this PR in the API / godoc?  The commit message reads
```
This commit adds a new `experimentalflags` package that can be used to enable
experimental features in the image-builder ecosystem.
```
but that's not entirely true, is it?  It's more accurate to say:
```
This commit adds a way of reading an env var which clients should use to
determine if experimental features should be enabled.
```
and the documentation would be something like
```
// Package experimentalflags provides functionality for reading options defined
// in an environment variable named IMAGE_BUILDER_EXPERIMENTAL. These functions
// should be used to determine, in a common way, if experimental features should
// be enabled when using the libarary.
```

mvo5:
> Blocking for discussion:
> 
> How is this going to be used exactly? I'm asking because currently this is a convenience function for reading env vars from a predefined key. That's fine, no issues there, but I wonder if this is the right place to do it.
[..]

Answering this first, the current idea for usage is in the follow code
* https://github.com/osbuild/image-builder-cli/pull/136/files#diff-a15366436dafd5b402932194d7c809db3f2c060440bee74f69e7d009e965d3dfR154
* https://github.com/osbuild/images/commit/ed3dfe466aa6237d112660a60887a52ea44dc577

I.e. the library/frontend will just ask if the given experimentalflag is set (with the bool convenience) helper. Its deliberately lightweight/low-key and uses the same (env) namespace to allow things like:
```
IMAGE_BUILDER_EXPERIMENTAL=skip-arch-checks,buildroot=ghcr.io/mvo5/fedora-buildroot:41 image-builder-cli --arch=riscv64 build...
``` 
instead of 
```
IMAGE_BUILDER_EXPERIMENTAL=skip-arch-checks IMAGES_LIBRARY_EXPERIMENTAL=buildroot=ghcr.io/mvo5/fedora-buildroot:41 image-builder-cli --arch=riscv64 build...
```
but of course  you raise important points, I will reply in more detail in a bit.

supakeen:
I'll add my use case first as well: I want to conditionally inject stages in pipelines based on environment variables. For example; to reset the machine-id. While still building images from some frontend application.

I'd like to do this during experimentation without having to plumb through options through the client *and* library :)

achilleas-k (approved):

PR #1249, opened 2025-02-24
repositories: add `f4{1,2}` base repos for riscv64

mvo5:
This commit adds the riscv64 base repository to the repositories f41,f42 repositories (in addition to the staging ones).

I think this changed recently on the fedora-riscv side, from full to incremental repos. But I could be wrong, the whole repo setup is very dynamic (and I expect more changes soon).

No test (sadly) because we still need to figure out the testing situation inside images for cross-arch builds like riscv


supakeen (approved):

data/repositories/fedora-42.json
```diff
@@ -60,6 +60,12 @@
     }
   ],
   "riscv64": [
+    {
+      "name": "fedora",
+      "baseurl": "http://fedora.riscv.rocks/repos-dist/f42/latest/riscv64",
```
ochosi, line 65: This one gives me a 404 :thinking:

mvo5: Yeah, this one is "aspirational", i.e. I expect it to appear soon(ish) but I can remove it again and only add it once the f42 repos appear.

PR #1250, opened 2025-02-24
[RFC] manifest: use experimentalflags.String("buildroot")

mvo5:
This commit allows to force a container based buildroot as part of images when setting:
```console
$ IMAGE_BUILDER_EXPERIMENTAL=buildroot=ghcr.io/mvo5/fedora-buildroot:41 \
   image-builder manifest --arch=riscv64 minimal-raw --distro=fedora-41
```
and it will replace the buildroot with the given container.

This will allow to do cross-arch building for riscv. The generated manifest can be build with osbuild (as long as qemu-user is installed) and will generate a minimal-raw for riscv64.

Cross-building is useful because qemu-user is a lot faster (and easier to setup) than full riscv system emulation. A minimal raw build on a really fast AWS machine takes about 100 minutes. A full minimal-raw build on my (moderate) PC takes about 20min - which is fast enough so that we can run this in GH workflows as part of the CI.

Long term I would love to see this generalized and that we would (optionally) support container based buildroots for most/all(?) our distro/arch combos. This could be done without too much maintenance burden on our side via e.g. just pointing to the official
 `ubi` containers for rhel that we then install extra rpms into
(like the mkfs.* tools). This would then allow cross-arch image building for e.g. a raspberry pi from a normal PC. But obviously this is all long term stuff that needs more design/discussion (I'm just mentioning it to make clear that this is not a dead-end but has more potential).

This would enable a relatively clean version of https://github.com/osbuild/image-builder-cli/pull/136/files#diff-dd961e42853d4345e2cf6a29f188f80386ed6b583056d7078a38c0c099c97c26R99, i.e. a (relatively) nice integration test that confirms riscv is building and also "image-builder build --arch=riscv64 ..." support (with the right experimental options set)

### DIFF
--- a/pkg/manifest/build.go
+++ b/pkg/manifest/build.go
@@ -3,7 +3,9 @@ package manifest
 import (
 	"fmt"
 
+	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/container"
+	"github.com/osbuild/images/pkg/experimentalflags"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/rpmmd"
 	"github.com/osbuild/images/pkg/runner"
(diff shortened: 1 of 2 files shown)


mvo5:
Closing in favor of the cleaner and nicer https://github.com/osbuild/images/pull/1263

PR #1251, opened 2025-02-24
fedora: reset machine-id for disk images

supakeen:
Fedora resets the `/etc/machine-id` file for [any disk image](https://pagure.io/fedora-kiwi-descriptions/blob/960270bb861a0e54bc837f7feeb9c3a9b9a18661/f/config.sh#_117); let's do the same here.

Enables `ConditionFirstBoot` which is/was also part of a [request](https://github.com/fedora-minimal/distribution-minimal/issues/3) for Fedora Minimal to support `systemd-firstboot`.

### DIFF
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -345,6 +345,9 @@ func diskImage(workload workload.Workload,
 
 	img.Filename = t.Filename()
 
+	d := t.arch.distro
+	img.FirstBoot = common.VersionGreaterThanOrEqual(d.osVersion, VERSION_FIRSTBOOT)
+
 	return img, nil
 }
 
--- a/pkg/manifest/os_test.go
+++ b/pkg/manifest/os_test.go
@@ -230,3 +230,21 @@ func testTomlPkgsFor(t *testing.T, os *OS) {
 		assert.Contains(t, buildPkgs, tc.expectedTomlPkg)
 	}
 }
+
+func TestFirstBootIncludesMachineIdStage(t *testing.T) {
+	os := NewTestOS()
+
+	os.FirstBoot = true
+
+	pipeline := os.serialize()
+	st := findStage("org.osbuild.machine-id", pipeline.Stages)
+	require.NotNil(t, st)
+}
+
+func TestFirstBootDoesNotIncludeMachineIdStage(t *testing.T) {
+	os := NewTestOS()
+
+	pipeline := os.serialize()
+	st := findStage("org.osbuild.machine-id", pipeline.Stages)
+	require.Nil(t, st)
+}
--- a/pkg/osbuild/machine_id_stage.go
+++ b/pkg/osbuild/machine_id_stage.go
@@ -1,0 +1,22 @@
+package osbuild
+
+type MachineIdStageOptions struct {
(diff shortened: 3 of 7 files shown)


pkg/osbuild/machine_id_stage.go
```diff
@@ -0,0 +1,22 @@
+package osbuild
+
+type MachineIdStageOptions struct {
+	// Determines the state of `/etc/machine-id`, valid values are
+	// `yes` (reset to `uninitialized`), `no` (empty), `preserve` (keep).
+	FirstBoot string `json:"first-boot"`
+}
```
achilleas-k, line 7: A set of constants for the stage would be useful here, but not necessary.

achilleas-k (approved):
Nice.  LGTM.

Small note inline, but wont block.

Similar to my note about removing the rescue kernels (that I asked on the matrix channel): let's find out if we can make this change for current Fedora versions as well.

pkg/manifest/os_test.go
```diff
@@ -230,3 +230,21 @@ func testTomlPkgsFor(t *testing.T, os *OS) {
 		assert.Contains(t, buildPkgs, tc.expectedTomlPkg)
 	}
 }
+
+func TestFirstBootIncludesMachineIdStage(t *testing.T) {
```
mvo5, line 234: Thanks for the tests \o/

(nitpick) maybe a single test with a loop to test the two cases? but this if fine too of course
```go
func TestFirstBootIncludesMachineIdStage(t *testing.T) {
	os := NewTestOS()

	for _, firstBoot := range []bool{false, true} {
		os.FirstBoot = firstBoot
		pipeline := os.serialize()
		st := findStage("org.osbuild.machine-id", pipeline.Stages)
		if firstBoot {
			require.NotNil(t, st)
		} else {
			require.Nil(t, st)
		}
	}
}
````

File: pkg/osbuild/machine_id_stage.go (hunk 1 above)
Comment (mvo5) [reply pkg/osbuild/machine_id_stage.go:7]: Fwiw, I agree, having a type and const (like e.g. in qemu_stage.go and the QEMUFormat here) would be slightly nicer as it would give us (more) type safety.

File: pkg/distro/fedora/images.go [hunk 3]
```diff
@@ -345,6 +345,9 @@ func diskImage(workload workload.Workload,
 
 	img.Filename = t.Filename()
 
+	d := t.arch.distro
+	img.FirstBoot = common.VersionGreaterThanOrEqual(d.osVersion, VERSION_FIRSTBOOT)
```
mvo5, line 349: I wish we had an easy way to unit test this as well but we don't currently so its fine.

pkg/osbuild/machine_id_stage_test.go
```diff
@@ -0,0 +1,30 @@
+package osbuild
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewMachineIdStageOptions(t *testing.T) {
```
mvo5, line 9: Thanks for the tests (and the other ones!)

mvo5 (approved):
Thank you! This is very nice. I think its fine to go in as is, I would love to see const/a custom type for the fistboot options but it could be a followup.

PR #1252, opened 2025-02-24
fedora: allow disabling weak dependencies

supakeen:
Weak dependency resolution is hardcoded per imagetype. This commit changes it so weak dependency resolution can be turned off per image.

We guard on the Fedora version to ensure we don't accidentally change older images.

This commit also disables weak dependency resolution for Fedora Minimal which I have mentioned [here](https://github.com/fedora-minimal/distribution-minimal/issues/13).

### DIFF
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -110,6 +110,10 @@ func osCustomizations(
 		osc.Hostname = "localhost.localdomain"
 	}
 
+	if imageConfig.InstallWeakDeps != nil {
+		osc.InstallWeakDeps = *imageConfig.InstallWeakDeps
+	}
+
 	timezone, ntpServers := c.GetTimezoneSettings()
 	if timezone != nil {
 		osc.Timezone = *timezone
@@ -334,7 +338,7 @@ func diskImage(workload workload.Workload,
 	img.Compression = t.compression
 	if bp.Minimal {
 		// Disable weak dependencies if the 'minimal' option is enabled
-		img.InstallWeakDeps = common.ToPtr(false)
+		img.OSCustomizations.InstallWeakDeps = false
 	}
 	// TODO: move generation into LiveImage
 	pt, err := t.getPartitionTable(bp.Customizations, options, rng)
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -455,6 +455,7 @@ func mkMinimalRawImgType(d distribution) imageType {
 				// Overwrite the default Grub2 timeout value.
 				Timeout: 5,
 			},
+			InstallWeakDeps: common.ToPtr(common.VersionLessThan(d.osVersion, VERSION_MINIMAL_WEAKDEPS)),
 		},
 		rpmOstree:              false,
 		kernelOptions:          defaultKernelOptions,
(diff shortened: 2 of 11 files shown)


schuellerf:
The inverted logic is necessary for "backwards compatibility" or just to map the default behavior better? I.e. disableWeakDeps=false is actually weakDeps=true

supakeen:
> The inverted logic is necessary for "backwards compatibility" or just to map the default behavior better? I.e. disableWeakDeps=false is actually weakDeps=true

I did it because the `imageType` struct has no constructor and the default value for a boolean is `false`. It also maps pretty well on the behavior I'd like; fetch weak deps by default but explicitly disallow them per-image :)

supakeen:
I'll be moving this to draft so it can be rewritten to use image config(s) as per @thozza / @achilleas-k 's request.

supakeen:
Undrafting again. I gave it some thought and I'd prefer to merge as-is unless they are strong objections.

Moving this to image configs in a *nice* way requires a bunch of thought and it'd be nice to build images more quickly in the shorter term, especially with risc-v cross compilation.

supakeen:
> 
> This is simply a notice. It will not block this PR from being merged.

Huh, why? :)

pkg/distro/fedora/images.go
```diff
@@ -329,12 +329,21 @@ func diskImage(workload workload.Workload,
...
+
 	if bp.Minimal {
 		// Disable weak dependencies if the 'minimal' option is enabled
 		img.InstallWeakDeps = common.ToPtr(false)
+	} else {
+		// We only support disabling weak dependencies starting at Fedora 43, this is to
+		// ensure that we don't break older images when/if this toggle is added to an image.
+		if common.VersionGreaterThanOrEqual(d.osVersion, VERSION_WEAKDEPS) && t.disableWeakDeps {
+			img.InstallWeakDeps = common.ToPtr(false)
+		}
```
thozza, line 346: I don't like these conditionals in the `ImageKind` generator function. Especially since this will become a problem once we consolidate the implementation across fedora and rhel.

I'd suggest to move this to the `ImageConfig` and instead set it in the image type's default `ImageConfig`.

thozza (commented):
I'd suggest to move the option to `ImageConfig` and have the logic to not install weak deps in the image type generator function (which does not exist yet in Fedora).

pkg/distro/fedora/images.go
supakeen: I've adjusted this PR to move `InstallWeakDeps` into the `ImageConfig` instead.

supakeen:
So, I'm a bit confused here @thozza this PR *seems* like it really shouldn't touch RHEL yet we now have a missing package (or script) on RHEL which makes it seem that `crypto-policies-scripts` is no longer installed.

The same happens for Fedora. I'm not entirely sure because the *default* is still the same and I'm only setting this to false on minimal?

supakeen:
I've adjusted the code to *always* take `InstallWeakDeps` from `ImageConfig`. The only failures left now are RHEL 8.9, and RHEL 8.10 in Schutzbot.

supakeen:
Cool, all is good in this PR now as far as I can see :)

achilleas-k (approved):

achilleas-k:
Thanks again.  This got a bit messy last week but glad to see it's been sorted out :)
Nice work

thozza (approved):
LGTM, thanks!

PR #1257, opened 2025-02-25
Repos: add RHEL 10.1 and 9.7 repositories

thozza:
SSIA

/jira-epic COMPOSER-2435


mvo5 (commented):
Quick drive-by meta comment - should this be two ~PRs~commits? one for the gitlab-ci stuff and one that actually adds the json files to repositores data?

[edit: sorry, meant to say commits here]

bcl:
Do you mean two PRs or two commits? :) I'd agree with two commits.

supakeen (approved):
This is fine to me as a single commit; splitting them doesn't add much, you'd get "add repos", "use repos".

thozza:
> Quick drive-by meta comment - should this be two PRs? one for the gitlab-ci stuff and one that actually adds the json files to repositores data?

source prep will fail if one adds repos without regenerating .gitlab-ci.yml...

mvo5:
> > Quick drive-by meta comment - should this be two PRs? one for the gitlab-ci stuff and one that actually adds the json files to repositores data?
> 
> source prep will fail if one adds repos without regenerating .gitlab-ci.yml...

Thank you for splitting. I like that the auto-generated stuff is in its own commit. I also like it when the commit message explains what was run to auto-generate the things (so that a simple git log or git blame will help understand how the chanages came into existance) but AIUI this is already explained in the test/README.md already so probably no need to put here.

mvo5 (approved):

PR #1258, opened 2025-02-25
 many: move machine id first boot to image config

supakeen:
This moves the machine id handling from specific to Fedora to the image config.

---

Had a chat with @achilleas-k and @thozza today who pointed out that they'd prefer this to live in the image config. Is this the right approach?

### DIFF
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -134,8 +134,9 @@ func mkIotCommitImgType(d distribution) imageType {
 			osPkgsKey: packageSetLoader,
 		},
 		defaultImageConfig: &distro.ImageConfig{
-			EnabledServices: iotServicesForVersion(&d),
-			DracutConf:      []*osbuild.DracutConfStageOptions{osbuild.FIPSDracutConfStageOptions},
+			EnabledServices:        iotServicesForVersion(&d),
+			DracutConf:             []*osbuild.DracutConfStageOptions{osbuild.FIPSDracutConfStageOptions},
+			MachineIdUninitialized: common.ToPtr(false),
 		},
 		rpmOstree:              true,
 		image:                  iotCommitImage,
@@ -153,6 +154,9 @@ func mkIotBootableContainer(d distribution) imageType {
 		mimeType: "application/x-tar",
 		packageSets: map[string]packageSetFunc{
 			osPkgsKey: packageSetLoader,
+		},
+		defaultImageConfig: &distro.ImageConfig{
+			MachineIdUninitialized: common.ToPtr(false),
 		},
 		rpmOstree:              true,
 		image:                  bootableContainerImage,
@@ -176,8 +180,9 @@ func mkIotOCIImgType(d distribution) imageType {
 			},
 		},
 		defaultImageConfig: &distro.ImageConfig{
-			EnabledServices: iotServicesForVersion(&d),
-			DracutConf:      []*osbuild.DracutConfStageOptions{osbuild.FIPSDracutConfStageOptions},
+			EnabledServices:        iotServicesForVersion(&d),
+			DracutConf:             []*osbuild.DracutConfStageOptions{osbuild.FIPSDracutConfStageOptions},
+			MachineIdUninitialized: common.ToPtr(false),
 		},
 		rpmOstree:              true,
 		bootISO:                false,
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -488,6 +488,7 @@ var defaultDistroImageConfig = &distro.ImageConfig{
 	Locale:                 common.ToPtr("C.UTF-8"),
 	DefaultOSCAPDatastream: common.ToPtr(oscap.DefaultFedoraDatastream()),
 	InstallWeakDeps:        common.ToPtr(true),
+	MachineIdUninitialized: common.ToPtr(true),
 }
 
 func defaultDistroInstallerConfig(d *distribution) *distro.InstallerConfig {
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -238,6 +238,10 @@ func osCustomizations(
 		osc.CACerts = ca.PEMCerts
 	}
 
+	if imageConfig.MachineIdUninitialized != nil {
+		osc.MachineIdUninitialized = *imageConfig.MachineIdUninitialized
+	}
+
 	return osc, nil
 }
 
@@ -348,9 +352,6 @@ func diskImage(workload workload.Workload,
 	img.PartitionTable = pt
 
 	img.Filename = t.Filename()
-
-	d := t.arch.distro
-	img.FirstBoot = common.VersionGreaterThanOrEqual(d.osVersion, VERSION_FIRSTBOOT)
 
 	return img, nil
 }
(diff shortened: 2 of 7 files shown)


pkg/distro/fedora/distro.go
```diff
@@ -473,6 +473,7 @@ var defaultDistroImageConfig = &distro.ImageConfig{
 	Timezone:               common.ToPtr("UTC"),
 	Locale:                 common.ToPtr("en_US"),
 	DefaultOSCAPDatastream: common.ToPtr(oscap.DefaultFedoraDatastream()),
+	FirstBootStrategy:      common.ToPtr(osbuild.MachineIdFirstBootYes),
```
thozza, line 476: You lost the conditional use of it `common.VersionGreaterThanOrEqual(d.osVersion, VERSION_FIRSTBOOT)`, but I'm not sure if it was intentional.

pkg/distro/image_config.go
```diff
@@ -94,6 +94,9 @@ type ImageConfig struct {
 	LockRootUser *bool
 
 	IgnitionPlatform *string
+
+	// How to handle the /etc/machine-id file
+	FirstBootStrategy *osbuild.MachineIdFirstBoot
```
thozza, line 99: In general, we want to avoid using `osbuild` stage options or types as types in `ImageConfig`.

I'm also not too keen on the name. `FirstBootStrategy` sounds way too general, while all it does is manage `machine-id`... :thinking:

thozza (commented):
Thank you!

Yes, this is the correct approach from my point of view, but I've added a few comments.

pkg/distro/fedora/distro.go
supakeen: It was; we considered it a bug (mostly because it was too much work to also turn `defaultDistroConfigs` in Fedora into functions).

pkg/distro/image_config.go
supakeen: `MachineIdStrategy`?

pkg/distro/fedora/distro.go
supakeen: Now that I look at this; it isn't too much work to change. Do you feel strongly to reintroduce the version guard? :)

thozza: I don't feel strongly about it. I was just curious...

pkg/distro/image_config.go
thozza: `MachineIdFirstBoot`? :roll_eyes:  I like your suggestion, but the values make it weird.

I'll let others comment since there isn't a good option.

thozza: > Yes, but the status quo right now is to use them. Do I need to introduce another type and then directly convert/map that onto osbuild types?

It is, but we want to move away from it. So, it seems natural not to use it for newly added options. So yes, ideally, you should add a new type to https://github.com/osbuild/images/tree/main/pkg/customizations if it can't be represented by a primitive type.

On the other hand, we do not need to necessarily map the stage options 1:1 to the `ImageConfig` option. Maybe adding just something like `MachineIdUninitialized bool` would be good enough? And the `OS` pipeline would just set the correct stage option if `true`.

achilleas-k: > On the other hand, we do not need to necessarily map the stage options 1:1 to the `ImageConfig` option. Maybe adding just something like `MachineIdUninitialized bool` would be good enough? And the `OS` pipeline would just set the correct stage option if `true`.

This should cover everything we're doing with the stage for now and we can expand on it later if we need to.

pkg/distro/image_config.go
```diff
@@ -94,6 +94,9 @@ type ImageConfig struct {
 	LockRootUser *bool
 
 	IgnitionPlatform *string
+
+	// How to handle the /etc/machine-id file
+	MachineIdUninitialized *bool
```
achilleas-k, line 99: It'd be useful to expand on this.  Since the option is a boolean "how to handle ..." doesn't say much.  Let's clarify what happens when it's enabled, like "When enabled, sets the machine ID to `uninitialized`, which causes first boot operations to run."
(or something like that)

achilleas-k (approved):
LGTM!

Small docstring comment but wont block.

achilleas-k (approved):

supakeen:
Several of the IoT images are failing due to:

```
error: Finalizing rootfs: During kernel processing: invalid machine ID 'uninitialized
'
```

I've addressed this in a new commit (I hope). And also addressed @achilleas-k's comment on the comment(s) since I was force pushing anyhow.

supakeen:
This seems to finally be OK but I think it rebuilt images that I didn't expect to be rebuilt?

pkg/distro/fedora/images.go
```diff
@@ -234,6 +234,10 @@ func osCustomizations(
 		osc.CACerts = ca.PEMCerts
 	}
 
+	if imageConfig.MachineIdUninitialized != nil {
+		osc.MachineIdUninitialized = *imageConfig.MachineIdUninitialized
```
mvo5, line 238: Do we need `mageConfig.MachineIdUninitialized` to be a pointer? Given that `osc.MachineIdUninitialized ` is a bool and always defaults to false it seems `imageConfig.MachineIdUninitialized` could also just be a bool (not a pointer)? This would also resolve the confusion of having a var with the same name in one place that is a bool and another that is a *bool

supakeen: I don't know. I kept the consistent idiom of pointer-to in `ImageConfig` and direct in `OSCustomizations` which was already there for other `bool` and `string` types.

thozza (approved):
LGTM.

pkg/distro/fedora/images.go
mvo5: @thozza My main point specific to this is that everything that is not a `*T` in OSCustomizations does not need to be a `*T` in ImageConfig (unless I miss something of course). I.e. in OSCustomizations we have some bools (like NoBLS) and some *bool (like SELinuxForceRelabel). And for cases like "NoBLS" (and now MachineIdUninitialized) I don't see why they need to be *bool if they are collapsed into a bool anyway.

Unrelated to the above I did  https://github.com/osbuild/images/pull/860 for the points vs type thing a while ago, it never got traction so I closed it (and its value can be debated, I personally like it but I see the argument that this is just unneeded abstraction).

supakeen:
Force pushed to resolve conflicts.

pkg/distro/fedora/images.go
supakeen: I'm fine with either of these approaches but I'd like to do it as a follow-up so I can take a look at all the properties in `ImageConfig` + `OSCustomizations` to see which ones need, and need not be nullable.

achilleas-k: > I don't see why they need to be *bool if they are collapsed into a bool anyway.

The unset value is treated differently from the default (`0`, `false`, etc) when merging image configs in `InheritFrom()`.  In other words,
```
<image config with Uninitialized=nil>.InheritFrom(<image config with Uninitialized=default>)
```
with return an image config with `Uninitialized=default`.

```
<image config with Uninitialized=&false>.InheritFrom(<image config with Uninitialized=default>)
```
will return an image config with `Uninitialized=&false`.

achilleas-k (approved):

pkg/distro/fedora/images.go
mvo5: Thanks, the InheritFrom() mechanism was the missing piece for me, this is obviously how it should be. Sorry for the noise!

thozza: Sorry, the link in my first response pointed to a comment inside the `InheritFrom()` method, but it was probably not so obvious what I meant by that.

We never merge or inherit `OSCustomizations`, so we don't need to use pointers there.

thozza (approved):

PR #1261, opened 2025-02-25
fedora: iot: Package cleanups

nullr0ute:
Some cleanups for Fedora IoT to align with the rest of deliverables.


supakeen:
Converting to draft so this doesn't accidentally get merged. These changes were also made in [Fedora](https://pagure.io/fedora-iot/ostree/commits/f42) but they only apply to Fedora 42 and up. The changes as done in this PR would also apply to Fedora 41.

@nullr0ute do you want me to push a commit or do you want to try your hand at that?

supakeen:
Undrafted, I added a commit that conditionalizes the changes. For packages that was easy because of the cool YAML. For services that was a bit ... messy? I must be missing something, cc @thozza.

pkg/distro/fedora/distro.go
```diff
@@ -700,6 +681,12 @@ func newDistro(version int) distro.Distro {
 	minimalrawZstdImgType.payloadPipelines = []string{"os", "image", "zstd"}
 	minimalrawZstdImgType.exports = []string{"zstd"}
 
+	// set iot services (these are version-dependent)
+	iotCommitImgType.defaultImageConfig.EnabledServices = iotServicesForVersion(&rd)
+	iotOCIImgType.defaultImageConfig.EnabledServices = iotServicesForVersion(&rd)
+	iotInstallerImgType.defaultImageConfig.EnabledServices = iotServicesForVersion(&rd)
+	iotSimplifiedInstallerImgType.defaultImageConfig.EnabledServices = iotServicesForVersion(&rd)
```
thozza, line 688: Please, turn the image type definitions into a generator function and move the conditionals there. SImilar to https://github.com/osbuild/images/blob/9371af5db3bc8e8e9c4a620d29166a9149fdd35d/pkg/distro/rhel/rhel9/edge.go#L263-L299.

I don't like this approach, because the default image configuration is scattered in multiple places, which is not great for readability.

I'd like to consolidate this anyway once we move pkg definitions in RHEL to YAML and consolidate the distro interface implementation across all distros.

thozza (changes requested):
Thanks!

The package changes look fine.

For the version-specific image config, please use the approach from `rhel` distro (specific link provided as an in-line comment). I know it will be a bit more work, but we'd have to do it anyway and it will at least prevent a mess with scattered image configs...

pkg/distro/fedora/distro.go
supakeen: I've rebased on `main` where we now have the `mk*` functions so the above code is gone.

achilleas-k (approved):
Splendid!

PR #1262, opened 2025-02-25
fedora: remove rescue kernels from minimal

supakeen:
Let's remove the rescue kernels from all versions. This could be considered a bug or regression so turn it off.

Helps speed up the work on RISC-V due to saving time in cross-compilation as well.

---

Initially turned off for Fedora >= 43 in: #1245.


achilleas-k (approved):
:tada:

ondrejbudai (approved):
Thanks!

PR #1263, opened 2025-02-25
manifest: use experimentalflags.String("bootstrap")

mvo5:
This commit allows to use a container to bootstrap the buildroot via the experimtnal "bootstrap" flag:
```console
$ IMAGE_BUILDER_EXPERIMENTAL=bootstrap=ghcr.io/mvo5/fedora-buildroot:41 \
   image-builder manifest --arch=riscv64 minimal-raw --distro=fedora-41
```
and it will use the given container to bootstap the buildroot.

This will allow to do cross-arch building for riscv. The generated manifest can be build with osbuild (as long as qemu-user is installed) and will generate a minimal-raw for riscv64.

Cross-building is useful because qemu-user is a lot faster (and easier to setup) than full riscv system emulation. A minimal raw build on a really fast AWS machine takes about 100 minutes. A full minimal-raw build on my (moderate) PC takes about 20min - which is fast enough so that we can run this in GH workflows as part of the CI.

The approach with the bootstrap buildroot [0] allows us to generalize this later so that each distro would (in addition to the repos) an upstream container ref. With that we could auto-detect cross builds and we would just add this upstream container to bootstrap the buildroot.

[0] Thanks to Ondrej, Achilleas, Thozza, Simon and Florian.

P.S. With that we can also do:
```
$ IMAGE_BUILDER_EXPERIMENTAL=bootstrap=registry.fedoraproject.org/fedora-toolbox:41 image-builder manifest --arch=aarch64 --distro fedora-41 minimal-raw
# or
IMAGE_BUILDER_EXPERIMENTAL=bootstrap=registry.access.redhat.com/ubi9/ubi:latest image-builder manifest --arch=ppc64le --distro centos-9 qcow2
```
so we will not have to maintain our own containers (there are still some assumption like that the container contains python3 for the runners).

### DIFF
--- a/pkg/manifest/build.go
+++ b/pkg/manifest/build.go
@@ -3,7 +3,9 @@ package manifest
 import (
 	"fmt"
 
+	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/container"
+	"github.com/osbuild/images/pkg/experimentalflags"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/rpmmd"
 	"github.com/osbuild/images/pkg/runner"
@@ -57,6 +59,11 @@ func NewBuild(m *Manifest, runner runner.Runner, repos []rpmmd.RepoConfig, opts 
 		repos:              filterRepos(repos, name),
 		containerBuildable: opts.ContainerBuildable,
 	}
+
+	// This allows to bootstrap the buildroot with a custom container
+	// for e.g. cross-arch-build experiments,
+	maybeAddExperimentalContainerBootstrap(m, runner, opts, pipeline)
+
 	m.addPipeline(pipeline)
 	return pipeline
 }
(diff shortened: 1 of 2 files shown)


pkg/manifest/build.go
```diff
@@ -149,6 +156,33 @@ func (p *BuildrootFromPackages) getSELinuxLabels() map[string]string {
 	return labels
 }
 
+// maybeAddExperimentalContainerBootstrap will return a container buildroot
+// if if the "IMAGE_BUILDER_EXPERIMENTAL=bootstrap=<container-ref>" is
```
bcl, line 160: typo 'if if'

bcl (dismissed):
Looks good to me other than the small typo.
What's the plan for documenting the experimental flags? Is the commit message + function docstring enough or should there be a central document listing them all?

supakeen:
> Looks good to me other than the small typo. What's the plan for documenting the experimental flags? Is the commit message + function docstring enough or should there be a central document listing them all?

As we don't want to share these to the public in general (unless we specifically request someone to tinker) I think just internalized knowledge of them is fine :)

pkg/manifest/build.go
mvo5: Thank you! Fixed now (and while reading this again expanded about what a bootstrap container actually needs).

mvo5:
> Looks good to me other than the small typo. What's the plan for documenting the experimental flags? Is the commit message + function docstring enough or should there be a central document listing them all?

For now I agree with @supakeen - we can keep this internal. I would love to see this generalized eventually (see also the commit message) so that a user just types (on their x86_64 box) `image-builder --arch=aarch64 minimal-raw --distro fedora-41` and we detect internally that its a cross arch build and we would add the bootstrap container `registry.fedoraproject.org/fedora-toolbox:41` automatically. But baby-steps :)

pkg/manifest/build.go
```diff
@@ -149,6 +156,40 @@ func (p *BuildrootFromPackages) getSELinuxLabels() map[string]string {
...
 }
 
+// maybeAddExperimentalContainerBootstrap will return a container buildroot
+// if the "IMAGE_BUILDER_EXPERIMENTAL=bootstrap=<container-ref>" is
+// defined. This allows us to do cross-arch build experimentation.
+//
+// A "bootstrap" container has only these requirements:
+// - python3 for the runners
+// - rpm so that the real buildroot rpms can get instaleld
+// - setfiles so that the selinux stage for the real buildroot can run
```
ondrejbudai, line 166: The second commit invalidates this.

pkg/manifest/build.go
```diff
@@ -149,6 +156,40 @@ func (p *BuildrootFromPackages) getSELinuxLabels() map[string]string {
 	return labels
 }
 
+// maybeAddExperimentalContainerBootstrap will return a container buildroot
+// if the "IMAGE_BUILDER_EXPERIMENTAL=bootstrap=<container-ref>" is
+// defined. This allows us to do cross-arch build experimentation.
+//
+// A "bootstrap" container has only these requirements:
+// - python3 for the runners
+// - rpm so that the real buildroot rpms can get instaleld
```
ondrejbudai, line 165:
```suggestion
// - rpm so that the real buildroot rpms can get installed
```

ondrejbudai (commented):
Awesome! :) It might be a good idea to also cover the selinux special case: a test that checks that the bootstrap pipeline doesn't have an selinux stage.

achilleas-k (approved):
This is great!  I really like how straightforward it looks.

mvo5:
> This looks amazingly clean and simple... Nice work.
> 
> I have only one nitpick for the 1st commit's message:
> 
> ```
> Cross-building is useful because qemu-user is a lot faster
> (and easier to setup) than full riscv system emulation.
> A minimal raw build on a really fast AWS machine takes about
> 100 minutes. A full minimal-raw build on my (moderate) PC takes
> about 20min - which is fast enough so that we can run this in
> GH workflows as part of the CI.
> ```
> 
> I assume that it should be `100 seconds`, instead of `100 minutes`, unless your moderate PC is better than the really fast AWS machine 😅

Uh, thanks for pointing this out, what this should say is "100min on the fastest AWS with full system riscv64 emulation. But only 20min on a moderate desktop with qemu-user based (non-full-system) riscv64 emulation". Sorry that this is not clear enough.

PR #1264, opened 2025-02-26
many: `systemd-firstboot` support

supakeen:
This PR introduces a bunch of preliminary work to get `systemd-firstboot` to work with the Fedora Minimal spin though it is set up that it could be configured with other image types as well.

For the eventual interface I *think* we don't want each image type to define all the parts separately but to instead have a single flag in the image config to 'SystemdFirstBootable' or such.

The work is not yet done but this PR serves for documentation of encountered problems and such.

---

## Problems, challenges, fun, quirks

1. `systemd-firstboot --image=...` works great, everything is configurable.

However when booting an image more software runs and things start to get funky.

1. `systemd-tmpfiles` is configured on Fedora to create (amongst others) the `/etc/locale.conf`, `/etc/vconsole.conf` files (see: `/usr/lib/tmpfiles.d/etc.conf`). `systemd-tmpfiles` is explicitly requested to run *before* `systemd-firstboot`. This needs to be changed either in the Fedora-shipped RPMs or it needs to be worked around at build time (patch the `tmpfiles.d` configurations? change the order?) otherwise those options cannot be configured.
2. The `systemd-firstboot` prompt is 'broken'. To start with we get a bunch of output that goes straight through the prompt, perhaps some ordering issue; for now a `PreExecStart=/bin/sleep 1` fixes this but is not the right thing to do. 
3. The `systemd-firstboot` prompt is 'broken'. Initial keypresses do not work on the console. Potentially we'd want to wait for `udev-settle` or not at all.
4. The `root` user needs to be handled, currently it is created by either `systemd-tmpfiles` or `systemd-sysusers` (haven't looked yet) *before* `systemd-firstboot` runs since.

### DIFF
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -9,7 +9,6 @@ import (
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/internal/environment"
 	"github.com/osbuild/images/pkg/arch"
-	"github.com/osbuild/images/pkg/customizations/fsnode"
 	"github.com/osbuild/images/pkg/customizations/oscap"
 	"github.com/osbuild/images/pkg/datasizes"
 	"github.com/osbuild/images/pkg/distro"
@@ -464,9 +463,6 @@ func mkMinimalRawImgType(d distribution) imageType {
 		},
 		defaultImageConfig: &distro.ImageConfig{
 			EnabledServices: minimalServicesForVersion(&d),
-			// NOTE: temporary workaround for a bug in initial-setup that
-			// requires a kickstart file in the root directory.
-			Files: []*fsnode.File{initialSetupKickstart()},
 			Grub2Config: &osbuild.GRUB2Config{
 				// Overwrite the default Grub2 timeout value.
 				Timeout: 5,
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -474,6 +474,8 @@ func mkMinimalRawImgType(d distribution) imageType {
 			InstallWeakDeps: common.ToPtr(common.VersionLessThan(d.osVersion, VERSION_MINIMAL_WEAKDEPS)),
 			// Unset hostname so it isn't written (necessary for systemd-firstboot).
 			Hostname: common.ToPtr(""),
+			// Unset locale so it isn't written (necessary for systemd-firstboot).
+			Locale: common.ToPtr(""),
 		},
 		rpmOstree:              false,
 		kernelOptions:          defaultKernelOptions(),
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -472,6 +472,8 @@ func mkMinimalRawImgType(d distribution) imageType {
 				Timeout: 5,
 			},
 			InstallWeakDeps: common.ToPtr(common.VersionLessThan(d.osVersion, VERSION_MINIMAL_WEAKDEPS)),
+			// Unset hostname so it isn't written (necessary for systemd-firstboot).
+			Hostname: common.ToPtr(""),
 		},
 		rpmOstree:              false,
 		kernelOptions:          defaultKernelOptions(),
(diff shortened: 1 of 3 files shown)


pkg/manifest/os_test.go
```diff
@@ -248,3 +248,21 @@ func TestFirstBootDoesNotIncludeMachineIdStage(t *testing.T) {
...
+	require.NotNil(t, st)
+}
+
+func TestLanguageDoesNotIncludeLocaleStage(t *testing.T) {
+	os := NewTestOS()
+
+	pipeline := os.serialize()
+	st := findStage("org.osbuild.locale", pipeline.Stages)
+	require.Nil(t, st)
+}
```
achilleas-k, line 268: This commit can be squashed with the first one.  They even have almost the same message already.

pkg/distro/fedora/distro.go
```diff
@@ -442,6 +442,8 @@ var (
 				// Overwrite the default Grub2 timeout value.
 				Timeout: 5,
 			},
+			// Unset hostname so it isn't written (necessary for systemd-firstboot).
+			Hostname: common.ToPtr(""),
```
achilleas-k, line 442: Wouldn't it be more correct to set it to `nil` (or keep it unset)?

pkg/distro/fedora/distro.go
```diff
@@ -444,6 +444,8 @@ var (
 			},
 			// Unset hostname so it isn't written (necessary for systemd-firstboot).
 			Hostname: common.ToPtr(""),
+			// Unset locale so it isn't written (necessary for systemd-firstboot).
+			Locale: common.ToPtr(""),
```
achilleas-k, line 444: Same as previous comment about hostname.

Setting this to `ptr("")` will make the stage run and write `""`, no?

achilleas-k (commented):
This is really nice.  Good cleanup and unification of option handling.

Couple of comments below.

Any reason it's kept draft?  I think we can merge it and continue working on the firstboot process from here.

pkg/manifest/os_test.go
supakeen: Done.

pkg/distro/fedora/distro.go
supakeen: Setting it to `nil` does the wrong thing when the merge between image configs gets done; it then gets turned into the distro default.

pkg/distro/fedora/distro.go
supakeen: Setting it to `nil` does the wrong thing when the merge between image configs gets done; it then gets turned into the distro default.

The `os` pipeline verifies that the value is non-empty (e.g. `if p.Language != ""`) as a conditional for writing the stage so it doesn't add the stage at all.

supakeen:
Rebased, addressed @achilleas-k's comments.



> Any reason it's kept draft? I think we can merge it and continue working on the firstboot process from here.

Everything before the *last* commit is mergeable. I can create another PR with all the commits minus the last one because I think those are fine. Would that be something you want?

The last commit makes people unable to actually configure their system on *boot* for the reasons listed in the topic post; you can't configure a root password and systemd-tmpfiles pre-writes these files. I haven't had time to jerryrig that temporarily and anything we do for that will need to go into Fedora :)

achilleas-k (approved):

supakeen:
@achilleas-k Thanks for the approval: this needs to be scoped so it only applies to Fedora 43 (rawhide) because it relies on some things that exist there, unless I can confirm it to work on Fedora 42. I hope to get to those tidbits tomorrow.

After that it's going to need approval from more people than just code owners, I think, mostly Peter, Paul, and Adam (not tagging on purpose). The first two due to it being a significant change in the minimal fedora artifact. The last one due to having to make sure that the QA systems can verify this, since minimal is release blocking in fedora.

supakeen:
Still active, I just need to find time to remap/version gate this still.

AdamWill:
Is this no longer planned? We have an [os-autoinst-distri-fedora ticket](https://forge.fedoraproject.org/quality/os-autoinst-distri-fedora/issues/367) that depends on it, so I'm checking in.

supakeen:
> Is this no longer planned? We have an [os-autoinst-distri-fedora ticket](https://forge.fedoraproject.org/quality/os-autoinst-distri-fedora/issues/367) that depends on it, so I'm checking in.

We can land this without this PR but I need to discuss with Peter when we want to enable this.

PR #1268, opened 2025-02-26
imagefilter: implement yaml output format

schuellerf:
This output is a machine readable format and compatible to the `/distributions` endpoint of the cloudAPI.

This PR is a followup of #1206

### DIFF
--- a/pkg/imagefilter/formatter.go
+++ b/pkg/imagefilter/formatter.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 
 	"github.com/osbuild/images/pkg/distrosort"
+	"gopkg.in/yaml.v3"
+
 	// we cannot use "maps" yet, as it needs go1.23
 	"golang.org/x/exp/maps"
 )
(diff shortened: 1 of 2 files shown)


mvo5 (changes requested):
Thanks for your PR - I will still block it for now because I think this needs more discussion and I don't want this to get merged accidentally. It is adding new API that we will have to support for a long time. When it comes to API I some guidelines I try to keep in mind (among others) are:
1. orthogonality
2. YAGNI

Unless I miss something it seems:
1. we already have a machine readable API via json output, adding a new one with a different format/format-language makes me wonder if we need it
2. which leads to YAGNI, i.e. what is the use-case that prompts adding the new API?

Having a simple linear text format that is easy to grep seems good, having a machine readable json seems good, having a "short" format as an entry point is (probably) also good. This one here I'm not sure about.

supakeen (changes requested):
Not necessary. We have JSON structured output that can be programmatically read and human-readable short output in #1206 (which I just approved).

schuellerf:
Thanks, I just implemented this as a suggestion as for me a short human readable output and `json` is enough, too.

PR #1270, opened 2025-02-26
Enable image builds that use systemd mount units and kargs instead of writing /etc/fstab

achilleas-k:
This PR enables the use of systemd .mount (and .swap) units to replace writing an `/etc/fstab` file for the filesystem configuration.  This is enabled for the Fedora `minimal-raw` image and its derivatives for Fedora 43 (rawhide) and above.

The feature will be useful for bootc-image-builder, where filesystem configurations can be written at different stages (base container image build or disk build) and creating a systemd unit for each mount is more straightforward than merging different entries in an `/etc/fstab` file.

The lack of an `/etc/fstab` requires a few changes to kernel commandline arguments and the partition table.  Each change is described in its respective commit message, but I also summarise here:
- Partitions need to have GUIDs in accordance with the Discoverable Partitions Specification (done in #1231).
- The `/usr` partition, if separate from `/`, must be defined in the kernel command line arguments.
- The system root needs to be mounted `rw` during initial boot, because the lack of `/etc/fstab` means that the systemd-remount-fs service will not discover the mount options for `/` (and `/usr`) to remount them.
- For the same reason, the `/` mount options (`rootflags`) need to be defined in the kernel command line arguments.

This PR requires osbuild/osbuild#2001.


pkg/manifest/os.go
```diff
@@ -881,6 +881,29 @@ func prependKernelCmdlineStage(pipeline osbuild.Pipeline, kernelOptions []string
...
+		// remounted, which means if they are mounted read-only in the initrd,
+		// they will remain read-only. Flip the option if we're using only
+		// mount units, otherwise the filesystems will stay mounted 'ro'.
+		//
+		// See https://www.freedesktop.org/software/systemd/man/latest/systemd-remount-fs.service.html
+		for idx := range kernelOptions {
+			if kernelOptions[idx] == "ro" {
+				kernelOptions[idx] = "rw"
+				break
+			}
```
cgwalters, line 896: It seems like it'd be cleaner to ensure that `rw` is injected from the start if mount units are in use.

pkg/manifest/os.go
```diff
@@ -881,6 +881,29 @@ func prependKernelCmdlineStage(pipeline osbuild.Pipeline, kernelOptions []string
...
+		}
+
+		// set the rootflags for the same reason as above
+		fsOptions, err := rootFs.GetFSTabOptions()
+		if err != nil {
+			panic(fmt.Sprintf("error getting filesystem options for / mountpoint: %s", err))
+		}
+		kernelOptions = append(kernelOptions, fmt.Sprintf("rootflags=%s", fsOptions.MntOps))
```
cgwalters, line 904: Cosmetic: Since *most* use cases don't use `rootflags` it'd be nice to not end up with `rootflags=default`, so we could special case that and not emit `rootflags` at all.

cgwalters (commented):
Thanks for working on this! I only gave a quick skim

supakeen:
I'll do a comment before I take a look at the code and say this is super nice and exciting to have.

With my Fedora Minimal hat on: while we do plan to eventually have no-`/etc/fstab` I'd like to do some testing with this first even before landing in `rawhide`.

Can this PR include `experimentalflags.String("no-fstab") != ""` in the version gate for Fedora Minimal?

pkg/manifest/os.go
achilleas-k: You're absolutely correct.  I don't like switching options this late in the manifest generation process either.  Making the change earlier in the process might need more refactoring though, so if it turns out to be too much I might leave it for a later cleanup.

pkg/manifest/os.go
achilleas-k: Yes, this is good advice.  We also end up with duplicate rootflags when the root is on btrfs or luks.  I'll clean it up.

achilleas-k:
The test failure is a Fedora 43 AMI with customizations.  I checked it with some local builds and discovered that the image kernel panics on boot.  Diffing the package set between that and the empty config (which does boot) revealed that the problematic package is probably `plymouth`.

Fun!

achilleas-k:
> The test failure is a Fedora 43 AMI with customizations. I checked it with some local builds and discovered that the image kernel panics on boot. Diffing the package set between that and the empty config (which does boot) revealed that the problematic package is probably `plymouth`.
> 
> Fun!

Confirmed.  Building a raw image with 
```json
{
  "name": "plymouth",
  "blueprint": {
    "packages": [
      {
        "name": "plymouth"
      }
    ]
  }
}
```

> The test failure is a Fedora 43 AMI with customizations. I checked it with some local builds and discovered that the image kernel panics on boot. Diffing the package set between that and the empty config (which does boot) revealed that the problematic package is probably `plymouth`.

Confirmed.  Building a Fedora 43 x86_64 raw image with the following config:
```json
{
  "name": "plymouth",
  "blueprint": {
    "packages": [
      {
        "name": "plymouth"
      }
    ]
  }
}
```
fails to boot with a kernel panic.

achilleas-k:
Checking `libevdev` as well, since it's the only extra package that gets pulled in as a dependency on top of the raw / ami base package set.

achilleas-k:
Adding just `libevdev` to the build doesn't cause any issues.

achilleas-k:
Last push is a simple rebase and conflict resolution.  I'll be addressing comments later.

pkg/manifest/os.go
achilleas-k: I'm going to leave this for another time, because `ro` is a very common option we statically add to a lot of image definitions.  Making it conditional on the mount units option from the top would need a lot of changes that would complicate (and perhaps delay) this PR.

cgwalters: I understand, it makes sense.

However, I would also say that I think the `ro` option doesn't make sense to use anymore *anywhere* (see also e.g. https://github.com/ostreedev/ostree/issues/3193#issuecomment-2578237486 )

achilleas-k: ah!  That's great to know.  Happy to drop it everywhere, though might have to do it carefully for RHEL in case there are any side effects that users rely on.

pkg/distro/rhel/rhel10/ami.go
```diff
@@ -157,10 +157,10 @@ func mkEC2SapImgTypeX86_64(osVersion string) *rhel.ImageType {
 // IMAGE CONFIG
 
 // TODO: move these to the EC2 environment
-const (
-	amiKernelOptions        = "console=tty0 console=ttyS0,115200n8 nvme_core.io_timeout=4294967295"
-	amiAarch64KernelOptions = amiKernelOptions + " iommu.strict=0"
-	amiSapKernelOptions     = amiKernelOptions + " processor.max_cstate=1 intel_idle.max_cstate=1"
+var (
```
supakeen, line 160: I'll comment on this change once (though it's done in many places), would it make sense to change this to a function that returns `[]string` instead of turning `const` to `var`? I'm always slightly worried about global mutable values.

pkg/manifest/common.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/osbuild/images/pkg/disk"
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+// filesystemConfigStages generates either an org.osbuild.fstab stage or a
+// collection of org.osbuild.systemd.unit.create stages for .mount and .swap
+// units (and an org.osbuild.systemd stage to enable them) depending on the
+// pipeline configuration.
+func filesystemConfigStages(pt *disk.PartitionTable, mountUnits bool) ([]*osbuild.Stage, error) {
```
supakeen, line 12: Perhaps it could make sense to pass in an interface here instead that this function can use to determine all it needs to create any filesystem configs. There's no need for it *now* so feel free to ignore.

pkg/manifest/os.go
```diff
@@ -853,13 +857,55 @@ func (p *OS) serialize() osbuild.Pipeline {
...
 	if rootFs == nil {
 		panic("root filesystem must be defined for kernel-cmdline stage, this is a programming error")
 	}
+
+	// if /usr is on a separate filesystem, it needs to be defined in the
+	// kernel cmdline options for autodiscovery (when there's no /etc/fstab)
+	// see:
+	//  - https://github.com/systemd/systemd/issues/24027
+	//  - https://github.com/systemd/systemd/pull/33397
+	if usrFs := pt.FindMountable("/usr"); usrFs != nil && mountUnits {
```
supakeen: It slightly easier to me to combine the `if mountUnits` here and on line 884 into one block with a nested `FindMountable` in it.

achilleas-k:
> Can this PR include `experimentalflags.String("no-fstab") != ""` in the version gate for Fedora Minimal?

Added this but it's a bit quick and dirty.  I want to follow up with a nicer way of handling this and add a convenient way of enabling them for tests (using the `test/configs/`).

pkg/osbuild/systemd_unit_create_stage.go
```diff
@@ -177,3 +229,112 @@ func NewSystemdUnitCreateStage(options *SystemdUnitCreateStageOptions) *Stage {
...
+		fsSpec := ent.GetFSSpec()
+		fsOptions, err := ent.GetFSTabOptions()
+		if err != nil {
+			return err
+		}
+
+		var filename string
+		var options *SystemdUnitCreateStageOptions
+		device := fmt.Sprintf("/dev/disk/by-uuid/%s", strings.ToLower(fsSpec.UUID))
+		if err := uuid.Validate(fsSpec.UUID); err != nil {
```
supakeen, line 250: This feels weird, are all potential errors from `uuid.Validate` still valid `by-uuid` identifiers?

pkg/osbuild/systemd_unit_create_stage.go
```diff
@@ -177,3 +229,112 @@ func NewSystemdUnitCreateStage(options *SystemdUnitCreateStageOptions) *Stage {
...
+	genOption := func(ent disk.FSTabEntity, path []disk.Entity) error {
+		fsSpec := ent.GetFSSpec()
+		fsOptions, err := ent.GetFSTabOptions()
+		if err != nil {
+			return err
+		}
+
+		var filename string
+		var options *SystemdUnitCreateStageOptions
+		device := fmt.Sprintf("/dev/disk/by-uuid/%s", strings.ToLower(fsSpec.UUID))
```
supakeen, line 249: I prefer `filepath.Join` which suits fine here.

pkg/osbuild/systemd_unit_create_stage.go
```diff
@@ -177,3 +229,112 @@ func NewSystemdUnitCreateStage(options *SystemdUnitCreateStageOptions) *Stage {
...
+		if err := uuid.Validate(fsSpec.UUID); err != nil {
+			// vfat IDs and other non-UUID identifiers aren't lowercased
+			device = fmt.Sprintf("/dev/disk/by-uuid/%s", fsSpec.UUID)
+		}
+		switch ent.GetFSType() {
+		case "swap":
+			filename = fmt.Sprintf("%s.swap", pathEscape(device))
+			options = &SystemdUnitCreateStageOptions{
```
supakeen, line 257: Can we move options up to above the switch and then only assign the `Config.Swap` and `Config.Mount` properties in the switch itself? Otherwise they're quite similar.

supakeen (commented):
Left some comments, nothing major. Very nice to have this capability and very well implemented IMHO.

pkg/distro/rhel/rhel10/ami.go
achilleas-k: I am too, and I agree.  I'd say I wanted to minimise the diff but tbh it was probably laziness...

pkg/manifest/os.go
achilleas-k: I considered it but I felt the separation made it easier to reason with.  I'm not against changing it though.

pkg/osbuild/systemd_unit_create_stage.go
achilleas-k: You're right, it is a bit weird.  It's difficult to tell, really.  I don't know what the rules are for the `by-uuid` links.  I think I'll add regex check for the FAT volid format instead.

pkg/distro/rhel/rhel10/ami.go
achilleas-k: Done!

pkg/osbuild/systemd_unit_create_stage.go
achilleas-k: Done.

pkg/osbuild/systemd_unit_create_stage.go
achilleas-k: Done.

supakeen: I think you missed the one here and did the one below?

supakeen (approved):
Very, very nice. I have one last remark but it was something that was missed from the previous remarks I think and also not super critical.

achilleas-k:
> The test failure is a Fedora 43 AMI with customizations. I checked it with some local builds and discovered that the image kernel panics on boot. Diffing the package set between that and the empty config (which does boot) revealed that the problematic package is probably `plymouth`.

Repo updates fixed this too :tada:

supakeen (approved):

ondrejbudai (approved):
Great work, thank you! 😍

pkg/osbuild/systemd_unit_create_stage.go
```diff
@@ -223,3 +228,98 @@ func NewSystemdUnitCreateStage(options *SystemdUnitCreateStageOptions) *Stage {
...
+
+		// this sorter is not guaranteed to be stable, but the unit Filenames
+		// are unique
+		switch {
+		case optsa.Filename < optsb.Filename:
+			return -1
+		case optsa.Filename > optsb.Filename:
+			return 1
+		}
+		panic(fmt.Sprintf("error sorting systemd unit mount stages: possible duplicate mount unit filenames: %q %q", optsa.Filename, optsb.Filename))
```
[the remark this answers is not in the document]
ondrejbudai: Right, that's true. Let's not worry about this too much then. It really shouldn't ever happen.

PR #1272, opened 2025-02-27
workflow: revert GOPROXY setting

mvo5:
This partially reverts c92f9c1168cd9211e8cd1fa7435c3fa5b3086566 because the issue with `go.opencensus.io` is fixed.

It cannot just be a full go revert because the commit also moves the fedora tests inside a container.
changes as part of the commit).


mvo5:
Hm, this has some unexpected side effects, closing to investigate.

schuellerf:
I think those look similar to the effects I saw, why we put GOPROXY in there initially…
which problem do you see with `go.opencensus.io`? How could I reproduce this?

PR #1273, opened 2025-02-27
fedora: `mkImage$Type` for fedora

supakeen:
Introduces the `mkImage$Type` format from the RHEL definitions to Fedora. This commit only adds the functions that return `imageType`'s directly. This allows other in-flight PRs to make use of these new functions for any versioning and such.


thozza (approved):
Thank you. This is a good start :+1:

PR #1274, opened 2025-02-27
osbuild: add insights-client config stage (HMS-5670)

rverdile:
Supports adding the insights-client config stage. Feedback and help much appreciated :)

stage added to osbuild here: https://github.com/osbuild/osbuild/pull/2030

JIRA: [HMS-5670](https://issues.redhat.com/browse/HMS-5670)

### DIFF
--- a/pkg/distro/rhel/images.go
+++ b/pkg/distro/rhel/images.go
@@ -228,6 +228,9 @@ func osCustomizations(
 	var subscriptionStatus subscription.RHSMStatus
 	if options.Subscription != nil {
 		subscriptionStatus = subscription.RHSMConfigWithSubscription
+		if options.Subscription.Proxy != "" {
+			osc.InsightsClientConfig = &osbuild.InsightsClientConfigStageOptions{Proxy: options.Subscription.Proxy}
+		}
 	} else {
 		subscriptionStatus = subscription.RHSMConfigNoSubscription
 	}
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -99,35 +99,36 @@ type OSCustomizations struct {
 	ShellInit []shell.InitFile
 
 	// TODO: drop osbuild types from the API
-	Firewall            *osbuild.FirewallStageOptions
-	Grub2Config         *osbuild.GRUB2Config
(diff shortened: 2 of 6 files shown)


pkg/manifest/os.go
```diff
@@ -634,7 +635,11 @@ func (p *OS) serialize() osbuild.Pipeline {
 	}
 
 	if p.SshdConfig != nil {
-		pipeline.AddStage((osbuild.NewSshdConfigStage(p.SshdConfig)))
+		pipeline.AddStage(osbuild.NewSshdConfigStage(p.SshdConfig))
```
achilleas-k, line 638: Oh wow.  This must have been here for _ages_!

achilleas-k (dismissed):
Thanks.  This LGTM.  What's the plan for enabling this now?  It's not wired up to anything externally that set the options yet.  Will it be a customization in the blueprint?

pkg/distro/rhel/images.go
```diff
@@ -228,6 +228,7 @@ func osCustomizations(
 	var subscriptionStatus subscription.RHSMStatus
 	if options.Subscription != nil {
 		subscriptionStatus = subscription.RHSMConfigWithSubscription
+		osc.InsightsClientConfig = &osbuild.InsightsClientConfigStageOptions{Proxy: options.Subscription.Proxy}
```
achilleas-k, line 231: Should this also check if `options.Subscription.Proxy != ""`?  Otherwise we'll be adding a stage that sets the proxy to "", or it will generate a stage with empty options, which will probably fail.

rverdile: that's a good point, pushed that fix, thanks!

achilleas-k (approved):
Thanks!  LGTM now.

FTR for anyone else reviewing this: This option will only be enabled automatically when an image is built in the staging service.  I think the best way to enable this would be using an osbuild-composer config key or an environment variable.  Then composer will have to set the subscription options itself before starting the manifest generation.

pkg/distro/rhel/images.go
achilleas-k: Ty!

supakeen (approved):

PR #1269, opened 2025-02-26
fedora: drop `firewalld` for minimal

supakeen:
Drops `firewalld` for Fedora Minimal in rawhide for now, see issue [1].

[1]: https://github.com/fedora-minimal/distribution-minimal/issues/1

PR #1276, opened 2025-03-01
fedora: disable firewalld for minimal >= 43

supakeen:
In [1] I removed the package for `firewalld` conditionally. However, the service enabling wasn't made conditional leading to build failures as the service file does not exist.

[1]: https://github.com/osbuild/images/pull/1269


achilleas-k (approved):

mvo5 (approved):
My usual pet-peeve, I whish this had a test but eventually this will all be yaml anyway :)

PR #1277, opened 2025-03-01
customization: sshkeys

supakeen:
On Matrix it was figured out that our documented `SSHKey` customization doesn't work. After taking a look it seems like it wasn't actually ever plumbed through...

Since we've documented it I assume we want to support it so here's an initial PR that does just that. This PR would close https://github.com/osbuild/image-builder-cli/issues/142

After this PR the stages get injected to create files and directories:

```console
$ cat sshkey.toml 
[[customizations.sshkey]]
user = "testuser"
key = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKNh/u8oWHfYwr01X8G8ijSC3hPfKfLpK8MISxg2mq1O user@muja.home.arpa"
$ ./image-builder manifest --distro fedora-43 --blueprint sshkey.toml minimal-raw-zst | jq . | rg testuser
                "path": "/home/testuser/.ssh",
                "to": "tree:///home/testuser/.ssh/authorized_keys",
                "to": "tree:///home/testuser/.ssh/authorized_keys",
```

Note that setting the SSH key for a nonexistent user will lead to a build failure.

---

As draft now as it needs tests, defensiveness in the customization parsing so we can fail earlier, and to move the filesystem generation to a helper function.

@achilleas-k how do we (or should we) handle the ostree case here? I see authorized keys is normally punted to a firstboot script.

### DIFF
--- a/pkg/blueprint/customizations.go
+++ b/pkg/blueprint/customizations.go
@@ -34,6 +34,7 @@ type Customizations struct {
 	RPM                *RPMCustomization              `json:"rpm,omitempty" toml:"rpm,omitempty"`
 	RHSM               *RHSMCustomization             `json:"rhsm,omitempty" toml:"rhsm,omitempty"`
 	CACerts            *CACustomization               `json:"cacerts,omitempty" toml:"cacerts,omitempty"`
+	SSHKey             []SSHKeyCustomization          `json:"sshkey,omitempty" toml:"sshkey,omitempty"`
 }
 
 type IgnitionCustomization struct {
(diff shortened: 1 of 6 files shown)


achilleas-k:
This was removed from images a while ago on purpose as a way to simplify the backend blueprint implementation.  It remains in osbuild-composer for backwards compatibility.

I don't know how I feel about bringing it back.  You can see the full reasoning here: https://github.com/osbuild/images/pull/928

The idea was that frontends should be responsible for defining what they support wrt the blueprint.  I would much rather we say this option is not supported anywhere (except in composer) than bring it back here.

supakeen:
> This was removed from images a while ago on purpose as a way to simplify the backend blueprint implementation. It remains in osbuild-composer for backwards compatibility.
> 
> I don't know how I feel about bringing it back. You can see the full reasoning here: #928
> 
> The idea was that frontends should be responsible for defining what they support wrt the blueprint. I would much rather we say this option is not supported anywhere (except in composer) than bring it back here.

The important bit here is that this is supported by the `user` customization; which will modify pre-existing users or create new ones when necessary so:

```toml
[[customizations.user]]
user = "name"
keys = ["ssh-edsa XXX"]
```

Will customize the user if it exists (and thus add the SSH keys) or create it if not (and add the SSH keys).

In that case the above is purely due to the documentation not being super clear about this so I'll create PRs there instead.

PR #1287, opened 2025-03-05
packagesets: introduce  `IMAGE_BUILDER_EXPERIMENTAL=yamldir=...`

mvo5:
This commit adds a new `IMAGE_BUILDER_EXPERIMENTAL=yamldir=...`
experimental flag to force setting the datadir for the yaml
directory to override e.g. packagesets.

Ideally we would pass this all the way down the stack via a
search path but this will require a bunch of more refactoring.
So to give people the ability to tweak packagesets today, start
here.

Thanks to Ondrej for the suggestion, see also https://github.com/osbuild/images/pull/1283 where this will become immediately useful when the centos folks start hacking on our package lists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,27 @@ Please refer to the [developer guide](https://www.osbuild.org/docs/developer-gui
 
 See also the [local developer documentation](./docs/developer) for useful information about working with this specific project.
 
+#### YAML based image definitions
+
+More and more parts of the library are converted to use yaml to define core
+parts of an image. See this [example](./pkg/distro/packagesets/fedora/)
+directory. For local development that just changes the YAML based
+definitions the library can be forced to use alternative yaml dirs.
+
+E.g. if there is a `./my-yaml/fedora/package-sets.yaml` then that can be
+used via:
+```console
+$ IMAGE_BUILDER_EXPERIMENTAL=yamldir=./my-yaml image-builder build minimal-raw --distro fedora-42
+```
+WARNING: this is an experimental feature and unsupported feature that should
+never be used in production and may change anytime.
+
+We do plan to eventually stabilize this so that it can be a switch for
+the `image-builder` tool but for now this environment option is
+required.
+
+#### Build requirements
+
 The build-requirements of the Go library for Fedora and rpm-based distributions are:
 
 - `go`
--- a/pkg/distro/packagesets/loader.go
+++ b/pkg/distro/packagesets/loader.go
@@ -4,6 +4,8 @@ import (
 	"embed"
 	"fmt"
 	"io/fs"
+	"os"
+
 	"path/filepath"
 	"strings"
 
@@ -11,6 +13,7 @@ import (
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/distro"
+	"github.com/osbuild/images/pkg/experimentalflags"
 	"github.com/osbuild/images/pkg/rpmmd"
 )
 
@@ -53,12 +56,19 @@ func Load(it distro.ImageType, overrideTypeName string, replacements map[string]
 	distroName := distroNameVer[:strings.LastIndex(distroNameVer, "-")]
 	distroVersion := distribution.OsVersion()
 
-	distroSets, err := DataFS.Open(filepath.Join(distroName, "package_sets.yaml"))
+	// XXX: this is a short term measure, pass a set of
+	// searchPaths down the stack instead
+	var dataFS fs.FS = DataFS
+	if overrideDir := experimentalflags.String("yamldir"); overrideDir != "" {
+		dataFS = os.DirFS(overrideDir)
+	}
+	f, err := dataFS.Open(filepath.Join(distroName, "package_sets.yaml"))
 	if err != nil {
 		return rpmmd.PackageSet{}, err
 	}
+	defer f.Close()
 
-	decoder := yaml.NewDecoder(distroSets)
+	decoder := yaml.NewDecoder(f)
 	decoder.KnownFields(true)
 
 	var pkgSets map[string]packageSet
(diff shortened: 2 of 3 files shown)


supakeen:
I'm fine with (something like) this, but is there really a pressing time concern to make this available *right now* behind a feature flag?

It'd be much nicer to have this live prettily in the frontend(s) and I don't think we're in a super rush to provide this? Perhaps I misunderstood the meeting yesterday.

mvo5:
> I'm fine with (something like) this, but is there really a pressing time concern to make this available _right now_ behind an experimental flag?
> 
> It'd be much nicer to have this live prettily in the frontend(s) and I don't think we're in a super rush to provide this? Perhaps I misunderstood the meeting yesterday.

We can do that too and wait for that. It did not seem like a huge detraction, the test will be reused and the code will be largely the same except of course the search path needs to travel some layers down. But fine to not do it now too.

supakeen:
> I wouldn't block on making it nice. I would love an MVP that allows me to alter image types without needing to touch go for simple changes as soon as possible. :)
> 
> Also, this sends a very positive message that we are committed to opening the defs more which feels pretty important in today's context.

I'll approve then. But we've previously said that we wouldn't document experimental features as things people can use/rely on.

supakeen (dismissed):

README.md
```diff
@@ -27,6 +27,23 @@ Please refer to the [developer guide](https://www.osbuild.org/docs/developer-gui
...
+definitions the library can be forced to use alternative yaml dirs.
+
+E.g. if there is a `./my-yaml/fedora/package-sets.yaml` then that can be
+used via:
+```console
+$ IMAGE_BUILDER_EXPERIMENTAL=yamldir=./my-yaml image-builder build minimal-raw --distro fedora-42
+```
+Note that this will eventually become a switch for the `image-builder` tool
+but for now this environment option is required.
+
```
bcl, line 44: I would expand this to include a very clear note that this will change in the future and is totally experimental and unsupported.

pkg/distro/packagesets/loader.go
```diff
@@ -34,15 +38,23 @@ func Load(it distro.ImageType, replacements map[string]string) rpmmd.PackageSet
 	archName := arch.Name()
 	distribution := arch.Distro()
 	distroNameVer := distribution.Name()
-	distroName := strings.SplitN(distroNameVer, "-", 2)[0]
```
bcl, line 37: This really seems like it ought to be a function provided by Distro at some point instead of needing to know about the formatting of the Name() result.

pkg/distro/packagesets/loader.go
```diff
@@ -34,15 +38,23 @@ func Load(it distro.ImageType, replacements map[string]string) rpmmd.PackageSet
...
+	// use rsplit() here, for "centos-stream-10" like distro names
+	distroName := distroNameVer[:strings.LastIndex(distroNameVer, "-")]
 	distroVersion := distribution.OsVersion()
 
-	distroSets, err := Data.Open(filepath.Join(distroName, "package_sets.yaml"))
+	// XXX: this is a short term measure, pass a set of
+	// searchPaths down the stack instead
+	var dataFS fs.FS = DataFS
+	if overrideDir := experimentalflags.String("yamldir"); overrideDir != "" {
```
bcl, line 48: This is something that it would be really nice to log somehow so that when debugging you can clearly tell it has been overridden. I don't see any provisions for logging anywhere near this code though :/

pkg/distro/packagesets/loader.go
mvo5: Yeah, totally agree. The distroidparser cannot be used unfortunately here because of circular imports. A refactor so that distro returns `distro.ID` instead of Name() might be a good first step but then we should try to also remove releaever and OsVersion along the way so it would require a bit of refactor.

pkg/distro/packagesets/loader.go
mvo5: Great point, I added simple logging.

README.md
mvo5: I reworded this, please let me know what you think and if it should still be stronger.

pkg/distro/packagesets/loader.go
```diff
@@ -34,15 +38,23 @@ func Load(it distro.ImageType, replacements map[string]string) rpmmd.PackageSet
...
+		dataFS = os.DirFS(overrideDir)
+	}
+	f, err := dataFS.Open(filepath.Join(distroName, "package_sets.yaml"))
 	if err != nil {
 		panic(err)
```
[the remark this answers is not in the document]
mvo5: Indeed, I refactored this now so that the error can be passed up.

README.md
```diff
@@ -27,6 +27,27 @@ Please refer to the [developer guide](https://www.osbuild.org/docs/developer-gui
...
+E.g. if there is a `./my-yaml/fedora/package-sets.yaml` then that can be
+used via:
+```console
+$ IMAGE_BUILDER_EXPERIMENTAL=yamldir=./my-yaml image-builder build minimal-raw --distro fedora-42
+```
+WARNING: this is an experimental feature and unsupported feature that should
+never be used in production and may change anytime.
+
+We do plan to eventually stabelize this so that it can be a switch for
```
achilleas-k, line 45:
```suggestion
We do plan to eventually stabilize this so that it can be a switch for
```

achilleas-k (approved):
Tiny typo.  Otherwise LGTM.

I'm looking forward to this being de-experimentalised!

supakeen (dismissed):

mvo5:
Thank you! I fixed the typo and rebased to master, this is ready to re-review

supakeen (approved):

ondrejbudai (approved):
Love this, thanks!

PR #1288, opened 2025-03-05
packagesets: add distro_name condition

mvo5:
This commit allows packagesets to be conditional on the distribution name. This is useful when we define packagesets for RHEL/Centos where we only want to pull in certain packages like "subscription-manager" when building images for RHEL.

Cleaned up split out of https://github.com/osbuild/images/pull/1283/commits

### DIFF
--- a/pkg/distro/packagesets/loader.go
+++ b/pkg/distro/packagesets/loader.go
@@ -3,6 +3,7 @@ package packagesets
 import (
 	"embed"
 	"fmt"
+	"io/fs"
 	"path/filepath"
 	"strings"
 
@@ -13,7 +14,9 @@ import (
 )
 
 //go:embed */*.yaml
-var Data embed.FS
+var data embed.FS
+
+var DataFS fs.FS = data
 
 type packageSet struct {
 	Include   []string    `yaml:"include"`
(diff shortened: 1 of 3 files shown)


supakeen (approved):
I feel slightly conflicted as I'm of the opinion that distributions should be separate things, a bit of duplication wouldn't hurt us too much (e.g. having `rhel` and `centos-stream` directories with each having their image types defined).

However, I also understand how we currently do these things and the code is fine.

pkg/distro/packagesets/loader.go
```diff
@@ -34,10 +38,12 @@ func Load(it distro.ImageType, replacements map[string]string) rpmmd.PackageSet
 	archName := arch.Name()
 	distribution := arch.Distro()
 	distroNameVer := distribution.Name()
-	distroName := strings.SplitN(distroNameVer, "-", 2)[0]
+	// we need to split from the right for "centos-stream-10" like
+	// distro names, sadly go has no rsplit() so we do it manually
+	distroName := distroNameVer[:strings.LastIndex(distroNameVer, "-")]
```
thozza, line 43: I think that the best approach would be to use [`distroidparser.Parser`](https://github.com/osbuild/images/blob/9eb2318dfadc1060751a59872734202d33bb81b9/pkg/distroidparser/idparser.go#L34-L51) instead of this code.  Or even better, extend the `distro.Distribution` interface to return `distro.ID`. Anyway, it's probably material for a follow-up.

thozza (approved):
Thanks., LGTM

pkg/distro/packagesets/loader.go
mvo5: Sadly distroidparser cannot be imported because of import cycles (I would like that fixed but that requires a bit more refactor). I like `distro.ID` in the distro.Distribution interface but for that I think we should remove Name/ReleaseVer/OsVersion along the way as its not orthogonal to ID() (and we need to make sure distro.ID covers all the cases thtat the other three cover (I think it does)). But that is probably a bigger discussion :)

PR #1289, opened 2025-03-06
Use grub2 for BIOS iso booting

bcl:
This adds support for using grub2 instead of syslinux (which is no longer updated) when booting an iso on BIOS systems. This matches how Lorax creates the boot.iso and live media images. This applies to BIOS bootin on the x86 architecture for all of the anaconda live install images created by osbuild-composer. It does not change how the ostree simplified image boots (it still uses syslinux) and it does not change how the coreos iso boots since it uses it's own self-contained stage.

### DIFF
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -737,6 +737,11 @@ func iotInstallerImage(workload workload.Workload,
 		img.RootfsType = manifest.SquashfsRootfs
 	}
 
+	// Enable grub2 BIOS iso on x86_64 only
+	if img.Platform.GetArch() == arch.ARCH_X86_64 {
+		img.ISOBoot = manifest.Grub2ISOBoot
+	}
+
 	if locale := t.getDefaultImageConfig().Locale; locale != nil {
 		img.Locale = *locale
 	}
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/internal/workload"
+	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/blueprint"
 	"github.com/osbuild/images/pkg/container"
 	"github.com/osbuild/images/pkg/customizations/anaconda"
@@ -417,6 +418,11 @@ func liveInstallerImage(workload workload.Workload,
 
 	if common.VersionGreaterThanOrEqual(img.OSVersion, VERSION_ROOTFS_SQUASHFS) {
 		img.RootfsType = manifest.SquashfsRootfs
+	}
+
+	// Enable grub2 BIOS iso on x86_64 only
+	if img.Platform.GetArch() == arch.ARCH_X86_64 {
+		img.ISOBoot = manifest.Grub2ISOBoot
 	}
 
 	if locale := t.getDefaultImageConfig().Locale; locale != nil {
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -526,6 +526,11 @@ func imageInstallerImage(workload workload.Workload,
 		img.RootfsType = manifest.SquashfsRootfs
 	}
 
+	// Enable grub2 BIOS iso on x86_64 only
+	if img.Platform.GetArch() == arch.ARCH_X86_64 {
+		img.ISOBoot = manifest.Grub2ISOBoot
+	}
+
 	return img, nil
 }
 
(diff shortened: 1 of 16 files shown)


pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -25,6 +25,15 @@ const ( // Rootfs type enum
 	ErofsRootfs                          // Create a plain erofs rootfs
 )
 
+type ISOBootType uint64
+
+// These constants are used by the ISO images to control the type of bootable iso
+const ( // ISOBoot type enum
+	UEFIOnlyISOBoot ISOBootType = iota // Only boot with grub2 UEFI
```
supakeen, line 32: Can we call this `Grub2ISOBootUefiOnly` or something, I know its a bunch longer but the other two here include the name of the bootloader used.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -145,6 +154,11 @@ func (p *AnacondaInstallerISOTree) getBuildPackages(_ Distro) []string {
 	default:
 	}
 
+	if p.ISOBoot == Grub2ISOBoot {
+		// Needed for the i386-pc directory of modules needed by grub2 BIOS booting
```
supakeen, line 158: Perhaps it is wise to assert architecture here as well? Not super necessary of course.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -342,8 +356,8 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 	default:
 	}
 
-	if p.ISOLinux {
-		isoLinuxOptions := &osbuild.ISOLinuxStageOptions{
+	if p.ISOBoot == SyslinuxISOBoot {
```
supakeen, line 359: And then I assume we don't need to do anything extra for the grub2 UEFI only version?

supakeen (commented):
Since I've done this on a few on my previous PRs, would it be nicer to pipe this through image configs (either through installer config, or a new type iso config or such). It'd let you have a lot less code inside the image and manifest parts.

Otherwise, it seems we don't case for the ISO in grub2 with UEFI only mode. Is that correct? :)

pkg/manifest/anaconda_installer_iso_tree.go
bcl: I assume the caller only sets it on x86. Otherwise we end up needing to figure out an error path, or... add a panic which I really really do not like.

bcl:
> Since I've done this on a few on my previous PRs, would it be nicer to pipe this through image configs (either through installer config, or a new type iso config or such). It'd let you have a lot less code inside the image and manifest parts.

I'm not actually sure what you mean by this. Can you point to an example of what you've done before?

> Otherwise, it seems we don't case for the ISO in grub2 with UEFI only mode. Is that correct? :)

Right, grub2 UEFI is always setup.

supakeen:
> > Since I've done this on a few on my previous PRs, would it be nicer to pipe this through image configs (either through installer config, or a new type iso config or such). It'd let you have a lot less code inside the image and manifest parts.
> 
> ~I'm not actually sure what you mean by this. Can you point to an example of what you've done before?~
> 
> Ah, now I see, I think. Like #1258 as an example? I don't think I want to try doing that with the PR at this point.

Let's leave it for a followup then.

pkg/manifest/anaconda_installer_iso_tree.go
bcl: Renamed it to `Grub2UEFIOnlyISOBoot` so it's consistent with the others.

supakeen (approved):

pkg/distro/fedora/images.go
```diff
@@ -419,6 +420,11 @@ func liveInstallerImage(workload workload.Workload,
 		img.RootfsType = manifest.SquashfsRootfs
 	}
 
+	// Enable grub2 BIOS iso on x86_64 only
+	if img.Platform.GetArch() == arch.ARCH_X86_64 {
```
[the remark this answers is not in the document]
bcl: Correct. Using syslinux was actually wrong. Fedora has been using grub2 for bios booting the iso since Fedora 37.

mvo5 (approved):
Thanks, this looks fine, one idea/suggestion inline for your consideration.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -342,8 +356,8 @@ func (p *AnacondaInstallerISOTree) serialize() osbuild.Pipeline {
 	default:
 	}
 
-	if p.ISOLinux {
```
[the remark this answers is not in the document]
bcl: I don't think it really adds anything. The ISOBoot is as close to an enum as we can get and shouldn't ever get set to anything wrong.

PR #1291, opened 2025-03-06
distro: tweak the packagesets.Loader() interface

mvo5:
The current `packagesets.Loader()` takes a single imagetype argument to determine what packagesets to load from YAML.

This generally works great but its a challenge for image types like the anaconda installer where we need a different package set for the iso and the os. In fedora we workaround this by re-using the minimal-raw package set for the anaconda installers "os" package set. However in Centos/RHEL this is less straightforward so having the ability to "override" the pkgset name is required.

Another trivial split out from https://github.com/osbuild/images/pull/1283

### DIFF
--- a/pkg/distro/fedora/distro.go
+++ b/pkg/distro/fedora/distro.go
@@ -69,11 +69,7 @@ func mkImageInstallerImgType(d distribution) imageType {
 		packageSets: map[string]packageSetFunc{
 			osPkgsKey: func(t *imageType) rpmmd.PackageSet {
 				// use the minimal raw image type for the OS package set
-				ft := &imageType{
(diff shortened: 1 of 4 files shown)


supakeen (approved):
Nice. This could lead to decoupling package set names from image types perhaps? That'd remove a bunch of not-really-necessary definitions in the YAML for Fedora as well.

achilleas-k (approved):
Approving, but I'd like to consider an alternative.  I think that it might be simpler and more flexible to have multiple package sets per image type, keyed by some keywords we define globally like "payload" and "installer".

I could be wrong about this being simpler though.  If we only have a couple of these cases, then the solution in this PR is preferable.

PR #1293, opened 2025-03-06
many: make bootstrap container usage a manifest option

mvo5:
[draft as a) it lacks tests b) there are multiple ways of doing this and this outlines the one I like best but that may not match with the others of course]

This commit adds the option to use a bootstrap container when
generating the manifest. This allows a seamless experience when
using e.g. image-builder:
```console
$ image-builder build --arch=riscv64 minimal-raw --distro fedora-42
```

The approach is to make the distro bootstrap container ref part of
the manifest struct just like we set "Distro" there. This is not
ideal, it would be much nicer if manifest would hold the full
`distro.Distro` and this way we could simply run something like
`distro.BootstrapContainerRef()` when needed. However because
of circular inputs that is not possible.

The other option is to make it part of every `imageFunc()` call,
this was suggested by Achilleas (thanks!) this requires some
refactor. Or we could make it part of `InstantiateManifest()`
and pass a new generalized `RepoAndContainersConfig` there that
would contain both the []rpmmd.RepoConfig and the BootstrapRef.

Note that I think eventually the bootstrap container ref will be
part of the "distro.yaml" and the currently (slightly ugly) helper
`bootstrapContainerFor()` will go away.

The matching and small image-builder-cli changes are in https://github.com/osbuild/image-builder-cli/compare/main...mvo5:cross-arch-for-all-2?expand=1 (and the cross-build test in there passes locally but takes ~30min for all 4 tested arches and it lacks a check that inside the generate container image we actually have binaries of the expected architecture)

Note that cross-arch building will not support selinux in the buildroot yet - see the second commit about how we can still archive this in the future.

### DIFF
--- a/pkg/distro/distro.go
+++ b/pkg/distro/distro.go
@@ -134,6 +134,8 @@ type ImageOptions struct {
 	Subscription     *subscription.ImageOptions `json:"subscription,omitempty"`
 	Facts            *facts.ImageOptions        `json:"facts,omitempty"`
 	PartitioningMode disk.PartitioningMode      `json:"partitioning-mode,omitempty"`
+
+	UseBootstrapContainer bool `json:"use_bootstrap_container,omitempty"`
 }
 
 type BasePartitionTableMap map[string]disk.PartitionTable
--- a/pkg/distro/fedora/imagetype.go
+++ b/pkg/distro/fedora/imagetype.go
@@ -310,6 +310,9 @@ func (t *imageType) Manifest(bp *blueprint.Blueprint,
 	}
 	mf := manifest.New()
 	mf.Distro = manifest.DISTRO_FEDORA
+	if options.UseBootstrapContainer {
+		mf.DistroBootstrapRef = bootstrapContainerFor(t)
+	}
 	_, err = img.InstantiateManifest(&mf, repos, t.arch.distro.runner, rng)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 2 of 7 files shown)


achilleas-k:
I think the bootstrap container ref should be treated like any other source during the manifest instantiation and serialization flow.  I don't understand why it needs to be part of the `InstantiateManifest()` call.

So, right now we have:
(setting aside our plans to simplify this flow for now)
```go
// options contains ostree refs, so ignore the inconsistencies for now
img, err := t.image(w, t, bp, options, staticPackageSets, containerSources, rng)

mf := manifest.New()
img.InstantiateManifest(&mf, repos, runner, rng)

// get source specs
pkgSets := mf.GetPackageSetChains()
containerSources := mf.GetContainers()
commitSources := mf.GetCommits()

// resolve each
packages := depsolve(packages)
containers := resolveContainers(containerSources)
commits := resolveCommits(commitSources)

finalmf := mf.Serialize(packages, containers, commits)
```

I think the bootstrap container should come from outside this flow, just like we do with repositories.  It can be a config file we embed and can override externally.  Exactly like we do with repositories.

Unlike repositories though, I like that it goes into the image function like you did in the PR.  In fact, maybe repositories should go into the image function too (but, again, let's ignore that for now).

mvo5:

mvo5:
Closing (at least for now) in favor of https://github.com/osbuild/images/pull/1342

PR #1294, opened 2025-03-06
packagesets: add yaml based merge support

mvo5:
This commit adds yaml based merge support for package sets by adding another level of indirection. This is already a useful feature so that we can e.g. avoid duplicating the `cloud_base` or `anaconda` package sets.

It works by adding another layer of map under the top-level image type key. This is typically just the image type name again. But it allows us to write:
```yaml
qcow2: &qcow2
  <<: *cloud_base
  qcow2:
    include:
      - "qemu-guest-agent"
```
and benefit from the yaml mapping merge support. It would be more natural to have a sequence here but merging of sequences is not support so we need this workaround.

Another alternative would be we introduce a new custom key next to `condition` like `merge` and to this ourselfs but that is more custom code.

Another splitout/cleanup of https://github.com/osbuild/images/pull/1283 where we will need this for e.g. the SAP packages.

### DIFF
--- a/pkg/distro/packagesets/fedora/package_sets.yaml
+++ b/pkg/distro/packagesets/fedora/package_sets.yaml
@@ -1,922 +1,605 @@
 ---
-cloud_base:  # not used directly; added for reference
-  include:
-    - "@Fedora Cloud Server"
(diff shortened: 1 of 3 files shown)


mvo5:

PR #1295, opened 2025-03-06
[RFC] packagesets: add basic imagetype yaml to merge pkgSets

mvo5:
[draft as this is one possible way of doing it, all fedora package_sets.yaml needs converting, I like this because its very symetric but it is not perfect, alternative version of https://github.com/osbuild/images/pull/1294]

It works by adding new toplevels "image_types" and "virtual_package_sets" in the package_sets.yaml and then using anchors/refs to refer to them in the image_type yaml.

It allows us to write:
```yaml
# this key is optional but would come in handy for cloud_base in fedora
virtual_package_sets:
  base: &base_pkgset
    include: [from-base-inc]
    exclude: [from-base-exc]
    condition:
      distro_name:
        test-distro:
          include: [from-base-condition-inc]
          exclude: [from-base-condition-exc]

image_types:
  other_type:
    package_sets:
      - &other_type_pkgset
        include: [from-other-type-inc]
        exclude: [from-other-type-exc]
  test_type:
    package_sets:
      - *base_pkgset
      - *other_type_pkgset
      - include: [from-type-inc]
        exclude: [from-type-exc]
        condition:
          distro_name:
            test-distro:
              include: [from-condition-inc]
              exclude: [from-condition-exc]
```
The downside of this approach is that:
1. it requires the anchor to be there before the ref, which can be confusing (we could solve this later via includes)
2. if the anchor is forgotten the error becomes very obscure
3. its a bit "hardcore" yaml

The upside is that it is very symetrical and actually feels quite nice.

### DIFF
--- a/pkg/distro/packagesets/loader.go
+++ b/pkg/distro/packagesets/loader.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/fs"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/osbuild/images/internal/common"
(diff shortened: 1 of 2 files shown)


supakeen:
I like this one more than #1296, I'll take a look at the implementation tomorrow.

mvo5:

PR #1296, opened 2025-03-06
[RFC] packagesets: add basic imagetype yaml to merge pkgSets (non-symetric)

mvo5:
[draft as this is one possible way of doing it, all fedora package_sets.yaml needs converting, not a huge fan because it is not symmetrical but it is very simple. maybe different naming would make it feel less asymetric.  alternative version of https://github.com/https://github.com/osbuild/images/pull/1294 and https://github.com/osbuild/images/pull/1295]

It works by adding a new toplevel "image_types" in the package_sets.yaml.

It allows us to write:
```yaml
image_types:
 qcow2:
  package_sets:
   - cloud_base_pkgset
   - qcow2_pkgset

package_sets:
  cloud_base_pkgset:
    include: [foo]
    exclude: [bar]
  qcow2_pkgset:
    include: [qemu-guest-agent]
```
The downside of this approach is that it is nice to read but not symetrical, i.e. one cannot just include the pkgset verbatim in an `image_type`, it always needs to be defined inside the "package_sets" key. It also uses the same "package_sets" name for two different things which is also slightly not nice.

### DIFF
--- a/pkg/distro/packagesets/loader.go
+++ b/pkg/distro/packagesets/loader.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/fs"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/osbuild/images/internal/common"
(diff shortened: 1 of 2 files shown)


supakeen:
Stylistically I prefer #1295, it also seems like that one would more easily map on people who have knowledge of YAML as opposed to this one.

mvo5:

PR #1298, opened 2025-03-07
packagesets: add basic imagetype yaml to merge pkgSets

mvo5:
[this is my personal winner in the race for best-yaml-for-shareing-pkgsets, based on a new idea of @ondrejbudai (thanks!), see also https://github.com/osbuild/images/pull/1294 and https://github.com/osbuild/images/pull/1296 and https://github.com/osbuild/images/pull/1295]

It works by adding new toplevels "image_types" in the package_sets.yaml and then using anchors/refs to refer to them in the image_type yaml. To allow sharing yaml that is not part of an image_type it also introduces a top-level `.common` that can have `any` yaml in it.

It allows us to write:
```yaml
.common:
  base: &base_pkgset
    include: [from-base-inc]
    exclude: [from-base-exc]
    condition:
      distro_name:
        test-distro:
          include: [from-base-condition-inc]
          exclude: [from-base-condition-exc]

image_types:
  other_type:
    package_sets:
      - &other_type_pkgset
        include: [from-other-type-inc]
        exclude: [from-other-type-exc]
  test_type:
    package_sets:
      - *base_pkgset
      - *other_type_pkgset
      - include: [from-type-inc]
        exclude: [from-type-exc]
        condition:
          distro_name:
            test-distro:
              include: [from-condition-inc]
              exclude: [from-condition-exc]
```
The downside of this approach is that:
1. it requires the anchor to be there before the ref, which can be confusing (we could solve this later via includes)
2. if the anchor is forgotten the error becomes very obscure
3. its a bit "hardcore" yaml

The upside is that it is very symetrical and actually feels quite nice.

### DIFF
--- a/pkg/distro/packagesets/fedora/package_sets.yaml
+++ b/pkg/distro/packagesets/fedora/package_sets.yaml
@@ -1,922 +1,600 @@
 ---
-cloud_base:  # not used directly; added for reference
-  include:
-    - "@Fedora Cloud Server"
-    - "chrony"  # not mentioned in the kickstart anaconda pulls it when setting the timezone
-    - "langpacks-en"
-  exclude:
-    - "dracut-config-rescue"
-    - "firewalld"
-    - "geolite2-city"
-    - "geolite2-country"
-    - "plymouth"
+.common:
+  cloud_base_pkgset: &cloud_base_pkgset
+    include:
+      - "@Fedora Cloud Server"
+      - "chrony"  # not mentioned in the kickstart anaconda pulls it when setting the timezone
+      - "langpacks-en"
+    exclude:
+      - "dracut-config-rescue"
+      - "firewalld"
+      - "geolite2-city"
+      - "geolite2-country"
+      - "plymouth"
 
-qcow2: &qcow2
-  include:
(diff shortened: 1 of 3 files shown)


achilleas-k (approved):
We discussed this a bit offline but I'd like to repeat my thoughts here for posterity.

I like this!  At first I was a bit hesitant about using too many yaml features, but I think it makes sense.  It means we don't have to have obscure code constructs and internal rules for merging sets and if people want to define their sets without using all the references and fancy merging they can just write straight-up plain yaml without any of this and without needing to think about how it might interact with our internals.

supakeen (approved):
Yep, this is definitely the winner in my book. As close to "standards" as one can get with YAML I guess? :)

mvo5:
Thanks! I had to rebase because of conflicts but this should be good now.

supakeen (dismissed):

ondrejbudai (approved):
I love it! Feels very straightforward. :)

thozza (approved):
Looks great, I love it :rocket:

ondrejbudai (approved):
Go!

PR #1299, opened 2025-03-07
Add cloud init options to WSL images

croissanne:
This mirrors the Ubuntu images in the MS store. The WSL datasource looks for userdata on the Windows host.

---

also see https://cloudinit.readthedocs.io/en/latest/reference/datasources/wsl.html

### DIFF
--- a/pkg/osbuild/cloud_init_stage.go
+++ b/pkg/osbuild/cloud_init_stage.go
@@ -30,6 +30,7 @@ type CloudInitConfigFile struct {
 	Datasource     *CloudInitConfigDatasource `json:"datasource,omitempty"`
 	DatasourceList []string                   `json:"datasource_list,omitempty"`
 	Output         *CloudInitConfigOutput     `json:"output,omitempty"`
+	Network        *CloudInitConfigNetwork    `json:"network,omitempty"`
 }
 
 // Represents the 'system_info' configuration section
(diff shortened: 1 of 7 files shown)


supakeen (dismissed):
Sure, we could probably have a single place to define the WSL related cloudinit options and then refer to those in the image types to remove a bunch of duplicates? Not really a blocker as I also like locality of information :)

There are some testcases that are failing, so those should be addressed:

```
panic: datasource WSL is not allowed, only [Azure Ec2 None] are allowed [recovered]
	panic: datasource WSL is not allowed, only [Azure Ec2 None] are allowed
```

You can add it here: https://github.com/osbuild/images/blob/main/pkg/osbuild/cloud_init_stage.go#L92 it's already allowed in the stage itself.

croissanne:
oh woops, missed that

pkg/osbuild/cloud_init_stage.go
```diff
@@ -160,3 +170,10 @@ func (du CloudInitConfigDefaultUser) validate() error {
 	}
 	return nil
 }
+
+func (n CloudInitConfigNetwork) validate() error {
+	if n.Config == "disabled" {
```
supakeen, line 175: `!=`?

supakeen (dismissed):

croissanne:
ah I think I need to regen manifests?

supakeen:
> ah I think I need to regen manifests?

Don't think so no, it seems the tests are actually failing on a file not being available or at a different path from where the stage tries to copy it from.

croissanne:
accidentally closed this, reopened at https://github.com/osbuild/images/pull/1321 ; had to figure out some more package sets / excludes.

PR #1305, opened 2025-03-10
Update devcontainer configuration

kksat:
Use fedora:latest for devcontainer
Install project dependencies into devcontainer during build
Download all necessary go modules during devcontainer first run

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-  "name": "images",
+  "name": "Fedora",
   "build": {
     "dockerfile": "Dockerfile",
     "context": ".."
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -13,5 +13,6 @@
   "extensions": [
     "golang.Go",
     "GitHub.vscode-pull-request-github"
-  ]
+  ],
+  "postStartCommand": "go mod tidy"
 }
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "images",
+  "name": "Fedora",
```
achilleas-k, line 2: I don't use the devcontainer myself so I might be wrong about how this works, but doesn't this set the name for the container itself?  Is it a good idea to call it `Fedora` rather than something more specific, like the name of the project?

achilleas-k (approved):
LGTM, but small question inline.

.devcontainer/devcontainer.json
kksat: One of the places image name is used - when GitHub codespaces created. Project name is evident without additional indication, but what what is codespace is about is not. That is why I think naming container is beneficial in this case. See screenshot.
image

achilleas-k: This isn't a major issue, btw.  Just curious about best practices.  If you'd rather keep it named "Fedora", I'm good to merge.

kksat: I do not know best practices in regards to devcontainers unfortunately. Just relying on whatever seems reasonable.

schuellerf:
@supakeen I don't use the dev containers. In the "getting started usecase" I use containers to really run the application but not from within e.g. vscode.

PR #1300, opened 2025-03-07
 rhel10: extract Fedora package sets into a single YAML file (2)

mvo5:
An improved version of https://github.com/osbuild/images/pull/1283 based on https://github.com/osbuild/images/pull/1298

PR #1307, opened 2025-03-10
[RFC] test: add new manifest reference tests

mvo5:
*TODO* as a middle ground the idea of having just the list of hashes of the generated manifests was suggested and tooling to create diffs so that we can avoid carrying all the manifests in the git repo

[edit: as expected lots of resistance :) also one suggestion was to make the yaml instead of json to make them easier to diff/merge so less merge conflicts in e.g. the merge-queue (thanks @achilleas-k!)]

This PR adds new manifest reference tests to capture when we
core aspects of the generated manifests. This is inspired by the
work we did in `otk` [0].

The test is very fast, there is no depsolving or other network
activity, but because "fake" packages get generated we will
notice any changes in the packagelists.

This is very useful when working on the yamlification of the
package lists (and later the other parts) because this way
we will spot any differences right away.
See https://github.com/osbuild/images/pull/1300/ for how
this is used.

[0] https://github.com/osbuild/otk/blob/main/test/test_against_images_refs.py

### DIFF
--- a/pkg/osbuild/rpm_stage.go
+++ b/pkg/osbuild/rpm_stage.go
@@ -126,15 +126,23 @@ func NewRpmStageSourceFilesInputs(specs []rpmmd.PackageSpec) *RPMStageInputs {
 }
 
 func pkgRefs(specs []rpmmd.PackageSpec) FilesInputRef {
-	refs := make([]FilesInputSourceArrayRefEntry, len(specs))
-	for idx, pkg := range specs {
+	refs := make([]FilesInputSourceArrayRefEntry, 0, len(specs))
+
+	// skip duplicated pkgs
+	seenChksums := make(map[string]bool, len(specs))
+	for _, pkg := range specs {
+		if seenChksums[pkg.Checksum] {
+			continue
+		}
+		seenChksums[pkg.Checksum] = true
+
 		var pkgMetadata FilesInputRefMetadata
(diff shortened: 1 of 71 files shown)


supakeen:
This is using `pytest` style names and layouting for a Go project. None of the makefiles or CI is adjusted and the test cases won't get ran when I do the (expected) `go test ./...`.

I like the approach (it worked really well in `otk`) but can we please run this through a normal Go testcase instead of using `pytest` only to run a subprocess? :)

mvo5:
> This is using `pytest` style names and layouting for a Go project. None of the makefiles or CI is adjusted and the test cases won't get ran when I do the (expected) `go test ./...`.
> 
> I like the approach (it worked really well in `otk`) but can we please run this through a normal Go testcase instead of using `pytest` only to run a subprocess? :)

Yeah, happy to tweak this in whatever shape this is preferred. pytest was the quickest do demo the idea for me.

mvo5:
We had an (offline) discussion about this last week and there was strong disagreement about diffing manifests (even non-depsolved ones as in here) so I'm closing this.

We may explore options in the fuure:
1. a bot that posts the diff against the reference on PRs
2. putting the hashes of the manifests into git instead of the full manifest so that we can detect and ACK changes 

both need tooling which this PR does not provide so a new one will be need to be created.

PR #1309, opened 2025-03-10
rpmmd: add json tags to PackageSet struct

croissanne:
Added a comment to explain the weird values they carry.

---

The way to make them lowercase is to change the depsolve job struct in osbuild-composer so it is backwards compatible with the old (=current) struct, and then change the tags.

### DIFF
--- a/pkg/rpmmd/repository.go
+++ b/pkg/rpmmd/repository.go
@@ -139,11 +139,15 @@ func (pkg Package) ToPackageInfo() PackageInfo {
 // to exclude. The Repositories are used when depsolving this package set in
 // addition to the base repositories.
 type PackageSet struct {
-	Include         []string
-	Exclude         []string
-	EnabledModules  []string
-	Repositories    []RepoConfig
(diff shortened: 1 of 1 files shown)


supakeen (approved):

achilleas-k (approved):
LGTM but!

Should we make a plan to change this in composer?  Perhaps composer should define its own internal representation which can read the old stuff and then also introduce new snake_case_keys?

croissanne:
> LGTM but!
> 
> Should we make a plan to change this in composer? Perhaps composer should define its own internal representation which can read the old stuff and then also introduce new snake_case_keys?

yes to both

thozza (changes requested):
I'm not sure that this is a good idea.

> LGTM but!
> 
> Should we make a plan to change this in composer? Perhaps composer should define its own internal representation which can read the old stuff and then also introduce new snake_case_keys?

This sounds like the thing that should be implemented instead. What is the added value of adding tags here to a struct that should be kept internal and not used for serialization?

PR #1310, opened 2025-03-10
tools: add new "gen-manifests-diff" tool

mvo5:
gen-manifests-diff is a helper to diff the generated manifests against what the current "main" of the images library would do.

Alternatively a revision can be given as the first argument and with that the diff is calculated against that. All images types and architectures are used by default. It is still very fast (typically 3s).

Example output:
```diff
calculating diff against d285ec65bac83fad1b36e88b50cecd5f5a5612f9
found differences between /tmp/tmpbjhime7p/new and reference manifests:
diff -uNr /tmp/tmpbjhime7p/ref/centos_9-aarch64-edge_ami-edge_ostree_pull_user.json /tmp/tmpbjhime7p/new/centos_9-aarch64-edge_ami-edge_ostree_pull_user.json
--- /tmp/tmpbjhime7p/ref/centos_9-aarch64-edge_ami-edge_ostree_pull_user.json   2025-03-10 20:55:20.736477268 +0100
+++ /tmp/tmpbjhime7p/new/centos_9-aarch64-edge_ami-edge_ostree_pull_user.json   2025-03-10 20:55:22.294453956 +0100
@@ -161,7 +161,13 @@
             },
             "kernel_opts": [
               "luks.uuid=e3eb0343-83b0-441c-a8aa-6a715afc1d39",
-              "console=tty0 console=ttyS0,115200n8 net.ifnames=0 nvme_core.io_timeout=4294967295 modprobe.blacklist=vc4 rw coreos.no_persist_ip",
+              "console=tty0",
+              "console=ttyS0,115200n8",
+              "net.ifnames=0",
+              "nvme_core.io_timeout=4294967295",
+              "modprobe.blacklist=vc4",
+              "rw",
+              "coreos.no_persist_ip",
               "ignition.platform.id=metal",
               "$ignition_firstboot"
             ]
...
```

This is (hopefully) the uncontroversial part of
https://github.com/osbuild/images/pull/1307
and already useful when checking that a yamlification (or other change) is not resulting in any regressions. Also useful to see if a change results in the expected change(s).

I still think it would be nice to have this as part of our tests (like in #1307) but that is a bigger discussion of course :)

### DIFF
--- a/tools/gen-manifests-diff
+++ b/tools/gen-manifests-diff
@@ -1,0 +1,72 @@
+#!/usr/bin/python3
+#
+# gen-manifests-diff is a helper to diff the generated manifests
+# against what the current "main" of the images library would do.
+#
+# Alternatively a revision can be given as the first argument
+# and with that the diff is calculated against that. All images
+# types and architectures are used by default. It is still
+# very fast (typically 3s).
+
+import os
+import pathlib
+import subprocess
+import sys
+import tempfile
+
+# TODO: omit once we have a "riscv64" mirror and sources entry
+arches = ["x86_64", "aarch64", "ppc64el", "s390x"]
+
+
+def top_srcdir() -> pathlib.Path:
+    d = pathlib.Path(__file__).parent
+    while not (d / "cmd" / "gen-manifests").exists():
+        d = d.parent
+        if d == "/":
+            raise RuntimeError("cannot find gen-manifests dir")
+    return d
+
+
+def run_gen_manifests(cmd, output_dir):
+    env = os.environ.copy()
+    env["OSBUILD_TESTING_RNG_SEED"] = "0"
+    subprocess.run(cmd + [
+        "-packages=false",
+        "-metadata=false",
+        "-containers=false",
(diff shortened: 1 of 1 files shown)


supakeen:
Thank you, I think we all had our own implementations or workflows surrounding this.

supakeen (approved):

bcl (approved):
Nice!

achilleas-k:
I didn't get a chance to review this, but it's all good.  One thing to note though is that I think it should become part of the developer docs, especially the part that describes the manifest generation and diffing workflow: https://github.com/osbuild/images/blob/main/docs/developer/cmds.md#manifest-generation

mvo5:
> I didn't get a chance to review this, but it's all good. One thing to note though is that I think it should become part of the developer docs, especially the part that describes the manifest generation and diffing workflow: https://github.com/osbuild/images/blob/main/docs/developer/cmds.md#manifest-generation

Makes sense! I opened https://github.com/osbuild/images/pull/1320

PR #1312, opened 2025-03-11
osbuild: do not write duplicated package IDs

mvo5:
This commit is a very naive fix to avoid writing duplicated packages
into the osbuild manifest. Writing multiple identical package ids
into the manifest is no real problem but with the new manifest
compare tooling it creates unneeded diffs (if there is reason
to do/allow duplicated IDs and I'm missing that my apologies,
I will close this of course).

This shows up in e.g. https://github.com/osbuild/images/pull/1300

Note that this is incredible naive - the right thing to do here would
be to make the []rpmmd.PackageSpec a type that abstracts that
it is a list and has an "Add()" that will automatically de-dup. But
that would distract (a lot) from the yaml-ifaction so this naive
version first.

The PR also contains a test change in the second commit. This
is totally optional but I found it nicer to look at the expected data
structure to see what is going on and it feels nice to know exactly
what is going on and what is converted in the result. I.e. I was
surprised that we set "IgnoreSSL" to true in the test but the resulting
FilesInputSourceArrayRef does not contain this. In the old style of
the test less visible if it is not checked or not there. I don't see a
real downside, the previous test was also casting into the right types
already (so not using interfaces) but I will not fight over it and can
drop the commit.

### DIFF
--- a/pkg/osbuild/rpm_stage.go
+++ b/pkg/osbuild/rpm_stage.go
@@ -126,15 +126,23 @@ func NewRpmStageSourceFilesInputs(specs []rpmmd.PackageSpec) *RPMStageInputs {
 }
 
 func pkgRefs(specs []rpmmd.PackageSpec) FilesInputRef {
-	refs := make([]FilesInputSourceArrayRefEntry, len(specs))
-	for idx, pkg := range specs {
+	refs := make([]FilesInputSourceArrayRefEntry, 0, len(specs))
+
+	// skip duplicated pkgs
+	seenChksums := make(map[string]bool, len(specs))
+	for _, pkg := range specs {
+		if seenChksums[pkg.Checksum] {
+			continue
+		}
+		seenChksums[pkg.Checksum] = true
+
 		var pkgMetadata FilesInputRefMetadata
 		if pkg.CheckGPG {
 			pkgMetadata = &RPMStageReferenceMetadata{
 				CheckGPG: pkg.CheckGPG,
 			}
 		}
-		refs[idx] = NewFilesInputSourceArrayRefEntry(pkg.Checksum, pkgMetadata)
+		refs = append(refs, NewFilesInputSourceArrayRefEntry(pkg.Checksum, pkgMetadata))
 	}
 	return NewFilesInputSourceArrayRef(refs)
 }
--- a/pkg/osbuild/rpm_stage_test.go
+++ b/pkg/osbuild/rpm_stage_test.go
@@ -154,6 +154,20 @@ func TestNewRpmStageSourceFilesInputs(t *testing.T) {
 			CheckGPG:       true,
 			IgnoreSSL:      true,
 		},
+		// duplicated on purpose to ensure we only add the same
+		// ID only once to make the manifests more stable.
+		{
+			Name:           "package-with-md5-checksum",
+			Epoch:          1,
+			Version:        "3.4.2.",
+			Release:        "5.el9",
+			Arch:           "x86_64",
+			RemoteLocation: "https://example.com/repo/Packages/package-with-md5-checksum-4.3.2-5.el9.x86_64.rpm",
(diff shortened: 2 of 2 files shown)


thozza (changes requested):
Thanks for the PR.

I have two comments:

1. As far as I can tell, the duplication is not happening in practice. The duplication happens when we combine package sets (the base package set, extended with any packages needed for used customizations) with user-requested packages. However, this is the state only **before** we depsolve the package set. IOW, the depsolve request can contain the `firewalld` package multiple times, but the depsolving result will **never** contain duplicate packages. The reason that you are seeing this behavior is due to generating manifests without depsolving package sets in the `gen-manifests-diff` tool that you've recently introduced https://github.com/osbuild/images/blob/48252e2b3c67df5e92266e215552fd5088a42796/tools/gen-manifests-diff#L34

2. Diffing manifests without depsolving package sets is OK for a quick check, but not enough as proof that nothing changed. I know that depsolving can take significant time, but from my point of view, it is a must if used to check changes to generated manifests when changing image definitions. Omitting depsolving skips at least the following code paths:
   - Enablement of the package gpgcheck based on the repo config.
   - Weak dependencies behavior can be turned on/off in `ImageConfig` on the image type level or at the distro level.

My point is that skipping depsolving is sometimes acceptable and is always nice as a quick check. However, one should decide to (not) depsolve intentionally and consider the implications. IMHO, diffing with depsolved package sets should be done at least once, as the last check to verify one's changes.

The same applies to skipping the resolution of containers and commits, but these are not used in every image type and are not part of each manifest generation.

IMHO, the correct place to fix the ID duplication is in the `gen-manifests` tool, which is "mocking" the depsolving. So in https://github.com/osbuild/images/blob/main/cmd/gen-manifests/main.go#L312

mvo5:
Fair points, I will look into the mocked depsolver then

mvo5:
[..]
>     2. Diffing manifests without depsolving package sets is OK for a quick check, but not enough as proof that nothing changed. I know that depsolving can take significant time, but from my point of view, it is a must if used to check changes to generated manifests when changing image definitions. Omitting depsolving skips at least the following code paths:
>        
>        * Enablement of the package gpgcheck based on the repo config.
>        * Weak dependencies behavior can be turned on/off in `ImageConfig` on the image type level or at the distro level.
> 
> 
> My point is that skipping depsolving is sometimes acceptable and is always nice as a quick check. However, one should decide to (not) depsolve intentionally and consider the implications. IMHO, diffing with depsolved package sets should be done at least once, as the last check to verify one's changes.
[..]

@thozza Thanks for these interesting and valid points. I wonder however if we can overcome these limitations? In the mockdepsover in gen-manifests we encode the repositories as pseudo packages so we could encode more (like pgchecks and weak-dep hangling). This is not "nice" but does give us assurance that nothing has changed and it is incredible quick, as pointed out in https://github.com/osbuild/images/pull/1307 it takes only 3sec and 80mb of data which makes it practical to run as a unit test and as part of the normal dev iteration. Compared to the minutes/hours and 1GB+ of data the full run takes.  Maybe we can not get rid of the full run but it seems worth exploring how close we can get, wdyt?

thozza:
> @thozza Thanks for these interesting and valid points. I wonder however if we can overcome these limitations? In the mockdepsover in gen-manifests we encode the repositories as pseudo packages so we could encode more (like pgchecks and weak-dep hangling). This is not "nice" but does give us assurance that nothing has changed and it is incredible quick, as pointed out in #1307 it takes only 3sec and 80mb of data which makes it practical to run as a unit test and as part of the normal dev iteration. Compared to the minutes/hours and 1GB+ of data the full run takes. Maybe we can not get rid of the full run but it seems worth exploring how close we can get, wdyt?

Yes, I can see the benefits of this approach and think that it is worth exploring. :+1:

PR #1313, opened 2025-03-11
Add check-spec-deps-action action

thozza:
Add a composite action, which can be used by projects that depend on `osbuild/images` and are packaged as RPM, to check that the dependencies in their SPEC file specify the minimum versions as required by `osbuild/images`.

For now, only `osbuild` dependency is checked.

The goal is to use this action in `osbuild-composer` and `image-builder-cli` repositories to check the SPEC files.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,34 @@
+---
+name: Tests
+
+on:
+  pull_request:
+    branches:
+      - "*"
+  push:
+    branches:
+      - check-spec-deps-action
+
+jobs:
+  python-lint:
+    name: "🐍 Lint"
+    runs-on: ubuntu-24.04
+    container: registry.fedoraproject.org/fedora:latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Install job dependencies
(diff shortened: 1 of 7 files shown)


thozza:
I've tested this on my `osbuild-composer` fork -> https://github.com/thozza/osbuild-composer/actions/runs/13797231624/job/38591679732?pr=535

thozza:
> Is it correct that this PR is not targeting `main`?

It is intentional from my side since this is a composite GH action related to the repo, but I don't think it should live on `main` ( as it has its own README and actions). We have precedence for this approach in https://github.com/osbuild/release-action/tree/create-tag

thozza:
Snyk reports `Path Traversal` issue on line no. 64, but this is a false positive. The case is handled above in the function. There is apparently no easy way how to mark specific issues to be ignored by Snyk :shrug: (i.e. comment in the code, etc...).

achilleas-k (approved):
Pretty cool!!

achilleas-k:
Should we make different branch protection rules for this one?  Maybe we don't need two approvals here, for example.

supakeen (approved):

PR #1318, opened 2025-03-12
Introduce new Azure CVM image type

achilleas-k:
Add a new image type for Azure called 'vhd-cvm' (Confidential VM).  This is a UKI-based image.

New configs and functionality introduced to support this image:
- `platform.Bootloader`: an enum that takes one of four values: `None`, `Grub2`, `zipl`, and `UKI`.  Aarch64 and ppc64le always return `Grub2`, s390x always returns `zipl`.  x86_64 platforms that have `None` set will return `Grub2`.  This is temporary and meant to reduce the size of the patch.  I'm preparing a follow-up that replaces all bootloader-related decisions with a check against the `platform.Bootloader` value.  I will also make `None` the value used for images that don't configure bootloaders (containers, wsl).
- `ImageConfig.DefaultKernelName`: For now, this overrides any kernel customization.  This is fine for our current use case, but in the future we might want to control the fallback logic differently.

Some of the most interesting properties of this image:
- The DefaultKernelName is kernel-uki-virt.  This is currently the only image type that defines a DefaultKernelName.
- The image config for the new image type masks the firewalld service.  In the kickstart file used until now, the command 'firewall-offline-cmd --disabled' is used to disable the firewalld.service [1].  We could add an option to the firewall stage in osbuild, but this is simpler and more direct.  More importantly, masking makes sure the firewall doesn't get enabled by a firstboot process or a systemctl preset-all.
- Sets network bootproto to DHCP for eth0 for CVM image type.  We need to specify a device name to configure.  The kickstart file doesn't specify one, however this is considered deprecated behaviour [2].
- We need to set an environment variable, `$BOOT_ROOT`, to the root of the ESP during rpm installation.  See osbuild/osbuild#2045.
- The image definition writes a (UTF-16-LE-encoded) CSV file in the ESP which controls the fallback boot to the UKI.  Note that this is a temporary workaround.  We expect that the kernel-bootcfg command from the python3-virt-firmware package will gain the ability to write these files offline during the RHEL 9.7 / 10.1 development cycle.


mvo5 (commented):
I just read through it so far, looks very nice and straightforward, just one wondering inline (so far). 

And one question from running `gen-manifests-diff`. I noticed we create a `org.osbuild.fix-bls` entry and I wonder if we need this with the new BOOT_ROOT environemnet?

achilleas-k:
> noticed we create a `org.osbuild.fix-bls` entry and I wonder if we need this with the new BOOT_ROOT environemnet?

Oh, I missed that.  Thanks for catching it.

achilleas-k:
> > noticed we create a `org.osbuild.fix-bls` entry and I wonder if we need this with the new BOOT_ROOT environemnet?
> 
> Oh, I missed that. Thanks for catching it.

Fixed!

achilleas-k:
I disabled the fix-bls stage by setting `NoBLS = true` on the image config, but perhaps the UKI option in the platform can be used to control that instead.  So, for example, https://github.com/achilleas-k/images/blob/532f2f9d03b82da07074d4fcb204974e2112829f/pkg/manifest/os.go#L472 can become:
```go
	if !p.NoBLS && platform.Bootloader == platform.BOOTLOADER_GRUB2 {
```

achilleas-k:
I'm pushing image definition changes based on feedback from QE.

achilleas-k:
osbuild/osbuild#2045 has been merged.  I've updated the commit ID in the Schutzfile an I'm marking the PR ready.

internal/common/helpers.go
```diff
@@ -59,3 +61,15 @@ func ExecError(err error) error {
 	}
 	return err
 }
+
+// EncodeUTF16le encodes a source string to UTF-16 / UCS-2 (LE).
```
supakeen, line 65: Perhaps a nitpick but this is *not* UCS-2 so let's drop the reference to it.

internal/common/helpers_test.go
```diff
@@ -33,3 +34,34 @@ func TestSystemdMountUnit(t *testing.T) {
 		assert.Equal(t, tc.expectedName, name)
 	}
 }
+
+func TestEncodeUTF16le(t *testing.T) {
+	// the function was written to replicate 'iconv -f ASCII -t UCS-2' so these
```
supakeen, line 39: These should be generated with `iconv -f ASCII -t UTF-16` instead, if we're doing UTF-16.

pkg/distro/image_config.go
```diff
@@ -23,6 +23,9 @@ type ImageConfig struct {
 	DefaultTarget       *string
 	Sysconfig           []*osbuild.SysconfigStageOptions
 
+	// The name of the default kernel to use for the image type
```
supakeen, line 26: Perhaps a note about what a "default" kernel is :)

pkg/manifest/os.go
```diff
@@ -943,6 +914,96 @@ func usersFirstBootOptions(users []users.User) *osbuild.FirstBootStageOptions {
...
+// ukiBootCSVfile creates a file node for the csv file in the ESP which
+// controls the fallback boot to the UKI.
+// NOTE: This is a temporary workaround. We expect that the kernel-bootcfg
+// command from the python3-virt-firmware package will gain the ability to
+// write these files offline during the RHEL 9.7 / 10.1 development cycle.
+func ukiBootCSVfile(espMountpoint, kernelVer, vendor string) (*fsnode.File, error) {
+	kernelFilename := fmt.Sprintf("ffffffffffffffffffffffffffffffff-%s.efi", kernelVer)
+	data := fmt.Sprintf("shimx64.efi,%s,\\EFI\\Linux\\%s ,UKI bootentry\n", vendor, kernelFilename)
+
+	// TODO: what about other architectures
```
supakeen, line 975: Perhaps we could assume (can't find a definite quicky) that `aarch64` -> `BOOTAA64` and `x86` -> `BOOTX64`?

pkg/manifest/os.go
```diff
@@ -943,6 +914,96 @@ func usersFirstBootOptions(users []users.User) *osbuild.FirstBootStageOptions {
...
+	csvPath := filepath.Join(espMountpoint, "EFI", vendor, "BOOTX64.CSV")
+
+	return fsnode.NewFile(csvPath, nil, nil, nil, common.EncodeUTF16le(data))
+}
+
+func findESPMountpoint(pt *disk.PartitionTable) (string, error) {
```
supakeen, line 981: ❤️

pkg/platform/platform.go
```diff
@@ -19,6 +19,15 @@ const ( // image format enum
 	FORMAT_OVA
 )
 
+type Bootloader int
+
+const ( // bootloader enum
```
supakeen, line 24: I *love* this moving of bootloaders to platforms instead of toggling on architecture in the `os` pipeline so feel free to ignore this, but there are multiple stages of bootloaders and it seems like we're conflating them into a single bootloader enum?

supakeen (approved):
Very nice, I really only have circumstantial remarks on this from going through the code.

pkg/platform/platform.go
achilleas-k: > Feel free to ignore this, but there are multiple stages of bootloaders and it seems like we're conflating them into a single bootloader enum?

I don't want to ignore this, but maybe not something to address in this PR.  I would love to have a more _correct_ model of reality in the way the code and options are structured.  We should definitely come back to this.

pkg/manifest/os.go
achilleas-k: I looked around a bit and found it hard to find any good references, documentation, or specifications for the CSV file, so I didn't want to make too many assumptions.  But yes, we should parameterise the architecture here for the filename as well as the shim.

internal/common/helpers_test.go
achilleas-k: So I realised while writing this that there are differences but I wasn't entirely sure what they were.  Would love some guidance.

Going by https://www.rodsbooks.com/efi-bootloaders/fallback.html (which always pops up as the first result for this stuff):

> One key point is that BOOT.CSV must be a UCS-2 or UTF-16 file.
> ...
> `$ iconv -t UCS-2 < ~/boot.csv > ~/BOOT.CSV`

makes me think that the two are equivalent in all the ways that matter in this particular instance.  I assume this means, in other words, that no one should ever need to use any of the codepoints not supported by UCS-2.  But I do agree with your general point.  We should be proper and accurate here, so it's probably best to get rid of all mentions of UCS-2 and generate the test cases with `-t UTF-16` (which probably wont change any of the cases, but we would document it as such).

supakeen: In the practical case as long as the input doesn't contain stuff outside of the Basic Multilingual Plane (and thus needs no continuations) there will be no problems. It's just about UCS-2 != UTF-16 even though it might be a very long time (or never) for that to cause problems.

achilleas-k: Yup.  I'll remove any mentions of UCS-2.

pkg/distro/rhel/rhel9/azure.go
```diff
@@ -582,3 +637,94 @@ func defaultAzureImageConfig(rd *rhel.Distribution) *distro.ImageConfig {
...
+						FSTabOptions: "defaults,uid=0,gid=0,umask=077,shortname=winnt",
+						FSTabFreq:    0,
+						FSTabPassNo:  2,
+					},
+				},
+				{
+					Size: 5 * datasizes.GibiByte,
+					Type: disk.RootPartitionX86_64GUID,
+					Payload: &disk.Filesystem{
+						Type:         "xfs",
```
ondrejbudai, line 717: Shouldn't this be `ext4`? I think I saw it in the kickstarts that Vitaly gave us. I think it must be ext4 because the deployment script is shrinking this filesystem, so it can be put into a LUKS container with the same size (so the partition_size = luks_header + filesystem).

achilleas-k: Right!  Thanks for catching it.

ondrejbudai:
btw, we now live in the future, so the package sets need to be moved to the yaml definitions. Crazy, right?

achilleas-k:
Rebased, resolved conflicts, and addressed open comments.

supakeen (approved):

pkg/manifest/os.go
bcl: The BOOT EFI file names are defined in section 3.5.1.1 of the UEFI spec [version 2.11](https://uefi.org/sites/default/files/resources/UEFI_Spec_Final_2.11.pdf) where the arch names are the 'machine type short name' defined by PE format.

pkg/distro/image_config.go
```diff
@@ -23,6 +23,13 @@ type ImageConfig struct {
 	DefaultTarget       *string
 	Sysconfig           []*osbuild.SysconfigStageOptions
 
+	// The name of the default kernel to use for the image type.
+	// NOTE: Currently this overrides the kernel named in the blueprint. The
+	// only image type that uses it is the vhd-cvm, which doesn't allow kernel
+	// selection. The option should generally be a fallback for when the
+	// blueprint doesn't specify a kernel.
+	DefaultKernelName *string
```
mvo5, line 31: (potentially silly question/suggestion as it might be obvious for anyone familar with sysconfig/ukis) 

We set `ImageConfig.Sysconfig.Kernel.DefaultKernel` in many places currently to customize the kernel (e.g. rhel9/gce.go but many more places). Maybe worth a note of the relationship between this new option and the sysconf one? I.e. should setting this always also set the sysconfig option (c.f. also https://github.com/osbuild/images/pull/1350 that also pulls out the default kernel but uses it in sysconfig hence my question)

achilleas-k: Setting the sysconfig option will add the option to the sysconfig stage and we don't always want that.  I think this default kernel should become something we use in the package selection when the blueprint doesn't mention one.  Right now, we always select `kernel` and this is the first image type where that's not what we want.

I'll add a note clarifying that this option doesn't affect the sysconfig stage options in the ImageConfig.

Sidenote: this is a good example for stage options not being directly used in ImageConfig.  Ideally, what we would want to have here is an option that communicates "write sysconfig kernel" but the value for the `default_kernel` option would come from this option to make it impossible to have a different default kernel package and sysconfig `default_kernel`.

achilleas-k:
The last thing that's left for this PR is to add a kernel install script that's been patched upstream but wont make it into 9.6 and 10.0.  While working on that, I opened #1353 to make file creation a bit easier to work with, so I'm going to base this PR off that one.

achilleas-k:
The first five commits now are because this is based on #1353.  I'd like to get that merged separately if possible.

achilleas-k:
I'd like to know what people think about vendoring the kernel install script.  It might be a bit much and we would have to keep it around for as long as we support RHEL 10.0.  I considered writing something simpler, but I think adding the real script is safer.

There's also another issue I haven't considered until now: Updates to the uki-direct package (for example, from a distro upgrade) will install a new script in `/usr/lib/kernel/install.d/`, but any changes to that script will be overridden as long as our script in `/etc/kernel/install.d/` is still around.  I don't know if there's a good way around this issue.

achilleas-k:
Draft: working on hmac stage and waiting for #1353 to be finalised and merged.

achilleas-k:
hmac stage: https://github.com/osbuild/osbuild/pull/2055

pkg/manifest/os.go
```diff
@@ -702,63 +702,65 @@ func (p *OS) serialize() osbuild.Pipeline {
...
-				product := osbuild.GRUB2Product{
-					Name:    p.OSProduct,
-					Version: p.OSVersion,
-					Nick:    p.OSNick,
-				}
+		if p.platform.GetBootloader() == platform.BOOTLOADER_GRUB2 {
+			var bootloader *osbuild.Stage
+			switch p.platform.GetArch() {
+			case arch.ARCH_S390X:
+				bootloader = osbuild.NewZiplStage(new(osbuild.ZiplStageOptions))
```
achilleas-k: This is obviously wrong and makes it so the zipl stage is never added.

test/data/manifest-checksums.txt
```diff
@@ -48,6 +48,7 @@
 0662e3131ab360d21f6635eef4ce8ffabcee6408  rhel_9.0-aarch64-ami-partitioning_lvm.json
 0669a2576d6a0a025b57b61c53f3563da1097441  rhel_9.4-x86_64-edge_vsphere-ostree_filesystem_customizations.json
 070ac33d2263c4b5ab4ef31fa62d19cf17b98ac3  rhel_8.4-x86_64-edge_container-empty.json
+070e1f3c672df6c8c58d9da708d184cf47edaba6  rhel_10.0-s390x-qcow2-empty.json
```
achilleas-k: Bad change.

achilleas-k:
Rebased and fixed conflicts.  With the regular Azure work finishing up, I'll be getting back to this now.

pkg/distro/rhel/rhel10/ami.go
```diff
@@ -131,3 +131,24 @@ func mkEC2SapImgTypeX86_64(rd *rhel.Distribution) *rhel.ImageType {
 
 	return it
 }
+
+// Azure Confidential VM
+func mkAzureCVMImgType(rd *rhel.Distribution) *rhel.ImageType {
```
litian1992, line 136: Should this go to azure.go?

achilleas-k: It should, yes.  Thanks for catching it :)

achilleas-k:
Alright, I'm marking this READY.  Based on https://issues.redhat.com/browse/RHEL-79970, there should be only one change we might want to do (version lock `shim-x64`).  If we end up needing that, I'd like to do it in a separate PR.  It'll also probably need an osbuild stage.

But if we get this merged now, it would be easier to review modifications later.

achilleas-k:
@litian1992 I re-requested a review from everyone in the list of reviewers, which included you since you commented earlier.  You don't need to review the whole patch so feel free to ignore :)

bcl (dismissed):
Looks good.

pkg/distro/rhel/rhel10/azure.go
```diff
@@ -69,3 +69,24 @@ func mkAzureSapInternalImgType(rd *rhel.Distribution, a arch.Arch) *rhel.ImageTy
...
+		rhel.DiskImage,
+		[]string{"build"},
+		[]string{"os", "image", "vpc"},
+		[]string{"vpc"},
+	)
+
+	it.Bootable = true
+	it.DefaultSize = 32 * datasizes.GibiByte
+	it.DefaultImageConfig = imageConfig(rd, "x86_64", "azure-cvm")
+	it.BasePartitionTables = defaultBasePartitionTables
```
litian1992, line 89: The default partition table has xfs root filesystem. CVM requires ext4.

achilleas-k:
Changes in last rebase + push:
- Switched RHEL 10 CVM image root filesystem type to ext4
- Added `chrony` to the image type for both distros.  We want the package to be installed with the default config.  Previously, it was included because we had a chrony configuration set in the image config for CVM (inherited from the Azure base config), but since I removed the configuration line, it also dropped the package.  It's added back now.

supakeen (dismissed):
Had read through the code before and the changes that came after that are still OK.

achilleas-k:
FYI: QE signed off on the scratch image builds from this PR.

pkg/distro/defs/rhel-9/distro.yaml
```diff
@@ -1182,3 +1182,33 @@ image_types:
...
+            - "iwl*"
+            - "kernel"
+            - "kernel-core"
+            - "kernel-modules"
+            - "linux-firmware*"
+
+            - "grub2-common"
+            - "grub2-efi-x64"
+            - "grub2-tools"
+            - "grub2-tools-minimal"
```
thozza, line 1212: I would expect some (if not all) of these to come from the `platform` instead of being defined here.

pkg/distro/rhel/rhel9/distro.go
```diff
@@ -346,6 +346,22 @@ func newDistro(name string, major, minor int) *rhel.Distribution {
 			},
 			mkEC2ImgTypeAarch64(),
 		)
+
+		// CVM is only available starting from 9.6
+		if common.VersionGreaterThanOrEqual(rd.OsVersion(), "9.6") {
```
thozza, line 351: Do we really need this condition?

It is not like we will backport the image type to 9.6. What would be the problem if one builds 9.5 CVM image on 9.7? We can also restrict what is available on-prem and limiting our options for SaaS version seems unnecessary. Especially when this is not a change in behavior, partitioning, or configuration... :thinking:

pkg/manifest/os.go
```diff
@@ -400,6 +400,13 @@ func (p *OS) getBuildPackages(distro Distro) []string {
 		packages = append(packages, tomlPkgsFor(distro)...)
 	}
 
+	if p.platform.GetBootloader() == platform.BOOTLOADER_UKI {
+		// Only required if the hmac stage is added, which depends on the
+		// version of uki-direct. Add it conditioned just on the bootloader
+		// type for now, until we find a better way to decide.
+		packages = append(packages, "libkcapi-hmaccalc")
+	}
```
thozza, line 408: I have a wild idea: If we return the results of each transaction depsolving, in addition to the complete result (which we already discussed at some point), we could depsolve packages such as this one in a separate transaction. We would opportunistically depsolve additional options and include the additional package set in the one to be installed conditionally here.

pkg/distro/defs/rhel-10/distro.yaml
```diff
@@ -297,6 +298,193 @@
...
+          echo "Creating a dataloss warning file at ${MOUNTPATH}/DATALOSS_WARNING_README.txt"
+
+          cat <<'EOF' > "${MOUNTPATH}/DATALOSS_WARNING_README.txt"
+          WARNING: THIS IS A TEMPORARY DISK.
+
+          Any data stored on this drive is SUBJECT TO LOSS and THERE IS NO WAY TO RECOVER IT.
+
+          Please do not use this disk for storing any personal or application data.
+
+          EOF
```
thozza, line 338: Too bad that we lose the option to embed by moving to YAML :cry:

pkg/distro/defs/rhel-10/distro.yaml
```diff
@@ -297,6 +298,193 @@
...
+            After: ["multi-user.target", "cloud-final.service"]
+          "Service":
+            Type: "oneshot"
+            ExecStart: [*dataloss_script]
+            StandardOutput: "journal+console"
+          "Install":
+            WantedBy: ["default.target"]
+    keyboard:
+      keymap: "us"
+      "x11-keymap":
```
thozza, line 355: The keys don't need to be quoted, right? :thinking:

thozza (commented):
Wow, this must have been a lot of work... :heart: 

I've added a few comments and ideas (for the future), but in general, this looks very good :+1:

pkg/distro/rhel/rhel9/distro.go
achilleas-k: I think the uki kernel didn't exist in older versions.  Not entirely sure when it was added.

litian1992: 9.6 is the first supported version. Where 9.4 is missing the kernel-virt-uki-addons package and 9.5 is missing FIPS and so on.

pkg/distro/defs/rhel-9/distro.yaml
thozza: It would make sense to try to remove them from the excludes, since after the other commits in the PR, `grub2` should not be pulled in IIRC...

pkg/distro/rhel/rhel9/distro.go
thozza: ACK. It would be nice to add some additional context to the comment, mainly to ensure that we do not end up wondering in the future if we can't enable it for older RHEL-9 minor versions.

pkg/manifest/os.go
achilleas-k: Interesting.  Yeah, something to think about.

pkg/distro/defs/rhel-9/distro.yaml
achilleas-k: I'll move them around and see which ones are necessary to keep.

achilleas-k: I droppped most of the excludes but kept them in the image type package sets.  One issue is that the firmware exclusion is distro-specific right now, so we can't just put them all in the platform.  I'm not entirely sure if the rest belong in the platform now.

thozza (approved):
LGTM, thanks :+1:

pkg/distro/rhel/images.go
```diff
@@ -37,7 +37,16 @@ func osCustomizations(
...
 
 	if t.Bootable || t.RPMOSTree {
+		// TODO: for now the only image types that define a default kernel are
+		// ones that use UKIs and don't allow overriding, so this works.
+		// However, if we ever need to specify default kernels for image types
+		// that allow overriding, we will need to change c.GetKernel() to take
+		// an argument as fallback or make it not return the standard "kernel"
+		// when it's unset.
 		osc.KernelName = c.GetKernel().Name
+		if imageConfig.DefaultKernelName != nil {
```
mvo5, line 47: Do we need the coresponding change in pkg/distro/generic/images.go ?

pkg/distro/rhel/rhel10/azure.go
mvo5: Just FTR this is addressed via a custom partition table for CVM in the yaml config.

pkg/manifest/os_test.go
```diff
@@ -449,3 +451,106 @@ func collectCopyDestinationPaths(stages []*osbuild.Stage) []string {
 	}
 	return destinationPaths
 }
+
+func TestHMACStageInclusion(t *testing.T) {
```
mvo5, line 455: Thanks for this test!

mvo5 (approved):
This is very nice, thank you! Some tiny questins/ideas inline but no blockers (I think)

Issue #1323, opened 2025-03-13
Packages selected via blueprint customizations can conflict with the base package set excludes


achilleas-k:
I just ran into the following problem while adding `firewalld` to the package selection in [OS.getPackageSetChain()](https://github.com/osbuild/images/blob/d91c5c7f5bc6a7fa14d4b5c5a4897555d1ae4452/pkg/manifest/os.go#L221).

If we add `firewalld` to `packages` but the image type excludes the package in its base set (Fedora 43 minimal-raw), the depsolve fails.

Full description of problem:
A long time ago we (@thozza) introduced the concept of chained depsolving.  We depsolve the base package set first, then take those results and mark them as installed and depsolve the **user** package set afterwards.  This solves the problem of a user selecting a package that is in the image type's exclude list, or when a dependency of that package is in the exclude list.
The current problem is very similar to the above but instead of it being a package that the user selects, it's a package that gets selected from a user customization.

Here's how things work now:
- Image types have a [default `ImageConfig`](https://github.com/osbuild/images/blob/d91c5c7f5bc6a7fa14d4b5c5a4897555d1ae4452/pkg/distro/fedora/imagetype.go#L44).
- The `ImageConfig` and user-provided Blueprint are combined to create [`manifest.OSCustomizations`](https://github.com/osbuild/images/blob/d91c5c7f5bc6a7fa14d4b5c5a4897555d1ae4452/pkg/distro/fedora/images.go#L30-L34), which are part of any customizable ImageKind, for example [`image.DiskImage`](https://github.com/osbuild/images/blob/d91c5c7f5bc6a7fa14d4b5c5a4897555d1ae4452/pkg/image/disk.go#L24).
  - The `OSCustomizations` are copied to the appropriate pipeline, for example the [`OS` pipeline](https://github.com/osbuild/images/blob/d91c5c7f5bc6a7fa14d4b5c5a4897555d1ae4452/pkg/image/disk.go#L60).
- Then, [`OS.getPackageSetChain()`](https://github.com/osbuild/images/blob/d91c5c7f5bc6a7fa14d4b5c5a4897555d1ae4452/pkg/manifest/os.go#L221) returns a list of package sets, the chain to be depsolved.
  - The first package set is the combined set of the base image package set (`OSCustomizations.ExtraBasePackages`), the Environment and Platform packages, and **the packages required by all customizations**.
  - The second package set is just the list of packages defined in the blueprint.

[...]

PR #1324, opened 2025-03-13
Setting a firewall customization installs firewalld

achilleas-k:
Make sure firewalld is available in the image.  Note that org.osbuild.firewall runs 'firewall-offline-cmd' in the os tree using chroot, so we don't need a build package for this.

NOTE: The new test config will fail for Fedora 43 because of issue #1323.  I'm preparing a fix for that issue separately.


mvo5 (dismissed):

supakeen (dismissed):

achilleas-k:
Rebased on #1327

croissanne (approved):

thozza (approved):
Thanks, LGTM. We are finally taking advantage of the :package: :chains:  with more than two transactions :sweat_smile:

Issue #1322, opened 2025-03-13
Should gen-manifests-diff  run against the merge-base by default?


achilleas-k:
https://github.com/osbuild/images/blob/d91c5c7f5bc6a7fa14d4b5c5a4897555d1ae4452/tools/gen-manifests-diff#L64

`gen-manifests-diff` generates a diff between manifests at `HEAD` and `main` from the upstream repository.  This can cause unpredictable results when `main` changes while working on a branch.  For example, if I run `gen-manifests-diff` twice with some time apart but without making any changes, it's possible to get different diffs if a PR gets merged upstream between the two runs.  More importantly, I might get a diff from a change I didn't introduce.  For example, if I add a package or change a stage option, but upstream one or more PRs get merged, the merged PRs can (and probably will) make changes unrelated to my branch and I might end up chasing down the source of the unrelated change in my own branch.

I think it would be wiser for the rev argument to default to `git merge-base origin/main HEAD`.

CC @mvo5

PR #1325, opened 2025-03-14
tools: make `gen-manifests-diff` run against the merge base

mvo5:
This commit tweak `gen-manifests-diff` to run against the merge base of the upstream branch instead of whatever is upstream@main. This avoids potentially confusing output when upstream/main has moved forward since the branching and the diff would include unrelated upstream changes.

Thanks to Achilleas for suggesting this.

Closes: https://github.com/osbuild/images/issues/1322

### DIFF
--- a/tools/gen-manifests-diff
+++ b/tools/gen-manifests-diff
@@ -61,9 +61,21 @@ def manifests_diff(tmp_path, rev):
 
 
 def main():
-    rev = "main"
-    if len(sys.argv) > 1:
-        rev = sys.argv[1]
+    match len(sys.argv):
+        case 1:
+            print("no revision given as argument, running against merge base")
+            # merge base is calculated against upstream
+            # "github.com/osbuild/images" not against the users fork
+            rev = subprocess.run(
+                ["git", "merge-base", "upstream/main", "HEAD"],
+                capture_output=True, text=True, check=True,
+            ).stdout.strip()
+        case 2:
+            rev = sys.argv[1]
+        case _:
+            print("only a single revision argument can be passed")
+            sys.exit(1)
+
     print("Note that this diff does *not* include depsolved rpms,")
     print("so upstream dependency changes will not be caught\n")
     with tempfile.TemporaryDirectory() as tmpdir:


tools/gen-manifests-diff
```diff
@@ -61,9 +61,21 @@ def manifests_diff(tmp_path, rev):
 
 
 def main():
-    rev = "main"
-    if len(sys.argv) > 1:
-        rev = sys.argv[1]
+    match len(sys.argv):
```
supakeen, line 64: Note that this now requires Python >= 3.10.

supakeen (approved):

tools/gen-manifests-diff
mvo5: Fair point, it seems centos9 has 3.9 - if its a problem I can change it to if/elif/else

achilleas-k: Since it's a developer tool, I don't think this is an issue.
FTR, [`test/scripts/boot-image`](https://github.com/osbuild/images/blob/a52fe88357ed7beb1d03612949f8206d21280639/test/scripts/boot-image#L172) which runs in our GitLab tests, also uses `match`.

tools/gen-manifests-diff
```diff
@@ -61,9 +61,21 @@ def manifests_diff(tmp_path, rev):
...
+        case 1:
+            print("no revision given as argument, running against merge base")
+            # merge base is calculated against upstream
+            # "github.com/osbuild/images" not against the users fork
+            rev = subprocess.run(
+                ["git", "merge-base", "upstream/main", "HEAD"],
```
achilleas-k, line 70: This assumes the remote name, which might be different for every person.  Personally, I use `origin` for upstream and my name for my fork, for example.

It might be enough to just use `main`.  Assuming everyone starts every branch from a local copy of `main`, `git merge-base main HEAD` should be identical to `git merge-base upstream|origin/main HEAD`, if `upstream/main` is ahead of the local `main`.

Using plain `main` only becomes an issue (and would provide different results) if somehow the `main` has local modifications that don't exist upstream, but I think at that point, there's nothing we can do.

achilleas-k (changes requested):

tools/gen-manifests-diff
mvo5: Good point, I will close again for now then, if it becomes a problem we can make the code try to be smart and figure out how what remote name points to upstream

PR #1327, opened 2025-03-14
Depsolve customization packages separately

### LINKED ISSUE #1323: Packages selected via blueprint customizations can conflict with the base package set excludes

achilleas-k:
The OS package set chain now consists of three transactions (up from two).

The transactions contain the following packages:
1. Platform, Environment, Partition Table, and Base (static Image Type) packages, *with* Excludes from the Base package set.
    - Weak dependency installation is controlled by the Image Type's ImageConfig.
2. Customization packages: These are packages that are selected based on ImageConfig and Blueprint customizations (e.g. 'chrony' for NTP, 'firewalld' for firewall configuration).
    - Weak dependency installation is *disabled*.
3. Blueprint (Workload) packages: This is the same as the previous second transaction in the chain.

This change solves the problem where a package selected by a customization could conflict with a base package set exclude.

See linked issue for a full description of the problem.

This PR has no effect on generated manifests.

Fixes #1323

### DIFF
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -226,11 +226,8 @@ func (p *OS) getPackageSetChain(Distro) []rpmmd.PackageSet {
 		environmentPackages = p.Environment.GetPackages()
 	}
 
-	// If we have a logical volume we need to include the lvm2 package.
-	// OSTree-based images (commit and container) aren't bootable images and
-	// don't have partition tables.
 	var partitionTablePackages []string
-	if p.PartitionTable != nil && p.OSTreeRef == "" {
+	if p.PartitionTable != nil {
 		partitionTablePackages = p.PartitionTable.GetBuildPackages()
 	}
 
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -219,44 +219,48 @@ func NewOS(buildPipeline Build, platform platform.Platform, repos []rpmmd.RepoCo
 }
 
 func (p *OS) getPackageSetChain(Distro) []rpmmd.PackageSet {
-	packages := p.platform.GetPackages()
+	platformPackages := p.platform.GetPackages()
 
-	if p.KernelName != "" {
-		packages = append(packages, p.KernelName)
+	var environmentPackages []string
+	if p.Environment != nil {
+		environmentPackages = p.Environment.GetPackages()
 	}
 
 	// If we have a logical volume we need to include the lvm2 package.
 	// OSTree-based images (commit and container) aren't bootable images and
 	// don't have partition tables.
+	var partitionTablePackages []string
 	if p.PartitionTable != nil && p.OSTreeRef == "" {
-		packages = append(packages, p.PartitionTable.GetBuildPackages()...)
+		partitionTablePackages = p.PartitionTable.GetBuildPackages()
 	}
 
-	if p.Environment != nil {
-		packages = append(packages, p.Environment.GetPackages()...)
+	if p.KernelName != "" {
+		// kernel is considered part of the platform package set
+		platformPackages = append(platformPackages, p.KernelName)
 	}
 
+	customizationPackages := make([]string, 0)
 	if len(p.NTPServers) > 0 {
-		packages = append(packages, "chrony")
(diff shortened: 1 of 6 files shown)


supakeen (commented):
I quite like this, and the new explicitness of where each of the packages comes from. I have one question.

pkg/manifest/os.go
```diff
@@ -287,13 +305,15 @@ func (p *OS) getPackageSetChain(Distro) []rpmmd.PackageSet {
 			ps := rpmmd.PackageSet{
 				Include:      workloadPackages,
 				Repositories: append(osRepos, p.Workload.GetRepos()...),
+				// Although 'false' is the default value, set it explicitly to make
+				// it visible that we are not adding weak dependencies.
+				InstallWeakDeps: false,
```
supakeen, line 310: Is this in fact what we want? I *thought* we want workload packages (which include blueprint packages, right?) to be non-weak? :)

achilleas-k: The idea is that as long as blueprint users don't have a mechanism for excluding packages, installing weak dependencies gives them no way to trim their blueprint-installed packages down.

achilleas-k:
Closing since it's included in #1324.

PR #1330, opened 2025-03-17
data/repositories/rhel-8*.json: don't include the "auxiliary key 2" (RHEL-71397)

thozza:
The `osbuild/images` go module supports distro cross-building. This means that you can build images for a different distribution than the one you are running on. For example, you can build a RHEL-8 image on RHEL-9. Nevertheless, the supported combinations are controlled by the end users of the `osbuild/images` go module.

The `rhel-8*.json` repository definitions have been modified to contain only the [`4096R/199e2f91fd431d51 (2009-10-22): Red Hat, Inc. (release key 2) <security@redhat.com>`](https://security.access.redhat.com/data/fd431d51.txt) as specified at [Red Hat Product Signing keys page](https://access.redhat.com/security/team/key). The[`4096R/F76F66C3D4082792 (2018-06-27): Red Hat, Inc. (auxiliary key 2) <security@redhat.com>`](https://security.access.redhat.com/data/d4082792.txt) is completely omitted from the `rhel-8*.json` repository definitions. The reasons are:
 * The `Red Hat, Inc. (auxiliary key 2)` is signed only with RSA-SHA1, which is not allowed by the default crypto-policy on RHEL-9 and newer. This means that importing such key when bootstrapping the RHEL-8 buildroot on RHEL-9 and newer would fail.
 * The `Red Hat, Inc. (auxiliary key 2)` is not commonly used for signing the RHEL-8 RPMs. It is a disaster recovery key that would be used in the unlikely event Red Hat loses the ability to sign with the master hardware keys. The key file uses a sub-key, that would be most probably used to sign RPMs in this unlikely event, so that the main key can be stored offline. Previously, we used to ship only the auxiliary key, without the sub-key in `rhel-8*.json` configurations on RHEL-9, because only the sub-key's signature was causing issues when importing it on RHEL-9 with the default crypto-policy. However, given the purpose of the sub-key, there's no reason to include the main key in the `rhel-8*.json` configurations without the sub-key.
 * None of the projects that use the repository definitions in the `osbuild/images` go module (`osbuild-composer`, `image-builder-cli`, etc.) are being released or updated in RHEL-8. The `rhel-8*.json` repository definitions are used only for cross-building RHEL-8 images on RHEL-9 and newer.


supakeen (approved):
Some nice detective work (and documentation).

achilleas-k (approved):
Thank you!!  Great to finally know what's going on here :)

PR #1336, opened 2025-03-18
 distro: rename "packagesets.Load" to "defs.PackageSet"  (HMS-5819)

mvo5:
[draft as build on top of https://github.com/osbuild/images/pull/1335, only the last commit is new here]

This commit renames the existing `packagesets` module to `distro/defs`
to reflect that it will be about more than just the packagesets.

There are various other names we could use, I have no super strong
opinion but `distro/defs` is nice and short. Some more options:
- loader.{Load,}PackageSet
- yamlload.PackageSet
- distrodefs.PackageSet (distro/distrodefs feels a bit redundant)
- defload{,er}.PackageSet()

This prepares for the upcoming:
- defs.PartitionTable
- defs.ImageConfig
- defs.Platform
helpers.

---

defs: rename package_sets.yaml to distro.yaml

This reflects that the data we load will be no longer about
package_sets only.

JIRA: [HMS-5819](https://issues.redhat.com/browse/HMS-5819)


ondrejbudai:
I like the term "def", it's short, and it maps nicely to the term "definitions" that we have been using to describe how we define images for ages.

achilleas-k:
> * loader.{Load,}PackageSet
> * yamlload.PackageSet
> * defload{,er}.PackageSet()

I kinda like these two assuming it will be a package for loading the definitions only.  It sort of emphasises that the definitions are a separate concept from the code that loads them.

> * distrodefs.PackageSet (distro/distrodefs feels a bit redundant)

I agree with the (parens), this is redundant.
I'm not a huge fan of `defs`, but I could get used to it.

achilleas-k (approved):
Naming aside: LGTM

I'd like to consider using a name like `defloader` though.

supakeen (approved):
+1 to `defs`, thank you.

PR #1337, opened 2025-03-18
distro/fedora: move defaultImageConfig into YAML

mvo5:
Now that all the package sets are part of the YAML this moves
the default distro.ImageConfiguration into YAML.

Note that the individual imageType ImageConfiguration are not moved yet and that will (probably) still take a while.

### DIFF
--- a/pkg/distro/defs/rhel-8/distro.yaml
+++ b/pkg/distro/defs/rhel-8/distro.yaml
@@ -524,6 +524,18 @@
             - "insights-client"
             - "subscription-manager-cockpit"
     
+image_config:
+  default_kernel: "kernel"
+  # XXX: this needs to be conditional for centos and rhel
+  default_oscap_datastream: "/usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml"
+  install_weak_deps: true
+  kernel_options_bootloader: true
+  locale:   "en_US.UTF-8"
+  sysconfig:
+    networking: true
+    no_zero_conf: true
+  timezone: "America/New_York"
+  update_default_kernel: true
 
 image_types:
   # XXX: not a real pkgset but the "os" pipeline pkgset for image-installer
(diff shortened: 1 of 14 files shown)


achilleas-k (approved):

achilleas-k:
Something odd is happening here.  Manifest diffs without depsolving shows no changes, but when depsolved, I'm seeing package differences.  But the repositories haven't changed.

Confused.

achilleas-k:
Oh!  Weak deps!

achilleas-k:
I wrote #1379 to add depsolve configs to mocked manifests, to reflect things like repository selection and weak deps for each depsolve transaction in a chain.

Using it here, I see the first transaction of the OS pipeline disabled weak dependencies.

pkg/distro/defs/rhel-8/distro.yaml
```diff
@@ -524,6 +524,17 @@
...
+  timezone: "America/New_York"
+  locale:   "en_US.UTF-8"
+  sysconfig:
+    networking: true
+    no_zero_conf: true
+  default_kernel: "kernel"
+  update_default_kernel: true
+  kernel_options_bootloader: true
+  install_weak_deps: true
+  default_oscap_datastream: "/usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml"
```
ochosi, line 537: Currently not all items are consistently ordered, e.g. the `default_oscap_datastream`.

While the order may not matter strictly speaking, it might help with readability.

supakeen (approved):
Nice, is the next step to have per-type image configs here as well?

achilleas-k:
> Nice, is the next step to have per-type image configs here as well?

I keep pushing that back because I'm being annoying pushing to restructure them _before_ we pull them out into yaml.

achilleas-k (approved):

PR #1339, opened 2025-03-19
fedora,rhel: move partition tables into YAML

mvo5:
This WIP but already interesting to check the direction.

TODO: this diff contains a mix of "PartitionTable" and "PartitionTables" (singular/plural) that needs fixing.


mvo5:
Closing in favor of https://github.com/osbuild/images/pull/1372 which has easier to read YAMl

PR #1342, opened 2025-03-20
many: make bootstrap container usage a manifest option

mvo5:
Alternative version of https://github.com/osbuild/images/pull/1293

Based on feedback from @achilleas-k - if you like this version I will
add a few more tests and then I think this is ready to merge.

---

This commit adds the option to use a bootstrap container when
generating the manifest. This allows a seamless experience when
using e.g. image-builder:
```console
$ image-builder build --arch=riscv64 minimal-raw --distro fedora-42
```

The approach is to make the distro bootstrap container ref part of
the manifest struct just like we set "Distro" there. This is not
ideal, it would be much nicer if manifest would hold the full
`distro.Distro` and this way we could simply run something like
`distro.BootstrapContainerRef()` when needed. However because
of circular inputs that is not possible.

Note that I think eventually the bootstrap container ref will be
part of the "distro.yaml" and the currently (slightly ugly) helper
`bootstrapContainerFor()` will go away.

The matching and small image-builder-cli changes are in https://github.com/osbuild/image-builder-cli/compare/main...mvo5:cross-arch-for-all-2?expand=1 (and the cross-build test in there passes locally but takes ~30min for all 4 tested arches and it lacks a check that inside the generate container image we actually have binaries of the expected architecture)

Note that cross-arch building will not support selinux in the buildroot yet - see the second commit about how we can still archive this in the future.


pkg/distro/fedora/imagetype.go
```diff
@@ -514,3 +517,19 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
...
 }
+
+// XXX: this will become part of the yaml distro definitions, i.e.
+// the yaml will have a "bootstrap_ref" key for each distro/arch
+func bootstrapContainerFor(t *imageType) string {
+	arch := t.arch.Name()
+	distro := t.arch.distro
+
+	// XXX: remove once fedora containers are part of the upstream
+	// fedora registry (and can be validated via tls)
```
supakeen, line 528: Is there an upstream tracking ticket for this? I'm not *super* happy making your personal account load bearing even if it's just for `riscv64`.

pkg/manifest/build.go
```diff
@@ -36,12 +34,30 @@ type BuildrootFromPackages struct {
...
 
 	containerBuildable bool
+
+	// XXX: disableSelinux controlls if selinux should be disabled
+	// for the given buildroot. This is currently needed for
+	// bootstraped buildroot containers because most of the
+	// boostrap containers do not include setfiles(8). A
+	// reasonable fix is to teach osbuild to chroot into the
+	// buildroot itself when running setfiles. Once osbuild has
+	// this then this option would become "useChrootSetfiles"
```
supakeen, line 44: I'll assume we have requested `setfiles` to be included, if so, do we have tracking issues for this?

supakeen (approved):
Very cool; I'd like to have the tracking issues in the comments if they exist.

achilleas-k (approved):
Neat!  This LGTM.

achilleas-k:
There are two things here we need to resolve soon:
1. The bootstrap container source.
2. Setting SELinux labels in the buildroot.

There's a tiny bit of interdependence between these two issues depending on how we tackle it, right?  If we want to use upstream, official containers for the bootstrap pipeline, we're going to have to do something new to fix the labels (`setfiles` in a chroot, like this PR suggests).  I don't think getting `setfiles` into upstream containers is realistic.  It's not useful in containers normally, and I don't like the idea of asking for something as fundamental as a base distro container image to include something for our very specific use case.

I don't think having our own images is out of the question though.  A very simple container with rpm, python3, mount, and setfiles could be built on a weekly schedule and we could host it under the `osbuild` namespace to offer it as the official way to bootstrap your cross-arch builds.  It might even be nice to use osbuild to build it, so we have full control of what goes in it and avoid any surprises with upstream base containers changing (this would be for x86_64 and aarch64, which we can build easily).  I realise this whole endeavour started to solve the risc-v hardware availability issue, which this suggestion ignores, but I'd still like to explore the option for the others.

pkg/distro/fedora/imagetype.go
mvo5: I created https://github.com/osbuild/images/issues/1351 for us, we could move the repo to osbuild for now, its just a bunch of busywork - my hope was that we relatively soon see riscv in the upstream registry but maybe I'm too optimistic here.

pkg/manifest/build.go
mvo5: I created  https://github.com/osbuild/osbuild/issues/2052 for this - I am not sure how likely it is that we can convince upstreams but allowing to chroot in setfiles seems okay to me (and not too much work)

mvo5:
Fwiw, I'm fine having our own boostrap containers, I was  optimizing for the least amount of maintenance   but having our own means they will be way more targeted and smaller than the toolbox ones (which are relatively heavy). It would also solve the riscv problem and we could indeed just build them via ibcli in a github action - by using the cross build feature of ibcli to create them we would even test that this works correctly on a weekly basis :)

PR #1346, opened 2025-03-21
Add ro to FSTabOptions for IoT raw disk

djach7:
Adds `ro` to FSTabOptions for Fedora IoT raw disk images.

Fixes https://github.com/fedora-iot/iot-distro/issues/81

### DIFF
--- a/pkg/distro/fedora/partition_tables.go
+++ b/pkg/distro/fedora/partition_tables.go
@@ -396,7 +396,7 @@ var iotBasePartitionTables = distro.BasePartitionTableMap{
 					Type:         "ext4",
 					Label:        "root",
 					Mountpoint:   "/",
-					FSTabOptions: "defaults",
+					FSTabOptions: "defaults,ro",
 					FSTabFreq:    1,
 					FSTabPassNo:  1,
 				},


supakeen:
Should this be done unconditionally, e.g. will things still work in Fedora 40, or 41? Otherwise OK.

---

As an aside, we also have the option to not write `/etc/fstab` at all which I think is the eventual goal or is that only the `bootc`-case? cc @achilleas-k

djach7:
Looking into this, I think enabling it just conditionally might make more sense

achilleas-k:
> As an aside, we also have the option to not write `/etc/fstab` at all which I think is the eventual goal or is that only the `bootc`-case? cc @achilleas-k

The bootc disk images are now all fstab-less, yes.  I don't think there are plans to drop the file from RHEL images any time soon.  On the Fedora side, we'll only make any changes if the official images (and default Fedora setup) change.  So, for IoT images, I wouldn't drop it unless someone from the IoT team tells me to.

achilleas-k (approved):

mvo5 (approved):

PR #1349, opened 2025-03-24
disk: add new unmarshalYAMLviaJSON helper

mvo5:
Split out of https://github.com/osbuild/images/pull/1339 to make this PR more reviewable.

With partition tables moving into the distro YAML we need a way to write them in a "nice"
way. Yaml offers some helper with anchors and aliases here (shown as part of the tests)
but we still need to add json tags to all our (internal) disk structs. This of course means that
when we change our disk data structure we need to update the yaml. As long as we maintain
the distro YAML in our repo this should be fine. The alternative would be to use the advanced
partitioning but it is currently not able to express what we need and my worry is that tweaking
it to deal with what we need inside #1339 is going to slow us down significantly (but I'm open
to discuss this of course).

---

This commit adds a new 'unmarshalYAMLviaJSON` helper and uses
it (initially) for the `disk.PartitionTableType` and
`disk.Partition`.

This allows us to cheaply reuse all our custom json unmarshal
code when we define partition tables in yaml.

---

disk: give struct fields "nice" json names

This commit prepares for moving partition tables into the YAML
distro definitions by making the names "nice". This commit only
changes the common things so far which is partition-table,
partition and some payloads.

### DIFF
--- a/pkg/disk/filesystem.go
+++ b/pkg/disk/filesystem.go
@@ -9,18 +9,19 @@ import (
 
 // Filesystem related functions
 type Filesystem struct {
-	Type string
+	Type string `json:"type"`
+
 	// ID of the filesystem, vfat doesn't use traditional UUIDs, therefore this
 	// is just a string.
-	UUID       string
-	Label      string
-	Mountpoint string
+	UUID       string `json:"uuid,omitempty"`
+	Label      string `json:"label,omitempty"`
+	Mountpoint string `json:"mountpoint,omitempty"`
 	// The fourth field of fstab(5); fs_mntops
-	FSTabOptions string
+	FSTabOptions string `json:"fstab_options,omitempty"`
 	// The fifth field of fstab(5); fs_freq
-	FSTabFreq uint64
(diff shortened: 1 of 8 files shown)


pkg/disk/filesystem.go
```diff
@@ -9,18 +9,19 @@ import (
...
 type Filesystem struct {
-	Type string
 	// ID of the filesystem, vfat doesn't use traditional UUIDs, therefore this
-	// is just a string.
-	UUID       string
-	Label      string
-	Mountpoint string
+	Type string `json:"type"`
+
+	// UUID is just a string.
```
achilleas-k, line 15: The two-line comment here for UUID got broken into two.  `Type` should move one line up and the two-line comment should be above UUID.

achilleas-k (changes requested):
LGTM but there's a comment split that needs fixing.

pkg/disk/filesystem.go
mvo5: Thanks, nice catch! fixed.

achilleas-k (approved):

supakeen (approved):

PR #1350, opened 2025-03-24
Imageconfig no osbuild

mvo5:
In PR#1337 the desire was expressed to move all direct `osbuild` reference out of `distro.ImageConfig` to make it a more high level abstraction. This commit is the first step to show how this could be done with the simple `osbuild.WSLConfStageOptions`.

However doing this with the more complex ones will be quite a bit of busywork and reviewing. I am not sure we should do this before we do the yamlification as I'm worried converting the other ~26 structs that needs similar converstion will take some time.

The other concern with this is that we currently have json schema information for all osbuild options. So validating the yaml against the json-schema of osbuild seems feasible. If add this conversion stage we will need to also duplicate the json schema.

---

distro: move `osbuild.SysconfOptions` out of ImageConfig

This commit shows what needs to happen for a more complex option
in image config. With that we can/could continue with PR#1337.

Note that moving Kernel/UpdateDefaultKernel is based on an initial brainstorm with @achilleas-k but we could of course move it back into sysconfig (that would also fix the XXX that we have in gce.go). The downside of doing that is that we may conflate unrelated concerns and introduce bugs/confusion. Because we only do shallow merging it seems risky (c.f. the XXX in gce.go).

### DIFF
--- a/pkg/distro/image_config.go
+++ b/pkg/distro/image_config.go
@@ -64,7 +64,8 @@ type ImageConfig struct {
 	Firewall            *osbuild.FirewallStageOptions
 	UdevRules           *osbuild.UdevRulesStageOptions
 	GCPGuestAgentConfig *osbuild.GcpGuestAgentConfigOptions
-	WSLConfig           *osbuild.WSLConfStageOptions
+
+	WSLConfig *WSLConfig
 
 	Files       []*fsnode.File
 	Directories []*fsnode.Directory
@@ -110,6 +111,10 @@ type ImageConfig struct {
 	MountUnits *bool
 }
 
+type WSLConfig struct {
+	BootSystemd bool
+}
+
 // InheritFrom inherits unset values from the provided parent configuration and
 // returns a new structure instance, which is a result of the inheritance.
 func (c *ImageConfig) InheritFrom(parentConfig *ImageConfig) *ImageConfig {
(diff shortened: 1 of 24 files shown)


pkg/distro/image_config.go
```diff
@@ -110,6 +111,10 @@ type ImageConfig struct {
 	MountUnits *bool
 }
 
+type WSLConfig struct {
+	BootSystemd bool
+}
```
achilleas-k, line 120: This is probably the first instance of a substruct of ImageConfig being defined here and it makes me think we need to figure out a more general strategy.

See my main comment.

achilleas-k (commented):
I still keep going back and forth in my mind about the differences and similarities between ImageConfig and OSCustomizations.  While ImageConfig has a more general scope (it configures the whole image, while OSCustomizations are strictly limited to the OS pipeline), I'd like to explore the idea of having one be an extension of the other.

We could, for example, have an ImageConfig that includes a number of sub-configs for different pipelines or image types.
```go
type ImageConfig struct {
  OS *OSConfig  // previously OSCustomizations
  Installer *InstallerConfig
  Build *BuildConfig  // options that affect the build pipeline
  Container *ContainerConfig
}
```

And then, before generating a manifest (inside the `Manifest()` method, but before the `ImageFunc()`), we have the cascading merge:
```go
imageConfig := distroImageConfig.Merge(imageType.ImageConfig)
```
and then modify using the blueprint:
```go
finalConfig := imageConfig.Customize(blueprint)

```
and then this `finalConfig`, or parts of it, get passed down to the pipeline generators.

I'm aware that there's a pressure to get these structures externalised as quickly as possible, and we disagree on the order of tasks, but I still think it would be a better to experiment with these kinds of changes before we pull everything out.

achilleas-k (approved):
Approving this now.  The bigger discussion can happen elsewhere.

supakeen (approved):

PR #1353, opened 2025-03-24
manifest: addInlineDataAndStages() helper method for the OS and OSTreeDeployment pipelines

achilleas-k:
The new addInlineDataAndStages() method of the OS and OSTreeDeployment pipeline generators appends file creation, chmod, and chown stages to the pipeline and also adds the inline data to the inlineData property.  The inlineData list is used to return inline data that was collected during manifest generation.

Previously, we used the OSCustomizations.Files list in two ways:
- To define files to be created externally, during the initialisation of OSCustomizations (e.g. from a blueprint).
- To collect files that needed to be created under certain conditions or for certain stages while generating the pipeline.

This lead to inconsistencies.  The handling of the Files list was meant to always be near the end of the pipeline, before the selinux stage.  However, files were being added to the list after the Files list was handled, which meant that appending to the Files was done to generate the inline data, but not to generate stages.  This meant that appending to the stages was necessary in some cases and not others (depending on whether the files were added to the list before or after the stage generation from the OSCustomization.Files list), which could (and very likely did) lead to confusion.

This change also makes the usage of OSCustomizations more consistent.
During pipeline generation, it's better to treat OSCustomizations as read-only properties, initialised before serialization starts.  We might consider enforcing this in code.

### DIFF
--- a/pkg/manifest/export_test.go
+++ b/pkg/manifest/export_test.go
@@ -1,6 +1,7 @@
 package manifest
 
 import (
+	"github.com/osbuild/images/pkg/customizations/fsnode"
 	"github.com/osbuild/images/pkg/dnfjson"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/ostree"
@@ -58,6 +59,14 @@ func (p *OS) GetPackageSetChain(d Distro) []rpmmd.PackageSet {
 	return p.getPackageSetChain(d)
 }
 
+func (p *OS) AddInlineDataAndStages(pipeline *osbuild.Pipeline, files []*fsnode.File) {
+	p.addInlineDataAndStages(pipeline, files)
+}
+
+func (p *OS) GetInline() []string {
+	return p.getInline()
+}
+
 // NewTestOS is used in both internal and external package tests.
 // TODO: make all tests external and define this only in the manifest_test
 // package.
@@ -83,4 +92,12 @@ func (p *OSTreeDeployment) Serialize() osbuild.Pipeline {
 		Commits: []ostree.CommitSpec{{}},
 	})
 	return p.serialize()
+}
+
+func (p *OSTreeDeployment) AddInlineDataAndStages(pipeline *osbuild.Pipeline, files []*fsnode.File) {
+	p.addInlineDataAndStages(pipeline, files, "ostree/ref")
+}
+
+func (p *OSTreeDeployment) GetInline() []string {
+	return p.getInline()
 }
--- a/pkg/manifest/os.go
+++ b/pkg/manifest/os.go
@@ -792,15 +792,6 @@ func (p *OS) serialize() osbuild.Pipeline {
 			}))
 	}
 
-	// First create custom directories, because some of the custom files may depend on them
-	if len(p.OSCustomizations.Directories) > 0 {
-		pipeline.AddStages(osbuild.GenDirectoryNodesStages(p.OSCustomizations.Directories)...)
-	}
-
-	if len(p.OSCustomizations.Files) > 0 {
-		p.addInlineDataAndStages(&pipeline, p.OSCustomizations.Files)
-	}
-
 	// write modularity related configuration files
 	if len(p.moduleSpecs) > 0 {
 		pipeline.AddStages(osbuild.GenDNFModuleConfigStages(p.moduleSpecs)...)
@@ -826,6 +817,18 @@ func (p *OS) serialize() osbuild.Pipeline {
 
 		pipeline.AddStages(osbuild.GenDirectoryNodesStages([]*fsnode.Directory{failsafeDir})...)
 		p.addInlineDataAndStages(&pipeline, failsafeFiles)
+	}
+
+	// First create custom directories, because some of the custom files may depend on them
+	if len(p.OSCustomizations.Directories) > 0 {
+		pipeline.AddStages(osbuild.GenDirectoryNodesStages(p.OSCustomizations.Directories)...)
+	}
+
+	// Custom files (from the blueprint) are often used to create systemd
+	// units, so let's make sure they get created before the systemd stage that
+	// will probably want to enable them
+	if len(p.OSCustomizations.Files) > 0 {
+		p.addInlineDataAndStages(&pipeline, p.OSCustomizations.Files)
 	}
 
 	enabledServices := []string{}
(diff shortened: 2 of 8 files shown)


bcl:
Does this need any new tests, or is all this behavior covered by existing ones?

achilleas-k:
> Does this need any new tests, or is all this behavior covered by existing ones?

It's indirectly covered by a test in #1318, the main PR from which this one was spun-out.  But I think it would be good to have a more explicit one.

pkg/manifest/os_test.go
```diff
@@ -369,3 +370,63 @@ func TestHostnameDoesNotIncludeHostnameStage(t *testing.T) {
 	st := manifest.FindStage("org.osbuild.hostname", pipeline.Stages)
 	require.Nil(t, st)
 }
+
+func TestAddFilesOS(t *testing.T) {
+	os := manifest.NewTestOS()
```
mvo5, line 375: This is (almost) identical to the OS test is there potential for code sharing here?

pkg/manifest/os.go
```diff
@@ -807,15 +804,24 @@ func (p *OS) serialize() osbuild.Pipeline {
...
+	// First create custom directories, because some of the custom files may depend on them
+	if len(p.Directories) > 0 {
+		pipeline.AddStages(osbuild.GenDirectoryNodesStages(p.Directories)...)
+	}
+
+	// Custom files (from the blueprint) are often used to create systemd
+	// units, so let's make sure they get created before the systemd stage that
+	// will probably want to enable them
+	if len(p.Files) > 0 {
+		p.addFiles(&pipeline, p.Files)
```
mvo5, line 824: I wonder if we should tweak the name `addFiles()` so close to `p.Files` sounds like the Files are added to `p.Files` but they are added to the inline data - so maybe `addFileStageAndInlineData`, `addFileStages``addFilePlumbing` or `addToInlineData` or something slightly more descriptive? I guess `p.addFiles(pipline, p.customizations.Files)` would have been slightly easier for me (i.e. not embedding the os customizations) as the connection would have been easier this way (is there a deeper reason for the embedding?)

pkg/manifest/os_test.go
```diff
@@ -369,3 +370,63 @@ func TestHostnameDoesNotIncludeHostnameStage(t *testing.T) {
...
+
+	require := require.New(t)
+
+	// add some files to the OSCustomizations which are included near the end
+	// of the pipeline
+	fileOne, err := fsnode.NewFile("/etc/test/one", nil, nil, nil, []byte("test 1"))
```
mvo5, line 381: (super nitpick) this could be `fileOne := common.Must(fsnode.NewFile(...` but not sure if that would improve readability much

mvo5 (approved):
Thanks, this looks fine, some suggestions inline for your consideration but no blockers.

pkg/manifest/os.go
achilleas-k: > (is there a deeper reason for the embedding?)

Nope.  It's just always been there.  You're right, it'd be easier to read if we made it a regular property.

pkg/manifest/os_test.go
achilleas-k: Might be nicer to  get an error from `require` than a `panic()`, but `fsnode.NewFile()` shouldn't return an error unless the path is bad (and the user and group have bad types), so it's probably fine to use `Must()`.

pkg/manifest/os.go
achilleas-k: > it'd be easier to read if we made it a regular property.

but that's for another time.

achilleas-k: > I wonder if we should tweak the name `addFiles()` so close to `p.Files` sounds like the Files are added to `p.Files` but they are added to the inline data - so maybe `addFileStageAndInlineData`,

I think what's more important here is that `p.Files` shouldn't be modified during serialization.  `addFiles()` fits with the `AddStages()`, but one works on the OS object and the other on the (serialized) `osbuild.Pipeline`.  I think there's something we can learn here about what information ends up where, like maybe we should be returning both sources and pipelines from `serialize()` functions, so instead of needing to do:

```go
os := NewOS()
pipeline := os.serialize()
inlines := os.getInline()
```

we could do
```go
os := NewOS()
pipeline, sources := os.serialize()
```

and `os`'s internal state never gets modified by the serialization.

pkg/manifest/os_test.go
achilleas-k: Made two new functions, one for creating the test files and another for collecting the copy stage destination paths.

pkg/manifest/os_test.go
achilleas-k: Done.

pkg/manifest/os.go
mvo5: I agree that the key is that `p.File` is not modified and with the rest of the comments. For me it is still confusing to have `p.Files` and a `p.addFiles(&pipeline, p.Files)` helper that does not actually add files to `p`, if it was `pipline.addFiles(p.Files)` (which of course it can't be) it would be a non-issue, but as is as a casual reading I get the wrong impression (but that might be just me). I will not press this point though, if you feel this is consistent with the rest of the helper naming just mark this as resolved.

achilleas-k: oh, right.  Yeah you make a good point.  Sorry, I addressed all the other issues directly, but completely skipped renaming the method.  Let me do a quick refactor.  It's more important to be clear than here, I think.

achilleas-k: Renamed to `addInlineDataAndStages()`.

pkg/manifest/export_test.go
```diff
@@ -58,6 +59,14 @@ func (p *OS) GetPackageSetChain(d Distro) []rpmmd.PackageSet {
 	return p.getPackageSetChain(d)
 }
 
+func (p *OS) AddFiles(pipeline *osbuild.Pipeline, files []*fsnode.File) {
```
mvo5, line 62: (super nitpick) this part did not get renamed but its fine to ignore as its just export_test

mvo5 (approved):
Thank you! One tiny suggestion inline, I really appreciate the rename, thanks for this!

pkg/manifest/export_test.go
achilleas-k: Woopsie

mvo5 (approved):
Thank you!

achilleas-k:
Rebased and fixed conflicts.

achilleas-k:
> Rebased and fixed conflicts.

For real this time.

achilleas-k:
Fixed manifest checksums too!

mvo5 (approved):

supakeen (approved):

PR #1354, opened 2025-03-25
docs/developer: details for changes not caught by gen-manifests-diff

achilleas-k:
Add some details about the kinds of changes and behaviours that are not caught by gen-manifests-diff.


mvo5 (approved):
Thank you!

supakeen (approved):

PR #1355, opened 2025-03-25
disk: add luks/lvm payload loading support via json/yaml

mvo5:
[split out of https://github.com/osbuild/images/pull/1339]

This commit adds support to load payloads for LUKS/LVM via
json/yaml. This is needed for PR#1339 where we extract the
partition tables into YAML. We have nested LUKS/LVM partitions
in e.g. the iot simplified installer.

This sadly involves some duplicated code in the Partition,
LUKSContainer, LVMLogicalVolume `UnmarshalJSON()` code.

Ideally we would write somethign like:
```go
func payloadUnmarshaller[T any](data []byte) (*T, error) {
        type alias T
        var withoutPayload struct {
                alias
                Payload     json.RawMessage `json:"payload"`
                PayloadType string          `json:"payload_type"`
        }
        if err := jsonUnmarshalStrict(data, &withoutPayload); err != nil {
                return nil, fmt.Errorf("cannot unmarshal %q: %w", data, err)
        }
        t := T(withoutPayload.alias)

        payload, err := unmarshalJSONPayload(data)
        t.Payload = payload
        return payload, err
}

func (lv *LVMLogicalVolume) UnmarshalJSON(data []byte) (err error) {
        *lv, err = payloadUnmarshaller[*LVMLogicalVolume](data)
        return err
}
```
but GO does not let us and sadly its pretty fundamental:
```
./lvm.go:248:13: cannot use a type parameter as RHS in type declaration
```
and without type aliases in the custom unmarshaler we run into
an infinite recursion. So unless go supports this we will have to
live with the duplication.

### DIFF
--- a/pkg/disk/unmarshal.go
+++ b/pkg/disk/unmarshal.go
@@ -1,8 +1,10 @@
 package disk
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
+	"reflect"
 )
 
 // unmarshalYAMLviaJSON unmarshals via the JSON interface, this avoids code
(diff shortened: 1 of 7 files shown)


supakeen (dismissed):

pkg/disk/unmarshal.go
```diff
@@ -22,3 +24,35 @@ func unmarshalYAMLviaJSON(u json.Unmarshaler, unmarshal func(any) error) error {
...
+	var payload struct {
+		Payload     json.RawMessage `json:"payload"`
+		PayloadType string          `json:"payload_type,omitempty"`
+	}
+	if err := json.Unmarshal(data, &payload); err != nil {
+		return nil, fmt.Errorf("cannot peek payload: %w", err)
+	}
+	if payload.PayloadType == "" {
+		if len(payload.Payload) > 0 {
+			return nil, fmt.Errorf("can build payload: empty payload type but payload is: %q", payload.Payload)
```
achilleas-k, line 38: `can build payload`?
I assume that's meant to be `cannot`.

achilleas-k (commented):
LGTM.  Typo in error message though.

pkg/disk/unmarshal.go
mvo5: Uh, silly me, thank you! Indeed, "cannot" - fixed now.

achilleas-k (approved):

mvo5:
This got merged as part of #1362 so closing it here

PR #1356, opened 2025-03-25
tools: make gen-manifests-diff run against the merge base

mvo5:
Improved version of #1325 that addresses https://github.com/osbuild/images/pull/1325#discussion_r1995375323

This commit tweak `gen-manifests-diff` to run against the merge base of the upstream branch instead of whatever is upstream@main. This avoids potentially confusing output when upstream/main has moved forward since the branching and the diff would include unrelated upstream changes.

Thanks to Achilleas for suggesting this.

Closes: osbuild#1322

### DIFF
--- a/tools/gen-manifests-diff
+++ b/tools/gen-manifests-diff
@@ -9,6 +9,7 @@
 # very fast (typically 3s).
 
 import os
+import re
 import pathlib
 import subprocess
 import sys
@@ -16,6 +17,18 @@ import tempfile
 
 # TODO: omit once we have a "riscv64" mirror and sources entry
 arches = ["x86_64", "aarch64", "ppc64le", "s390x"]
+
+
+def find_git_remote_upstream_name():
+    res = subprocess.run(
+        ["git", "remote", "-v"], text=True, capture_output=True, check=True)
+    for line in res.stdout.split("\n"):
+        if m := re.search(r'(\w+)\s+(git@|https:\/\/)github\.com(:|\/)osbuild\/images', line):
+            return m.group(1)
+    # if this becomes a frequent issue add the upstream remote, see
+    # https://github.com/osbuild/images/pull/1356#issuecomment-2750997780
+    # for details how to do this
+    raise RuntimeError(f"cannot find upstream github banch in {res.stdout}, please report as a bug")
 
 
 def top_srcdir() -> pathlib.Path:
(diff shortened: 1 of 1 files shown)


tools/gen-manifests-diff
```diff
@@ -18,6 +19,16 @@ import tempfile
 arches = ["x86_64", "aarch64", "ppc64le", "s390x"]
 
 
+def find_git_remote_upstream_name():
+    res = subprocess.run(
+        ["git", "remote", "-v"], text=True, capture_output=True, check=True)
+    for line in res.stdout.split("\n"):
+        if m := re.search(r'(\w+)\s+(git@|https:\/\/)github\.com(:|\/)osbuild\/images', line):
```
achilleas-k, line 26: Glorious!  :heart_eyes:

tools/gen-manifests-diff
```diff
@@ -18,6 +19,16 @@ import tempfile
...
+            return m.group(1)
+    # XXX: should we just return an error here?
+    return "uptream"
```
achilleas-k, line 29: Not sure.  If the user has an `upstream` remote, it will work fine.  If they don't, they'll get an error.  The problem is that if they do and it wasn't caught by the regex, then that means they have an `upstream` remote that is not the real upstream.  Perhaps it's fine to use it since that would mean they explicitly decided to have a repo they consider `upstream` that points to a fork or alternative repository.

achilleas-k (dismissed):
Very nice!

mvo5:
> I think the only alternative to parsing the remote URLs would be to add one and then remove it:
> 
> ```shell
> #!/bin/bash
> 
> set -euo pipefail
> 
> remote_name="upstream-$(uuidgen)"
> cleanup() {
>     git remote rm "${remote_name}"
> }
> 
> git remote add --fetch "${remote_name}" https://github.com/osbuild/images
> merge_base=$(git merge-base "${remote_name}/main" HEAD)
> 
> # ... do manifest diff
> ```
> 
> The advantage is that it's basically fool proof (as long as you're online). If the remote already exists, the `fetch` will be very fast since it wont have to download anything new. If the remote doesn't exist at all, it'll work, but it might be slow to `fetch`.

Yeah, I was thinking about this too, definitely an option. The downside is that we need to be careful about the cleanup, i.e. I really would like to avoid leaving remotes behind in the users config (that is why I did not pursue it further) but  I guess its straightforward enough and the case where the machine is resatrted/crahsed in the middle of a run would be so rare that it does not matter.

achilleas-k:
> The downside is that we need to be careful about the cleanup

I think we can be pretty good about this.  Like you said, the worst case is leaving behind remotes in the user's local checkout when the script or the machine crashes.  Maybe a bit of an extreme scenario, but definitely possible.

It might be enough to print a message "Adding temporary remote <name> <URL>", so the user knows we're fiddling with their remote configs.
It's definitely more invasive than the current approach though, so I'll leave the final decision to you.

supakeen:
> > The downside is that we need to be careful about the cleanup
> 
> I think we can be pretty good about this. Like you said, the worst case is leaving behind remotes in the user's local checkout when the script or the machine crashes. Maybe a bit of an extreme scenario, but definitely possible.
> 
> It might be enough to print a message "Adding temporary remote ", so the user knows we're fiddling with their remote configs. It's definitely more invasive than the current approach though, so I'll leave the final decision to you.

I guess if we really want to guard against this we can/could copy the repository into a temporary directory and do our work there; though that also has plenty of downsides and gotchas that will break.

thozza (dismissed):
LGTM, but I think that not finding the `upsteram` remote should be considered an error. :innocent:

tools/gen-manifests-diff
mvo5: Thanks, yeah, if two of the people using this do not have an "upstream" remote then this is clearly a bad assumption - I made it an error now and if we get this frequently (or even semi-frequently) I think we should implement the idea to just add the remote ourselfs (c.f. https://github.com/osbuild/images/pull/1356#issuecomment-2750997780) (still feeling a bit weary to do this by default because of the changing of user config (even if only temporary)).

thozza (approved):
:+1:

PR #1357, opened 2025-03-25
workflow: do not run unit test with -v

mvo5:
Do not run our unit tests with `-v` as the signal to noise ratio on failure is really bad. In my case [0] it produced 22607(!) lines and the failure is in line 192.

If there is a deeper reason to run with with `-v` please let me know and I will change this to add a comment why we need it :)

[0] https://github.com/osbuild/images/actions/runs/14057755420/job/39360970265?pr=1355


achilleas-k (approved):
I was looking at that exact job log thinking the same a few minutes ago :)

Issue #1281, opened 2025-03-03
Drop `ro` kernel command line option


achilleas-k:
[...] I would also say that I think the `ro` option doesn't make sense to use anymore *anywhere* (see also e.g. https://github.com/ostreedev/ostree/issues/3193#issuecomment-2578237486 )

_Originally posted by @cgwalters in https://github.com/osbuild/images/pull/1270#discussion_r1977901665_

PR #1360, opened 2025-03-26
fedora: kernel opts to be `rw` for minimal

supakeen:
Currently on Fedora 43 we only write mount units, without an fstab. This causes the system to boot with read-only `/`. I haven't quite figured out the why (systemd perhaps looks at the kernel options, perhaps its the default in mount units and needs to be explicitly enabled).

Regardless, it seems that we want the kernel opts to be `rw` and to not rely on things to remount `/` later on.

---

See also #1281. Alternatively to explicitly defining `rw` we could omit?


achilleas-k (approved):
Thanks for the fix!

ondrejbudai (approved):
Thanks! ❤️

PR #1361, opened 2025-03-26
fedora: add ro to FSTabOptions for IoT raw disk (aarch64)

mvo5:
Followup for https://github.com/osbuild/images/pull/1346 where we only added `ro` to the x86_64 partition table.

This commit also adds the same change to the `aarch64` partition table.


achilleas-k (approved):

bcl (approved):
Ack

PR #1365, opened 2025-03-27
Fix podman network backend fallback on RHEL 9 (HMS-8783)

thozza:
RHEL 9 Podman falls back to the legacy 'cni' network backend when it finds pre-existing container images in storage, interpreting them as a migration from an older version. Since the 'cni' packages are not installed by default, this silently breaks networking for images with embedded containers. Fix this by writing a `defaultNetworkBackend` file into container storage during image build, forcing Podman to use 'netavark'. Add host-config checks to detect this inconsistency and verify container embedding at boot time.

## Architectural Changes

Introduce a `container.NetworkBackend` type and a `GenDefaultNetworkBackendFile` helper in `pkg/container` to generate the `/var/lib/containers/storage/defaultNetworkBackend` file. This file is a Podman-native mechanism that overrides backend auto-detection. The helper accepts a custom storage path to support OSTree images, which relocate container storage to `/usr/share/containers/storage`. The new `PodmanDefaultNetBackend` field on `ImageConfig` drives this behavior declaratively from the distro definition YAML.

## Key Changes

- Add `GenDefaultNetworkBackendFile` in `pkg/container/podman.go` to produce the backend override file, supporting both default and relocated storage paths
- Add `PodmanDefaultNetBackend` option to `ImageConfig` and wire it into `osCustomizations`, gated on both the option being set and containers being present
- Set `podman_default_net_backend: "netavark"` in the RHEL 9 distro definition; Fedora and RHEL 10 are intentionally left unset since they don't have the fallback bug
- Add `container-embedding` host-config check that verifies blueprint containers are present in booted image podman storage
- Add `podman-network-backend` host-config check that detects rootful/rootless backend mismatches

## Breaking Changes

This PR is fully backward compatible.

## Testing

- Table-driven unit tests for `GenDefaultNetworkBackendFile` covering netavark/cni backends and default/custom storage paths
- `osCustomizations` tests verifying the backend file is generated only when both containers and the option are present, including the OSTree relocated-path case
- Distro-level cross-checks asserting RHEL 9 has the backend set to netavark while Fedora and RHEL 10 leave it nil
- Host-config checks tested with mock exec, covering container name matching (including podman's `localhost/` normalization), backend consistency, and error paths


achilleas-k (dismissed):
Really nice!  Thank.

The last commit message is a bit too long.  Can you make it shorter?

thozza:
> Really nice! Thank.
> 
> The last commit message is a bit too long. Can you make it shorter?

Will do.

I hoped that GH will mark this PR as a Draft. :innocent:  My plan is first to see where the consistency check fails, before pushing any fix for RHEL-9.

achilleas-k:
Do we still want this?

cmd/check-host-config/check/podman_network_backend_test.go
```diff
@@ -0,0 +1,103 @@
...
+			containers: []blueprint.Container{
+				{Source: "registry.example.com/test"},
+			},
+			mockExec: map[string]ExecResult{
+				"sudo podman info --format json": {
+					Err: errors.New("podman not found"),
+				},
+			},
+			wantErr: check.ErrCheckFailed,
+		},
```
brlane-rht, line 80: This should also test the empty backend string / unknown cases.
Also, what's the expected result if both are unknown? Should that be a pass or a fail?

brlane-rht:
I know this is still a draft, and I'm curious to see what Schutzbot thinks of it. It looks pretty nice to me.

PR #1366, opened 2025-03-27
osbuild: create .mount and .swap units in /etc/systemd

achilleas-k:
Create .mount and .swap units, which replace /etc/fstab, under /etc/systemd instead of /usr/lib/systemd.

This is needed for ostree-based images.
It's also probably more correct to put the files in /etc in general.

### DIFF
--- a/pkg/osbuild/systemd_unit_create_stage.go
+++ b/pkg/osbuild/systemd_unit_create_stage.go
@@ -244,6 +244,7 @@ func GenSystemdMountStages(pt *disk.PartitionTable) ([]*Stage, error) {
 		}
 
 		options := &SystemdUnitCreateStageOptions{
+			UnitPath: EtcUnitPath, // create all mount units in /etc/systemd/
 			Config: SystemdUnit{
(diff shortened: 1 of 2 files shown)


mvo5 (approved):

supakeen (approved):
I'd prefer this to be selectable in `ImageConfig` but don't hold that opinion very strongly.

achilleas-k:
> I'd prefer this to be selectable in `ImageConfig` but don't hold that opinion very strongly.
> 
> Will this work if/when there's a separate `/usr`-mount which doesn't contain the information on how to mount `/`?

Afaict, the manually written (as opposed to the auto-generated) `-.mount` is a bit useless.  As I discovered while working on #1270, without an `/etc/fstab`, `systemd-remount-fs` doesn't do anything, so the mount options for root are whatever's in the kernel args (see main PR comment).

achilleas-k:
There's a dependency chain of PRs starting with this one right now, so I'm going to merge to unblock stuff.

PR #1368, opened 2025-03-27
loader: generalize partitionTablesOverrideOp.Apply()

mvo5:
This commit removes the hardcoded `size`, `fstab_options` and moves them to a generic reflect based override implementation.

### DIFF
--- a/pkg/distro/defs/loader.go
+++ b/pkg/distro/defs/loader.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"reflect"
 
 	"path/filepath"
 	"sort"
@@ -63,7 +64,7 @@ type pkgSetConditions struct {
 }
 
 type partitionTablesOverrides struct {
-	Conditional *partitionTablesOverwriteConditional `yaml:"condition"`
+	Conditional partitionTablesOverwriteConditional `yaml:"condition"`
 }
 
 func (po *partitionTablesOverrides) Apply(it distro.ImageType, pt *disk.PartitionTable, replacements map[string]string) error {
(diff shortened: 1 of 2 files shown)


pkg/distro/defs/loader.go
```diff
@@ -83,27 +84,99 @@ type partitionTablesOverwriteConditional struct {
...
-		fs, ok := part.Payload.(*disk.Filesystem)
-		if !ok {
-			return fmt.Errorf("override %q part %v for fstab_options expecting filesystem got %T", op, selectPart, part)
+	pt.Partitions[selectPart] = part
+
+	// try to apply to payload
+	var consumedPayload []string
+	switch payload := part.Payload.(type) {
+	case nil:
+		// nothing to d
```
achilleas-k, line 168:
```suggestion
		// nothing to do
```
^_^

pkg/distro/defs/loader.go
```diff
@@ -83,27 +84,99 @@ type partitionTablesOverwriteConditional struct {
...
+	case *disk.Filesystem:
+		val := reflect.ValueOf(&payload).Elem().Elem()
+		consumedPayload, err = op.applyTo(val)
```
achilleas-k, line 171: Will this be extended to cover more types when we need them?

achilleas-k (dismissed):

pkg/distro/defs/loader.go
mvo5: Yeah, I can do it now too, it should be pretty trivial, i can also look into detecting this via code so that the type select is not needed (but reflect is sometimes tricky so I took this shortcut). OTOH YAGNI and all that, so far we only have overrides in partition/filesystem - I could add a comment that extending is (should be?) easy, idk :)

achilleas-k: nah, leave it.  Was just wondering if we think we might need it and if we want to generalise ahead of time.  But let's leave it as is for now.

pkg/distro/defs/loader.go
mvo5: Thanks!

mvo5: Fixed

mvo5:
Had to force push to fix conflicts.

achilleas-k (approved):

mvo5:
With the approach taken in https://github.com/osbuild/images/pull/1390 we won't needt this - might be good to keep it in our backpockets though in case we decode to use it (some stuff gets simpler with it, some harder)

PR #1369, opened 2025-03-28
osbuild: add new stage org.osbuild.hmac (HMS-5924)

achilleas-k:
Add osbuild stage type for the `org.osbuild.hmac` stage introduced in https://github.com/osbuild/osbuild/pull/2055

JIRA: [HMS-5924](https://issues.redhat.com/browse/HMS-5924)


pkg/osbuild/hmac_stage.go
```diff
@@ -0,0 +1,44 @@
...
+
+type HMACStageOptions struct {
+	Paths     []string `json:"paths"`
+	Algorithm string   `json:"algorithm"`
+}
+
+func (o *HMACStageOptions) isStageOptions() {}
+
+func (o *HMACStageOptions) validate() error {
+	if len(o.Paths) == 0 {
```
bcl, line 17: This should check `o` for nil first, or change `NewHMACStage` to check options for nil.

pkg/osbuild/hmac_stage_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHMACStageValidation(t *testing.T) {
+	type testCase struct {
+		options *HMACStageOptions
+		expErr  string
+	}
+
+	testCases := map[string]testCase{
```
bcl, line 15: Add a test with options not set.

bcl (changes requested):
Couple small tweaks

achilleas-k:
> Couple small tweaks

Good catch.  Thanks!  Amended.

pkg/osbuild/hmac_stage.go
```diff
@@ -0,0 +1,47 @@
...
+	if o.Algorithm == "" {
+		return fmt.Errorf("'algorithm' is a required property")
+	}
+
+	algorithms := []string{
+		"sha1",
+		"sha224",
+		"sha256",
+		"sha384",
+		"sha512",
```
thozza, line 32: Nitpick: It may be good to define these values as constants, so that it is clear what are the valid options.

PR #1372, opened 2025-03-31
rhel8,9: move partitioning into YAML

mvo5:
An updated WIP branch that uses the new `partition_table_override` idea which seems much easier to read then the previous attempt that used `partition_table_conditional` (c.f. https://github.com/osbuild/images/pull/1339). Still very messy but another (small) step into the right direction I think. Opening mostly get give ppl the chance to follow the current evolution of this work.


mvo5:
Closing for now in favor of https://github.com/osbuild/images/pull/1388

PR #1375, opened 2025-04-01
distro: almalinux and almalinux_kitten

supakeen:
Adds support for Alma Linux 8 and 9 based on the downstream patches currently carried there [1]. Also adds Alma Linux Kitten under the `kitten` name; this is the 'streaming' version of Alma Linux 10 [2].

[1]: https://git.almalinux.org/autopatch/osbuild-composer/src/branch/a10s/files
[2]: https://almalinux.org/blog/2024-10-22-introducing-almalinux-os-kitten/

---

Support for using Alma Linux as the host is landing in: https://github.com/osbuild/osbuild/pull/2060

### DIFF
--- a/pkg/distro/host.go
+++ b/pkg/distro/host.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"strconv"
 	"strings"
 )
 
@@ -24,8 +25,41 @@ func GetHostDistroName() (string, error) {
 	if _, ok := osrelease["ID"]; !ok {
 		return "", errors.New("cannot get the host distro name: missing ID field in os-release")
 	}
+
 	if _, ok := osrelease["VERSION_ID"]; !ok {
 		return "", errors.New("cannot get the host distro name: missing VERSION_ID field in os-release")
+	}
+
+	// In some cases (Alma Linux Kitten) `ID + VERSION_ID` is not the correct
+	// approach to determine the distribution. For Kitten the `/etc/os-release`
+	// file contains a `VERSION_ID` without a minor. We need to special case this:
+	if osrelease["ID"] == "almalinux" {
+		versionParts := strings.Split(osrelease["VERSION_ID"], ".")
+
+		if len(versionParts) > 2 {
+			return "", errors.New("failed to parse version from os-release, too many dotted parts")
+		}
+
+		majorVersion, err := strconv.Atoi(versionParts[0])
+		if err != nil {
+			return "", errors.New("failed to parse major version from os-release")
+		}
+
+		minorVersion := -1
+		if len(versionParts) > 1 {
+			minorVersion, err = strconv.Atoi(versionParts[1])
+			if err != nil {
+				return "", errors.New("failed to parse minor version from os-release")
+			}
+		}
+
+		// If `/etc/os-release` is `almalinux-10` then we select
+		// `almalinux_kitten` as our host distro. Note that it might be possible to
+		// not verify the major version but it's not guaranteed that with an eventual
+		// version 11 we will need to do this swap as well.
+		if majorVersion == 10 && minorVersion == -1 {
+			osrelease["ID"] = "almalinux_kitten"
+		}
 	}
 
 	name := osrelease["ID"] + "-" + osrelease["VERSION_ID"]
--- a/pkg/distro/host_test.go
+++ b/pkg/distro/host_test.go
@@ -24,6 +24,7 @@ func TestGetHostDistroName(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "toucanOS-42", name)
 }
+
 func TestGetHostDistroNameUnhappy(t *testing.T) {
 	backup := getHostDistroNameTree
 	defer func() { getHostDistroNameTree = backup }()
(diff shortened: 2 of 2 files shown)


Conan-Kudo:
AlmaLinux Kitten 10 is the continuous release and its repositories are available.

AlmaLinux OS 10 wll not have its first release until later this year, and the beta release is only present in the Vault.

That said, does osbuild have support for mirrorlist or metalink? For AlmaLinux, mirrorlist is preferred anyway.

pkg/distro/rhel/distribution.go
```diff
@@ -142,6 +146,25 @@ func NewDistribution(name string, major, minor int) (*Distribution, error) {
...
 			runner:           &runner.CentOS{Version: uint64(major)},
 		}
+	case "almalinux":
+		if major < 0 {
+			return nil, errors.New("Invalid AlmaLinux major version (must be positive)")
+		}
+
+		if minor < 0 {
+			return nil, errors.New("AlmaLinux requires a minor version")
+		}
```
Conan-Kudo, line 156: Kitten doesn't have a minor version.

pkg/distro/rhel/rhel10/distro.go
```diff
@@ -293,6 +297,11 @@ func ParseID(idStr string) (*distro.ID, error) {
 		return nil, fmt.Errorf("rhel requires minor version, but got: %d", id.MinorVersion)
 	}
 
+	// So does AlmaLinux
+	if id.Name == "almalinux" && id.MinorVersion == -1 {
+		return nil, fmt.Errorf("almalinux requires minor version, but got: %d", id.MinorVersion)
+	}
```
Conan-Kudo, line 303: This has to account for Kitten too.

pkg/distro/rhel/rhel10/distro.go
```diff
@@ -40,10 +40,14 @@ var (
 func distroISOLabelFunc(t *rhel.ImageType) string {
 	const RHEL_ISO_LABEL = "RHEL-%s-%s-0-BaseOS-%s"
 	const CS_ISO_LABEL = "CentOS-Stream-%s-BaseOS-%s"
+	const ALMALINUX_ISO_LABEL = "AlmaLinux-%s-%s-%s-dvd"
```
Conan-Kudo, line 43: We need a Kitten label too.

Conan-Kudo (changes requested):

supakeen:
@Conan-Kudo I didn't include the Kitten patch (yet) as per our chat I wanted to figure out if it's expected that there are no Alma Linux 10.x repositories, even though the downstream patches refer to them.

It seems like we should have Alma Linux 8.x and 9.x (but no 10.x until it gets released and the repositories are in their normal places) and then a separate Alma Linux Kitten which is 10.x?

Conan-Kudo:
Yep, that's right.

supakeen:
I've updated this PR to also include Alma Linux Kitten (called `kitten`, as per downstream patches). I need to do some verification of my own before I'll undraft this PR for review.

Conan-Kudo (changes requested):

pkg/distro/rhel/distribution.go
```diff
@@ -101,6 +101,10 @@ func (d *Distribution) IsAlmaLinux() bool {
 	return strings.HasPrefix(d.name, "almalinux")
 }
 
+func (d *Distribution) IsKitten() bool {
+	return strings.HasPrefix(d.name, "kitten")
+}
```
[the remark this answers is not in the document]
supakeen: Right; but we only have `kitten` as referring to Alma Linux so I don't think there's a need for an additional check here; see `pkg/distro/rhel/distribution.go` unless I'm missing something else? :)

Conan-Kudo:
That patch is because the way to identify Kitten is by having `almalinux-X` instead of `almalinux-X.Y` for the `$ID-$VERSION_ID` pair in `os-release(5)`.

supakeen:
> That patch is because the way to identify Kitten is by having `almalinux-X` instead of `almalinux-X.Y` for the `$ID-$VERSION_ID` pair in `os-release(5)`.

I'm debating (internally) if it's necessary to also add this patch as it only affects distribution selection based on the host (e.g. when you don't specify a distro which distro should be built) and probably for runner selection, the latter is where it probably doesn't matter all that much.

The former I need to confirm/test a bit more before I include it :)

pkg/distro/rhel/distribution.go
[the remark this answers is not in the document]
supakeen: Going to track separately here: https://github.com/osbuild/images/issues/1377

supakeen:
My internal debate is over and I've included (a *slightly* rewritten) version of the `GetHost` function, including tests to see if it correctly swaps distros.

Conan-Kudo (approved):

achilleas-k (commented):
I did a quick read through and this looks good, but I'd like to do a more thorough review before approving.

supakeen:
> I did a quick read through and this looks good, but I'd like to do a more thorough review before approving.

Most attention should probably be aimed at the `GetHost` conditionals (though I did add testcases), the rest of it is (hopefully) pretty generic 'names' :)

ondrejbudai (approved):
lgtm, thanks!

Conan-Kudo:
> I'd also be partial to calling `kitten`, `almalinux_kitten` instead which will need some special care in sorting prefixes.

I would prefer `almalinux_kitten` if we can do that.

supakeen:
> > I'd also be partial to calling `kitten`, `almalinux_kitten` instead which will need some special care in sorting prefixes.
> 
> I would prefer `almalinux_kitten` if we can do that.

Let's do that then; marking as draft so this doesn't gain more approvals and an accidental merge in the meanwhile :)

supakeen:
I've adjusted the name to be `almalinux_kitten`; available versions for Alma Linux now read  like this:

```console
$ go run ./cmd/image-builder list --filter type:qcow2 --filter distro:'alma*' --filter arch:x86_64
almalinux-8.4 type:qcow2 arch:x86_64
almalinux-8.5 type:qcow2 arch:x86_64
almalinux-8.6 type:qcow2 arch:x86_64
almalinux-8.7 type:qcow2 arch:x86_64
almalinux-8.8 type:qcow2 arch:x86_64
almalinux-8.9 type:qcow2 arch:x86_64
almalinux-8.10 type:qcow2 arch:x86_64
almalinux-9.0 type:qcow2 arch:x86_64
almalinux-9.1 type:qcow2 arch:x86_64
almalinux-9.2 type:qcow2 arch:x86_64
almalinux-9.3 type:qcow2 arch:x86_64
almalinux-9.4 type:qcow2 arch:x86_64
almalinux-9.5 type:qcow2 arch:x86_64
almalinux_kitten-10 type:qcow2 arch:x86_64
```

Conan-Kudo (approved):

Conan-Kudo:
Is there anything holding this up to land?

supakeen:
@Conan-Kudo Reviews/acks. Since it's easter holidays there aren't many people around. I'll ask a few people to take a look later today.

ondrejbudai (approved):
Go!

Conan-Kudo:
Awesome, I see it's merged now. When should I expect it to show up in a release?

supakeen:
> Awesome, I see it's merged now. When should I expect it to show up in a release?

I'll do an additional release of `image-builder` either today or tomorrow as there is some other things I'd like to land.

supakeen:
> Awesome, I see it's merged now. When should I expect it to show up in a release?

The updates are now in pending state on bodhi.

Conan-Kudo:
Will we see it for the EPEL builds too? I only see Fedora ones...

PR #1376, opened 2025-04-01
distro: add `ImageType.BasePartitionTable()`

mvo5:
Ondrej suggested to have `image-builder describe-image` include the base partition table and it seems this is needed to make this reality (but maybe I'm missing something?).

---

This commit exposes the base partition table for `ImageType`s.
The rational is that we want to expose those via:
```console
$ image-bulder describe-image qcow2
```
so that the yaml of the partition table is visible. This helps
developing our image definitions wihtout having to look at the
code and as added benefit it gives a "flattened" version of the
partition YAML.

As part of the work in fedora the `basePartitionTablesFunc` was
removed as it is now fully based on YAML. Once all partition
tables in RHEL are also in YAML [0] we can do a similar
simplification there.

[0] https://github.com/osbuild/images/pull/1372

---

defs: add error types when an partition table is not found

This commit allows to distinguish errors when loading a partition
table. It adds:
- ErrImageTypeNotFound
- ErrNoPartitionTableForImgType
- ErrNoPartitionTableForArch
So that the high level helpers can distinguish if there was a
error with e.g. yaml loading or just a missing partition table
(which for e.g. container types is not an error).


achilleas-k (dismissed):
Cheers!

achilleas-k (approved):

supakeen (approved):

PR #1379, opened 2025-04-01
gen-manifests: show depsolve config in mock depsolve

achilleas-k:
Add another fake package to each package list returned from the mock depsolver that reflects the depsolve config.  The name of each package consists of:
- The transaction index
- The name of each repository used in that transaction
- The word "weak" if weak dependencies are enabled


cmd/gen-manifests/main.go
```diff
@@ -363,6 +363,32 @@ func mockDepsolve(packageSets map[string][]rpmmd.PackageSet, repos []rpmmd.RepoC
 
 				specSet = append(specSet, spec)
 			}
+
+			// generate pseudo packages for the config of each transaction
+			var setRepoNames []string
+			for _, setRepo := range pkgSet.Repositories {
+				setRepo.Hash()
```
supakeen, line 370: Should we be doing something with the hash here?

achilleas-k: It calculates and sets the Id property

achilleas-k: Ugh, sorry, we actually don't need it.

I initially tested using the Id instead of the name but it was too long and harder to read.  Dropping the line.

achilleas-k: Also I was wrong, it doesn't set the Id, it just returns it. :upside_down_face:

supakeen (approved):

mvo5 (approved):
Love it!

PR #1382, opened 2025-04-02
build(deps): ensure toolbox is the latest

lzap:
So we are fighting with dep bots, the problem is often that `toolchain` is different. Everytime you run a tool from [Go toolchain](https://go.dev/doc/toolchain) starting from Go 1.21 it will record the version in `go.mod`. It seems it is only upgrading this number, so here is an idea.

This patch detects latest and greatest upstream Go version and pins the `toolchain` setting to that version so it is always up-to-date. If my theory is correct, this will prevent Go bots from touching this value ever again (up until the point new Go version is released and we run the script again to bump it).

This also removes the need of running or installing `go1.XX` separate command because this is exactly what `toolchain` does. If you run any `go` command in a project that has `toolchain` set, Go will automatically download and use this particular version for all the commands.

This will ultimately do the same thing - we will all use the latest and greatest Go version for development, testing or running `go mod` commands but we will keep `go` major version at lower level. Currently it is set to 1.22.9, this also means we will not allow updates of this major version until it is available in RHEL. This is the limitation we cannot workaround in any way, sometimes bots will fail on this and there is nothing we can do about it.

I also moved `go mod tidy` as the last command, it must be the last as I learned the other day (e.g. running `go generate` after it will cause issues).

Note: when you initialy run this branch it will be downloading the latest 1.24.2 it takes a minute, then it is fast. Also note that from now on, everyone will use 1.24.2 for all the local work - that is not an issue it is the `go 1.22.X` statement in `go.mod` what makes the compatibility promise.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -172,9 +172,9 @@ jobs:
         run: sudo apt install -y libgpgme-dev libbtrfs-dev libdevmapper-dev
 
       - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v6
+        uses: golangci/golangci-lint-action@v7
         with:
-          version: v1.61.0
+          version: latest
           args: --verbose --timeout 5m0s
 
   shellcheck:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,29 +1,40 @@
----
-linters-settings:
-  govet:
-    disable:
-      - shadow  # default value recommended by golangci
-      - composites
-  gosec:
-    excludes:
-      - G101
-
-linters:
-  enable:
-    - gosec
-
+version: "2"
 run:
   build-tags:
     - integration
-  timeout: 5m
-
+linters:
+  enable:
+    - gosec
+  settings:
+    gosec:
+      excludes:
+        - G101
+    govet:
+      disable:
+        - shadow
+        - composites
+  exclusions:
+    generated: lax
+    presets:
+      - comments
+      - common-false-positives
+      - legacy
+      - std-error-handling
+    rules:
+      - linters:
+          - gosec
+        path: ^(images/)?(pkg|internal)/.*_test\.go$
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
 issues:
-  # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
   max-issues-per-linter: 0
-
-  # Maximum count of issues with the same text. Set to 0 to disable. Default
-  # is 3.
   max-same-issues: 0
-  exclude-rules:
-    - path: ^(images/)?(pkg|internal)/.*_test\.go$
-      linters: ["gosec"]
+formatters:
+  exclusions:
+    generated: lax
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
--- a/tools/prepare-source.sh
+++ b/tools/prepare-source.sh
@@ -2,20 +2,22 @@
 
 set -eux
 
+# Figure out the latest and greatest Go version
+GO_LATEST=$(curl -s https://endoflife.date/api/go.json | jq -r '.[0].latest')
+
 # Go version must be consistent with image-builder which uses UBI
 # container that is typically few months behind
 GO_VERSION=1.22.9
-GO_BINARY=$(go env GOPATH)/bin/go$GO_VERSION
 
-# this is the official way to get a different version of golang
-# see https://go.dev/doc/manage-install
-go install golang.org/dl/go$GO_VERSION@latest
-$GO_BINARY download
-
-# Ensure that go.mod and go.sum are up to date.
-$GO_BINARY mod tidy
+# Pin Go and toolchain versions at a reasonable versions
+go get "go@$GO_VERSION" "toolchain@$GO_LATEST"
 
 # Ensure the code is formatted correctly.
-$GO_BINARY fmt ./...
+go fmt ./...
 
+# Update go.mod and go.sum
+go mod tidy
+
+# Generate CI
 ./test/scripts/generate-gitlab-ci ./.gitlab-ci.yml
+
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/osbuild/images
 
-go 1.22.8
+go 1.22.9
 
-toolchain go1.22.11
+toolchain go1.24.2
 
 require (
 	cloud.google.com/go/compute v1.33.0


tools/prepare-source.sh
```diff
@@ -2,20 +2,22 @@
...
 
 # Ensure the code is formatted correctly.
-$GO_BINARY fmt ./...
+go fmt ./...
 
+# Generate CI
 ./test/scripts/generate-gitlab-ci ./.gitlab-ci.yml
+
+# Update go.mod and go.sum (keep it as the last)
+go mod tidy
```
achilleas-k, line 22: The `generate-gitlab-ci` runs `./cmd/list-images`, so it should be after the `go mod tidy`, otherwise it might fail and complain about the `go.mod` and `go.sum` being inconsistent.

achilleas-k (commented):
The linter seems to be blowing up.  Presumably because we're running a pinned version of the linter with go1.22.
What's the proper way to handle this?  Should we be linting with a version that matches the toolchain?

lzap:
That is very likely what is happening, more recent versions of the linter do properly state what is going on.

```
Error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24.2)
Failed executing command with error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24.2)
```

Is there any specific reason not to use the most recent linter version? It is backwards compatible so it should not blow up linting 1.22 code. Let me try to change the configuration to use the latest version. It also requires migration of the configuration to the new V2 format.

.golangci.yml
```diff
@@ -1,29 +1,40 @@
...
-  exclude-rules:
-    - path: ^(images/)?(pkg|internal)/.*_test\.go$
-      linters: ["gosec"]
+formatters:
+  exclusions:
+    generated: lax
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
lzap, line 40: I used `golangci-lint migrate` for this, can put the two comments back if you think it holds any value, I am not fan of repeating what is in documentation. It was also inconsistent - most of other config values had no comments anyway.

achilleas-k: I don't think we need to keep those comments.

Where did these exclusion paths come from though?  Probably built-in defaults?  Let's drop them and add any we might need later.  Right now, we want all `.go` files to be considered.

.github/workflows/tests.yml
```diff
@@ -172,9 +172,9 @@ jobs:
         run: sudo apt install -y libgpgme-dev libbtrfs-dev libdevmapper-dev
 
       - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v6
+        uses: golangci/golangci-lint-action@v7
         with:
-          version: v1.61.0
+          version: latest
```
achilleas-k, line 177: Let's not do this.  This makes it unpredictable when the linter version will change, which means we can get unrelated linting errors on PRs and not know why.  It also forces unrelated changes into PRs.  We could be making a small docs PR, get a new `latest` linter version, and suddenly have to start changing code to get it through.

achilleas-k (changes requested):

.github/workflows/tests.yml
lzap: Well, I do not see occasional linter error which can be solved by bumping your local linter to the newest possible version as such a drag compared to the problem I am trying to solve, but sure. If you don't like this then I believe this PR is a dead end. Closing.

achilleas-k: What about updating the action to v2 though?  I think that's certainly worth doing, I just don't like the idea of a "floating" version for the linter itself.

lzap: Sure I can do that, but in a different PR.

lzap:
So I tried to upgrade the linter but for some weird reason it blows up the same way. I do not understand these nonsense results, so I guess we just wait with the upgrade?

```
$ golint
INFO golangci-lint has version 2.0.2 built with go1.24.1 from 2b224c2c on 2025-03-25T21:36:18Z
INFO [config_reader] Config search paths: [./ /app / /root]
INFO [config_reader] Used config file .golangci.yml
INFO maxprocs: Leaving GOMAXPROCS=10: CPU quota undefined
INFO [goenv] Read go env for 3.584958ms: map[string]string{"GOCACHE":"/root/.cache/go-build", "GOROOT":"/usr/local/go"}
INFO [lintersdb] Active 6 linters: [errcheck gosec govet ineffassign staticcheck unused]
INFO [loader] Using build tags: [integration]
INFO [loader] Go packages loading at mode 8767 (compiled_files|deps|exports_file|files|name|imports|types_sizes) took 24.390562897s
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 25.316125ms
INFO [linters_context/goanalysis] analyzers took 37.373782251s with top 10 stages: buildir: 1.507337562s, gosec: 544.200485ms, unused: 406.718282ms, S1008: 372.652409ms, S1035: 353.727624ms, S1032: 349.868525ms, S1038: 345.643569ms, S1005: 298.118349ms, SA9007: 297.098997ms, SA1012: 291.502182ms
INFO [runner/exclusion_paths] Skipped 0 issues by pattern "examples$"
INFO [runner/exclusion_paths] Skipped 0 issues by pattern "third_party$"
INFO [runner/exclusion_paths] Skipped 0 issues by pattern "builtin$"
INFO [runner/exclusion_rules] Skipped 0 issues by rules: [Path: "^(images/)?(pkg|internal)/.*_test\\.go$", Linters: "gosec"]
INFO [runner] Issues before processing: 2258, after processing: 11
INFO [runner] Processors filtering stat (in/out): exclusion_paths: 2156/2156, generated_file_filter: 2156/2156, uniq_by_line: 2156/11, max_per_file_from_linter: 11/11, path_prettifier: 11/11, cgo: 2258/2258, filename_unadjuster: 2258/2258, diff: 2156/2156, fixer: 2156/2156, max_from_linter: 11/11, source_code: 11/11, path_shortener: 11/11, sort_results: 11/11, path_relativity: 2156/2156, max_same_issues: 11/11, severity-rules: 11/11, path_absoluter: 2258/2258, invalid_issue: 2258/2156, exclusion_rules: 2156/2156, nolint_filter: 2156/2156
INFO [runner] processing took 1.258616ms with stages: path_relativity: 434.183µs, source_code: 191.503µs, exclusion_paths: 100.503µs, invalid_issue: 73.684µs, exclusion_rules: 71.453µs, cgo: 68.262µs, filename_unadjuster: 68.247µs, uniq_by_line: 64.459µs, nolint_filter: 62.677µs, path_absoluter: 56.297µs, generated_file_filter: 53.328µs, sort_results: 4.951µs, path_shortener: 4.487µs, max_per_file_from_linter: 1.778µs, max_same_issues: 911ns, fixer: 733ns, path_prettifier: 571ns, diff: 279ns, max_from_linter: 224ns, severity-rules: 86ns
INFO [runner] linters took 1.64980537s with stages: goanalysis_metalinter: 1.648445997s
pkg/distro/fedora/distro.go:536:66: not enough arguments in call to common.Must
	have (unknown type)
	want (T, error) (typecheck)
		defaultImageConfig: common.Must(defs.DistroImageConfig(nameVer)),
		                                                               ^
pkg/distro/rhel/rhel10/bare_metal.go:7:2: "github.com/osbuild/images/pkg/rpmmd" imported and not used (typecheck)
	"github.com/osbuild/images/pkg/rpmmd"
	^
pkg/distro/rhel/rhel10/distro.go:53:53: not enough arguments in call to common.Must
	have (unknown type)
	want (T, error) (typecheck)
	return common.Must(defs.DistroImageConfig(d.Name()))
	                                                   ^
pkg/distro/rhel/rhel7/distro.go:16:53: not enough arguments in call to common.Must
	have (unknown type)
	want (T, error) (typecheck)
	return common.Must(defs.DistroImageConfig(d.Name()))
	                                                   ^
pkg/distro/rhel/rhel8/bare_metal.go:7:2: "github.com/osbuild/images/pkg/rpmmd" imported and not used (typecheck)
	"github.com/osbuild/images/pkg/rpmmd"
	^
pkg/distro/rhel/rhel8/distro.go:40:53: not enough arguments in call to common.Must
	have (unknown type)
	want (T, error) (typecheck)
	return common.Must(defs.DistroImageConfig(d.Name()))
	                                                   ^
pkg/distro/rhel/rhel8/edge.go:12:2: "github.com/osbuild/images/pkg/rpmmd" imported and not used (typecheck)
	"github.com/osbuild/images/pkg/rpmmd"
	^
pkg/distro/rhel/rhel9/bare_metal.go:8:2: "github.com/osbuild/images/pkg/rpmmd" imported and not used (typecheck)
	"github.com/osbuild/images/pkg/rpmmd"
	^
pkg/distro/rhel/rhel9/distro.go:56:53: not enough arguments in call to common.Must
	have (unknown type)
	want (T, error) (typecheck)
	return common.Must(defs.DistroImageConfig(d.Name()))
	                                                   ^
pkg/distro/rhel/rhel9/edge.go:14:2: "github.com/osbuild/images/pkg/rpmmd" imported and not used (typecheck)
	"github.com/osbuild/images/pkg/rpmmd"
	^
pkg/reporegistry/repository_test.go:15:2: "github.com/osbuild/images/internal/common" imported and not used (typecheck)
	"github.com/osbuild/images/internal/common"
	^
11 issues:
* typecheck: 11
INFO File cache stats: 11 entries of total size 97.2KiB
INFO Memory: 262 samples, avg is 70.4MB, max is 755.3MB
INFO Execution took 26.070142897s
```

achilleas-k:
In my experience, those import errors happen when the required go version for the dependency is higher than the one in the project's `go.mod`.

PR #1384, opened 2025-04-03
manifestgen: extra io.Writer for depsolver

supakeen:
Let's pass along a DepsolveOutput field all the way to the depsolver so we can use it for the stderr of the `dnf-json` command. This then allows us to capture its output (mostly: package in group but excluded warnings) in, for example, `image-builder-cli`.

The depsolver already had provisions for passing in a writer for `os.Stderr`.


pkg/manifestgen/manifestgen_test.go
```diff
@@ -136,7 +136,7 @@ func TestManifestGeneratorWithOstreeCommit(t *testing.T) {
 	assert.Contains(t, osbuildManifest.String(), expectedSha256)
 }
 
-func fakeDepsolve(cacheDir string, packageSets map[string][]rpmmd.PackageSet, d distro.Distro, arch string) (map[string]dnfjson.DepsolveResult, error) {
+func fakeDepsolve(cacheDir string, warningsOutput io.Writer, packageSets map[string][]rpmmd.PackageSet, d distro.Distro, arch string) (map[string]dnfjson.DepsolveResult, error) {
```
mvo5, line 139: Let's add a test for the new depsolveOuput too:
1. AIUI warningsOutput is depsolveOutput, lets tweak the name to avoid confusion with the existing warningsOutputs which contolls something else
2. in the fakeDepsolve, lets write something to the depsolveOuptut (if non-nil)
3. and a test that passes a buffer where we check that the output ends up. 
4. it does not test much (but it also shouldn't be much work) but its still useful to ensure the plumbing is correct, i.e. l115,l169

mvo5 (commented):
Thank you! I like this, lets a add a small test and then this should go in, it will be a nice UX improvement!

pkg/manifestgen/manifestgen_test.go
supakeen: Yes, let's add a test. I must've forgotten the name because initially I was passing along `warningsOutput` instead of having a separate one (which was bad).

supakeen: I've added the test (I think mostly as you described) :)

mvo5 (approved):
Thank you, this looks fine and I like the fact that we can now discard the (confusing) warnings.

I was wondering if we need to capture the logs from dnfjson to put them into our buildlogs but that can/should all be followup work. I also wonder if we should capture the output in case of errors but then it seems that this really should be the job ob dnfjson itself so I think we are good for now.

supakeen:
Had to force push to appease the linter (return value from `.Write`).

pkg/manifestgen/manifestgen.go
```diff
@@ -49,6 +49,11 @@ type Options struct {
 	// generate an error.
 	WarningsOutput io.Writer
 
+	// DepsolveOutput will receive any warnings that are part of
+	// the depsolving step. If it is unset output ends up on the
+	// default stdout/stderr.
+	DepsolveOutput io.Writer
```
thozza, line 55: Nitpic: for consistency, would it make sense to call this `DepsolveWarnings` or `DepsolveWarningsOutput` (which is quite long :confused: ) instead? The actual depsolve output is where we get the response from the depsolve tool.

thozza (dismissed):
I have one naming suggestion, but overall, the PR looks good.

pkg/manifestgen/manifestgen.go
supakeen: Yea I kinda debated that, I dunno. If others feel strongly I'm fine with renaming it. The `*Warnings` things also carry (some) meaning in that in other places those would be direct `error`'s if not redirected.

thozza: Sure, but ultimately, the writer will be used for the solver's `Stderr` output, not its `Stdout`. This is my main concern with naming it `...Output`.

supakeen: So you feel strongly. It's been renamed.

mvo5 (approved):

thozza (approved):
Thank you!

PR #1385, opened 2025-04-03
ci: dependency updater via gobump

lzap:
Please review carefully, I did not test this locally as I have no idea how to do that. My assumption is we would merge this and then test it by manually triggering it?

To get more info about `gobump` please read: https://github.com/lzap/gobump

I can either move the project into osbuild GH org or give maintainer permissions to anyone interested just in case. My account is 2FA protected.

### DIFF
--- a/.github/workflows/gobump.yml
+++ b/.github/workflows/gobump.yml
@@ -1,0 +1,43 @@
+---
+name: "Updates Go dependencies via gobump"
+
+on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
+  schedule:
+    # Every Sunday at 15:00
+    - cron: "0 15 * * 0"
+
+jobs:
+  update-and-push:
+    runs-on: ubuntu-latest
+    container: registry.fedoraproject.org/fedora:41
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+
+      - name: Update go.mod
+          sudo dnf -y install gpgme-devel btrfs-progs-devel krb5-devel
+          echo -e '## gobump output:\n\n```' > github_pr_body.txt
+          go run github.com/lzap/gobump@latest -exec "go build ./..." -exec "go test ./..." 2>&1 | tee -a github_pr_body.txt
+          echo -e '\n```' >> github_pr_body.txt
+          ./tools/prepare-source.sh
+
+      - name: Open PR
+        env:
+          GITHUB_TOKEN: ${{ secrets.SCHUTZBOT_GITHUB_ACCESS_TOKEN }}
+        run: |
+          if git diff --exit-code; then echo "No changes"; exit 0; fi
+          git config --unset-all http.https://github.com/.extraheader
+          git config user.name "schutzbot"
+          git config user.email "schutzbot@gmail.com"
+          branch="schutz-gobump-$(date -I)"
+          git checkout -b "${branch}"
+          git add -A
+          git commit -m "build(deps): Update dependencies via gobump"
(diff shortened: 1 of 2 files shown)


.github/workflows/gobump.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  update-and-push:
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+
+      - name: Update go.mod
+        run: |
+          go install github.com/lzap/gobump@latest
+          gobump -exec "go build ./pkg/..." -exec "go test ./pkg/osbuild" 2>&1 | tee gobump.log
```
achilleas-k, line 19: Why are we only building `./pkg` and testing `./osbuild`?  Is it too much to run with `./...`?  I suppose everything in `./internal` is covered by imports from `./pkg`, but there's a lot that would be missed by an update breaking something in `./cmd`.

achilleas-k (commented):
Thanks!  Question inline about the `-exec` commands.

.github/workflows/gobump.yml
lzap: No particular reason other than time and output size, the thing is packages are vendored and I believe `./...` is testing or building all directories including `vendor/`? So let me add more explicit directories there.

I am bit worried about length of the output tho as it goes as PR input, particularly `go test` leaves ton of output behind which cannot be turned off. Let's see how it looks like first and then I can fine-tune it. If this works I will probably add a markdown output to the tool so we get nicely formatted report of an update.

lzap:
Rebased.

.github/workflows/gobump.yml
achilleas-k: Dependencies aren't vendored in this repository.

achilleas-k:
> Rebased. Added a bit of markdown formatting so the output of the command is verbosely rendered.

If the output from the tool is too noisy, we could have it only print on failure, so successful runs are silent, but errors are visible.

achilleas-k:
Alternatively, the PR body could be a short report of what changed (though, the diff would also provide that same information), and the log can be collapsed

<details>
<summary>Like this</summary>

```
noisy output log here
```
</details>

.github/workflows/gobump.yml
lzap: Ah good point changed both to `./...` then there is plenty of time on Sunday midnight on CICD so we can make sure everything tests correctly.

lzap:
I tested this locally and found out it commits the `gobump.log` because it was not in `gitignore`. Also there was a quoting problem seems like '"' does interpret backticks. Do not review or merge until I finish off testing.

lzap:
So I tested this locally, it works but it will not on GHA since they only provide ubuntu images for actions. This probably needs fedora only? Building and running tests requires CGO and deps which is not possible.

lzap:
Okay just found out one can use containers on GHA, so let's do that. I tested everything locally except PR push which I assume works since this was copy-pasted from a live action. I also found how to decrease verbosity of `go test` as well so the output will be reasonable.

lzap:
So do we want to try this out? @achilleas-k @supakeen

achilleas-k (changes requested):
Ok, let's give this a shot.
But let's please clean up the commits in the PR.  The first commit modifies go.sum and the second reverts it back.  Please squash the two commits so it's only the github action and the docs change.

If you're feeling extra generous, maybe make the docs change in a separate commit from the action as well.

lzap:
Squashed.

achilleas-k (approved):
Alright.  Looking forward to seeing this in action.
Let's trigger a manual run of the action as soon as it's merged.

PR #1386, opened 2025-04-03
manifest,image: stop embedding `OSCustomizations` in `manifest.OS`

mvo5:
This commit is very mechanical, it stops embedding the `manifest.OSCustomizations` directly inside `manifest.OS`.

This makes it more explicit what comes from a OSCustomization and what is just part of OS which makes it easier to refactor `ImageConfig` and `OSCustomizations` as there is currently much manual copying.


thozza (approved):
:heart_eyes:

achilleas-k (approved):
In my excitement, I forgot to approve.

PR #1387, opened 2025-04-03
[RFC] distro: unify OsCustomizations()

mvo5:
[draft as it needs comment and is build on top of https://github.com/osbuild/images/pull/1386]

This is an RFC to tweak the way we do the manifest.OSCustomizations and as a first
step I would like to remove the duplicated generators we currently have. Note that this
step is quite mechanical too and tries to keep the delta as smal as possible to make
reviewing this easier. It may go too far (see the invidual commit descriptions) but I think
its a step in the right direction (if that looks reasonable the next step is to do the same
for `ostreeDeploymentCustomizations)

---

distro: unify OsCustomizations()

This commit removes the two (almost) identical copies of
`OsCustomizations()` in {fedora,rhel} in favor of a single
implementation that takes the new ImageTypeConfig in additon
to the existing parameters.

This (hopefully) goes into the direction that Achilleas suggested
where we have a new single SuperConfig that is produced via
a smart merge of `ImageConfig`, `ImageTypeConfig`, `DistroConfig`,
Blueprints etc. But in order to make progress on this journey
we need to remove some duplication first :)

Along the way a subtle difference in the hostname handling was
unified and RHEL now also understands about `MachineIdUninitialized`
(which was previously only available for fedora probably because
of drift).

There is a tradeoff here of course, its slightly^W ugly that now
distro.OsCustomizations needs to understand about the details
of *all* distros/imageTypes we support. We could explore extracing
a common helper and then let the distro deal with specialized stuff
like subscriptions. But moving into this (other) extreme seems
a slightly better trade-off than the current duplicated ~300 lines
that we need to keep drift-free.
[edit: also OSCustomizations already contains fields for RHEL specific 
stuff so probably fine in this context]

---

distro: introduce new shared `distro.ImageTypeConfig` and use

This commit is a first step to eliminate the duplicated
`osCustomizations()` methods in fedora/rhel by extracting a new
`distro.ImageTypeConfig` that contains the configuration parts
of the distro image type.

### DIFF
--- a/cmd/osbuild-playground/my-container.go
+++ b/cmd/osbuild-playground/my-container.go
@@ -51,7 +51,7 @@ func (img *MyContainer) InstantiateManifest(m *manifest.Manifest,
 
 	// create a minimal non-bootable OS tree
 	os := manifest.NewOS(build, &platform.X86{}, repos)
-	os.BasePackages = []string{"@core"}
+	os.OSCustomizations.BasePackages = []string{"@core"}
 	os.OSCustomizations.Language = "en_US.UTF-8"
 	os.OSCustomizations.Hostname = "my-host"
 	os.OSCustomizations.Timezone = "UTC"
--- a/cmd/osbuild-playground/my-image.go
+++ b/cmd/osbuild-playground/my-image.go
@@ -46,8 +46,8 @@ func (img *MyImage) InstantiateManifest(m *manifest.Manifest,
 
 	// create a minimal bootable OS tree
 	os := manifest.NewOS(build, platform, repos)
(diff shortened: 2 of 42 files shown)


thozza:
Just a quick drive-by comment. I think that this is a bit of a step in the side. IMHO, we should instead unify the `Distro` interface implementations in Fedora and RHEL and really have only one. This will naturally de-duplicate also `OsCustomizations()` function.

I'm not against abstracting the image type configuration further, but it seems that we may as well get away with the existing structures after the merge of RHEL and Fedora.

mvo5:
Closing based on the comment from thozza - this will be done differently.

PR #1394, opened 2025-04-07
distro/rhel9/azure: blacklist more modules on 9.6+ (RHEL-79065)

### LINKED ISSUE #1398: Blacklist `intel_uncore` and `acpi_cpufreq` in addition to the existing modules.

achilleas-k:
Added intel_uncore and acpi_cpufreq to the list of blacklisted modules for Azure on RHEL 9.6+.

### DIFF
--- a/pkg/distro/rhel/rhel9/azure.go
+++ b/pkg/distro/rhel/rhel9/azure.go
@@ -474,6 +474,23 @@ func defaultAzureImageConfig(rd *rhel.Distribution) *distro.ImageConfig {
 
 	if rd.IsRHEL() {
 		ic.GPGKeyFiles = append(ic.GPGKeyFiles, "/etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release")
+		if common.VersionGreaterThanOrEqual(rd.OsVersion(), "9.6") {
+			ic.Modprobe = append(
+				ic.Modprobe,
+				&osbuild.ModprobeStageOptions{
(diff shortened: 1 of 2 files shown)


supakeen (approved):

thozza (commented):
Shouldn't this change be applied also to `c9s` Azure images? :thinking:

achilleas-k:
Good question.  They should probably be aligned, but I didn't spend too much time thinking about it because this (and a lot more changes that are coming in the next few days) are a priority with pretty a short timeframe.

mvo5 (approved):
This is fine with me, I think the question about centos-9 is valid, but iirc (from working on this as part of the YAMLification we have a few more places where we do version checks without also checking for centos where centos-9 is effecitvely handled as rhel-9.0. So maybe worth  a comment but I expect this to get noticed/cleaned up as part of the YAML work

thozza:
> centos-9 is effecitvely handled as rhel-9.0.

lzap:
Just for the record, I do not see these two in

https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/deploying_and_managing_rhel_on_microsoft_azure/deploying-a-rhel-image-as-an-compute-instance-on-azure

But I will carry over them to bootc derived container, this is harmless and while I could not find this in the Jira ticket, I guess this is just to silence some kernel warning during boot. These should be irrelevant for a VM.

PR #1405, opened 2025-04-08
distro/rhel9/azure: disable microcode.service on 9.6+

achilleas-k:


pkg/distro/rhel/rhel9/azure.go
```diff
@@ -490,6 +490,11 @@ func defaultAzureImageConfig(rd *rhel.Distribution) *distro.ImageConfig {
 					},
 				},
 			)
+
+			ic.DisabledServices = append(
+				ic.DisabledServices,
+				"microcode.service",
+			)
```
achilleas-k, line 497: Welp.  We're gonna need an architecture switch.  This is x86 only.

achilleas-k: Actually, I think we could (or should?) be excluding `microcode_ctl` entirely from Azure images.

mvo5 (commented):
This seems fine, what is keeping this in draft? is it unclear if we really need it?

achilleas-k:
I was asking around for confirmation that it is, in fact, the right thing to do.  Marking as ready.

mvo5 (approved):

PR #1406, opened 2025-04-08
tools/gen-manifest-checksums: sort by image name

achilleas-k:
Instead of sorting the file by checksum, sort by the image name (`--key 2`, the second field of each line).

This makes it easier to read diffs because the filenames don't move around when their checksum changes.  It makes it much easier to distinguish cases where image configurations are added or removed, as opposed to images changing.

NOTE: Merging this PR will create issues with any other open PRs that change image definitions, so let's maybe hold off on merging it for a bit.  Let me know what you think of the idea though.


achilleas-k:
Weird that it's failing the check.

achilleas-k:
Oooooooh.  Locale!

supakeen (approved):

mvo5 (approved):
Thank you

PR #1414, opened 2025-04-10
rhsm: support for podman secrets

lzap:
Image builder CLI currently does not build RHEL images and errors out with missing subscription error. This patch extends RHSM secrets utility function to search for entitlement certificate in runtime directory and then falls back to regular directory.

When a RHEL container is executed via podman, it automatically mounts subscription files making them available for RHSM running in UBI containers so it can access Red Hat repositories. Images can leverage this and simply use the X509 entitlement cert to access RHEL bits on CDN automatically.

When this is merged:

* [ ] Update ibcli README with information about this
* [ ] Create new option in ibcli which allows overriding repositories with own ones including GPG keys and basic config data (this will be useful for Satellite users)

```
# find /run/secrets/
/run/secrets/
/run/secrets/etc-pki-entitlement
/run/secrets/etc-pki-entitlement/941581227453314952-key.pem
/run/secrets/etc-pki-entitlement/941581227453314952.pem
/run/secrets/redhat.repo
/run/secrets/rhsm
/run/secrets/rhsm/ca
/run/secrets/rhsm/ca/redhat-entitlement-authority.pem
/run/secrets/rhsm/ca/redhat-uep.pem
/run/secrets/rhsm/rhsm.conf
/run/secrets/rhsm/syspurpose
/run/secrets/rhsm/syspurpose/syspurpose.json
```

### DIFF
--- a/pkg/rhsm/secrets.go
+++ b/pkg/rhsm/secrets.go
@@ -42,18 +42,26 @@ type ConsumerSecrets struct {
 }
 
 func getRHSMSecrets() (*RHSMSecrets, error) {
-	keys, err := filepath.Glob("/etc/pki/entitlement/*-key.pem")
-	if err != nil {
-		return nil, err
+	// search /etc first to allow container users to override the entitlements
+	globs := []string{
+		"/etc/pki/entitlement/*-key.pem",             // for regular systems
+		"/run/secrets/etc-pki-entitlement/*-key.pem", // for podman containers
 	}
-	for _, key := range keys {
-		cert := strings.TrimSuffix(key, "-key.pem") + ".pem"
-		if _, err := os.Stat(cert); err == nil {
-			return &RHSMSecrets{
-				SSLCACert:     "/etc/rhsm/ca/redhat-uep.pem",
-				SSLClientKey:  key,
-				SSLClientCert: cert,
-			}, nil
+
+	for _, glob := range globs {
+		keys, err := filepath.Glob(glob)
+		if err != nil {
+			return nil, err
+		}
+		for _, key := range keys {
+			cert := strings.TrimSuffix(key, "-key.pem") + ".pem"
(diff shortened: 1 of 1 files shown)


mvo5 (approved):
Thank you! Having a test would be nice but given that there is non already and this is pretty straightforward I think its fine as is and it should unblock https://github.com/osbuild/image-builder-collection/issues/3

lzap:
I just realized the order should be reverse so we allow container users to subscribe and "override" the podman-provided credentials. Amending.

supakeen (dismissed):
Looks OK to me, the explicit ordering of the filesystem paths could use a comment in the code.

supakeen:
> * Create new option in ibcli which allows overriding repositories with own ones including GPG keys and basic config data (this will be useful for Satellite users)

We have the option to override repositories, we are lacking the GPG keys. I believe the idea was that `--{force,extra}-repo` accept paths to repo configuration files (cc @mvo5).

pkg/rhsm/secrets.go
```diff
@@ -42,18 +42,25 @@ type ConsumerSecrets struct {
 }
 
 func getRHSMSecrets() (*RHSMSecrets, error) {
-	keys, err := filepath.Glob("/etc/pki/entitlement/*-key.pem")
-	if err != nil {
-		return nil, err
+	globs := []string{
```
mvo5, line 45: Maybe worth a comment here that the order is deliberate, something like: `// search /etc/ first to allow container users to override the entitlements` or something

mvo5 (dismissed):
Thank you! One tiny and optional suggestion inline.

thozza (dismissed):

lzap:
Added the missing comment, rebased.

supakeen (approved):

PR #1419, opened 2025-04-11
test: kernel name in base-host-check (COMPOSER-2498)

achilleas-k:
**test: kernel name in base-host-check**

---

**test: add the kernel.name customization to all-customizations**

This enables the kernel name customization test since we run
base-host-check on AMI builds and all-customizations is used with AMIs.

---


test/scripts/base-host-check.sh
```diff
@@ -169,4 +169,12 @@ if (( $# > 0 )); then
     if jq -e '.blueprint.enabled_modules' "${config}" || jq -e '.blueprint.packages[] | select(.name | startswith("@") and contains(":")) | .name' "${config}"; then
         check_modularity "${config}"
     fi
+
+    if jq -e '.blueprint.customizations.hostname' "${config}"; then
+        [[ "$(hostname)" == "$(jq -r '.blueprint.customizations.hostname' "${config}")" ]]
+    fi
+
+    if jq -e '.blueprint.customizations.kernel.name' "${config}"; then
```
mvo5, line 177: (super nitpick) maybe a small comment that the kernel-rt,debug customization sets this (its not part of all-configs) but maybe/probably overkill

mvo5 (approved):
Thank you

test/scripts/base-host-check.sh
achilleas-k: Not sure what you mean.  Yeah, we don't set any alt kernels in all-customizations.  Maybe we should make it `kernel-debug` though?

We have a `kernel-debug` config for `iot-commit`.  Maybe we should add `ami` to that config too to cover it.

achilleas-k: I added `kernel-debug` as the kernel name in `all-customizations`, making the name of the test a little more accurate.

achilleas-k:
There's some really interesting (and inconsistent) failures in the tests.
I think some of them are simply that we're hitting limits with the number of AWS instances we're running.  But I'm also seeing some hostname checks and failures to boot.
But that's Monday's problem

test/scripts/base-host-check.sh
```diff
@@ -169,4 +169,12 @@ if (( $# > 0 )); then
...
+        [[ "$(uname -s)" == "$(jq -r '.blueprint.customizations.kernel.name' "${config}")" ]]
```
thozza, line 178: This is probably not the correct approach.

```console
[root@vm-10-0-184-196 ~]# cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.7 Beta (Plow)
[root@vm-10-0-184-196 ~]# uname -s
Linux
```

thozza (changes requested):
Thanks for adding these tests.

The kernel name test is IMO not correct.

In addition, I would prefer not to use shortcuts to make the test script fail, but instead print a message indicating which check failed. This information is currently lost, and the script just "exits with 1"...

test/scripts/base-host-check.sh
achilleas-k: yeah I don't know what I was thinking.  It turns out this is pretty tricky to check.  I gave it a shot though.

achilleas-k:
Ideally, a failed check would not `exit 1` and we would print error messages and then exit if there's at least one, but doing this in bash is annoying so let's keep it like it is and rewrite it in a real language later.

achilleas-k:
Ok, now we're seeing a failure that I was sort of expecting.  The hostname doesn't match because cloud-init (I think it's cloud-init?) sets it again on boot.  I think this is a good opportunity to re-examine this.  Either find the best way to stop this from happening if and when a user sets a hostname on an AMI build, or inform users (through docs) that hostnames are overridden when launching on EC2 with default settings and how to disable it on the platform.

(Note: there's a bit of speculation about the inner workings here on my part; I'd like to dig deeper at some point)

supakeen:
> Ok, now we're seeing a failure that I was sort of expecting. The hostname doesn't match because cloud-init (I think it's cloud-init?) sets it again on boot. I think this is a good opportunity to re-examine this. Either find the best way to stop this from happening if and when a user sets a hostname on an AMI build, or inform users (through docs) that hostnames are overridden when launching on EC2 with default settings and how to disable it on the platform.
> 
> (Note: there's a bit of speculation about the inner workings here on my part; I'd like to dig deeper at some point)

@lucasgarfield and I discussed this in our chat; because I found out the same in #1432 where I do print the values if things failed :)

supakeen:
Obsoleted by #1432. If you can drop the hostname check from this one then we can merge the kernel-change.

achilleas-k:
Haven't forgotten about this but I'll get back to it when I have time, unless someone finds some time to clean it up and only do the kernel name test.

mvo5 (approved):
I can

thozza (approved):
LGTM, Thanks!

achilleas-k:
Rebased and fixed conflicts.

thozza (dismissed):
LGTM

mvo5 (approved):

supakeen:
The boot tests are failing (but it seems unrelated to this change?) and the manifest checksums are clashing. Otherwise looks good :)

achilleas-k:
> The boot tests are failing (but it seems unrelated to this change?) and the manifest checksums are clashing. Otherwise looks good :)

Yeah, not sure what's up there.  I'll rebase and fix the conflict and check back.

thozza (approved):

achilleas-k:
Some tests are failing rather consistently on the boot check.  I don't think it's an issue with this change, unless `kernel-debug` is causing issues.  I'd like to keep this PR open and see if I can add some stability to the tests (at least some retry logic?) in another PR.

PR #1423, opened 2025-04-11
Shrink the current installer images a bit

bcl:
While working on adding a new netboot image type I came across a few things.
- dracut setup was totally different from lorax/livemedia-creator (see commit and comments)
- installer images don't need docs or weak deps
- lz4 compression of the image isn't as good as xz and the bcj filter

So these commits change those things, hopefully not breaking anything, leading up to a netboot (a large one) next.

For the existing Fedora 42 image-installer these changes shrink it by 938MiB

Currently a draft to see what CI thinks  of it.


pkg/distro/fedora/images.go
```diff
@@ -508,6 +508,9 @@ func imageInstallerImage(workload workload.Workload,
 		img.AdditionalDrivers = append(img.AdditionalDrivers, installerConfig.AdditionalDrivers...)
 	}
 
+	// On Fedora anaconda needs dbus-broker, but isn't added when dracut runs.
```
supakeen, line 511: Is there a reason for this; are we missing a package in the set? Probably not since enabling the module doesn't fail but it *feels* weird :)

supakeen (approved):
Looks good, I like the single-dracutOptions :)

bcl:
Hm, anyone know why the manifest checksums are failing? I ran the generate script and the new checksums are in the last commit.

supakeen:
> Hm, anyone know why the manifest checksums are failing? I ran the generate script and the new checksums are in the last commit.

It runs it on every commit, so if you have the checksums as a separate commit it'll fail on the previous ones. I learned this yesterday as well, @achilleas-k told me to do:

```
git rebase -i origin/main --exec ./tools/gen-manifest-checksums.sh
```

Which opens every commit in your branch for amending and runs the `gen-manifest-checksums` after which you can add and continue. Probably drop the separate checksum commit first.

pkg/distro/fedora/images.go
supakeen: Understood, let's leave it as is then :)

achilleas-k:
Also it's not a required check.  I made the test check every commit as a bit of an experiment to see how annoying it would be.  The idea is that it helps reviewers or git history readers know when a commit had an effect on generated manifests, without needing to have whole manifests checked in like we used to.

So if it's really annoying, I'm open to discussing its usefulness.

supakeen (approved):

achilleas-k (approved):
Great cleanup.  Thank you!!
LGTM

thozza (approved):
LGTM

PR #1432, opened 2025-04-15
base-host-check: expand to cover more (COMPOSER-2498)

supakeen:
Verifies the hostname configuration, if set.


supakeen:
TODO:

- [ ] Properties of users (group, authorized_keys, password).
- [ ] Check masked systemd services.
- [ ] Check ports in the firewall as well.

supakeen:
@lucasgarfield / @croissanne, when using the hostname customization for an AMI, and booted on EC2 the hostname gets overwritten by `cloud-init` when the machine is brought online.

How do we want to handle this. This seems like a behavioral issue (we need to not allow hostname customizations when `cloud-init` will take care of it, *or* we need to disable `cloud-init`'s hostname stuff when it is customized)?

supakeen:
Looking at the results here it seems like `masked` services are not actually being masked; see example output:

```
+ echo '📗 Checking masked services'
++ jq -rc '.blueprint.customizations.services.masked[]' all-customizations.json
+ services_expected='nfs-server
rpcbind
nftables'
+ echo 'nfs-server
rpcbind
nftables'
+ read -r service_expected
++ systemctl list-unit-files --state=masked
+ states='UNIT FILE STATE PRESET
0 unit files listed.'
```

achilleas-k:
> This seems like a behavioral issue (we need to not allow hostname customizations when `cloud-init` will take care of it, _or_ we need to disable `cloud-init`'s hostname stuff when it is customized)?

I think it might be more reasonable to modify the cloud-init config to set `preserve_hostname: true` when a user sets a hostname.  I'm not sure though.  Maybe people would want to set a base hostname in their golden image but also override it on boot for some images.

lucasgarfield:
> @lucasgarfield / @croissanne, when using the hostname customization for an AMI, and booted on EC2 the hostname gets overwritten by `cloud-init` when the machine is brought online.
> 
> How do we want to handle this. This seems like a behavioral issue (we need to not allow hostname customizations when `cloud-init` will take care of it, _or_ we need to disable `cloud-init`'s hostname stuff when it is customized)?

For now, it seems like testing this in AWS is impractical. And of course, it also seems pointless to expose this to users for AWS. We'll handle all of that in the front end.

In the future, we'll probably need to extend these boot tests for all targets. Maybe we can just disable the hostname test for targets that are initialized with cloud-init.

test/scripts/base-host-check.sh
```diff
@@ -140,6 +140,21 @@ check_hostname() {
...
+    services_expected=$(jq -rc '.blueprint.customizations.services.enabled[]' "${config}")
+
+    echo "$services_expected" | while read -r service_expected; do
+        state=$(systemctl is-enabled "${service_expected}")
+        if [[ "${state}" == "enabled" ]]; then
+            echo "Service was enabled service=${service_expected} state=${state}"
+        else
+            echo "❌ Service was not enabled service=${service_expected} state=${state}"
+        fi
+    done
```
thozza, line 156: Shouldn't the check `exit 1` in case there are any failures? :thinking:

test/scripts/base-host-check.sh
```diff
@@ -155,6 +155,21 @@ check_services_enabled() {
...
+    services_expected=$(jq -rc '.blueprint.customizations.services.disabled[]' "${config}")
+
+    echo "$services_expected" | while read -r service_expected; do
+        state=$(systemctl is-enabled "${service_expected}" || true)
+        if [[ "${state}" == "disabled" ]]; then
+            echo "Service was disabled service=${service_expected} state=${state}"
+        else
+            echo "❌ Service was not disabled service=${service_expected} state=${state}"
+        fi
+    done
```
thozza, line 171: ditto: shouldn't the check `exit 1`?

test/scripts/base-host-check.sh
```diff
@@ -170,6 +170,21 @@ check_services_disabled() {
...
+    services_expected=$(jq -rc '.blueprint.customizations.services.masked[]' "${config}")
+
+    echo "$services_expected" | while read -r service_expected; do
+        states=$(systemctl list-unit-files --state=masked)
+        if echo "${states}" | grep -q "${service_expected}"; then
+            echo "Service was masked service=${service_expected}"
+        else
+            echo "❌ Service was not masked service=${service_expected} output=${states}"
+        fi
+    done
```
thozza, line 186: ditto: shouldn't the check `exit 1` in case there are any failures?

test/scripts/base-host-check.sh
```diff
@@ -125,6 +125,151 @@ check_modularity() {
...
+    expected_hostname=$(jq -r '.blueprint.customizations.hostname' "${config}")
+    actual_hostname=$(hostname)
+
+    # we only emit a warning here since the hostname gets reset by cloud-init and we're not
+    # entirely sure how to deal with it yet on the service level
+    if [[ $actual_hostname != "${expected_hostname}" ]]; then 
+        echo "🟡 Hostname was not set: hostname=${actual_hostname} expected=${expected_hostname}"
+    else
+        echo "Hostname was set"
+    fi
```
thozza, line 141: Could the check warn only if `cloud-init` service is enabled and fail in case there is no `cloud-init`?

thozza (commented):
Thank you for extending the test coverage, much appreciated.

I've added a few comments.

On a general note, we should consider not failing fast in checks that are checking multiple values in a loop, but failing after all values are checked.

test/scripts/base-host-check.sh
supakeen: Only if the service is enabled, if the package is installed, or if it is configured in a certain way? There were a lot of ways and since we don't *currently* test in any ways where `cloud-init` isn't being used I felt it wasn't necessary.

test/scripts/base-host-check.sh
supakeen: Done.

test/scripts/base-host-check.sh
supakeen: Done.

test/scripts/base-host-check.sh
supakeen: Done.

test/scripts/base-host-check.sh
thozza: My idea was to warn only if the cloud-init service is enabled, which implies that it is installed. It could be configured not to set the hostname, but let's not complicate it with this option.

You are, of course, right that the effect in our current testing matrix is none. I was thinking more in the context of us not having to remember to fix this test case when we ever end up boot-testing image types without `cloud-init`.

FWIW, I'm OK with keeping the test as is for now.

thozza (approved):
LGTM

PR #1433, opened 2025-04-15
chore: update Go to 1.23

lzap:
Relaxing pinned Go version to 1.23.6 which is the latest release which has been recently added into UBI9:

https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?architecture=amd64&image=67f7e5d0639042b2fe77b475&container-tabs=gti&gti-tabs=registry-tokens

Then I used the gobump tool to upgrade as many dependencies as possible but keep the Go to the pinned version:

https://github.com/lzap/gobump


achilleas-k:
Thanks!!

There's some Go version pinning in the test workflow too: https://github.com/osbuild/images/blob/a33dc725f80ea75d08f86f52abcbd53f2186b0ce/.github/workflows/tests.yml

Annoying, I know.  Having a single place where it's defined where all sources could read it would be nice, but I don't know if parsing the go.mod or something similar is the nicest way to do it.
For now let's just bump the version there as well and think about a better solution another time.

thozza:
> Annoying, I know. Having a single place where it's defined where all sources could read it would be nice, but I don't know if parsing the go.mod or something similar is the nicest way to do it. For now let's just bump the version there as well and think about a better solution another time.

I thought about this in the past and I have two ideas:

- Define a variable with the Go version at the top of the workflow file and use it everywhere. Bumping it then means changing it only in a single place.
- Actually extract the version from `go.mod` in a job and set a variable with the value. All other jobs should then use that to setup Go. I tried this in the past elsewhere and it is possible. The best part is that this means updating Go version only in `go.mod`...

achilleas-k:
> > Annoying, I know. Having a single place where it's defined where all sources could read it would be nice, but I don't know if parsing the go.mod or something similar is the nicest way to do it. For now let's just bump the version there as well and think about a better solution another time.
> 
> I thought about this in the past and I have two ideas:
> 
> * Define a variable with the Go version at the top of the workflow file and use it everywhere. Bumping it then means changing it only in a single place.
> * Actually extract the version from `go.mod` in a job and set a variable with the value. All other jobs should then use that to setup Go. I tried this in the past elsewhere and it is possible. The best part is that this means updating Go version only in `go.mod`...

I suppose reading from the go.mod is best since that's the one authoritative source of truth.

achilleas-k:
> Ok I am trying this new feature: https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

Oh, this is perfect!  Nice!!

lzap:
Had to put setup-go after checkout, obviously it needs `go.mod` to be present.

lzap:
So "Unit tests" actions were completely missing `setup-go` so I guess it was running with a "random" Go version that gets installed into Ubuntu image. Added at both places to workaround:

```
Run go test -race  ./...
go: go.mod requires go >= 1.23.6 (running go 1.23.4; GOTOOLCHAIN=local)
```

lzap:
Just realized we need to keep the Go version as low as the oldest supported OS, Go was only updated to 1.23 in UBI9 but not in RHEL9 so I guess this is a dead end for me for now.

lzap:
Wait COS9 has Go 1.23 is it is just Fedora I assume.

lzap:
I will rebase this once we remove F40.

Does anyone know what this error message is:

```
Run go test -race ./... -test.skip 'TestBlockingResolverLocalManifest|TestResolverLocalManifest'
OCI runtime exec failed: exec failed: unable to start container process: exec: "sh": executable file not found in $PATH: unknown
```

lzap:
So to sum it up:

* Fedora 40: Go 1.22 (we are dropping it soon)
* Fedora 41: Go 1.23.4
* CentOS 9: Go 1.23.7
* CentOS 10: Go 1.23.1

So the final Go version will be: Go 1.23.1 🎉

achilleas-k:
So we keep this open until we're ready to drop F40?

lzap:
> So we keep this open until we're ready to drop F40?

Looks like a plan, converted to draft for the time being.

lzap:
Closing for now until we get rid of F40.

PR #1434, opened 2025-04-15
RHEL 9.6+/Azure: systemd service and script for dataloss warning on temporary resource disk

### LINKED ISSUE #1426: systemd service for Azure temporary resource disk dataloss warning
- Requires adding a custom script to `/usr/local/sbin/` and a systemd service to run it.
- The systemd service requires `StandardOutput`, which was added recently as part of this group of tasks (https://github.com/osbuild/osbuild/pull/2067).

achilleas-k:
This PR adds a script and systemd service in RHEL 9.6+ Azure images.  The script runs on boot, looks for the mountpoint of a temporary resource disk, identified by the device node (`/dev/disk/cloud/azure_resource-part1`), and adds a warning file at the root of the filesystem to warn against using it for any non-ephemeral data.

Closes #1426.

Note: I'm not a huge fan of having this script embedded in our code for this one specific quirk and carrying it with us forever.  I'm proposing we add this now and find a better way to handle it later (osbuild stage?  rpm?).


thozza (approved):
All looks good! Thanks for renaming the constants in the `osbuild` package :heart_eyes:

PR #1437, opened 2025-04-15
github: don't run manifest checksum validation on main

achilleas-k:
Move the test to its own workflow file and configure it to run only on PRs and the merge queue.  It doesn't make sense to run it on pushes to main.


ondrejbudai (approved):
yup yup, thanks!

PR #1440, opened 2025-04-15
fedora: bump branched and rawhide

supakeen:
Fedora 42 is final, no longer branched. Rawhide should've been bumped ages ago but wasn't affecting any real artifacts as the condition is only used on the live installer to gate webui.


achilleas-k (approved):
Cheers!

achilleas-k:
Oh this affects the live installer `final` like you said, so the manifest checksums need updating.

supakeen:
> Oh this affects the live installer `final` like you said, so the manifest checksums need updating.

And the other installers since we toggle final in the bootstamp as well; I've regenerated the checksums and force-pushed.

ondrejbudai (approved):
Thanks!

achilleas-k (approved):

PR #1442, opened 2025-04-16
ci: dependency updater gobump logging

lzap:
So `gobump` itself works fine but the subsequent action "Push a PR" does error out with:

```
warning: Not a git repository. Use --no-index to compare two paths outside a working tree
usage: git diff --no-index [<options>] <path> <path>
```

I have no idea why it does not work, it seems the working directory changes? Adding few debugging statements.


lzap:
I don't understand, we use `container` in other GHA and it seems that it should work seamlessly. Why would `git` fail like that.

lzap:
Found a solution possibly: https://stackoverflow.com/questions/72575700/running-docker-from-github-actions-cant-find-file-added-during-previous-step/72595316

Rebasing.

achilleas-k (approved):
The `run` block is getting a bit too long, but it's fine for now.  It might be worth making a GH action that wraps gobump and does all of this at some point.

lzap:
OMG `git diff` exits with: `Error: Process completed with exit code 5.`

Why is that? The line is:

```
if git diff --exit-code; then echo "No changes"; exit 0; fi
```

How on Earth this can exit with 5. Any idea?

PR #1443, opened 2025-04-16
manifest: register insights to template on boot (HMS-5994)

rverdile:
Supports registering with a content template on bootup

For RHEL 9.6 and 10, the rhc connect command can be used with the template to register. For RHEL 8, a curl command is used to associate the system to the template through the patch API

JIRA: [HMS-5994](https://issues.redhat.com/browse/HMS-5994)

### DIFF
--- a/pkg/manifest/subscription.go
+++ b/pkg/manifest/subscription.go
@@ -3,6 +3,7 @@ package manifest
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/osbuild/images/pkg/customizations/fsnode"
 	"github.com/osbuild/images/pkg/customizations/subscription"
@@ -116,12 +117,24 @@ func subscriptionService(subscriptionOptions subscription.ImageOptions, serviceO
 
 	var commands []string
 	if subscriptionOptions.Rhc {
+		var curlToAssociateSystem string
 		// Use rhc for registration instead of subscription manager
-		commands = []string{fmt.Sprintf("/usr/bin/rhc connect --organization=${ORG_ID} --activation-key=${ACTIVATION_KEY} --server %s", subscriptionOptions.ServerUrl)}
+		rhcConnect := fmt.Sprintf("/usr/bin/rhc connect --organization=${ORG_ID} --activation-key=${ACTIVATION_KEY} --server %s", subscriptionOptions.ServerUrl)
+		if subscriptionOptions.TemplateUUID != "" {
+			curlToAssociateSystem = getCurlToAssociateSystem(subscriptionOptions)
+		} else if subscriptionOptions.TemplateName != "" {
+			rhcConnect += fmt.Sprintf(" --content-template %s", subscriptionOptions.TemplateName)
+		}
+		commands = append(commands, rhcConnect)
 		// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
 		commands = append(commands, "restorecon -R /root/.gnupg")
 		// execute the rhc post install script as the selinuxenabled check doesn't work in the buildroot container
 		commands = append(commands, "/usr/sbin/semanage permissive --add rhcd_t")
+		// register to template if template uuid is specified
+		if curlToAssociateSystem != "" {
+			commands = append(commands, curlToAssociateSystem)
+			commands = append(commands, "/usr/sbin/subscription-manager refresh")
+		}
 		if insightsOnBoot {
 			icDir, icFile, err := runInsightsClientOnBoot()
 			if err != nil {
@@ -138,6 +151,12 @@ func subscriptionService(subscriptionOptions subscription.ImageOptions, serviceO
 			commands = append(commands, "/usr/bin/insights-client --register")
 			// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
 			commands = append(commands, "restorecon -R /root/.gnupg")
+			// register to template if template is specified
+			if subscriptionOptions.TemplateUUID != "" {
+				curlToAssociateSystem := getCurlToAssociateSystem(subscriptionOptions)
+				commands = append(commands, curlToAssociateSystem)
+				commands = append(commands, "/usr/sbin/subscription-manager refresh")
+			}
 			if insightsOnBoot {
 				icDir, icFile, err := runInsightsClientOnBoot()
 				if err != nil {
@@ -228,4 +247,14 @@ ExecStartPre=
 [Install]
 WantedBy=multi-user.target
 `
+}
+
+func getCurlToAssociateSystem(subscriptionOptions subscription.ImageOptions) string {
+	patchURL := strings.TrimSuffix(subscriptionOptions.PatchURL, "/")
+	addTemplateURL := fmt.Sprintf("%s/templates/%s/subscribed-systems", patchURL, subscriptionOptions.TemplateUUID)
+	curlToAssociateSystem := fmt.Sprintf("curl -v --retry 5 --cert /etc/pki/consumer/cert.pem --key /etc/pki/consumer/key.pem -X PATCH %s", addTemplateURL)
+	if subscriptionOptions.Proxy != "" {
+		curlToAssociateSystem += fmt.Sprintf(" --proxy %s", subscriptionOptions.Proxy)
+	}
+	return curlToAssociateSystem
 }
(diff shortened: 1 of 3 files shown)


pkg/manifest/subscription.go
```diff
@@ -138,6 +142,13 @@ func subscriptionService(subscriptionOptions subscription.ImageOptions, serviceO
 			commands = append(commands, "/usr/bin/insights-client --register")
 			// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
 			commands = append(commands, "restorecon -R /root/.gnupg")
+			// register to template if template is specified. For Prod only.
+			if subscriptionOptions.TemplateUUID != "" {
+				patchURL := fmt.Sprintf("%s/api/patch/v3/templates/%s/subscribed-systems", "https://cert.console.redhat.com", subscriptionOptions.TemplateUUID)
+				curlToAssociateSystem := fmt.Sprintf("curl -v --retry --proxy %s %s --cert /etc/pki/consumer/cert.pem  --key /etc/pki/consumer/key.pem -X PATCH", subscriptionOptions.Proxy, patchURL)
```
thozza, line 148: I think that it would be desirable to make this deployment-agnostic if possible and not mention prod at all. Would it make sense to move the URL to the `subscription.ImageOptions` and have it be specified by the library caller, instead of hardcoding it in the library itself? I have the same question for the API endpoint template, but I expect that to be already deployment-agnostic and stable.

thozza (commented):
Thanks for the PR and added unit tests.

I have one question in-line.

On a general note, could you please add the reasoning for the change (basically the PR description) to the commit message? In general, we want as much information as possible captured in the commit message. Git history is provider agnostic, while PRs are not, so we may eventually lose the information.

pkg/manifest/subscription.go
rverdile: I think that makes sense. I pushed a change to that effect. If you had something else in mind I'm happy make other changes.

thozza (dismissed):
LGTM, thank you!

rverdile:
@thozza Apologies, I missed something. This also needs to handle the case of RHEL <9.6 using rhc, which would need to use the curl command. I pushed up a new change with a new test.

I'm assuming we can choose to send the template name from image builder specifically for the case of RHEL 9.6 and 10, and otherwise send the template UUID. Happy to make any changes if this doesn't make sense.

thozza:
> @thozza Apologies, I missed something. This also needs to handle the case of RHEL <9.6 using rhc, which would need to use the curl command. I pushed up a new change with a new test.
> 
> I'm assuming we can choose to send the template name from image builder specifically for the case of RHEL 9.6 and 10, and otherwise send the template UUID. Happy to make any changes if this doesn't make sense.

If I understand you correctly, I would say yes. The logic in the image definitions code can make the right choice based on what is set in the `subscription.ImageOptions`. Then, the image-builder-crc (or the API client) should be able to put the correct options based on the RHEL version being built.

pkg/manifest/subscription.go
```diff
@@ -116,12 +116,29 @@ func subscriptionService(subscriptionOptions subscription.ImageOptions, serviceO
...
+		}
+		commands = append(commands, rhcConnect)
 		// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
 		commands = append(commands, "restorecon -R /root/.gnupg")
 		// execute the rhc post install script as the selinuxenabled check doesn't work in the buildroot container
 		commands = append(commands, "/usr/sbin/semanage permissive --add rhcd_t")
+		// register to template if template uuid is specified
+		if curlToAssociateSystem != "" {
+			commands = append(commands, curlToAssociateSystem)
+			commands = append(commands, "/usr/sbin/subscription-manager refresh")
```
thozza, line 140: Please move these to the `if subscriptionOptions.TemplateUUID != "" ` statement block, which is the only case when `curlToAssociateSystem` is actually set. You have it done this way in the case when `rhc` is not used.

pkg/manifest/subscription.go
```diff
@@ -138,6 +155,17 @@ func subscriptionService(subscriptionOptions subscription.ImageOptions, serviceO
...
+			// register to template if template is specified
+			if subscriptionOptions.TemplateUUID != "" {
+				patchURL := subscriptionOptions.PatchURL
+				if patchURL[len(patchURL)-1] == '/' {
+					patchURL = patchURL[:len(patchURL)-1]
+				}
+				addTemplateURL := fmt.Sprintf("%s/templates/%s/subscribed-systems", patchURL, subscriptionOptions.TemplateUUID)
+				curlToAssociateSystem := fmt.Sprintf("curl -v --retry --proxy %s %s --cert /etc/pki/consumer/cert.pem  --key /etc/pki/consumer/key.pem -X PATCH", subscriptionOptions.Proxy, addTemplateURL)
+				commands = append(commands, curlToAssociateSystem)
+				commands = append(commands, "/usr/sbin/subscription-manager refresh")
```
thozza, line 167: This code block is identical to the one used with `rhc`. Please de-duplicate it into a function and call it from both places instead.

pkg/manifest/subscription.go
```diff
@@ -138,6 +155,17 @@ func subscriptionService(subscriptionOptions subscription.ImageOptions, serviceO
...
 			// insights-client creates the .gnupg directory during boot process, and is labeled incorrectly
 			commands = append(commands, "restorecon -R /root/.gnupg")
+			// register to template if template is specified
+			if subscriptionOptions.TemplateUUID != "" {
+				patchURL := subscriptionOptions.PatchURL
+				if patchURL[len(patchURL)-1] == '/' {
+					patchURL = patchURL[:len(patchURL)-1]
+				}
+				addTemplateURL := fmt.Sprintf("%s/templates/%s/subscribed-systems", patchURL, subscriptionOptions.TemplateUUID)
+				curlToAssociateSystem := fmt.Sprintf("curl -v --retry --proxy %s %s --cert /etc/pki/consumer/cert.pem  --key /etc/pki/consumer/key.pem -X PATCH", subscriptionOptions.Proxy, addTemplateURL)
```
thozza, line 165: I just noticed that `-proxy` is always added to the command. However, I expect that specifying the proxy is not mandatory. Would it make more sense to add it to the command conditionally, in case `subscriptionOptions.Proxy` is not empty?

thozza (changes requested):

pkg/manifest/subscription.go
rverdile: Good point

pkg/manifest/subscription.go
rverdile: I agree it feels  a little messy, but I did it this way to ensure the curl command runs after `rhc connect`. The template needs to be associated to the system with the patch API after the system is registered.

pkg/manifest/subscription.go
rverdile: agreed, that is likely the better practice

pkg/manifest/subscription.go
thozza: Thanks for the additional context.

If `TemplateUUID` and `TemplateName` are mutually exclusive, as they are in your implementation, the code is fine from my PoV.

thozza (commented):
Thanks for the changes and additional context. I have one small suggestion. Otherwise the PR LGTM :+1:

pkg/manifest/subscription.go
```diff
@@ -229,3 +247,18 @@ ExecStartPre=
 WantedBy=multi-user.target
 `
 }
+
+func getCurlToAssociateSystem(subscriptionOptions subscription.ImageOptions) string {
+	patchURL := subscriptionOptions.PatchURL
+	if patchURL[len(patchURL)-1] == '/' {
+		patchURL = patchURL[:len(patchURL)-1]
+	}
```
achilleas-k, line 255: This could be simplified to:
```
	patchURL = strings.TrimSuffix(patchURL, "/")
```

achilleas-k, line 255: This could be simplified to:
```
	patchURL = strings.TrimSuffix(patchURL, "/")
```

rverdile:
thanks @achilleas-k @thozza! pushed those changes

thozza (approved):

achilleas-k (approved):
LGTM.  Thank you!

Issue #1396, opened 2025-04-08
Azure (azure-rhui) image definition updates for RHEL 9.6+


achilleas-k:
I am working on update the Azure image definition for RHEL 9.6.  There's quite a few independent tasks, so I'm listing them all here for anyone to pick up and work in parallel.

- [ ] Disable subscription manager yum plugin.  Set `enabled=0` in `/etc/yum/pluginconf.d/subscription-manafer.conf`
    - Will probably require new osbuild stage options for the `org.osbuild.yum.config` stage.
- [ ] Disable the auto-enabling of yum plugins for subscription manager: `subscription-manager config --rhsm.auto_enable_yum_plugins=0`
    - Will probably require a new osbuild stage.
- [ ] Blacklist `intel_uncore` and `acpi_cpufreq` in addition to the existing modules.
    - Already done at time of writing.  Listing here for completeness.
- [ ] Kernel cmdline options:
    - `loglevel=3`
- [ ] Disable `microcode.service`
- [ ] Enable `nftables.service`
- [ ] Chrony config `/etc/chrony.conf`: `refclock PHC /dev/ptp_hyperv poll 3 dpoll -2 offset 0`
    - Requires new stage options in the `org.osbuild.chrony` stage.
- [ ] LVM config `/etc/lvm/lvm.conf`: `use_devicesfile=0`
    - Will probably require a new osbuild stage that modifies the config file in-place.
- [ ] Remove kdump entries from `/boot`
    - We want `/boot` to be free of kdump entries.  In our images, there should be none to begin with because we don't install `krump-utils`.
- [ ] systemd service for Azure temporary resource disk dataloss warning
    - Requires adding a custom script to `/usr/local/sbin/` and a systemd service to run it.
    - The systemd service requires `StandardOutput`, which was added recently as part of this group of tasks.

----

The following changes *might* be required.  Confirmation is pending.
- [ ] Kernel cmdline options:
    - `nvme_core.io_timeout=240`
- [ ] grub configs (`/etc/default/grub`):
    - `GRUB_TIMEOUT_STYLE=countdown`
    - `GRUB_TERMINAL="serial console"`
    - `GRUB_SERIAL_COMMAND="serial --speed=115200 --unit=0 --word=8 --parity=no --stop=1"`
    - We have stage options for these already.
- [ ] Network configuration `/etc/sysconfig/network-scripts/ifcfg-eth0`:
    - `PEERDNS=yes`
[...]

PR #1447, opened 2025-04-17
RHEL 9.6+ Azure: enable the dataloss warning service and add unmanaged devices to the NetworkManager config

achilleas-k:
1. Enable the dataloss warning service
2. Add org.osbuild.nm.conf stage
3. Add NetworkManager config to set unmanaged devices

Part of #1396
These came up after a review of the currently merged changes from QE.


pkg/osbuild/nm_conf_stage.go
```diff
@@ -0,0 +1,117 @@
...
+		settings.Keyfile == nil {
+		return fmt.Errorf("%s: at least one setting must be set", nmConfStageType)
+	}
+
+	if main := settings.Main; main != nil {
+		if err := validateArrayHasItems(main.NoAutoDefault); err != nil {
+			return fmt.Errorf("%s: main.no-auto-default %s", nmConfStageType, err)
+		}
+		if err := validateArrayHasItems(main.Plugins); err != nil {
+			return fmt.Errorf("%s: main.no-auto-default %s", nmConfStageType, err)
```
ondrejbudai, line 72: Copilot found out that this error message should mention `main.plugins`. It also suggested a test for it, as the `empty-main` test case only covers the first condition.

ondrejbudai (commented):
A minor nitpick, otherwise lgtm. :)

pkg/osbuild/nm_conf_stage.go
achilleas-k: Thanks!  Fixed.

thozza (approved):

ondrejbudai (approved):
Thanks!

PR #1453, opened 2025-04-22
Add netinst iso

bcl:
This builds off of the dracut changes in #1423


pkg/distro/fedora/distro.go
```diff
@@ -487,6 +484,32 @@ func mkMinimalRawImgType(d distribution) imageType {
 	return it
 }
 
+// mkNetinstISOImgType adds a netinst image type that creates an Anaconda boot.iso
+func mkNetinstISOImgType(d distribution) imageType {
+	return imageType{
+		name:        "netinst",
```
supakeen, line 490: I suppose we could call this `everything-netinst` to follow more closely what 'variant' it belongs to; I tried to do some work around this in #1233 so it might be nice to do this per-emptively here as well.

supakeen (commented):
Just a comment on the image type name.

pkg/distro/fedora/distro.go
bcl: Our support for variants doesn't really do much, and the boot.iso isn't tied to a specific variant, I just picked 'Everything' because 'Unknown' seems wrong to expose to users. Maybe keep netinst as an alias?

bcl:
Rebased this on top of the new main, added checksums to all the commits.
But it keeps failing tests in the RHEL9 commit, it times out in `TestOSTreeOptionsErrorForNonOSTreeImgTypes/rhel-9.2/x86_64/edge-container ` (I think) and I can't figure out why. Let's see what happens with CI...

bcl:
Rebased on top of new yaml changes, left the RHEL yaml off this version for the moment.

pkg/distro/defs/loader.go
```diff
@@ -189,7 +189,7 @@ func versionLessThanSortedKeys[T any](m map[string]T) []string {
 func DistroImageConfig(distroNameVer string) (*distro.ImageConfig, error) {
 	toplevel, err := load(distroNameVer)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w while loading %s", err, distroNameVer)
```
mvo5, line 192: :heart: thank you

bcl:
Still draft because I wanted to wait for the yaml dust to settle, and I'm exploring getting lorax's runtime-cleanup template script working.

bcl:
Rebased on the new generic code, I have another patch to move the hard-coded build packages to use the platform build_packages but that should wait until the rhel distros have been adapted to the new ways.

I think this is ready, I don't think it should be exposed in the service yet since it only adds Fedora and it doesn't include running the lorax cleanup script (I'll do that as a separate PR).

One thing to note is that I added the firmware to the anaconda-tree, since this is what actually boots the installer it should also have the firmware packages available in order to support the hardware at the same level as the installed system.

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -85,11 +85,35 @@
       image_format: "raw"
       packages:
         <<: *x86_64_bios_platform_packages
-        firmware:
+        booting:
```
mvo5, line 88: :heart:

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -133,6 +178,37 @@
           - "grub2-ppc64le"
           - "grub2-ppc64le-modules"
       bootloader: "grub2"
+    ppc64le_installer_platform: &ppc64le_installer_platform
+      <<: *ppc64le_bios_platform
+      image_format: "qcow2"
+      packages:
+        <<: *ppc64le_bios_platform_packages
+        booting:
+
```
mvo5, line 187: (super nitpick) is the extra newline here on purpose?

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -1796,3 +1870,121 @@ image_types:
...
+            - mtr
+            - vconfig
+            - spice-vdagent
+            - gdisk
+            - hexedit
+            - sg3_utils
+          condition:
+            architecture:
+              x86_64:
+                include:
```
mvo5, line 1983: (maybe a silly question) given that this image-type only uses platforms "x86_64" and "aarch64", do we need this conditional? Or should we just add `dmidecode` to the "default" package-set? I guess if the plan is that netinst becomes available on more arches it makes sense to already have the condition here.

pkg/manifest/anaconda_installer.go
```diff
@@ -455,6 +462,8 @@ func (p *AnacondaInstaller) dracutStageOptions() *osbuild.DracutStageOptions {
 			"dmsquash-live",
 			"convertfs",
 		}...)
+	default:
+		panic("Unknown AnacondaInstallerType in dracutStageOptions")
```
mvo5, line 466: Thanks for adding this, maybe we could include the unkonwn type in the panic, i.e. `panic(fmt.Errorf("unknown AnacondaInstallerType %v in dracutStageOptions", p.Type))`  ? (and by making it an error we prepare for a potential future with less panicing :)

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -219,14 +219,18 @@ var installerBootExcludePaths = []string{
...
 			Filename: "images/install.img",
 		}
-	} else if p.anacondaPipeline.Type == AnacondaInstallerTypeLive {
+	case AnacondaInstallerTypeLive:
 		squashfsOptions = osbuild.SquashfsStageOptions{
 			Filename: "LiveOS/squashfs.img",
 		}
+	default:
+		// Shouldn't be possible, but catch it anyway
+		panic("Unknown AnacondaInstallerType for NewSquashfsStage")
```
mvo5, line 233: Same suggestion as about about including the unknown type in the panic.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -258,14 +262,18 @@ func (p *AnacondaInstallerISOTree) NewSquashfsStage() *osbuild.Stage {
...
 			Filename: "images/install.img",
 		}
-	} else if p.anacondaPipeline.Type == AnacondaInstallerTypeLive {
+	case AnacondaInstallerTypeLive:
 		erofsOptions = osbuild.ErofsStageOptions{
 			Filename: "LiveOS/squashfs.img",
 		}
+	default:
+		// Shouldn't be possible, but catch it anyway
+		panic("Unknown AnacondaInstallerType for NewErofsStage")
```
mvo5, line 276: Same suggestion as about about including the unknown type in the panic.

pkg/distro/generic/images.go
```diff
@@ -928,6 +928,80 @@ func iotSimplifiedInstallerImage(workload workload.Workload,
 	return img, nil
 }
 
+// Make an Anaconda installer boot.iso
```
supakeen, line 931: We could probably be a bit more verbose on what exactly is a `boot.iso` but that's really just docs.

pkg/image/anaconda_tar_installer.go
```diff
@@ -94,26 +94,27 @@ func (img *AnacondaTarInstaller) InstantiateManifest(m *manifest.Manifest,
...
 		// so let's make sure the kickstart path option is set
 		if img.Kickstart.Path == "" {
 			img.Kickstart.Path = osbuild.KickstartPathOSBuild
 		}
 	}
 
+	installerType := manifest.AnacondaInstallerTypeNetinst
+	if img.Kickstart != nil {
+		installerType = manifest.AnacondaInstallerTypePayload
+	}
```
supakeen, line 114: Why do we do this? We always switch to a payload installer when there's a kickstart?

pkg/manifest/anaconda_installer.go
achilleas-k: `p.Type` is an int, right?  Not very useful, but we could add it and mention it's an "invalid enum value".

pkg/manifest/anaconda_installer.go
```diff
@@ -179,6 +179,9 @@ func (p *AnacondaInstaller) getBuildPackages(Distro) []string {
 func (p *AnacondaInstaller) getPackageSetChain(Distro) []rpmmd.PackageSet {
 	packages := p.anacondaBootPackageSet()
 
+	// Install firmware packages and other platform specific packages
+	packages = append(packages, p.platform.GetPackages()...)
```
achilleas-k, line 183: :heart:

pkg/distro/defs/fedora/distro.yaml
achilleas-k, line 88: What does this separation do exactly?  Is it just for readability?

thozza (commented):
Nice work. As Simon noted, I'd also be in favor os adding a separate Image Kind for the Netinst ISO and not trying to fit it onto the tar installer.

pkg/distro/defs/fedora/distro.yaml
bcl: Nope :)

pkg/distro/generic/images.go
```diff
@@ -928,6 +928,80 @@ func iotSimplifiedInstallerImage(workload workload.Workload,
...
+		img.AdditionalDrivers = append(img.AdditionalDrivers, installerConfig.AdditionalDrivers...)
+	}
+
+	// On Fedora anaconda needs dbus-broker, but isn't added when dracut runs.
+	img.AdditionalDracutModules = append(img.AdditionalDracutModules, "dbus-broker")
+
+	d := t.arch.distro
+
+	img.Product = d.Product()
+	img.Variant = "Everything"
```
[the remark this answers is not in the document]
bcl: Good idea!

bcl:
I've addressed all the issues except the concern about reusing tar installer for this. Ideally this would be the base for all the other installer iso types, so I do agree :) But I'd also like to see this merged before rearranging everything since it does work as it stands.

supakeen:
> I've addressed all the issues except the concern about reusing tar installer for this. Ideally this would be the base for all the other installer iso types, so I do agree :) But I'd also like to see this merged before rearranging everything since it does work as it stands.

I'm fine with doing rearrangement in a follow up, especially *after* we see what we need from the DVD ISO.

supakeen (dismissed):

supakeen (dismissed):
Might need some validation from @pcdubs?

bcl:
New version, rebased on main, and moved the netinst anaconda code into its own file instead of fiddling with the anaconda tar installer. I still think the firmware changes are important so let's see what Paul thinks. Eventually I'd like to see all these anaconda installer variations converge on a single struct and manifest function, but that's not for now.

Still does not include RHEL support, adding that should be a yaml changes once the RHEL definitions are completely moved. This also doesn't include any minimization and F42 is currently 1.3G when I build it here :/

supakeen (dismissed):

bcl:
Now that generic RHEL support has been merged I'm working on rebasing on that and adding RHEL netinst.

supakeen:
> Now that generic RHEL support has been merged I'm working on rebasing on that and adding RHEL netinst.

I'm honestly fine with doing Fedora first; having that build a bit in CI, before we continue with RHEL in a followup? That's the same what I did when introducing some new image types :)

pcdubs:
Looks Ok to me. I'll add it to the agenda for the iot-meeting to see if there is any other feedback.

supakeen:
For some reason (some of) the ostree builds in CI keep failing on full disks, cc @achilleas-k.

I'd like to land this because it's been open for ages and must be a huge chore to keep up to date.

thozza (approved):
LGTM, let's get this in...

EDIT: once all conflicts are resolved :grimacing:

supakeen (approved):

pkg/image/anaconda_net_installer.go
```diff
@@ -0,0 +1,151 @@
...
+	bootTreePipeline.KernelOpts = kernelOpts
+
+	// Payload type uses an OS Pipeline for the payload
+	var osPipeline *manifest.OS
+	if anacondaPipeline.Type == manifest.AnacondaInstallerTypePayload {
+		osPipeline = manifest.NewOS(buildPipeline, img.Platform, repos)
+		osPipeline.OSCustomizations = img.OSCustomizations
+		osPipeline.Environment = img.Environment
+		osPipeline.Workload = img.Workload
+	}
```
thozza, line 126: This is technically dead code, since `anacondaPipeline.Type` is always set to `manifest.AnacondaInstallerTypeNetinst` above.

thozza (approved):
LGTM. My inline comment can be a follow-up.

supakeen (dismissed):

PR #1456, opened 2025-04-23
many: propagate masked services

supakeen:
Masked services were not propagated through the workload onto images, meaning they never ended up in the manifest. Let's set them on the workload, and take them from the workload in the `os` pipeline.


ondrejbudai (approved):
Awesome, thanks!

thozza (approved):
LGTM, thanks!

achilleas-k (approved):

achilleas-k:
The failed jobs were caught by some manual test cache deletion that I did to unblock #1233

PR #1457, opened 2025-04-23
RHEL 9 Azure kernel command line arguments

### LINKED ISSUE #1455: RHEL 9.6 Azure kernel command line arguments
The following changes to the kernel args need to be made to align with the current marketplace images:
- Add `console=ttyAMA0` for ARM images.
- Remove `console=tty1`, `console=ttyS0`, `earlyprintk=ttyS0`, `rootdelay=300` from ARM images.
- Add `net.ifnames=0` to all images.

achilleas-k:
**distro/rhel9/azure: add net.ifnames=0 kernel arg**

Aligning kernel args with existing marketplace images for 9.6+.
Partially addresses #1455

---

**distro/rhel9/azure: change console kernel args for aarch64**

This aligns the aarch64 Azure images with the current marketplace images
for RHEL 9.5 and older.  The change drops console=tty1, console=ttyS0,
earlyprintk=ttyS0, and rootdelay=300 from the ARM (aarch64) kernel args
and adds console=ttyAMA0.

Unlike other changes we've been making to the Azure images, this is not
version-gated to 9.6+.  Our previous kernel args cause issues with
console output for aarch64 images, so it's better to change them across
the whole range.

We usually don't have image type initialisers that take an architecture
argument, so this approach of passing the architecture directly to the
initialiser is a bit quick and simple.  A more general solution, one
which might also support arch-specific image configs would be nicer.
Something similar to the partition table selection, which is a function
property of the image type, might be needed.

Partially addresses #1455.

---

Closes #1455


supakeen (dismissed):

ondrejbudai (dismissed):
Thank you!

supakeen (approved):

ondrejbudai (approved):
Thanks!

thozza (approved):
LGTM

PR #1462, opened 2025-04-24
WIP: move imageConfig into YAML (fedora, rhel-10)

mvo5:
This is a preview for moving ImageConfig into YAML  - this code moves fedora and rhel10 to get a feel for what issue there are.

Some of the challenges we need to resolve:
1. a bunch of missing custom unmarshallers custom types in YAML (smop)
2. inconsistency in our osbuild options, most common is "snake_case", but we use many different other styles, including: `ResourceDisk.Format`, `x11-keyboard` and singular/plural is often a problem.
3. in one case we put a variable "osVersion" into the imageConfig (sap.go). this is not directly mappable in yaml so we need a indirection here

But overall it is pretty straightforward.

JIRA: HMS-5284

### DIFF
--- a/pkg/distro/defs/loader.go
+++ b/pkg/distro/defs/loader.go
@@ -78,7 +78,9 @@ type imageType struct {
 type ImagePlatform struct {
 	Arch string
 
-	BIOS       bool
+	BIOS bool
+	Zipl bool
+
 	UEFIVendor string `yaml:"uefi_vendor"`
 	platform.BasePlatform
 }
--- a/pkg/distro/defs/loader.go
+++ b/pkg/distro/defs/loader.go
@@ -64,6 +64,7 @@ type imageConfig struct {
 }
 
 type conditionsImgConf struct {
+	Architecture    map[string]*distro.ImageConfig `yaml:"architecture,omitempty"`
 	DistroName      map[string]*distro.ImageConfig `yaml:"distro_name,omitempty"`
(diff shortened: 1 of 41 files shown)


pkg/distro/rhel/rhel10/sap.go
```diff
@@ -103,6 +103,11 @@ func sapImageConfig(osVersion string) *distro.ImageConfig {
 			),
 		},
 		// E4S/EUS
+		//
+		// XXX: this must become a declrative option like
```
mvo5, line 107: This can/will be fixed once https://github.com/osbuild/images/pull/1464 is merged

pkg/distro/defs/loader.go
```diff
@@ -54,6 +55,45 @@ type imageType struct {
...
+	Bootable      bool
+	DefaultSize   uint64 `yaml:"default_size"`
+	// the image func name: disk,container,live-installer,...
+	Image                  string
+	BuildPipelines         []string `yaml:"build_pipelines"`
+	PayloadPipelines       []string `yaml:"payload_pipelines"`
+	Exports                []string
+	RequiredPartitionSizes map[string]uint64 `yaml:"required_partition_sizes"`
+
+	Platforms map[string]ImagePlatform
```
mvo5, line 75: Fww, this is wrong, platforms must be a list instead of a map[arch]ImagePlatform as it is defined here. We could have mutliple aarch64 images for example with different platform config/boot files etc

mvo5:

PR #1464, opened 2025-04-24
distro: make `DNFConfig` more declarative

mvo5:
This commit removes the `DNFConfigStageOptions` from `ImageConfig` in favor of a simpler `DNFConfig` struct.

The rational is that `DNFConfigStageOptions` current gets the `osVersion` passed in directly. When we move to a YAML based ImageConfig this will be problematic as we do not have access to variables like this. By making this a declarative option we sidestep the problem.

This  generates no manifest diff.

Needed for https://github.com/osbuild/images/pull/1462

### DIFF
--- a/pkg/distro/image_config.go
+++ b/pkg/distro/image_config.go
@@ -57,7 +57,6 @@ type ImageConfig struct {
 	Tmpfilesd           []*osbuild.TmpfilesdStageOptions
 	PamLimitsConf       []*osbuild.PamLimitsConfStageOptions
 	Sysctld             []*osbuild.SysctldStageOptions
-	DNFConfig           []*osbuild.DNFConfigStageOptions
 	SshdConfig          *osbuild.SshdConfigStageOptions
 	Authconfig          *osbuild.AuthconfigStageOptions
 	PwQuality           *osbuild.PwqualityConfStageOptions
@@ -72,6 +71,7 @@ type ImageConfig struct {
 	NetworkManager      *osbuild.NMConfStageOptions
 
 	WSLConfig *WSLConfig
+	DNFConfig *DNFConfig
 
 	Files       []*fsnode.File
 	Directories []*fsnode.Directory
@@ -121,6 +121,11 @@ type WSLConfig struct {
 	BootSystemd bool
 }
 
+type DNFConfig struct {
+	SetReleaseVerVar bool
+	ForceIPv4        bool
+}
+
 // InheritFrom inherits unset values from the provided parent configuration and
 // returns a new structure instance, which is a result of the inheritance.
 func (c *ImageConfig) InheritFrom(parentConfig *ImageConfig) *ImageConfig {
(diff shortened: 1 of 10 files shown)


pkg/distro/image_config.go
```diff
@@ -146,6 +151,34 @@ func (c *ImageConfig) InheritFrom(parentConfig *ImageConfig) *ImageConfig {
...
+		res = append(res, &osbuild.DNFConfigStageOptions{
+			Config: &osbuild.DNFConfig{
+				Main: &osbuild.DNFConfigMain{
+					IPResolve: "4",
+				},
+			},
+		})
+	}
+	return res
+}
```
thozza, line 180: I'm not sure if I like this approach.

FWIW, I do understand that setting the `releasever` from the YAML to `osVersion` would require introducing some magic value that we would need to replace. Which is a can of worms, because then we get into templating the YAML. This solution is simple from that PoV.

What I don't like is that this limits our options in the `ImageConfig` significantly going forward. My position is that we should keep `ImageConfig` as powerful as possible, which means keeping the ability to take advantage of the stage options that we have. With this change, we lose the option to define arbitrary variables for DNF easily. Additionally, if we ever add any additional DNF config options, I don't want us to be adding booleans for every possible combination.

I'd like us to use abstracted options only for complex things that would require touching multiple configurations. Or at the minimum, keep our options open and don't reduce what we can define via `ImageConfig`.

What about a compromise such as:

- Keep the `DNFConfigStageOptions` in `ImageConfig` and expose it in YAML. Set the `IPResolve` value explicitly in the YAML.
- Add a `bool` option specifically just for `SetReleaseVerVar` and in such case, set the `releasever` to `osVersion` in the DNFConfigStageOptions (potentially combining it with values and other variables set in the YAML). I understand that there is still the corner case of someone defining `releasever` in the YAML as well as setting the `bool` option, but that could be an error.

pkg/distro/image_config_test.go
```diff
@@ -166,3 +166,71 @@ func TestImageConfigInheritFrom(t *testing.T) {
...
+		DNFConfig: &DNFConfig{
+			SetReleaseVerVar: true,
+		},
+	}
+	assert.Equal(t, []*osbuild.DNFConfigStageOptions{
+		{
+			Variables: []osbuild.DNFVariable{
+				{
+					Name:  "releasever",
+					Value: "centos-9",
```
thozza, line 187: Nitpick: I know that these are just unit tests, but the `releasever` is ever set only to `X.Y` in RHEL images (and for centos it is `X-stream`). This mechanism allows you to pin the host repositories to a specific Extended Update Support (EUS) version of RHEL. Using `centos-9` may give one the wrong impression of its purpose.

pkg/distro/image_config_test.go
```diff
@@ -166,3 +166,71 @@ func TestImageConfigInheritFrom(t *testing.T) {
 		})
 	}
 }
+
+func TestImageConfiDNCConfigNone(t *testing.T) {
```
ondrejbudai, line 170:
```suggestion
func TestImageConfigDNFConfigNone(t *testing.T) {
```
AI found a double-typo here. :)

mvo5:
Closing in favor of https://github.com/osbuild/images/pull/1465

pkg/distro/image_config_test.go
mvo5: Thanks, fixed in the other PR

pkg/distro/image_config.go
mvo5: Thanks, I opened #1465 with this alternative approach.

PR #1465, opened 2025-04-25
distro: add DNFSetReleaseVerVar to ImageConfig

mvo5:
This commit adds a new `DNFSetReleaseVerVar` to
`ImageConfig` and a new `DNFConfigOptions() helper.

The rational is that `DNFConfigStageOptions` current gets the `osVersion` passed in directly. When we move to a YAML based ImageConfig this will be problematic as we do not have access to variables like this. By making this new declarative option we sidestep the problem.

This option is used in the rhel sap images.

Its slightly unfortunate that we need to change two places (images.go in rhel/fedora) to use the new
helper (pr#1387 would fix this) but with more YAML work eventually this should converge.

Thanks to thozza for suggesting this simplified approach.

JIRA: HMS-5284

### DIFF
--- a/pkg/distro/image_config.go
+++ b/pkg/distro/image_config.go
@@ -44,20 +44,22 @@ type ImageConfig struct {
 
 	// for RHSM configuration, we need to potentially distinguish the case
 	// when the user want the image to be subscribed on first boot and when not
-	RHSMConfig          map[subscription.RHSMStatus]*subscription.RHSMConfig
-	SystemdLogind       []*osbuild.SystemdLogindStageOptions
-	CloudInit           []*osbuild.CloudInitStageOptions
-	Modprobe            []*osbuild.ModprobeStageOptions
-	DracutConf          []*osbuild.DracutConfStageOptions
-	SystemdDropin       []*osbuild.SystemdUnitStageOptions
-	SystemdUnit         []*osbuild.SystemdUnitCreateStageOptions
-	Authselect          *osbuild.AuthselectStageOptions
-	SELinuxConfig       *osbuild.SELinuxConfigStageOptions
-	Tuned               *osbuild.TunedStageOptions
-	Tmpfilesd           []*osbuild.TmpfilesdStageOptions
-	PamLimitsConf       []*osbuild.PamLimitsConfStageOptions
-	Sysctld             []*osbuild.SysctldStageOptions
+	RHSMConfig    map[subscription.RHSMStatus]*subscription.RHSMConfig
+	SystemdLogind []*osbuild.SystemdLogindStageOptions
+	CloudInit     []*osbuild.CloudInitStageOptions
+	Modprobe      []*osbuild.ModprobeStageOptions
+	DracutConf    []*osbuild.DracutConfStageOptions
+	SystemdDropin []*osbuild.SystemdUnitStageOptions
+	SystemdUnit   []*osbuild.SystemdUnitCreateStageOptions
+	Authselect    *osbuild.AuthselectStageOptions
+	SELinuxConfig *osbuild.SELinuxConfigStageOptions
+	Tuned         *osbuild.TunedStageOptions
+	Tmpfilesd     []*osbuild.TmpfilesdStageOptions
+	PamLimitsConf []*osbuild.PamLimitsConfStageOptions
+	Sysctld       []*osbuild.SysctldStageOptions
+	// Do not use DNFConfig directly, call "DNFConfigOptions()"
 	DNFConfig           []*osbuild.DNFConfigStageOptions
+	DNFSetReleaseVerVar *bool
 	SshdConfig          *osbuild.SshdConfigStageOptions
 	Authconfig          *osbuild.AuthconfigStageOptions
 	PwQuality           *osbuild.PwqualityConfStageOptions
@@ -144,6 +146,35 @@ func (c *ImageConfig) InheritFrom(parentConfig *ImageConfig) *ImageConfig {
 		}
 	}
 	return &finalConfig
+}
+
+func (c *ImageConfig) DNFConfigOptions(osVersion string) []*osbuild.DNFConfigStageOptions {
+	if c.DNFSetReleaseVerVar == nil || !*c.DNFSetReleaseVerVar {
+		return c.DNFConfig
+	}
+
+	// We currently have no use-case where we set both a custom
+	// DNFConfig and DNFSetReleaseVerVar. If we have one this needs
+	// to change and we need to decide if we want two dnf
+	// configurations or if we want to merge the variable into all
+	// existing once (exactly once) and we need to consider what to
+	// do about potentially conflicting (manually set) "releasever"
+	// values by the user.
+	if c.DNFConfig != nil {
+		err := fmt.Errorf("internal error: currently DNFConfig and DNFSetReleaseVerVar cannot be used together, please reporting this as a feature request")
+		panic(err)
+	}
+	return []*osbuild.DNFConfigStageOptions{
+		osbuild.NewDNFConfigStageOptions(
+			[]osbuild.DNFVariable{
+				{
+					Name:  "releasever",
+					Value: osVersion,
+				},
+			},
+			nil,
+		),
+	}
 }
 
 func (c *ImageConfig) WSLConfStageOptions() *osbuild.WSLConfStageOptions {
--- a/pkg/distro/fedora/images.go
+++ b/pkg/distro/fedora/images.go
@@ -227,7 +227,7 @@ func osCustomizations(
 	osc.Tmpfilesd = imageConfig.Tmpfilesd
 	osc.PamLimitsConf = imageConfig.PamLimitsConf
 	osc.Sysctld = imageConfig.Sysctld
-	osc.DNFConfig = imageConfig.DNFConfig
+	osc.DNFConfig = imageConfig.DNFConfigOptions(t.arch.distro.osVersion)
 	osc.SshdConfig = imageConfig.SshdConfig
 	osc.AuthConfig = imageConfig.Authconfig
 	osc.PwQuality = imageConfig.PwQuality
--- a/pkg/distro/image_config_test.go
+++ b/pkg/distro/image_config_test.go
@@ -166,3 +166,32 @@ func TestImageConfigInheritFrom(t *testing.T) {
 		})
 	}
 }
+
+func TestImageConfigDNFSetReleaseVerNotSet(t *testing.T) {
+	var expected []*osbuild.DNFConfigStageOptions
+	cnf := &ImageConfig{}
+	assert.Equal(t, expected, cnf.DNFConfigOptions("9-stream"))
+
+	cnf.DNFSetReleaseVerVar = common.ToPtr(false)
+	assert.Equal(t, expected, cnf.DNFConfigOptions("9-stream"))
+}
+
+func TestImageConfigDNFConfigOptionsPreExisting(t *testing.T) {
+	cnf := &ImageConfig{
+		DNFConfig: []*osbuild.DNFConfigStageOptions{
+			{
+				Config: &osbuild.DNFConfig{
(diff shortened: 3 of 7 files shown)


achilleas-k:
> When we move to a YAML based ImageConfig this will be problematic as we do not have access to variables like this.

Well, now you got me thinking about having pre-defined variables in our image definition yaml files.  But no, let's not go down that path just yet please :)

achilleas-k (approved):
LGTM, thanks!

pkg/distro/image_config.go
```diff
@@ -44,20 +44,22 @@ type ImageConfig struct {
...
+	SystemdUnit   []*osbuild.SystemdUnitCreateStageOptions
+	Authselect    *osbuild.AuthselectStageOptions
+	SELinuxConfig *osbuild.SELinuxConfigStageOptions
+	Tuned         *osbuild.TunedStageOptions
+	Tmpfilesd     []*osbuild.TmpfilesdStageOptions
+	PamLimitsConf []*osbuild.PamLimitsConfStageOptions
+	Sysctld       []*osbuild.SysctldStageOptions
+	// Do not use DNFConfig directly, call "DNFConfigOptions()"
 	DNFConfig           []*osbuild.DNFConfigStageOptions
+	DNFSetReleaseVerVar *bool
```
thozza, line 62: Perhaps we could make these non-exported and export only `DNFConfigOptions()`. This, however, would require a custom Unmarshaler.

pkg/distro/image_config.go
```diff
@@ -146,6 +148,22 @@ func (c *ImageConfig) InheritFrom(parentConfig *ImageConfig) *ImageConfig {
...
+	return append(c.DNFConfig, osbuild.NewDNFConfigStageOptions(
+		[]osbuild.DNFVariable{
+			{
+				Name:  "releasever",
+				Value: osVersion,
+			},
+		},
+		nil,
+	))
+}
```
thozza, line 165: I have two comments:

1. The result of this change is that two DNF config stages may be added to the manifest, instead of a single one. Probably not a big deal, but it is a behavior change.
2. I'm not sure how pedantic we should be, because if two DNF config stages are added to the manifest, the one not created from the newly introduced `bool` option can also define the `releasever` variable. Then we will end up with one overriding the other. Would it be worth it to handle such a situation?

thozza (commented):
Thank you for this new version. I like it a lot more. I've added two comments. The question about not exporting the DNF config-related options in `ImageConfig` is probably something for the future, but the other comment is something that I'd like to resolve before we merge this PR.

pkg/distro/image_config.go
mvo5: Right, the old behavior is to just override any existing DNFConfig and because we  do not do a deep merge inheriting also resulting silent overriding.

If we want to be as close as possible to the original behavior we could do:
```go
diff --git a/pkg/distro/image_config.go b/pkg/distro/image_config.go
index 39023874c..6fc0c46c7 100644
--- a/pkg/distro/image_config.go
+++ b/pkg/distro/image_config.go
@@ -153,15 +153,22 @@ func (c *ImageConfig) DNFConfigOptions(osVersi
on string) []*osbuild.DNFConfigSta
 		return c.DNFConfig
 	}
 
-	return append(c.DNFConfig, osbuild.NewDNFConfigStageOptions(
-		[]osbuild.DNFVariable{
-			{
-				Name:  "releasever",
-				Value: osVersion,
+	// YAGNI: support both DNFConfig and DNFSetReleaseVerVar onc
e we
+	// have a use-case
+	if c.DNFConfig != nil {
+		panic("internal error: DNFConfig and DNFSetReleaseVe
rVar cannot be used together")
+	}
+	return []*osbuild.DNFConfigStageOptions{
+		osbuild.NewDNFConfigStageOptions(
+			[]osbuild.DNFVariable{
+				{
+					Name:  "releasever",
+					Value: osVersion,
+				},
 			},
-		},
-		nil,
-	))
+			nil,
+		),
+	}
 }
 
 func (c *ImageConfig) WSLConfStageOptions() *osbuild.WSLConfStageOp
tions {
diff --git a/pkg/distro/image_config_test.go b/pkg/distro/image_conf
ig_test.go
index ad63570c6..a439df96f 100644
--- a/pkg/distro/image_config_test.go
+++ b/pkg/distro/image_config_test.go
@@ -190,14 +190,8 @@ func TestImageConfigDNFConfigOptionsPreExisting
(t *testing.T) {
 	}
 	assert.Equal(t, cnf.DNFConfig, cnf.DNFConfigOptions("9-strea
m"))
 
-	expected := append(cnf.DNFConfig, &osbuild.DNFConfigStageOpt
ions{
-		Variables: []osbuild.DNFVariable{
-			{
-				Name:  "releasever",
-				Value: "9-stream",
-			},
-		},
-	})
 	cnf.DNFSetReleaseVerVar = common.ToPtr(true)
-	assert.Equal(t, expected, cnf.DNFConfigOptions("9-stream"))
+	assert.Panics(t, func() {
+		cnf.DNFConfigOptions("9-stream")
+	})
 }
```
(the upside is that it becomes very clear when we need to handle this case which is not a bad property).

As for (2) - that is a good question. My strawman would be that we just leave it to the user for now, if they want to manually create a releasever that conflicts with the one we automatically set they can keep the pieces. If mean, if we expose all of the DNFConfig options they have the freedom to break things, if we don't want this, we probably need an approach close to the original PR that restricts the settings. Maybe we could add a comment to the code saying something like `// users could create DNFConfig stages with conflicting releasever values maually, for now we ignore this failure scenario as there are more things a user can do that might break the DNFConfig stages` (or something along these lines).

thozza: > Right, the old behavior is to just override any existing DNFConfig and because we do not do a deep merge inheriting also resulting silent overriding.

I'm not sure what do you mean by silent overriding. `ImageConfig` can only inherit unset values from another `ImageConfig` instance. The behavior is very much intentional, because doing a deep merge would become a nightmare to get right for any reasonably complex stage options (especially if we want to support "unseting" values).

Nevertheless, you brought up an important case that I didn't consider before - `ImageConfig` inheritance.

The desire is for `ImageConfig` on lower level (i.e. distro-level vs. image-type-level) not to inherit values that are set on the lower level. However, by splitting the DNF config into two `ImageConfig` properties, one needs to set both values in the lower-level `ImageConfig`. I'm wondering if it wouldn't be better to nest all "DNF config options" (meaning the stage options + the new bool option) under a common `struct`. In this particular case, the use case for it is probably not that prominent, but we should keep this in mind to keep the `ImageConfig` inheritance model sane at least for the time being.

> 
> If we want to be as close as possible to the original behavior we could do:
> 
> ```go
> diff --git a/pkg/distro/image_config.go b/pkg/distro/image_config.go
> index 39023874c..6fc0c46c7 100644
> --- a/pkg/distro/image_config.go
> +++ b/pkg/distro/image_config.go
> @@ -153,15 +153,22 @@ func (c *ImageConfig) DNFConfigOptions(osVersi
> on string) []*osbuild.DNFConfigSta
>  		return c.DNFConfig
>  	}
>  
> -	return append(c.DNFConfig, osbuild.NewDNFConfigStageOptions(
> -		[]osbuild.DNFVariable{
> -			{
> -				Name:  "releasever",
> -				Value: osVersion,
> +	// YAGNI: support both DNFConfig and DNFSetReleaseVerVar onc
> e we
> +	// have a use-case
> +	if c.DNFConfig != nil {
> +		panic("internal error: DNFConfig and DNFSetReleaseVe
> rVar cannot be used together")
> +	}
> +	return []*osbuild.DNFConfigStageOptions{
> +		osbuild.NewDNFConfigStageOptions(
> +			[]osbuild.DNFVariable{
> +				{
> +					Name:  "releasever",
> +					Value: osVersion,
> +				},
>  			},
> -		},
> -		nil,
> -	))
> +			nil,
> +		),
> +	}
>  }
>  
>  func (c *ImageConfig) WSLConfStageOptions() *osbuild.WSLConfStageOp
> tions {
> diff --git a/pkg/distro/image_config_test.go b/pkg/distro/image_conf
> ig_test.go
> index ad63570c6..a439df96f 100644
> --- a/pkg/distro/image_config_test.go
> +++ b/pkg/distro/image_config_test.go
> @@ -190,14 +190,8 @@ func TestImageConfigDNFConfigOptionsPreExisting
> (t *testing.T) {
>  	}
>  	assert.Equal(t, cnf.DNFConfig, cnf.DNFConfigOptions("9-strea
> m"))
>  
> -	expected := append(cnf.DNFConfig, &osbuild.DNFConfigStageOpt
> ions{
> -		Variables: []osbuild.DNFVariable{
> -			{
> -				Name:  "releasever",
> -				Value: "9-stream",
> -			},
> -		},
> -	})
>  	cnf.DNFSetReleaseVerVar = common.ToPtr(true)
> -	assert.Equal(t, expected, cnf.DNFConfigOptions("9-stream"))
> +	assert.Panics(t, func() {
> +		cnf.DNFConfigOptions("9-stream")
> +	})
>  }
> ```
> 
> (the upside is that it becomes very clear when we need to handle this case which is not a bad property).
> 
> As for (2) - that is a good question. My strawman would be that we just leave it to the user for now, if they want to manually create a releasever that conflicts with the one we automatically set they can keep the pieces. If mean, if we expose all of the DNFConfig options they have the freedom to break things, if we don't want this, we probably need an approach close to the original PR that restricts the settings. Maybe we could add a comment to the code saying something like `// users could create DNFConfig stages with conflicting releasever values maually, for now we ignore this failure scenario as there are more things a user can do that might break the DNFConfig stages` (or something along these lines).

For now, the `user` in this context is someone defining an image type or a distribution. If we ever expose the defining DNF configuration as a BP customization, it should override any settings set in the `ImageConfig`.

That said, I'm mostly worried about our image definitions silently overriding configuration values in the resulting image. So from that point of view, we should ideally catch such issues at compile time or at startup (when reading the definition from the YAML). Your suggestion is OK from my PoV. An alternative would be to check for the existence of the `releasever` variable in the config when `DNFSetReleaseVerVar` is set. I'm fine with deferring this to the time when we will need to set both, the `DNFSetReleaseVerVar` value as well as some other DNF config values.

mvo5: > > Right, the old behavior is to just override any existing DNFConfig and because we do not do a deep merge inheriting also resulting silent overriding.
> 
> I'm not sure what do you mean by silent overriding. `ImageConfig` can only inherit unset values from another `ImageConfig` instance. The behavior is very much intentional, because doing a deep merge would become a nightmare to get right for any reasonably complex stage options (especially if we want to support "unseting" values).
[..]

Sorry, I agree with the exiting merging behavior of imageConfig. I just wanted to mention that it can be surprising in this particular context: if we would create a `gce-sap` image today just lke we have azure-sap and ec2-sap we would end up with something like:
[gce-sap.diff.txt](https://github.com/user-attachments/files/19941368/gce-sap.diff.txt), the relevant part:
```go
+func mkGCESapImageType(osVersion string) *rhel.ImageType {
...
+       it.DefaultImageConfig = sapImageConfig(osVersion).InheritFrom(baseGCEImageConfig())
...
```
but that would not include the `baseGCEImageConfig.DNFConfig` that sets `IPResolve: 4` as the sap image config overrides this (silently) with the DNFConfig to set the releasever. (just did this diff to illustrate what I meant with the inheriting/silent override comment).

mvo5: @thozza Fwiw, I updated the code so that it errors when both a manual DNFConfig is set and DNFSetReleaseVerVar is requested.

pkg/distro/image_config.go
mvo5: Thanks for the suggestion, I'm inclined to do this in a followup.

pkg/distro/image_config.go
thozza: Thanks. 

Can you please group the two DNF config options in a single struct in a follow-up?

thozza: Yes, exactly like this. Thank you.

PR #1469, opened 2025-04-28
distro/rhel10/ami: unset authselect in ImageConfig (COMPOSER-2517)

achilleas-k:
Unset the authselect profile in the ImageConfig for all AWS images. This will set the authselect profile to 'local', which is the package default, and should be used in all base images instead of the current 'sssd'.

COMPOSER-2517


mvo5 (approved):

achilleas-k:
Rebased and fixed conflict.

bcl (approved):
Ack

mvo5 (approved):

PR #1470, opened 2025-04-28
distro: move `kernelOptions` into `ImageConfig`

mvo5:
[draft until https://github.com/osbuild/images/pull/1472 is merged to not cause conflicts there]

This commit moves the `imageType.kernelOptions` option into `distro.ImageConfig`. The rational is that when we move to YAML we will need to be able to override the kernel options based on various conditions.

Here is an example from mkMinimalRawImgType():
```go
if common.VersionGreaterThanOrEqual(d.osVersion, "43") {
	it.kernelOptions = []string{"rw"}
}
```
The conditions will be implemented on the imageConfig but having conditions when defining image types would be a bit stange. The easiest fix is to move kernelOptions into ImageConfig. Fwiw, I think conceptually it also makes sense there so this is probably a worthwhile chage on its own but it will make the YAML work in
https://github.com/osbuild/images/pull/1462
easier.

### DIFF
--- a/pkg/distro/rhel/rhel10/distro_test.go
+++ b/pkg/distro/rhel/rhel10/distro_test.go
@@ -405,7 +405,10 @@ func TestRhel10_KernelOption_NoIfnames(t *testing.T) {
 			for _, imgTypeName := range arch.ListImageTypes() {
 				imgType, err := arch.GetImageType(imgTypeName)
 				assert.NoError(t, err)
-				assert.NotContains(t, imgType.(*rhel.ImageType).KernelOptions, "net.ifnames=0", "type %s contains unwanted net.ifnames=0", imgType.Name())
+				imgCfg := imgType.(*rhel.ImageType).DefaultImageConfig
+				if imgCfg != nil {
+					assert.NotContains(t, imgCfg.KernelOptions, "net.ifnames=0", "type %s contains unwanted net.ifnames=0", imgType.Name())
+				}
 			}
 		}
 	}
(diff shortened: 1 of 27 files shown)


pkg/distro/rhel/rhel10/distro_test.go
```diff
@@ -405,7 +405,7 @@ func TestRhel10_KernelOption_NoIfnames(t *testing.T) {
 			for _, imgTypeName := range arch.ListImageTypes() {
 				imgType, err := arch.GetImageType(imgTypeName)
 				assert.NoError(t, err)
-				assert.NotContains(t, imgType.(*rhel.ImageType).KernelOptions, "net.ifnames=0", "type %s contains unwanted net.ifnames=0", imgType.Name())
+				assert.NotContains(t, imgType.(*rhel.ImageType).DefaultImageConfig.KernelOptions, "net.ifnames=0", "type %s contains unwanted net.ifnames=0", imgType.Name())
```
thozza, line 408: I think that the issue here is that `DefaultImageConfig` does not need to be set. You should probably call `imgType.(*rhel.ImageType).getDefaultImageConfig()`. However, note that this file uses a different package name, so calling the unexported method won't work.

thozza (commented):
Unit tests are failing due to one issue. The rest looks good to me.

thozza (approved):

mvo5:
Achilleas asked me to set it to blocked until https://github.com/osbuild/images/pull/1472 is mered.

achilleas-k (dismissed):
Nice and simple.  LGTM!
Thanks for blocking it for the other PR.  It'll probably create conflicts, so we'll need to re-approve, but I'm approving the current state anyway.

achilleas-k (approved):

PR #1471, opened 2025-04-28
test/base-host-check.sh: fix disabled firewall services check

croissanne:
`firewall-cmd` returns 1 if the service is disabled, so just rely on that instead of its stdout.


achilleas-k (approved):
Thanks for the quick fix.

achilleas-k:
Maybe we should check if the exit code is actually exactly 1 instead of just non-zero.
The manpage says
```
       [--permanent] [--zone=zone] [--permanent] [--policy=policy] --query-service=service
           Return whether service has been added. Returns 0 if true, 1 otherwise.
```
but there are a bunch of other exit codes (2 - 254) that report other types of failures that we would probably want to catch.

thozza (dismissed):
LGTM

croissanne:
> Maybe we should check if the exit code is actually exactly 1 instead of just non-zero. The manpage says
> 
> ```
>        [--permanent] [--zone=zone] [--permanent] [--policy=policy] --query-service=service
>            Return whether service has been added. Returns 0 if true, 1 otherwise.
> ```
> 
> but there are a bunch of other exit codes (2 - 254) that report other types of failures that we would probably want to catch (and fail).

good point

thozza (approved):

achilleas-k (approved):

PR #1477, opened 2025-04-29
many: add custom unmarshalers for osbuild, platform

mvo5:
This PR adds custom unmarshalers to various parts of osbuild and platform. We will need to be able to unmarshal osbuild options that are used in the ImageConfig. And we will need to be able to set platform information for image types in YAML too so the platform.ImageFormat also needs to be handed.

Split out from https://github.com/osbuild/images/pull/1462

### DIFF
--- a/internal/common/unmarshal.go
+++ b/internal/common/unmarshal.go
@@ -1,0 +1,24 @@
+package common
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// UnmarshalYAMLviaJSON unmarshals via the JSON interface, this avoids code
+// duplication on the expense of slightly uglier errors
+func UnmarshalYAMLviaJSON(u json.Unmarshaler, unmarshal func(any) error) error {
+	var data any
+	if err := unmarshal(&data); err != nil {
+		return fmt.Errorf("cannot unmarshal to any: %w", err)
+	}
+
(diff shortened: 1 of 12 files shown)


mvo5:
> LGTM.

Thank you! 
 
> What's the long term plan? Will we need to have basically the same custom YAML unmarshaller on every osbuild stage options type that we have in the ImageConfig? Is there any way this can be deduplicated? (I don't see an obvious way, but I haven't thought about it much).

Yes, we will need YAML supportin all osbuild stages that are used. In most cases its just adding a matching `yaml:"tag"` but when custom unmarshaling is needed (like in the cases of the PR) this needs to be added. So far its not too many but it is definitely a bit of work.

The alternative would be to build an abstraction in ImageConfig that does not directly expose an osbuild stage and is then converted (a bit like what we did with https://github.com/osbuild/images/pull/1350) and in those we can optimize for simplicity. It can be a mix as well of course, i.e. we just convert some of them. 

But honestly I think its not too bad, its a bit of busywork to add these unmarshalers.

thozza (approved):
Thanks, LGTM.

FWIW, I think that unmarshalling stage options to JSON and YAML in the same way is better for consistency and removes any additional knowledge when writing the YAML when one already knows about osbuild stage options. Nevertheless, I do agree that it is a busy work.

PR #1478, opened 2025-04-29
github: don't run manifest checksum validation on the merge queue

achilleas-k:
On the merge queue, the validate checksum fails because we use the base branch to start checking.  We could check if the checksums changed from the rebase, but that wouldn't be useful since the check isn't required anyway.

For now, let's make the check run only on PRs.  This might result in inconsistencies in main, but we're still experimenting with this test anyway, so it'll be interesting to see how often that happens and how confusing it might be.


schuellerf (approved):
LGTM

bcl (approved):
Ack!

PR #1479, opened 2025-04-29
fsnode: add JSON/YAML support for File/Directory

mvo5:
More unmarshal support in preparation for the  ImageConfig work in https://github.com/osbuild/images/pull/1462 - split to make reviews easier, the fsnode ones are a bit more tricky.

--- 

fsnode: validate unknown keys in fsnode.{Dir,File}

When unmarshaling fsnode.{Dir,File}, also validate that no
extra keys are used.

---

fsnode: add JSON/YAML support for File/Directory

---

fsnode: use `path/filepath` instead of `path`

When validating paths in the fsnode module, use the `filepath`
module instead of the (more generic) `path`.

### DIFF
--- a/pkg/customizations/fsnode/dir.go
+++ b/pkg/customizations/fsnode/dir.go
@@ -1,6 +1,13 @@
 package fsnode
 
-import "os"
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/osbuild/images/internal/common"
+)
 
 type Directory struct {
 	baseFsNode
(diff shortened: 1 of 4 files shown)


achilleas-k (approved):
Minor comment/question.  Otherwise, LGTM.

lzap (changes requested):
Okay, this is not nice but it is what it is. I am not too familiar with this code yet.

But the helper function might be improved, it almost looks like you want it to be generic so a good contract and maybe separating it into its own file (with a test) might be a good option?

mvo5:
Thanks for the feedback - I agree that the map[string]any unmarshaller is not nice (it was quick and expedient as its really just a distraction from my main goal to get the image config into YAML). But I agree we should not compromise on quality!

I opened https://github.com/osbuild/images/pull/1500 with an alternative approach that use more buld-in go featurs and will close this one (at least for now).

PR #1481, opened 2025-04-30
refactor: replace logrus with stdlib log package

lzap:
This PR replaces outdated `logrus` with just `log` from standard library which is super easy to work with. But since `images` is a library it is good manners to allow to configure logging as needed for users (e.g. composer or CLI) a package called `olog` for all logging is created. It is super simple - it only holds the main logger variable, but it also contains a proxy functions for all functions so instead of:

    mylog.Logger.Printf("test")

one can just write:

    olog.Printf("test")

The logger itself can be passed to other libraries if needed either by itself (`olog.Default()` this is used in the patch itself) or by `Writer` interface (`olog.Default().Writer()`).

The same approach is considered for `image-builder-cli`: https://github.com/osbuild/image-builder-cli/pull/214 where "o" in `olog` stands for "option", in this case "o" stands more for "osbuild" because all logging is done in there and `images` library itself does not care too much if the logs are enabled or disabled. Well except one test.


cmd/osbuild-upload-gcp/main.go
```diff
@@ -57,59 +57,59 @@ func main() {
 	case "rhel-9":
 		guestOSFeatures = gcp.GuestOsFeaturesRHEL9
 	default:
-		logrus.Fatalf("[GCP] Unknown OS Family %q. Use one of: 'rhel-8', 'rhel-9'.", osFamily)
+		olog.Fatalf("[GCP] Unknown OS Family %q. Use one of: 'rhel-8', 'rhel-9'.", osFamily)
```
lzap, line 60: I wanted to keep things simple but it is super easy to create a prefix support:

```go
olog.WithPrefix("GCP").Fatalf("...")
```

Maybe in a followup, or perhaps it is too much. The simplicity of this API is its best strength I think.

pkg/upload/koji/koji.go
```diff
@@ -514,15 +514,15 @@ func customCheckRetry(ctx context.Context, resp *http.Response, err error) (bool
...
+		olog.Println("Not retrying: ", resp.Status)
 	}
 
 	return shouldRetry, retErr
 }
 
 func createCustomRetryableClient() *rh.Client {
 	client := rh.NewClient()
-	client.Logger = rh.LeveledLogger(&LeveledLogrus{logrus.StandardLogger()})
+	client.Logger = olog.Default()
```
lzap, line 525: The adapter is no longer needed, this library accepts standard logger directly.

cmd/osbuild-upload-gcp/main.go
achilleas-k: That's not a bad idea.  Let's consider it for a follow-up.

pkg/upload/koji/rh-logrus-adapter.go
```diff
@@ -1,42 +0,0 @@
...
-func (l *LeveledLogrus) Info(msg string, keysAndValues ...interface{}) {
-	l.WithFields(fields(keysAndValues...)).Info(msg)
-}
-func (l *LeveledLogrus) Debug(msg string, keysAndValues ...interface{}) {
-	if strings.Contains(msg, monitoringKeyword) {
-		l.WithFields(fields(keysAndValues...)).Info(msg)
-	} else {
-		l.WithFields(fields(keysAndValues...)).Debug(msg)
-	}
-}
```
achilleas-k, line 38: Tracing the history of this package and function brought me to this issue: https://issues.redhat.com/browse/COMPOSER-1394

We should probably keep this in mind in case the issue resurfaces now that we're simplifying the logger.  If it does, we should handle it in osbuild-composer, not here.

achilleas-k (dismissed):
Thank you!  I really like this.

Small note about the koji logging adapter, something to keep in mind for later.

Also, we might want to initialise and use a simple default logger in the cmd binaries we use for testing (`cmd/build`, `cmd/gen-manifests`), both as a way to see how the logger behaves and for catching errors in testing.

pkg/upload/koji/rh-logrus-adapter.go
lzap: We are not taking those lines out, they will be present. And since images library does not support passing context/trace_id we are not loosing anything. The caller (composer) should make sure it sets up the `olog` package writer properly and redirect logs to its (`log/slog`) logger.

Tho if we want logs to be correlated by trace_id from `slog`, that cannot be done without passing context. One idea is to copy the `slog` API by creating additional functions in the package: `PrintfContext(ctx, ...)` so external libraries could add a context hook and extract correlation id from there.

I will work on this only when I get here, there is bunch of work in composer ahead of me.

lzap:
> Also, we might want to initialise and use a simple default logger in the cmd binaries we use for testing (`cmd/build`, `cmd/gen-manifests`), both as a way to see how the logger behaves and for catching errors in testing.

Ah this is a problem, you cannot reliably enable a global variable in tests other than creating an `init` function for every single test file. That is because Go test tool compiles only those tests which you want to execute. So the only option is to set the default to the standard logger.

I am a not huge fan of writing any logs during tests, there is a `t.Log` and `t.Logf` that is only shown when a test fails. That is how logging should be done ideally. Also checking a code was executed via storing log lines into a buffer and testing against that is another story - API should be broken up properly to test it.

Anyways, if discard logging by default is what we want, this PR can be merged. Other than that, I will rebase and set to standard logger by default. I will rebase later in that case.

pkg/upload/koji/rh-logrus-adapter.go
achilleas-k: > I will work on this only when I get here, there is bunch of work in composer ahead of me.

Yup.  Agreed.

achilleas-k:
> Ah this is a problem, you cannot reliably enable a global variable in tests

I didn't mean unit tests.  I meant for the build tests in gitlab, which use `cmd/gen-manifests` (e.g. [1]) and `cmd/build` [2].  Those are run in the gitlab pipelines to generate manifests and build images, so initialising a simple logger for those cmds would be useful for troubleshooting.

[1] https://github.com/osbuild/images/blob/bc166716596c334e84149fcf8404fdfcac988fb5/.gitlab-ci.yml#L3038
[2] https://github.com/osbuild/images/blob/bc166716596c334e84149fcf8404fdfcac988fb5/test/scripts/build-image#L29-L33

lzap:
> I didn't mean unit tests. I meant for the build tests in gitlab, which use `cmd/gen-manifests` (e.g. [1]) and `cmd/build` [2]. Those are run in the gitlab pipelines to generate manifests and build images, so initialising a simple logger for those cmds would be useful for troubleshooting.

Oh, Go by default has the `log` package initialized to print everything to standard error. Is that not enough? Or do you want to suppress the messages?

pkg/olog/proxy.go
```diff
@@ -0,0 +1,83 @@
+package olog
+
+import "io"
+
+// Print is a proxy for log.Print.
+func Print(v ...any) {
+	Default().Print(v...)
+}
+
+// Printf is a proxy for log.Printf.
```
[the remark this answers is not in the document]
lzap: My reasoning is that you can pass the value to any interface. Since Go implicitly implements interfaces, many projects define their own copies. Some create an interface that only has `Printf` while others do have an interface that requires all the methods to be present. This will fulfill them all.

pkg/distro/defs/loader.go
```diff
@@ -304,7 +304,7 @@ func load(distroNameVer string) (*toplevelYAML, error) {
 	// searchPaths down the stack instead
 	var dataFS fs.FS = DataFS
 	if overrideDir := experimentalflags.String("yamldir"); overrideDir != "" {
-		logrus.Warnf("using experimental override dir %q", overrideDir)
+		olog.Printf("using experimental override dir %q", overrideDir)
```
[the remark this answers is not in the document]
lzap: Sure. Adding.

lzap:
Added `WARNING` prefixes, rebased.

mvo5 (approved):
Thank you

supakeen (approved):

achilleas-k:
> > I didn't mean unit tests. I meant for the build tests in gitlab, which use `cmd/gen-manifests` (e.g. [1]) and `cmd/build` [2]. Those are run in the gitlab pipelines to generate manifests and build images, so initialising a simple logger for those cmds would be useful for troubleshooting.
> 
> Oh, Go by default has the `log` package initialized to print everything to standard error. Is that not enough? Or do you want to suppress the messages?

ah no, that's all good.  Thanks!

PR #1482, opened 2025-04-30
repositories: Set 6h expire on main repo, use default for updates

bcl:
The default timeout for osbuild-depsolve-dnf is 20s so previously this was checking the main release repository frequently, and not checking the updates repo for 6 hours. The updates repo has a higher chance of changing so it should be checked more often, and the main repo generally isn't changed after release so 6h (or longer) is fine.


Conan-Kudo (approved):

achilleas-k (approved):
LGTM!

PR #1483, opened 2025-04-30
test: Run more distro tests in parallel

bcl:
Run TestImageTypePipelineNames and
TestOSTreeOptionsErrorForNonOSTreeImgTypes in parallel. Saves about 20s on my local machine.


bcl:
FWIW I also tried adding Parallel to the fedora and rhel distro_test.go moduled, but it didn't improve the runtime.

bcl:
Doing this seems to have revealed a problem with `OSCustomizations.Grub2Config` being reused. I think I've fixed it, but the massive use of pointers in this codebase leaves me feeling queasy. There may be more.

mvo5 (approved):
Thank you! This is nice!

supakeen (approved):

PR #1489, opened 2025-05-05
osbuild: add JSON/YAML unmarshal to UdevRulesStageOptions

mvo5:
When we move `ImageConfig` into YAML we will need support for the `UdevRulesStageOptions`.

This is split out from https://github.com/osbuild/images/pull/1462

### DIFF
--- a/pkg/osbuild/udev_rules_stage.go
+++ b/pkg/osbuild/udev_rules_stage.go
@@ -1,8 +1,11 @@
 package osbuild
 
 import (
+	"encoding/json"
 	"fmt"
 	"regexp"
+
+	"github.com/osbuild/images/internal/common"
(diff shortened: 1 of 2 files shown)


achilleas-k (commented):
The unmarshaller is a bit hard to read (understandably, since this stage is a bit too flexible).  Does it cover all forms of the stage options, or is it limited to only the ones we use in image config?

If I'm reading it right, it doesn't support the [UdevRuleKeyArg](https://github.com/osbuild/images/blob/5436fb4811230f5194a3deade6ef122ead8a89ae/pkg/osbuild/udev_rules_stage.go#L218-L221) form, does it?

supakeen (approved):

PR #1490, opened 2025-05-05
[RFC] customizations: drop `FsNode` interface as it is not used

mvo5:
This commit removes the unused `customizations.FsNode` interface.

AFAICT this is unused and composer is also not using it (it does have an identical copy though). Opening to see if really all tests pass wihtout it.


bcl (approved):
Looks ok to me, less code, less bugs :)

PR #1497, opened 2025-05-06
fsnode: add custom unmarshalers for File/Dir

mvo5:
Alternative implementation for https://github.com/osbuild/images/pull/1479 that changes the structure of the fsnode as suggested by @achilleas-k and @lzap 

I have no super strong opinion. One nice property of the original code with the getters/setters is that after NewFile() it would always be valid. This approach with the exported fields does not protect against changes to invalid data after the creation (or by creation from hand). It also adds a bunch of (extra) churn. 

---

fsnode: add custom unmarshalers for File/Dir

This commit adds custom unmarshalers for fsnode.{File,Directory}
that also do validation. For convenience when unmarshaling a
`fsnode.File` we support an extra `text: foo` field so that
in an ImageConfig YAML we can write:
```yaml
minimal_raw: &minimal_raw:
   image_config:
     files:
       - path: "/root/anaconda-ks.cfg"
         user: "root"
         group: "root"
         text: |
           # Run initial-setup on first boot
           # Created by osbuild
           firstboot --reconfig
```
The user can also use `data:` to put a base64 encoded string
in for binary files.

---

fsnode: remove Getters, export field to make JSON friendly

This can be done in multiple ways, one is to add a custom
unmarshaller. This is done in PR#1479 but because of the
way the code was written the custom unmarshaling is rather
ugly.

This commit is an alternative version. As a starting point
the struct fields are simply exported and the getters of
the same name removed. Then we can add a custom unmarshal
that just validates.

### DIFF
--- a/pkg/customizations/fsnode/dir.go
+++ b/pkg/customizations/fsnode/dir.go
@@ -1,21 +1,25 @@
 package fsnode
 
 import (
+	"bytes"
+	"encoding/json"
 	"os"
+
+	"github.com/osbuild/images/internal/common"
 )
 
 type Directory struct {
-	Path  string
-	Mode  *os.FileMode
-	User  any
-	Group any
+	Path  string       `json:"path,omitempty"`
+	Mode  *os.FileMode `json:"mode,omitempty"`
+	User  any          `json:"user,omitempty"`
+	Group any          `json:"group,omitempty"`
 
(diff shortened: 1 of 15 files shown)


mvo5:
I also close this one (for now) in favor of https://github.com/osbuild/images/pull/1500 which is less code and (arguably) cleaner.

PR #1498, opened 2025-05-06
distro/rhel9: make /boot 1 GiB everywhere

ondrejbudai:
Users are still running into issues with too small /boot on older EL9 minor versions. Let's just bump the size retroactively, and make all EL9 images have 1 GiB /boot. I know that this is a somewhat surprising change, but the too small /boot is an extremely hard thing to fix in existing images, so I think it's justified for.


mvo5 (dismissed):
I love it! If we can change older rhel versions partition tables like this then I'm all for it, makes the definitions a lot simpler.

supakeen (dismissed):

supakeen:
Needs fixes to the testcases that expect exact sizes.

thozza (commented):
It makes sense, and LGTM, just the unit tests, need fixing, as pointed out by Simon.

achilleas-k (approved):
Thanks!!

thozza (approved):

mvo5:
Meh, sorry, there is an extra complication here. The original change does two things:
1. increase the size of /boot to 1GiB everywhere
2. add /boot to rhel-9.{0,2}

With (2) we need to change the boot mode from "BOOT_LEGACY" to "BOOT_HYBRID" for rhel-9.{0,2} and update the tests. I add the two commit for this here for context but I'm not sure we should do this, maybe we need to get back to the original commit and tweak it so that rhel-9.{0,2} are left as they are (i.e. no /boot, BOOT_LEGACY). WDYT?

[0001-distro-update-EC2-partition-tests-for-boot-size.patch.txt](https://github.com/user-attachments/files/20160559/0001-distro-update-EC2-partition-tests-for-boot-size.patch.txt)
[0002-rhel-make-rhel-9.0-hybrid-bootable-images.patch.txt](https://github.com/user-attachments/files/20160557/0002-rhel-make-rhel-9.0-hybrid-bootable-images.patch.txt)

thozza:
> With (2) we need to change the boot mode from "BOOT_LEGACY" to "BOOT_HYBRID" for rhel-9.{0,2} and update the tests. I add the two commit for this here for context but I'm not sure we should do this, maybe we need to get back to the original commit and tweak it so that rhel-9.{0,2} are left as they are (i.e. no /boot, BOOT_LEGACY). WDYT?

Sounds good to me. :+1:

mvo5:
> > With (2) we need to change the boot mode from "BOOT_LEGACY" to "BOOT_HYBRID" for rhel-9.{0,2} and update the tests. I add the two commit for this here for context but I'm not sure we should do this, maybe we need to get back to the original commit and tweak it so that rhel-9.{0,2} are left as they are (i.e. no /boot, BOOT_LEGACY). WDYT?
> 
> Sounds good to me. 👍

I updated this now as discussed above, rhel-9.{0,2} increase the /boot size still but will no longer add /boot/efi. If tests are happy we could squash merge it, I did not merge my commit into Ondrejs as that felt a bit inappropriate (and harder to review) but in the final merge a squash-merge might be fine (or we keep it, I don't really mind).

mvo5 (approved):

supakeen (approved):

PR #1499, opened 2025-05-06
disk: make auto-generated /boot 1 GiB big

ondrejbudai:
It looks like all distributions use at least 1 GiB /boot if there's such a partition. I checked this on Fedora Workstation, and RHEL. We have a similar experience: a smaller /boot than 1 GiB tends to break kernel updates, or installing kernel-debug (which is quite big).

So let's make sure that the /boot partition we auto-create in disk (if one is required, e.g. for / on LBM) is 1 GiB big.


mvo5 (dismissed):

supakeen:
Needs fixes to testcases that require specific sizes.

ondrejbudai:
rebased and hopefully fixed the tests 🤞

achilleas-k (approved):

PR #1500, opened 2025-05-06
fsnode: add JSON/YAML support for File/Directory

mvo5:
Another alternative version of https://github.com/osbuild/images/pull/1479 - I think I like this one best so far

---

---

fsnode: use `path/filepath` instead of `path`

When validating paths in the fsnode module, use the `filepath`
module instead of the (more generic) `path`.

### DIFF
--- a/pkg/customizations/fsnode/fsnode.go
+++ b/pkg/customizations/fsnode/fsnode.go
@@ -1,34 +1,42 @@
 package fsnode
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
+
+	"github.com/osbuild/images/internal/common"
 )
 
 const usernameRegex = `^[A-Za-z0-9_.][A-Za-z0-9_.-]{0,31}$`
 const groupnameRegex = `^[A-Za-z0-9_][A-Za-z0-9_-]{0,31}$`
 
+type baseFsNodeJSON struct {
+	Path  string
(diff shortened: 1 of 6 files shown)


pkg/customizations/fsnode/fsnode.go
```diff
@@ -1,58 +1,68 @@
...
+	"path/filepath"
 	"regexp"
+
+	"github.com/osbuild/images/internal/common"
 )
 
 const usernameRegex = `^[A-Za-z0-9_.][A-Za-z0-9_.-]{0,31}$`
 const groupnameRegex = `^[A-Za-z0-9_][A-Za-z0-9_-]{0,31}$`
 
+type baseFsNodeJSON struct {
```
mvo5, line 17: We could also call this `baseFsNodeData` might be a bit clearer

lzap: This was exactly my idea when I read your concern about exposing fields.

Lesson learned, for the new blueprint I want to avoid getters/setters, so I think the new blueprint struct must be immutable. There might be some unmarshal hooks, but once everything is loaded up no changes should be allowed, no getters or anything. I really hope we can achieve this, then all fields can be exported. Well except unions which are not exported and one has to use `AsXXX` "getter" but that is the only way how to use it.

lzap (approved):
I think this is miles cleaner than the reflection magic you were proposing initially.

mvo5:

PR #1503, opened 2025-05-07
osbuild: Add config to InsightsClientConfigStageOptions

rverdile:
Fixes a bug because the insights_client_config stage expects the options to be nested under a config object.


mvo5 (commented):
Thanks, this looks good and we can merge it quickly if it blocks a deploy. 

If we have a bit more time I would love to see a minimal test for the json marshaling, here is a minimal example: https://github.com/osbuild/images/blob/main/pkg/osbuild/tar_stage_test.go#L82
and maybe something similar to https://github.com/osbuild/images/blob/main/pkg/osbuild/tar_stage_test.go#L10 - but that could also be a followup by us if you are busy.

PR #1504, opened 2025-05-07
arch: add unmarshal support for arch.Arch

mvo5:
This is another split out from #1462 - we will want to define the platforms for the image types in YAML too. Because the architecture(s) are part of that platform definition we need it to be unmarshal-able. 

This is part of https://github.com/osbuild/images/compare/main...mvo5:platform_yaml?expand=1 which I will open after (and all part of the larger 1462 effort).

---

arch: add unmarshal support for arch.Arch

We will define the platform in YAML so we need a way to
unmarshal the architecture (that is part of the platform).

---

arch: add architecture to "unsupported architecture" errors

The error so far could be more helpful, so far we do:
```
unsupported architecture
```
instead we now error with the unsupported architecture as
part of the error message.
```
unsupported architecture <unsupported-arch>
```

---

arch: make arch.FromString() return an error on invalid input

The previous code was panic()ing on invalid input. This is no
longer an option as we will define platforms in YAML and here
we need to be able to handle incorrect input more gracefully
than a panic. The callers are also updated.

### DIFF
--- a/cmd/otk/osbuild-gen-partition-table/main.go
+++ b/cmd/otk/osbuild-gen-partition-table/main.go
@@ -245,7 +245,10 @@ func genPartitionTable(genPartInput *Input, rng *rand.Rand) (*Output, error) {
 	architecture := arch.ARCH_UNSET
 	if genPartInput.Properties.Type == otkdisk.PartTypeGPT {
 		// GPT partition table generation requires an architecture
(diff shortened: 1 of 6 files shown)


supakeen (approved):
Thanks. Love the error.

mvo5:

PR #1505, opened 2025-05-08
distro: make `defs.PackageSets()` return a map

mvo5:
Another requirement for #1462

---

distro: update distro YAML to new packagesSet map

This commit is very mechnical. It updates the package
sets that used to be "flat" into a map. For most image
types this just means adding "os" but for some (like
the installers) it adds the second pipeline.

---

distro: make `defs.PackageSets()` return a map

This commit changes the defs.PackageSet loader to return
the full map of all static package sets defined for the
pipelines. This is needed when we want to move all of
the image types into YAML. We currently define in code
what image types load from os, container or installer
pipelines but that does not translate to pure YAML.

Cleanup of https://github.com/osbuild/images/pull/1291


mvo5:

PR #1506, opened 2025-05-08
Add support for specifying an selinux policy in the build pipeline

alexlarsson:
We already support specifying a selinux policy in the image itself, but we need to do it for the build pipeline as well.

This will be needed for a followup commit in bootc-image-builder.

### DIFF
--- a/pkg/manifest/build.go
+++ b/pkg/manifest/build.go
@@ -43,6 +43,8 @@ type BuildrootFromPackages struct {
 	// buildroot itself when running setfiles. Once osbuild has
 	// this then this option would become "useChrootSetfiles"
 	disableSelinux bool
+
+	selinuxPolicy string
 }
 
 type BuildOptions struct {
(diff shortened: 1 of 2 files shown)


alexlarsson:
This is needed for https://github.com/osbuild/bootc-image-builder/pull/920

pkg/manifest/build.go
```diff
@@ -152,6 +159,13 @@ func (p *BuildrootFromPackages) serialize() osbuild.Pipeline {
 	return pipeline
 }
 
+func (p *BuildrootFromPackages) getSELinuxPolicy() string {
+	if p.selinuxPolicy != "" {
```
supakeen, line 163: Are you using fully custom SELinux policy packages or could this validate against the set of 'usual suspects'?

alexlarsson:
https://github.com/osbuild/images/pull/1507

PR #1507, opened 2025-05-09
BootcDiskImage: Support using a custom container for the build pipeline

alexlarsson:
The automotive project wants to build minimal bootc images which will not contain tools like dnf, mkfs.ext, etc. We support this by allowing the container used in the build pipeline to come from another (but related) container image.

Note: This depends on, and includes the commits from, https://github.com/osbuild/images/pull/1506

Question: This just changes the API, what are the policy here? Is this fine, or should I duplicate NewBootcDiskImage() to NewBootcDiskImageWithBuild() so the old calls keep working?

### DIFF
--- a/pkg/image/bootc_disk.go
+++ b/pkg/image/bootc_disk.go
@@ -53,7 +53,11 @@ func (img *BootcDiskImage) InstantiateManifestFromContainers(m *manifest.Manifes
 	runner runner.Runner,
 	rng *rand.Rand) error {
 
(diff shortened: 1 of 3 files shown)


alexlarsson:
This is used in https://github.com/osbuild/bootc-image-builder/pull/921

supakeen:
> Question: This just changes the API, what are the policy here? Is this fine, or should I duplicate NewBootcDiskImage() to NewBootcDiskImageWithBuild() so the old calls keep working?

For this specific function it is (IMHO) fine to change the API as there is only one consumer (`bootc-image-builder`).

alexlarsson:
Closed in favor of https://github.com/osbuild/images/pull/1519

PR #1494, opened 2025-05-05
fedora: move per image type image config into YAML (HMS-6019)

mvo5:
[draft until https://github.com/osbuild/images/pull/1500 is merged]

This a cleaned up split-out of https://github.com/osbuild/images/pull/1462 - it moves the image configuration for fedora into YAML. It only needs #1500 as its prerequisite.

---

This commit moves the `ImageConfig` of each package into the
distro.yaml.

---

defs: implement `defs.ImageConfig()` for image config loading

This commit adds a new `defs.ImageConfig()` to load per image
type specific image config loading. This allows to write:
```yaml
image_types:
  test_type:
    image_config:
      locale: "C.UTF-8"
      timezone: "DefaultTZ"
      condition:
        version_less_than:
          "2":
            timezone: "OverrideTZ"

/jira-epic HMS-6019

PR #1508, opened 2025-05-09
distro: move disk/container image types into pure YAML (COMPOSER-2533)

mvo5:
This PR moves the disk/container image types in fedora into pure YAML. That means that there is no go code anymore that defines the converted image types and adding new image types to distro.yaml adds them dynamically.

Here is an example:
```yaml
  container: &container
    filename: "container.tar"
    mime_type: "application/x-tar"
    image_func: "container"
    bootable: false
    build_pipelines: ["build"]
    payload_pipelines: ["os", "container"]
    exports: ["container"]
    platforms:
      - arch: "x86_64"
      - arch: "aarch64"
      - arch: "ppc64le"
      - arch: "s390x"
      - arch: "riscv64"
    image_config: &image_config_container
      no_selinux: true
      exclude_docs: true
      locale: "C.UTF-8"
      timezone: "Etc/UTC"  
    package_sets:
      os:
        - include:
            - "bash"
...
```

This focused on fedora (for now) to show the concepts and to not get too distracted (as converting the code to YAML is a lot of busywork).

The iot and installer images for fedora still need to be done, but with this PR merge this should be just busywork. 

This started out split into a bunch of smaller PRs but I decided to open a huge one instead (which is not what I usually do) because a) it provides the needed context b) rebaseing big yaml trees is a pain c) slow traction on the small PRs so I try a big one now :)

Smaller PRs we could open (again):
- https://github.com/osbuild/images/pull/1505
- https://github.com/osbuild/images/pull/1504
- https://github.com/osbuild/images/pull/1500
- https://github.com/osbuild/images/pull/1494
- https://github.com/osbuild/images/pull/1462 (to some extend)

Note that there is some useful stuff in https://github.com/osbuild/images/pull/1462/commits to scavenge, especially https://github.com/osbuild/images/pull/1462/commits/62e5cbbffc28b3838410d840626ed06a74a44b3a


pkg/customizations/fsnode/fsnode.go
```diff
@@ -1,58 +1,68 @@
...
+	Group interface{}
+}
+
 type baseFsNode struct {
-	path  string
-	mode  *os.FileMode
-	user  interface{}
-	group interface{}
+	baseFsNodeJSON
 }
```
achilleas-k, line 26: I'm having a bit of trouble understanding the reason for this change.  Is it impossible (or messy) to do the unmarshalling into the `File` struct without the nesting?
Could we change the two fsnodes so that they can be unmarshalled unambiguously?  Or perhaps we could have a slightly different serialised format from the internal?

pkg/distro/defs/loader.go
```diff
@@ -349,3 +362,38 @@ func load(distroNameVer string) (*toplevelYAML, error) {
...
 	return &toplevel, nil
 }
+
+// ImageConfig returns the image type specific ImageConfig
+func ImageConfig(distroNameVer, archName, typeName string, replacements map[string]string) (*distro.ImageConfig, error) {
+	toplevel, err := load(distroNameVer)
+	if err != nil {
+		return nil, err
+	}
+	typeName = strings.ReplaceAll(typeName, "-", "_")
```
achilleas-k, line 415: Will this cause confusion, the mismatch between the name in the yaml and how it appears in the CLI and API?  Having names-with-dashes is possible with quotes, so maybe we could make it a habit or convention to define image type names that way?  Or would that be too ugly?

What I'm worried about here is that, if I was a user defining my own image type externally, I might name it "my-image-type" in the yaml, and then trying to `ib build my-image-type` wouldn't work.

pkg/distro/defs/loader.go
```diff
@@ -431,11 +431,17 @@ func ImageTypes(distroNameVer string) (map[string]ImageTypeYAML, error) {
...
 		return nil, err
 	}
+
+	// We have a bunch of names like "server-ami" that are writen
+	// in the YAML as "server_ami" so we need to normalize
+	imgTypes := make(map[string]ImageTypeYAML, len(toplevel.ImageTypes))
 	for name := range toplevel.ImageTypes {
 		v := toplevel.ImageTypes[name]
+		// normalize the name
+		name := strings.ReplaceAll(name, "_", "-")
```
achilleas-k, line 509: See previous comment about this.

pkg/distro/defs/loader.go
```diff
@@ -429,6 +441,43 @@ func ImageConfig(distroNameVer, archName, typeName string, replacements map[stri
 	return imgConfig, nil
 }
 
+// InstallerConfig returns the InstallerConfig for the given imgType
+// Note that on conditions the InstallerConfig is fully replaced, do
+// any merging in YAML
+func InstallerConfig(distroNameVer, archName, typeName string, replacements map[string]string) (*distro.InstallerConfig, error) {
```
achilleas-k, line 459: The commit message for this mentions `ImageConfig` instead of `InstallerConfig`.

achilleas-k (commented):
Generally, LGTM.  Some questions and comments inline.
Not approving because of the commit message that needs fixing (`InstallerConfig`) and my question about the FileNode struct and yaml format.

Also, I didn't check each line of the conversions to yaml in detail.  I trust the manifest checksums will take care of validating that.

pkg/distro/defs/loader.go
achilleas-k: Another idea might be to make image types an array in yaml with a `name` parameter, just like we do internally, and build the map as we load.

pkg/distro/defs/loader.go
mvo5: Thanks, silly me! Fixed (locally)

pkg/customizations/fsnode/fsnode.go
mvo5: Sorry, the commit message should probably be better here. The reason I did it this way is simply that there is already a "File.Path()" function. So just unmarshaling into "File" would require to rename the function and add an exported "File string". I did this in https://github.com/osbuild/images/pull/1497 but the downside of this appraoch is that its now trvial to create invalid node because there is no getters/setters anymore. The original simple map based approach was also not much liked (https://github.com/osbuild/images/pull/1479) and https://github.com/osbuild/images/pull/1500 got the most love so far. But happy to look into more directions here if there are new ideas (fwiw, I can open this one separtely again if you feel it would benefit from a more targeted discussion).

pkg/distro/defs/loader.go
mvo5: Adding quotes around the map seems okay to me, I did a commit with that and pushed.

achilleas-k:
Something broke.

pkg/customizations/fsnode/fsnode.go
achilleas-k: Thanks.  The explanation clarifies everything (I didn't connect it to the previous versions and I missed some of the discussion on those PRs).

I think the current solution is fine.  The only alternative I can think of is to basically disconnect the two, so unmarshal into `baseFsNodeJSON` and then copy everything to `baseFsNode` for internal use.  But that's pretty much the same effort and amount of code probably.

achilleas-k (approved):
Very nice!!

LGTM

achilleas-k:
I cloned this PR locally to poke around some of the YAML files and noticed some trailing whitespace on some lines.  Since we're going all in on YAML, we should eventually add yamllint to our PR checks to keep things clean.

mvo5:
> I cloned this PR locally to poke around some of the YAML files and noticed some trailing whitespace on some lines. Since we're going all in on YAML, we should eventually add yamllint to our PR checks to keep things clean.

Thanks, that is an excellent point. I added a yamllint check in https://github.com/osbuild/images/pull/1514 (and fixed the issues of course)

pkg/customizations/fsnode/fsnode.go
mvo5: Yeah, they are probably about the same. I /think/ I even had a version with a `baseFsNodeJSON` and that copied into `baseFsNode` - I don't remember why I decided on this in the end. I don't think go has a really nice pattern for this :/

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -12,6 +12,181 @@
...
+      - "console=ttyS0,115200n8"
+      - "biosdevname=0"
+      - "net.ifnames=0"
+    ostree_deployment_kernel_options: &ostree_deployment_kernel_options
+      - "modprobe.blacklist=vc4"
+      - "rw"
+      - "coreos.no_persist_ip"
+
+  environments:
+    kvm: &kvm_env
```
supakeen, line 30: Consistency wise, should this key be called `kvm_env`?

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -12,6 +12,181 @@
...
+    iot: &image_config_iot
+      <<: *image_config_iot_enabled_services
+      keyboard:
+        keymap: "us"
+      locale: "C.UTF-8"
+      ostree_conf_sysroot_readonly: true
+      lock_root_user: true
+
+  disk_sizes:
+    default_required_partition_sizes: &default_required_dir_sizes
```
supakeen, line 186: Anchor name should probably be `&default_required_partition_sizes`?

supakeen (approved):

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -493,438 +497,452 @@ image_types:
...
-                - "dmidecode"
-                - "grub2-tools-efi"
-                - "memtest86+"
-            aarch64:
-              include:
-                - "dmidecode"
+      os:
+        - *minimal_raw_pkgset
+      installer:
+        - &anaconda_pkgset
```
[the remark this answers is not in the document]
supakeen: Not particularly fond of "FQDN"s. We can visit this problem once we're also ready to split out the pipelines a bit more (e.g. separate customization pipeline, etc).

PR #1514, opened 2025-05-12
fedora: move all image types into pure YAML

mvo5:
This PR moves all image types into pure YAML.

### DIFF
--- a/pkg/distro/defs/fedora/distro.yaml
+++ b/pkg/distro/defs/fedora/distro.yaml
@@ -1226,7 +1226,8 @@ image_types:
                 include:
                   - "anaconda-webui"
     platforms:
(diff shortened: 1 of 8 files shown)


supakeen (approved):
Awesome stuff. Looks good to me. Needs current conflicts and the ones that will appear after #1523 merges fixed.

supakeen (approved):

bcl (approved):
Looks good to me. One small complaint about the new image fields is that `image_func` implies a function name, which it isn't. And if it was I'd complain that was exposing internal functions to the poor users :) I'd have named it `image_type` instead. And the struct name should also be ImageType (I had to search to figure out where .Image was coming from).

PR #1525, opened 2025-05-15
Fixes for ESP partition: Make optional, set label

alexlarsson:
I'm doing some experiments with disk customizations, and I'd like to create my own ESP partition, which means we need to not add it if it exists, and I also noticed that we don't support setting the label on vat filesystems atm.


alexlarsson:
I have one question here btw. Currently this label is "EFI-SYSTEM", but on my fedora system the label used is ESP. While we introduce this, maybe we should sync these?

alexlarsson:
Label `ESP` would match what automotive-image-builder uses too.

achilleas-k:
So we never had a label on the ESP it seems (I went all the way back to the first manifests we added in osbuild-composer).  I agree, since this is an introduction of the label, there's no fear of breaking any workflows that rely on a specific label.  Let's align it with Fedora.

achilleas-k:
Default ESP label aside: LGTM! :)

alexlarsson:
Added a commit that renames this to ESP

bcl:
FWIW nothing should be using the label to find the ESP, it should be using the partition type for that so setting it should not break anything that's behaving properly :)

achilleas-k:
> FWIW nothing should be using the label to find the ESP

Of course, but you never know what kind of things people get up to :laughing:

achilleas-k:
Restarting test failures.  Seems like we're hitting some test runner limits when there are too many rebuilds.

achilleas-k (approved):
Thanks!  LGTM

pkg/disk/partition_table.go
```diff
@@ -1083,6 +1083,20 @@ func addBootPartition(pt *PartitionTable, bootFsType FSType) error {
 	return nil
 }
 
+func hasESP(disk *blueprint.DiskCustomization) bool {
```
bcl, line 1086: This seems like a good candidate for a struct method on DiskCustomization since it only depends on info from the struct in question.

alexlarsson: I'm not sure its really worth making such a small helper be part of the disk customization object directly. Its quite specific. Having it as a regular function is similar to e.g. needsBoot() in the same file.

supakeen:
@achilleas-k there's errors in test coverage here, unrelated to the PR. How can I fix them?

achilleas-k:
> @achilleas-k there's errors in build coverage here, seems unrelated to the PR. How can I fix them?

Very odd.  Perhaps it was some kind of race issue?  Cache cleanup while something else was building and then it passed?
I'm going to ignore it for now and see if something similar happens again.

PR #1526, opened 2025-05-15
Cleanup of new APIs (mkfs options and build container)

alexlarsson:
This is some follow ups from  review in https://github.com/osbuild/images/pull/1519

### DIFF
--- a/pkg/disk/filesystem.go
+++ b/pkg/disk/filesystem.go
@@ -1,11 +1,56 @@
 package disk
 
 import (
+	"encoding/json"
+	"fmt"
 	"math/rand"
 	"reflect"
 
 	"github.com/google/uuid"
 )
+
+type MkfsOption int
+
+const ( // MkfsOption type enum
+	MkfsVerity MkfsOption = iota // Enable fs-verity option if needed (typically for EXT4)
+)
+
+func getMkfsOptionMapping() []string {
+	return []string{"verity"}
+}
+
+// ToString converts MkfsOption into a human readable string
+func (option MkfsOption) ToString() string {
+	return getMkfsOptionMapping()[int(option)]
+}
+
+func unmarshalHelper(data []byte, mapping []string) (int, error) {
+	var stringInput string
+	err := json.Unmarshal(data, &stringInput)
+	if err != nil {
+		return 0, err
+	}
+	for n, str := range mapping {
+		if str == stringInput {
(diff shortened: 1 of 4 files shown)


pkg/disk/filesystem.go
```diff
@@ -23,7 +68,7 @@ type Filesystem struct {
 	// The sixth field of fstab(5); fs_passno
 	FSTabPassNo uint64 `json:"fstab_passno,omitempty" yaml:"fstab_passno,omitempty"`
 	// Custom mkfs options
-	MkfsOptions []string `json:"mkfs_options,omitempty" yaml:"mkfs_options,omitempty"`
+	MkfsOptions []MkfsOption `json:"mkfs_options,omitempty" yaml:"mkfs_options,omitempty"`
```
achilleas-k, line 71: I suspect we might end up reverting this change when we decide to support arbitrary mkfs options, but for now I think let's keep it and see how it goes.

achilleas-k (approved):
LGTM

supakeen (approved):

pkg/disk/filesystem.go
```diff
@@ -1,12 +1,57 @@
...
+const ( // MkfsOption type enum
+	MkfsVerity MkfsOption = iota // Enable fs-verity option if needed (typically for EXT4)
+)
+
+func getMkfsOptionMapping() []string {
+	return []string{"verity"}
+}
+
+// ToString converts MkfsOption into a human readable string
+func (option MkfsOption) ToString() string {
```
mvo5, line 23: I'm a bit late to the party, sorry! I was thinking maybe `String()` is slightly better here, I opened https://github.com/osbuild/images/pull/1545

pkg/disk/filesystem.go
```diff
@@ -1,12 +1,57 @@
...
+	for n, str := range mapping {
+		if str == stringInput {
+			return n, nil
+		}
+	}
+	return 0, fmt.Errorf("invalid mkfsoption: %s", stringInput)
+}
+
+// UnmarshalJSON converts a JSON string into an MkfsOption
+func (option *MkfsOption) UnmarshalJSON(data []byte) error {
```
mvo5, line 42: I know its a bit silly for trivial things like this but I do feel better having tests for these helpers, I did this boring work in https://github.com/osbuild/images/pull/1545

PR #1538, opened 2025-05-16
distro/rhel: remove the user/group warnings for edge-commits

achilleas-k:
A long time ago, we decided that user and group creation for ostree-based images should happen in the deployment (disk image creation or via an installer).  The plan was to deprecate the feature of user and group creation in ostree commits (the base, immutable part of the system) for a few minor versions and then remove the feature completely.

Normally, this kind of thing would make sense to change in a major release (RHEL 10), but with the move to bootc and image mode in RHEL 10, we're not deprecating any features but the whole thing is changing.

The decision has changed now: There will be no removal of the feature in edge for the lifetime of RHEL 9.

See comment on [THEEDGE-4235](https://issues.redhat.com/browse/THEEDGE-4235)

See also https://github.com/osbuild/osbuild-composer/issues/3106 for some more historical context.


bcl (approved):
Ack

PR #1542, opened 2025-05-19
rhel10: move ImageConfig into pure YAML

mvo5:
This commit moves the rhel10 ImageConfig into pure YAML.


supakeen (approved):
We require additional YAML.

pkg/osbuild/chrony_stage.go
```diff
@@ -66,6 +69,65 @@ type ChronyConfigRefclock struct {
...
+	c.Offset = d.Offset
+
+	var peek struct {
+		Name string
+	}
+	if err := json.Unmarshal(*d.Driver, &peek); err != nil {
+		return err
+	}
+	switch peek.Name {
+	case "PPS":
```
bcl, line 96: I wonder if this should be case-insensitive? The docs use upper case for these, but users don't always follow directions. Do you happen to know what chrony does if they are lower case?

bcl: Nevermind -- chronyd insists on upper case names :)

bcl (approved):
Looks good to me.

PR #1543, opened 2025-05-19
Add support for setting partition uuid and label

alexlarsson:
As discussed in https://github.com/bootc-dev/bootc/issues/926, we want to be able to set more GPT partition details in the customization.

### DIFF
--- a/pkg/blueprint/disk_customizations.go
+++ b/pkg/blueprint/disk_customizations.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"slices"
 	"strings"
+	"unicode/utf16"
 
 	"github.com/google/uuid"
 	"github.com/osbuild/images/pkg/datasizes"
@@ -71,6 +72,14 @@ type PartitionCustomization struct {
 	// If not set, the type will be automatically set based on the mountpoint
 	// or the payload type.
 	PartType string `json:"part_type,omitempty" toml:"part_type,omitempty"`
+
+	// The partition label for GPT partitions, not supported for dos partitions.
+	// Note: This is not the same as the label, which can be set in "Label"
+	PartLabel string `json:"part_label,omitempty" toml:"part_label,omitempty"`
+
+	// The partition GUID for GPT partitions, not supported for dos partitions.
+	// Note: This is the unique uuid, not the type guid, that is PartType
+	PartUUID string `json:"part_uuid,omitempty" toml:"part_uuid,omitempty"`
 
 	BtrfsVolumeCustomization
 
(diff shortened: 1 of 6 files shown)


pkg/blueprint/disk_customizations.go
```diff
@@ -72,6 +73,14 @@ type PartitionCustomization struct {
 	// or the payload type.
 	PartType string `json:"part_type,omitempty" toml:"part_type,omitempty"`
 
+	// The partition label for GPT partitions, not supported for dos partitions.
+	// Note: This is not the same as the label, which can be set in "Label"
+	PartLabel string `json:"part_label,omitempty" toml:"part_label,omitempty"`
+
+	// The partition GUID for GPT partitions, not supported for dos partitions.
+	// Note: This is the uniqe uuid, not the type guid, that is PartType
+	PartId string `json:"part_id,omitempty" toml:"part_id,omitempty"`
```
achilleas-k, line 82: Would calling this `PartUUID` be less ambiguous?

achilleas-k (dismissed):
Thanks.  LGTM!

One small comment but I don't have a strong opinion so I wont block on it.

pkg/blueprint/disk_customizations.go
alexlarsson: I dunno, no matter what you do the two uuids are confusing. Pick whatever you think is best.

bcl: the Part prefixes for the fields is already kinda redundant since it's inside `PartitionCustomization` but whatever :) FWIW in the output from parted --json we use `type-uuid` to indicate the GPT type and `uuid` for the partition's actual unique ID so I'd go with `UUID` here. Or `PartUUID` :)

alexlarsson: The thing is that `PartitionCustomization` also contain an anonymous  FilesystemTypedCustomization which adds fields related to filesystems to the same namespace. Currently this already contains `Label` which means the FS label, and that means I *have* to use PartLabel for the partition label (as this particular prefix matches the already existing PartType). There is currently no "UUID" in the filesystem part, but technically there could be, because filesystems have UUIDs too. So a plain "UUID" could conflict in the future. Better to prefix it similar to the other fields.

But, I agree, PartUUID seems like a better name than PartId. Also it matches what disk.Partition already uses (UUID).

alexlarsson: I updated it to part_uuid/PartUUID

supakeen (approved):

achilleas-k (approved):
Thanks!

pkg/blueprint/disk_customizations.go
achilleas-k: Thank you.  Yeah, I'm convinced now that PartUUID is a better name.

bcl: Ah, that makes sense, thanks!

PR #1544, opened 2025-05-20
fedora: implement setting of the RootfsType via YAML

mvo5:
fedora: implement setting of the RootfsType via YAML

This commit moves the setting of the rootfs type into the
ImageConfig and by implication into pure YAML.

The relevant image defintions are now also updated. This
also means that one can define `iso_rootfs_type: "erofs"`
trivially now per image type.

Note that this also shows an inconsistency with the old
code, some of the iso images have been moved to the new
SquashfsRootfs but other have not. We should probably
look into this too but for now this PR opts to not change
the behavior just moves the place how the RootfsType for
the iso is set.

---

manifest: make `RootfsType` marshallable via JSON/YAML

This commit makes the `manifest.RootfsType` marshallable via
JSON/YAML so that we can use it in the image defintions.

This will be useful for:
https://github.com/osbuild/images/pull/1239
https://github.com/osbuild/images/pull/1537

### DIFF
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -1,6 +1,7 @@
 package manifest
 
 import (
+	"encoding/json"
 	"fmt"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


supakeen (dismissed):

bcl:
FWIW the reason why I didn't change them all (eg. coreos_installer) is because they do things differently than the anaconda iso so I didn't want to disturb whatever special things they expect in their images.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -25,6 +26,29 @@ const ( // Rootfs type enum
 	ErofsRootfs                          // Create a plain erofs rootfs
 )
 
+func (r *RootfsType) UnmarshalJSON(data []byte) error {
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+	switch s {
+	case "ext4", "":
```
bcl, line 35: I think this should be `squashfs-ext4` to make sure there is no confusion thinking it's plain ext4 or that arbitrary filesystem types can be used. This is also consistent with what you can pass to lorax's `--rootfs-type`

bcl (changes requested):

pkg/manifest/anaconda_installer_iso_tree.go
mvo5: Thanks, yeah, that makes sense. Updated.

mvo5:
> differently than the anaconda iso so I didn't want to disturb whatever special things they expect in their images.

Thanks! This is also excellent feedback, I updated the (actually slightly negative comment about iso_rootfs_type, sorry for that) question/XXX into a proper explanation.

supakeen (approved):

PR #1548, opened 2025-05-20
`vagrant-libvirt` implementation (HMS-6116)

supakeen:
Initial implementation of a new `vagrant-libvirt` image type. Necessary for CentOS. This implements the current version of the stage in `osbuild` to generate an archive with the appropriate files.

We'll be adding another variant (`vagrant-virtualbox`) after I clean up and merge the [osbuild PR for it](https://github.com/osbuild/osbuild/pull/1136).


bcl (approved):
Looks good to me

supakeen:
Ok, cleaned this up a bit and ready for review (again) :)

pkg/osbuild/vagrant_stage.go
```diff
@@ -12,10 +12,6 @@ type VagrantStageOptions struct {
 
 func (VagrantStageOptions) isStageOptions() {}
 
-func (o *VagrantStageOptions) validate() error {
-	return nil
-}
-
```
thozza: Wouldn't it make sense to move the check from `NewVagrantStageOptions()` to this method instead?

pkg/image/disk.go
```diff
@@ -80,6 +80,17 @@ func (img *DiskImage) InstantiateManifest(m *manifest.Manifest,
 		qcow2Pipeline := manifest.NewQCOW2(buildPipeline, rawImagePipeline)
 		qcow2Pipeline.Compat = img.Platform.GetQCOW2Compat()
 		imagePipeline = qcow2Pipeline
+	case platform.FORMAT_VAGRANT_LIBVIRT:
+		qcow2Pipeline := manifest.NewQCOW2(buildPipeline, rawImagePipeline)
+		qcow2Pipeline.Compat = img.Platform.GetQCOW2Compat()
+
+		vagrantPipeline := manifest.NewVagrant(buildPipeline, qcow2Pipeline)
```
thozza, line 87: Wouldn't it make sense to explicitly pass the backend to `manifest.NewVagrant()` since the format is `VAGRANT_LIBVIRT`?

thozza (commented):
Thanks. In general the PR LGTM. I've added two questions inline.

pkg/image/disk.go
supakeen: That is done in #1556 (the followup to this PR) which actually adds a second provider.

pkg/osbuild/vagrant_stage.go
supakeen: It does, I've changed it and force pushed.

bcl (dismissed):
Ack

thozza (approved):
:+1:

PR #1550, opened 2025-05-20
fedora/wsl: include `wsl-setup` (HMS-8573)

supakeen:
In newer versions of WSL there's a new out-of-the-box experience. This is packaged in Fedora under `wsl-setup`. Let's include the package.

### DIFF
--- a/pkg/distro/defs/fedora/distro.yaml
+++ b/pkg/distro/defs/fedora/distro.yaml
@@ -1518,10 +1518,10 @@ image_types:
     name_aliases: ["server-wsl"]
     filename: "wsl.tar"
     mime_type: "application/x-tar"
-    image_func: "container"
+    image_func: "tar"
     build_pipelines: ["build"]
-    payload_pipelines: ["os", "container"]
-    exports: ["container"]
+    payload_pipelines: ["os", "archive"]
+    exports: ["archive"]
     required_partition_sizes: *default_required_dir_sizes
     platforms:
       - arch: "x86_64"
--- a/pkg/distro/defs/fedora/distro.yaml
+++ b/pkg/distro/defs/fedora/distro.yaml
@@ -1527,25 +1527,24 @@ image_types:
       - arch: "x86_64"
     image_config:
       <<: *image_config_container
-      cloud_init:
-        - filename: "99_wsl.cfg"
-          config:
-            datasource_list:
-              - "WSL"
-              - "None"
-            network:
-              config: "disabled"
       condition:
         version_less_than:
           "42":
             wsl_config:
               boot_systemd: true
+            cloud_init:
+              - filename: "99_wsl.cfg"
+                config:
+                  datasource_list:
+                    - "WSL"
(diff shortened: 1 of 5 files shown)


supakeen:
> Looks fine to me, why is it draft?

Because I need to find a Windows machine to try if it works.

supakeen:
Confirmed that this works in a VM, see screenshot (logo/name is set, installable by double-click).

Untitled

supakeen:
@mvo5 this is now ready for review.

pkg/distro/fedora/images.go
```diff
@@ -368,6 +368,31 @@ func diskImage(workload workload.Workload,
 	return img, nil
 }
 
+func archiveImage(workload workload.Workload,
```
mvo5, line 371: In rhel this is called `TarImage`, lets be consistent here (c.f. https://github.com/osbuild/images/blob/main/pkg/distro/rhel/images.go#L791 ) and call it the same (either here or in the other place). Note that those will be unified hopefully not too far in the future and having similar names will make this easier.

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -1494,12 +1494,12 @@ image_types:
     # this is the eventual name, and `wsl` the alias but we've been
     # having issues with CI renaming it
     name_aliases: ["server-wsl"]
-    filename: "wsl.tar"
+    filename: "image.wsl"
```
mvo5, line 1497: In the rhel `mkWSLImgType()` the filename is `disk.tar.gz` - maybe worth a small comment why we diverge in the various place. Also will anything break if we change the filename? I.e. do our users expect the old one? Do we need to release-note this somehow?

mvo5 (commented):
Thanks for working on this, some small questions/suggestions inline

pkg/distro/defs/fedora/distro.yaml
supakeen: Since the previous WSL images for Fedora never worked at all (they cannot be imported, they were in an incorrect format) I don't suspect this to break for any users.

The reason for the name change and link to Microsoft documentation is listed in the commit message. WSLv2 images *must* have the `.wsl` file extension.

pkg/distro/fedora/images.go
supakeen: Done.

supakeen:
I've added a commit since the last review which removes the `cloud-init` configuration for the WSL images for Fedora. Fedora currently uses `wsl-setup`'s OOBE to configure the system on first "boot" instead of `cloud-init`.

This deviation between RHEL/CentOS and Fedora has been brought up and we'll likely start supporting `cloud-init` directly in `wsl-setup` at a later stage (perhaps for Fedora 43).

thozza (dismissed):
LGTM, I agree with @mvo5 suggestion to add a comment in the code about the filename difference.

bcl (approved):
Looks good!

PR #1555, opened 2025-05-21
bootc: Make installed rootfs configurable

mbarak:
Contributes towards: https://github.com/osbuild/bootc-image-builder/issues/924

Allow passing in the root filesystem format when generating an ISO.

Follow-up: https://github.com/osbuild/bootc-image-builder/pull/935

### DIFF
--- a/pkg/image/anaconda_container_installer.go
+++ b/pkg/image/anaconda_container_installer.go
@@ -10,6 +10,7 @@ import (
 	"github.com/osbuild/images/pkg/customizations/anaconda"
 	"github.com/osbuild/images/pkg/customizations/kickstart"
 	"github.com/osbuild/images/pkg/datasizes"
+	"github.com/osbuild/images/pkg/disk"
 	"github.com/osbuild/images/pkg/manifest"
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/platform"
@@ -58,6 +59,9 @@ type AnacondaContainerInstaller struct {
 	// Locale for the installer. This should be set to the same locale as the
 	// ISO OS payload, if known.
 	Locale string
+
+	// Filesystem type for the installed system as opposed to that of the ISO.
+	InstallRootfsType disk.FSType
 }
 
 func NewAnacondaContainerInstaller(container container.SourceSpec, ref string) *AnacondaContainerInstaller {
@@ -149,6 +153,8 @@ func (img *AnacondaContainerInstaller) InstantiateManifest(m *manifest.Manifest,
 	if img.FIPS {
 		isoTreePipeline.KernelOpts = append(isoTreePipeline.KernelOpts, "fips=1")
 	}
+
+	isoTreePipeline.InstallRootfsType = img.InstallRootfsType
 
 	isoPipeline := manifest.NewISO(buildPipeline, isoTreePipeline, img.ISOLabel)
 	isoPipeline.SetFilename(img.Filename)
--- a/pkg/manifest/anaconda_installer_iso_tree.go
+++ b/pkg/manifest/anaconda_installer_iso_tree.go
@@ -105,6 +105,8 @@ type AnacondaInstallerISOTree struct {
 	// Pipeline object where subscription-related files are created for copying
 	// onto the ISO.
 	SubscriptionPipeline *Subscription
+
+	InstallRootfsType disk.FSType
 }
 
 func NewAnacondaInstallerISOTree(buildPipeline Build, anacondaPipeline *AnacondaInstaller, rootfsPipeline *ISORootfsImg, bootTreePipeline *EFIBootTree) *AnacondaInstallerISOTree {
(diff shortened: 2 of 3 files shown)


pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -594,12 +596,29 @@ func (p *AnacondaInstallerISOTree) bootcInstallerKickstartStages() []*osbuild.St
...
-reqpart --add-boot
+	var rootFsType string
+	switch p.InstallRootfsType {
+	case "btrfs":
+		rootFsType = "btrfs"
+	case "xfs":
+		rootFsType = "xfs"
+	default:
+		rootFsType = "ext4"
+	}
```
supakeen, line 607: I'd prefer to not do a default inside `images` and to instead handle the default inside `bootc-image-builder` and `panic()`'ing here if an invalid `InstallRootfsType` is passed into it.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -594,12 +596,29 @@ func (p *AnacondaInstallerISOTree) bootcInstallerKickstartStages() []*osbuild.St
...
+	var hardcodedKickstartBits string
+	switch rootFsType {
+	case "btrfs":
+		hardcodedKickstartBits = `
+autopart --nohome --type=btrfs
+`
+	default:
+		hardcodedKickstartBits = fmt.Sprintf(`
+autopart --nohome --type=plain --fstype=%s
+`, rootFsType)
```
supakeen, line 618: I feel this could be an else/if instead since there's only two cases :)

pkg/image/anaconda_container_installer.go
```diff
@@ -58,6 +58,8 @@ type AnacondaContainerInstaller struct {
 	// Locale for the installer. This should be set to the same locale as the
 	// ISO OS payload, if known.
 	Locale string
+
+	InstallRootfsType string
```
supakeen, line 62: While I understand where this name comes from it feels slightly confusing with the other `RootfsType` in this struct. I don't have a better idea for a name myself though.

supakeen (commented):
Some remarks from a quick scan; don't act on them yet because I don't feel very strongly about them and others might disagree :)

mvo5 (commented):
Thank you! This looks nice, I made a few suggestions inline. Having a test would be really good, I looked into this but it seems we currently make it a bit hard to test for the right exta ks kickstart content, I will poke a bit more and get back if I have a good idea.

pkg/image/anaconda_container_installer.go
bcl: It should have a comment that clearly states which rootfs it applies to.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -594,12 +596,29 @@ func (p *AnacondaInstallerISOTree) bootcInstallerKickstartStages() []*osbuild.St
...
+		rootFsType = "ext4"
+	}
 
-part swap --fstype=swap --size=1024
-part / --fstype=ext4 --grow
+	var hardcodedKickstartBits string
+	switch rootFsType {
+	case "btrfs":
+		hardcodedKickstartBits = `
+autopart --nohome --type=btrfs
```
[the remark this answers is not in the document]
mbarak: I couldn't find where `autopart` is implemented to give you a definitive answer. It looks like it uses some heuristics to determine what to do. For BTRFS in particular it doesn't seem like it creates a swap at all.

Here's how the install went on my machine:
```
[mbarak@cloud mbarak]$ lsblk --fs
NAME        FSTYPE FSVER LABEL     UUID                                 FSAVAIL FSUSE% MOUNTPOINTS
nvme0n1                                                                                
├─nvme0n1p1 vfat   FAT32           300A-5E51                             591.3M     1% /boot/efi
├─nvme0n1p2 ext4   1.0             a3196507-d6f6-4351-83d8-c464e0088555  805.6M    10% /boot
└─nvme0n1p3 btrfs        fl_fedora babadc60-5d00-4845-80e5-ed2d2a0f2e5a  948.4G     0% /var/lib/containers/storage/overlay
                                                                                       /var
                                                                                       /sysroot/ostree/deploy/default/var
                                                                                       /etc
                                                                                       /sysroot
[mbarak@cloud mbarak]$ lsmem --summary
Memory block size:       128M
Total online memory:    31.9G
Total offline memory:      0B
```

pkg/image/anaconda_container_installer.go
```diff
@@ -58,6 +59,10 @@ type AnacondaContainerInstaller struct {
 	// Locale for the installer. This should be set to the same locale as the
 	// ISO OS payload, if known.
 	Locale string
+
+	// Filesystem type for the installed system as opposed to that of the ISO.
+	// The ISO filesystem type is hard coded to squashfs.
```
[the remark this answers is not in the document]
mbarak: Good catch. I'll update the comment to say something like: "ISO images have their own set of filesystem types"

mvo5:
> > The bigger question here is - this changes the way we install the system as we no longer provide an explicit swap size and leave that to "autopart" (and according to https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#id2 it will take care of swap). Which is probably a good thing actually but I would like to confirm with one of the more seasoned team members like @achilleas-k or @thozza or @bcl that this is okay.
> 
> I do like deferring such decisions to Anaconda. I think it aligns with people's expectations, if they have experience with the installer's defaults.

That is fine with me then, once the conflict and comment is updated I'm happy to approve (I can also help if you are busy or the conflict is gnarly)

pkg/image/anaconda_container_installer.go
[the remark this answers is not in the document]
mbarak: I opted to just delete the comment. It made it sound like there's a different said of filesystem types for the installed rootfs since the comment is above the variable `installRootfsType`.

mvo5:
> @mvo5 Thanks! Just checking in, I'm away until the weekend so I won't be able to do the final touches until then.
> 
> The conflict is simply more tests added to the end of the file.

Sorry for the delay, I was away for a few days, this looks great now!

thozza (approved):
LGTM

PR #1557, opened 2025-05-22
fedora/many: increase `/boot` to 1 GiB (HMS-8604)

supakeen:
@alexxa noticed that many of the Fedora `/boot` partitions are sized to 512 MiB while Fedora's images tend to have a 1 GiB size.

Let's align this with Fedora's official images.

---

This also changes the (previously custom) size for `/boot` on minimal-raw to now be the same as the base partition size :)


achilleas-k:
> Note that the `ppc64le` and `s390x` partition tables still use 512 MiB as they don't use the 'base' partition. Should I align those as well?

Yeah, might as well.

tabowling:
> @alexxa noticed that many of the Fedora `/boot` partitions are sized to 512 MiB while Fedora's images tend to have a 1 GiB size.
> 
> Let's align this with Fedora's official images.
> 
> This also changes the (previously custom) size for `/boot` on minimal-raw to now be the same as the base partition size :)

Did you intend to type "RHEL" in that opening sentence?

supakeen:
> > Note that the `ppc64le` and `s390x` partition tables still use 512 MiB as they don't use the 'base' partition. Should I align those as well?
> 
> Yeah, might as well.

Done.

achilleas-k (approved):
Nice and quick

ondrejbudai (approved):

PR #1558, opened 2025-05-22
fedora: add `anaconda.ModuleUsers` to ImageInstallerImage

mvo5:
This commit brings the fedora:imageInstallerImage() and the rhel:ImageInstallerImage() closer together. in RHEL we always add the `anaconda.ModuleUsers` but in fedora we don't do this.

By having them behave the same we will have an easier time unifying them when we move everything into YAML.

Note that I lack the domain knowledge to judge if this is the best solution. We could also do what the iot installer does instead:
```go
if len(img.Kickstart.Users)+len(img.Kickstart.Groups) > 0 {
	// only enable the users module if needed
	img.AdditionalAnacondaModules = append(img.AdditionalAnacondaModules, anaconda.ModuleUsers)
}
```
Suggestions welcome (but both helpers really should behave the same).

The resulting diff looks like this:
```json
diff -uNr /tmp/tmpn4o0xi2o/ref/fedora_43-x86_64-minimal_installer-empty_fedora.json /tmp/tmpn4o0xi2o/new/fedora_43-x86_64-minimal_installer-empty_fedora.json
--- /tmp/tmpn4o0xi2o/ref/fedora_43-x86_64-minimal_installer-empty_fedora.json	2025-05-22 22:00:05.469767621 +0200
+++ /tmp/tmpn4o0xi2o/new/fedora_43-x86_64-minimal_installer-empty_fedora.json	2025-05-22 22:00:12.744837844 +0200
@@ -582,7 +582,8 @@
             "activatable-modules": [
               "org.fedoraproject.Anaconda.Modules.Network",
               "org.fedoraproject.Anaconda.Modules.Payloads",
-              "org.fedoraproject.Anaconda.Modules.Storage"
+              "org.fedoraproject.Anaconda.Modules.Storage",
+              "org.fedoraproject.Anaconda.Modules.Users"
             ]
           }
         },

```


supakeen (approved):
This is fine to do; the Fedora image type this affects isn't built/delivered anywhere at the moment where this change would cause things to break.

bcl (approved):
Looks good.

PR #1564, opened 2025-05-27
image_config: allow specifying users

supakeen:
Allow to specify users directly in `ImageConfig`, for some image types such as Vagrant it might be nice to be able to do this.

---

This PR would allow #1562 to be implemented in a less abstract way. This PR is mostly for discussion though it does work.


lzap:
Out of curiousity, why adding users into image request when they are already defined in the customization? Is this some kind of override method?

PR #1572, opened 2025-05-30
test: install yamllint for tests

achilleas-k:
Install yamllint so that it's run by the unit test in the GH actions.

The test is skipped when the tool is not found.

https://github.com/osbuild/images/blob/2651be020a53db7f44d39ef028934fe55bc66d5d/pkg/distro/defs/loader_test.go#L48-L64


lzap (approved):
Or you can disable this pesky one globally. If it is not 79th, it is useless anyway :-)

PR #1577, opened 2025-06-02
distro: make "rhel-10" a generic distro (HMS-8743)

mvo5:
This PR makes "rhel-10" a generic distro that is fully defined in YAML.


pkg/distro/rhel/rhel10/options.go
```diff
@@ -1,108 +0,0 @@
...
-	"github.com/osbuild/images/pkg/blueprint"
-	"github.com/osbuild/images/pkg/customizations/oscap"
-	"github.com/osbuild/images/pkg/distro"
-	"github.com/osbuild/images/pkg/distro/rhel"
-	"github.com/osbuild/images/pkg/policies"
-)
-
-// checkOptions checks the validity and compatibility of options and customizations for the image type.
-// Returns ([]string, error) where []string, if non-nil, will hold any generated warnings (e.g. deprecation notices).
-func checkOptions(t *rhel.ImageType, bp *blueprint.Blueprint, options distro.ImageOptions) ([]string, error) {
```
mvo5, line 18: This needs double checking that really all options are ported over to "generic"

mvo5: We currently do not port the checks `UnsupportedPartitioningModes` - this is only needed in rhel-9,8, when we port the images to the generic distro this will happen there.

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -1572,6 +1572,13 @@ image_types:
     iso_label: "Unknown"
     # We don't know the variant that goes into the OS pipeline that gets installed
     variant: "Unknown"
+    # NOTE: this is not supported right now because the
+    # image-installer on Fedora isn't working when unattended.
+    # These options are probably necessary but could change.
+    # Unattended/non-interactive installations are better set to text
+    # time since they might be running headless and a UI is
```
achilleas-k, line 1579: I have no idea what "text time" is.  I don't know what I meant when [I wrote the original comment](https://github.com/osbuild/images/commit/e50f259522b020a00e64ece1e1b32a6875ca38c5).  I think I meant to just say "... set to text, since they might ...".

pkg/distro/generic/images.go
```diff
@@ -594,6 +594,13 @@ func imageInstallerImage(workload workload.Workload,
 		}
 	}
 
+	// Put the kickstart file in the root of the iso, some image
+	// types (like rhel10/image-installer) put them there, some
+	// others (like fedora/image-installer) do not and because
+	// its not uniform we need to make it configurable.
+	// XXX: unify with rhel-11 ? or rhel-10.x?
+	img.ISORootKickstart = t.ISORootKickstart
```
achilleas-k, line 602: Even in Fedora we put it in the root when we want the installation to be non-interactive.  IIRC, the reason we do this is because the correct way to set default values for Anaconda is in `/usr/share/anaconda/interactive-defaults.ks`, but you can't configure a fully automated installation with that file.

Do we can:
- Use the same logic in both Fedora and RHEL, though this might be considered a breaking change for RHEL users who modify the kickstart file after building one of our ISOs.
- Always put the kickstart file in the root.  This is what we used to do, but were asked to use interactive-defaults in Fedora (my memory is not 100% clear on this).
- Use [profiles](https://anaconda-installer.readthedocs.io/en/latest/developer/configuration-files.html#profile-configuration-files) for setting defaults, which we wanted to do at some point but never got around to it.

achilleas-k (dismissed):
:tada:

pkg/distro/generic/images.go
bcl: If people expect to be able to alter the kickstart it should be in the root of the iso, this makes it easy to use `mkksiso` to change them. Putting them into the installer root filesystem makes it much more difficult to make changes once the iso is built. Were these kickstarts incomplete? Anaconda used to support that, allowing the user to fill in missing details like the disk they wanted to install to.

pkg/distro/defs/loader_test.go
```diff
@@ -974,8 +976,25 @@ func TestDistrosLoadingFactoryCompat(t *testing.T) {
...
+	defer restore()
+
+	distro, err := defs.NewDistroYAML("fedora-43")
+	require.NoError(t, err)
+	// fedora-43 does not exclude this "test_type"
+	assert.Len(t, distro.ImageTypes(), 1)
+
+	distro, err = defs.NewDistroYAML("centos-10")
+	require.NoError(t, err)
+	// but centos-9 does exclude this "test_type"
```
bcl, line 996: Typo :) `centos-10` according to line 990

pkg/distro/defs/loader.go
```diff
@@ -214,6 +214,22 @@ func NewDistroYAML(nameVer string) (*DistroYAML, error) {
 			v := toplevel.ImageTypes[name]
 			v.name = name
 			foundDistro.imageTypes[name] = v
+			for idx := range v.Platforms {
+				templ, err := template.New("uefi-vendor").Parse(v.Platforms[idx].UEFIVendor)
+				if err != nil {
+					return nil, err
+				}
+				var buf bytes.Buffer
```
bcl, line 222: I think this should also check for an empty vendor string, if that were to happen the UEFI files would end up in the wrong place in the EFI directory.

bcl (changes requested):
Looks excellent, just one teeny typo and checking for empty uefi vendor strings.

bcl (dismissed):
Looks good to me, but now it has a conflict :)

pkg/distro/defs/fedora/distro.yaml
mvo5: I tweaked this now.

pkg/distro/generic/images.go
mvo5: Thanks for the extra context! I'm not sure if there is anything actionable for me here though, should I file an issue or do a followup ? It seems necessary to keep things as they are if changing is considered a breaking change in rhel, so the first opportunity for this would be rhel-11 I guess(?).

pkg/distro/generic/images.go
```diff
@@ -161,6 +173,9 @@ func osCustomizations(t *imageType, osPackageSet rpmmd.PackageSet, containers []
 	if containerStorage := c.GetContainerStorage(); containerStorage != nil {
 		osc.ContainersStorage = containerStorage.StoragePath
 	}
+	// set yum repos first, so it doesn't get overridden by
+	// imageConfig.YUMRepos
```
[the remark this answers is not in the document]
achilleas-k: I'm going to guess the comment in the original commit is wrong (or at least unclear, @kingsleyzissou?).  There's definitely something going on with the `YUMRepos` property being set twice and the order being important.  But maybe let's leave it for now and figure it out separately.

pkg/distro/generic/imagetype.go
```diff
@@ -489,7 +489,7 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint, options distro.ImageOp
 	}
 	if instCust != nil {
 		// only supported by the Anaconda installer
-		if slices.Index([]string{"iot-installer"}, t.Name()) == -1 {
+		if slices.Index([]string{"image-installer", "edge-installer", "live-installer", "iot-installer"}, t.Name()) == -1 {
```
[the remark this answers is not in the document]
mvo5: This is added now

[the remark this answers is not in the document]
mvo5: Fixed (as I had to edit this anyway)

pkg/distro/defs/loader.go
```diff
@@ -374,6 +361,34 @@ func (it *ImageTypeYAML) Name() string {
 	return it.name
 }
 
+func (it *ImageTypeYAML) runTemplates(distro *DistroYAML) error {
+	// fill the UEFI vendor string
+	var data any
```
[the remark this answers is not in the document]
mvo5: I moved it to a better place now :)

achilleas-k (approved):

pkg/manifest/manifest.go
```diff
@@ -32,8 +33,36 @@ const (
 	DISTRO_EL8
 	DISTRO_EL7
 	DISTRO_FEDORA
+	o
```
thozza, line 36: Not sure what this addition is for.... :thinking:

thozza (approved):
There's just one addition that I'm not sure about; the rest looks great. Thanks for all the changes, such as moving options to `InstallerConfig` and implementing shallow merge for it. 

I'm not sure if we still support regular "override" operation for any other structures. If not, we may need to fix the docs that you've recently added.

pkg/manifest/manifest.go
thozza: Thanks. At least it was "worth it" to wait for my review :sweat_smile:

thozza (approved):
:heart_eyes:

achilleas-k (approved):

PR #1580, opened 2025-06-04
distro: make "rhel-7" a "generic" distro (HMS-8595)

mvo5:
Move rhel-7 to a generic distro

/Jira-epic hms-8595


pkg/distro/defs/rhel-7/distro.yaml
```diff
@@ -134,15 +255,265 @@ image_config:
...
+        "no-subscription":
+          yum_plugin:
+            subscription_manager:
+              enabled: false
+          subman:
+            rhsmcertd:
+              auto_registration: true
+            rhsm:
+              manage_repos: false
+        "with-subscription":
```
mvo5, line 361: I think we do not generate manifests for this, i.e. I think this may not be tested

pkg/distro/generic/images.go
```diff
@@ -420,6 +428,18 @@ func diskImage(workload workload.Workload,
 
 	img.Filename = t.Filename()
 
+	img.VPCForceSize = t.ImageTypeYAML.DiskImageVPCForceSize
+
+	if img.OSCustomizations.NoBLS {
```
mvo5, line 433: Just FTR: this comes from https://github.com/osbuild/images/blob/v0.155.0/pkg/distro/rhel/images.go#L427

achilleas-k (approved):
Another one down!

PR #1581, opened 2025-06-05
osbuild/tar: explicit compression

supakeen:
In osbuild 151 the tar stage gained the ability to explicitly set the compression to be used. Previously this was always done based on the filename. This PR brings that ability into `images` as well where it's needed for the Vagrant `.box` format and the `.wsl` formats.


supakeen:
> LGTM.
> 
> Tiny nitpick: you added a newline in `data/dependencies/osbuild` in the first commit and removed it in the second commit. All changes to `data/dependencies/osbuild` should be in a single commit.

Oh, that's a mistake; I'll rebase that in a bit :)

pkg/osbuild/tar_stage.go
```diff
@@ -70,6 +85,25 @@ func (o TarStageOptions) validate() error {
...
+		}
+		valid := false
+		for _, value := range allowedArchiveCompressionValues {
+			if o.Compression == value {
+				valid = true
+				break
+			}
+		}
+		if !valid {
+			return fmt.Errorf("'compression' option does not allow %q as a value", o.Compression)
```
lzap, line 103: Wrapping errors without values does not allow the caller to use `errors.Is`, ideally there should be a new error value `ErrUnuspportedCompression` wrapped into this. But probably not big deal here since the caller will just error out.

lzap:
Dropped some random comments. I will not be doing formal reviews for a while until I get more confident with this repo :-)

pkg/osbuild/tar_stage.go
```diff
@@ -70,6 +85,25 @@ func (o TarStageOptions) validate() error {
...
+			TarArchiveCompressionGzip,
+			TarArchiveCompressionZstd,
+		}
+		valid := false
+		for _, value := range allowedArchiveCompressionValues {
+			if o.Compression == value {
+				valid = true
+				break
+			}
+		}
```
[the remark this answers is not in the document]
supakeen: Right, but it's consistent with the implementation of the `TarArchiveFormat` type above it; and in other files in this package?

pkg/osbuild/tar_stage.go
thozza: Personally, I do not see any value in wrapping the error at the moment just because it is possible. We never inspect the error type returned from stage options validation to take a specific action based on it.

PR #1582, opened 2025-06-05
fedora/wsl: xz compression (HMS-8573)

supakeen:
This PR builds on top of #1581 which now allows selecting the compression for the tar stage. It extends the archive image type to have a compression field which is set based on the value in the image type YAML.

I'm not entirely happy with the situation in `archive.go`, it is consistent with the handling in `disk.go`.

I've spoken with @achilleas-k and @mvo5 about this and it would maybe be possible to introduce a common subset enum of compression supported between both the disk, and archive, image types. However; they already deviate (archive supports gzip, disk does not) so this will require additional stages and pipelines to support gzip on disk images as well.

We can discuss this (or better) approaches in this PR.

### DIFF
--- a/pkg/image/archive.go
+++ b/pkg/image/archive.go
@@ -1,12 +1,14 @@
 package image
 
 import (
+	"fmt"
 	"math/rand"
 
 	"github.com/osbuild/images/internal/environment"
 	"github.com/osbuild/images/internal/workload"
 	"github.com/osbuild/images/pkg/artifact"
 	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/platform"
 	"github.com/osbuild/images/pkg/rpmmd"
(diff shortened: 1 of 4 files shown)


pkg/image/archive.go
```diff
@@ -41,6 +43,11 @@ func (img *Archive) InstantiateManifest(m *manifest.Manifest,
 
 	tarPipeline := manifest.NewTar(buildPipeline, osPipeline, "archive")
 	tarPipeline.SetFilename(img.Filename)
+
+	tarPipeline.Compression = osbuild.TarArchiveCompression(img.Compression)
+	if err := tarPipeline.Compression.Validate(); err != nil {
+		return nil, err
+	}
```
achilleas-k, line 50: This isn't entirely necessary, is it?  It'll be caught by the stage creation later anyway.  Doesn't hurt though.

achilleas-k (approved):
LGTM

supakeen:
Force pushed to drop the osbuild has commit; this is already in `main` and was causing conflicts.

PR #1584, opened 2025-06-05
distro: followups for reivew feedback in PR#1563

mvo5:
Followups for some outstanding review points from https://github.com/osbuild/images/pull/1563

---

distro: rename `ImgTypeLabel` to `ISOLabel`

This commit renames the variable `ImgTypeLabel` in the
`iso_label_tmpl` YAML to `ISOLabel` which reflects what
it actually is.

We should probably rethink/rework this a bit in any case
because even `{{.ISOLabel}}` is a bit strange compared to
doing somehting like `{{iso_label}}` which maps directly
to the YAML and hence is easier for a user to figure out.

Thanks to Achilleas for spotting this.

--- 

distro: fix double `yamllint disable`

Tiny fix for a rebase artifact that resulted in having
`# yaml disable` twice in the fedora/distro.yaml.

Thanks to Achilleas for noticing.

---

distro: use `d.Product()` consistently in the generic distro

This commit uses `d.Product()` consistently when accessing
the `Prodcut` data from the DistroYAML.

Followup for https://github.com/osbuild/images/pull/1563

Thanks to Tomáš for the suggestion.

### DIFF
--- a/pkg/distro/defs/distros.yaml
+++ b/pkg/distro/defs/distros.yaml
@@ -7,7 +7,7 @@ distros:
     module_platform_id: platform:f43
     product: "Fedora"
     ostree_ref_tmpl: "fedora/43/%s/iot"
-    iso_label_tmpl: "{{.Product}}-{{.OsVersion}}-{{.ImgTypeLabel}}-{{.Arch}}"
+    iso_label_tmpl: "{{.Product}}-{{.OsVersion}}-{{.ISOLabel}}-{{.Arch}}"
     default_fs_type: "ext4"
     defs_path: fedora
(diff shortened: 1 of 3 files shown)


supakeen (approved):

supakeen:
Failing on:

```
error: osbuild version "150" is lower than the minimum required version "151"
```

achilleas-k:
Huh, yeah interesting little edge case here.  #1581 bumped the required osbuild version but didn't need to rebuild anything, so it worked fine and got merged.  We should have a check somewhere that makes sure the pinned osbuild version in the Schutzfile is compatible with the minimum version we specify in images.

PR #1588, opened 2025-06-08
rhel8: move imageconfig into YAML

mvo5:
This PR moves the rhel8 image config into YAML.

This PR also shows some complexity with complex
image types that inherit like ec2-sap, we need to
create weird workarounds, we can hold this off
and chat how to make this cleaner. Note however
that the previous (in code) inheritance also has
their pitfalls as ImageConfig.InheritFrom() will
also do no merging but there are no conditions
there so the structure is simpler.

---

rhel8: move imageconfig into YAML

This commit moves all the image config of rhel-8 into pure
YAML.

---

rhel8: remove leftover partition loader coder

There was some leftover code that still defined the
azure-rhui partition table in code. This is all in
yaml and this commit removes it.

### DIFF
--- a/pkg/distro/defs/rhel-8/distro.yaml
+++ b/pkg/distro/defs/rhel-8/distro.yaml
@@ -138,6 +138,218 @@
             # actually added so we don't add it here either
             # - "rhc"
 
+  # Default AMI (custom image built by users) images config.
+  # The configuration does not touch the RHSM configuration at all.
+  # https://issues.redhat.com/browse/COMPOSER-2157
+  ami_image_config: &ami_image_config
+    timezone: "UTC"
+    time_synchronization:
+      servers:
+        - hostname: "169.254.169.123"
+          prefer: true
+          iburst: true
+          minpoll: 4
+          maxpoll: 4
+      # empty string will remove any occurrences of the option
+      # from the configuration
+      leapsectz: ""
+    keyboard:
+      keymap: "us"
+      "x11-keymap":
+        layouts: ["us"]
+    enabled_services:
+      - "sshd"
+      - "NetworkManager"
+      - "nm-cloud-setup.service"
+      - "nm-cloud-setup.timer"
+      - "cloud-init"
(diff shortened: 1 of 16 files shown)


pkg/distro/defs/rhel-8/distro.yaml
```diff
@@ -875,14 +1187,33 @@ image_types:
...
-    package_sets:
-      os:
-        - *ec2_common_pkgset
-
   "ec2-sap":
     <<: *ec2
+    # XXX: this is ugly, YAML only supports shallow merging of maps so
+    # we we want to merge e.g. the version conditions that sap and ec2
+    # have we need to manually "unroll" it. we need to think of a better
+    # way here.
```
thozza, line 1195: I agree that we need a better way to do this.

I think that if we could resolve the conditions before merging image configs, it would be easier. This would most probably mean having a special **key** with a list of image configs to inherit from, similar to what we used to do in the code. This would mean not doing all the merging in YAML. Trying to merge data structures with conditions makes this unnecessarily complex.

thozza (commented):
Thanks. This looks reasonably sane from my PoV.

You forgot to remove `sapImageConfig()`, which is now not used anymore. I also added some comments inline.

pkg/distro/defs/rhel-8/distro.yaml
mvo5: This becomes a lot nicer with PR#1600, c.f. https://github.com/osbuild/images/compare/main...mvo5:yaml-imageconfig-rhel8-starlark?expand=1#diff-c8017947ffd600efd01b571e33783872a8ae5eec8e7285b7563b6e567394e1ccR1074 which is this PR ported to this new mechanism (still not perfect but a lot better IMHO).

mvo5: This is now improved via pr#1608

supakeen (approved):
Cool :)

PR #1590, opened 2025-06-10
distro/rhel{8,9,10}: WSL extension and compression (HMS-8573)

supakeen:
As done earlier for Fedora newer Windows versions expect WSL images to be a tarball (potentially compressed) that has the `.wsl` extension.

---

This PR is the 'slightly controversial' part as there was some internal discussion within the team if renaming an artifact is to be considered a breaking change. I think in the end we decided it isn't (and the frontend can just serve it under a different name if it wants to).

For people who have automation and expected filenames, well this will break *eventually* unless we want to keep a 'wrong' and a 'right' image type in perpetuity under two similar names :(


bcl (approved):
Looks good to me. I consider the old name to be a bug which is now fixed FWIW :)
Looks like there's a problem with the unit tests though, it's complaining about old pkg/distro/fedora not being there.

supakeen:
> Looks good to me. I consider the old name to be a bug which is now fixed FWIW :) Looks like there's a problem with the unit tests though, it's complaining about old pkg/distro/fedora not being there.

Yea, this has been failing everywhere. It's an expected thing until `osbuild-composer` gets a PR to use the new generic paths I believe.

thozza (dismissed):
LGTM.

It may also be a good idea to move the WSL image from the `ubi.go` (if this would still be a thing after the YAMLification).

mvo5:
> Since I had to rebase anyways due to manifest checksums I've also added a commit that renames the `ubi.go` to `wsl.go` (@thozza)
> 
> This does now need new reviews :(

The ubi.go will go away when the yamlfication is done, adding wsl.go will actually create a conflcit with my in-progress branches. But its fine, conflicts are to be expected :)

mvo5 (approved):

thozza (dismissed):
Thanks for the rename, LGTM.

supakeen:
This had to be rebased again due to manifest checksums, please re-review again; @mvo5, @thozza.

mvo5 (approved):

PR #1591, opened 2025-06-11
distro/rhel-10: wsl-distribution config

supakeen:
Write (manually) the WSL configuration file that includes things to make using the produced image more user friendly. This includes a reference to an icon (in `redhat-logos`) and a friendly name.

In the future this will be replaced by the `wsl-setup` package.

There's a lot of duplication here between distributions because each one of them has a separate value for one of the keys. This might indicate that we should use a stage but I opted not to do so as this is only temporary.

---

Draft because I need to take a look at the state of templating to see if I can template in at least the version number here.


supakeen:
Created a PR to introduce a new stage in `osbuild` to write this file; after that lands I'll adjust this PR to make use of it.

supakeen:
This PR has been rewritten to use the (pending) stage addition in `osbuild`; I've adjusted image config to be able to set options for this new stage, and had to plumb through various bits and bobs to make sure they are available in archive image types. The end result is something like this:

```
+        {
+          "type": "org.osbuild.wsl-distribution.conf",
+          "options": {
+            "oobe": {
+              "default_uid": 1000,
+              "default_name": "RedHatEnterpriseLinux-10.1"
+            },
+            "shortcut": {
+              "enabled": true,
+              "icon": "/usr/share/pixmaps/fedora-logo.ico"
+            }
+          }
```

supakeen:
The stage has landed in `osbuild`, I've updated the Schutzfile here so this PR is now ready for CI and review.

pkg/distro/image_config.go
```diff
@@ -147,6 +147,26 @@ type WSLConfig struct {
...
+
+type WSLDistributionConfig struct {
+	OOBE     *WSLDistributionOOBEConfig     `yaml:"oobe,omitempty"`
+	Shortcut *WSLDistributionShortcutConfig `yaml:"shortcut,omitempty"`
+}
+
+type WSL struct {
+	Config             *WSLConfig             `yaml:"config,omitempty"`
+	DistributionConfig *WSLDistributionConfig `yaml:"distribution_config,omitempty"`
+}
```
thozza, line 168: I think that we should be moving these to `pkg/customizations` instead of adding them to the `distro` package. The same applies to the `Sysconfig` struct below, which creates the wrong precedence.

thozza (commented):
Besides my inline comment, the rest looks good.

pkg/distro/image_config.go
supakeen: By that you mean you would (eventually) want to expose them to (blueprint) users?

thozza: I understand that the directory name makes this impression. I mean, maybe, but that's not what's my intention. We started with osbuild stage options in the `ImageConfig` and wanted to add abstractions gradually, which were osbuild-agnostic. And definitions in the `pkg/customizations` are the place where we added them. These are not meant to be related to Blueprints, but in general to customizations done to images (whether via BP or in `ImageConfig`). The idea is that BP customizations get translated to these data structures, which then get merged with the `ImageConfig`.

supakeen: Ok, and do we already have the merging machinery?

supakeen: Thanks, then it's clear what I'll change :)

pkg/distro/rhel/images.go
```diff
@@ -282,10 +282,13 @@ func osCustomizations(
 	osc.WAAgentConfig = imageConfig.WAAgentConfig
 	osc.UdevRules = imageConfig.UdevRules
 	osc.GCPGuestAgentConfig = imageConfig.GCPGuestAgentConfig
-	osc.WSLConfig = imageConfig.WSLConfStageOptions()
-	osc.WSLDistributionConfig = imageConfig.WSLDistributionConfStageOptions()
 	osc.NetworkManager = imageConfig.NetworkManager
 
+	if imageConfig.WSL != nil {
+		osc.WSLConfig = osbuild.NewWSLConfStageOptions(imageConfig.WSL.Config)
+		osc.WSLDistributionConfig = osbuild.NewWSLDistributionConfStageOptions(imageConfig.WSL.DistributionConfig)
```
thozza, line 289: Nitpick: `OSCustomizations` could also use `wsl.Config` and `wsl.DistributionConfig` instead of the stage options. The conversion to stage options (or even directly to a stage) could be done when serializing the manifest, as it is done for other options.

thozza (approved):
LGTM, thank you.

kingsleyzissou (approved):
LGTM too, I agree with @thozza's nitpick but maybe could be a follow up

thozza:
> I assume this customization is not the customer-facing one, the one we want to unify.

I know this isn't very clear, but the PR has nothing to do with customizations in the context of Blueprins.

PR #1594, opened 2025-06-12
image: grow the anaconda rootfs image to 10 GiB

achilleas-k:
With newer versions of RHEL 8.10 packages, the anaconda-tree for the edge-installer has grown to be larger than 4 GiB.  This makes it impossible to build edge-installers with osbuild-composer on RHEL 8.10.

Grow the anaconda rootfs image to 10 GiB, which should be big enough to fit even future growth.  This shouldn't affect the size of the ISO since it's compressed using squashfs.

NOTE: I originally planned to do this in a quickfix branch to build unsigned RPMs in osbuild-composer with it, but @thozza suggested we release it properly in 8.10, so let's do this properly.

### DIFF
--- a/pkg/image/anaconda_ostree_installer.go
+++ b/pkg/image/anaconda_ostree_installer.go
@@ -98,7 +98,7 @@ func (img *AnacondaOSTreeInstaller) InstantiateManifest(m *manifest.Manifest,
 	isoLabel := fmt.Sprintf(img.ISOLabelTempl, img.Platform.GetArch())
 
 	rootfsImagePipeline := manifest.NewISORootfsImg(buildPipeline, anacondaPipeline)
-	rootfsImagePipeline.Size = 4 * common.GibiByte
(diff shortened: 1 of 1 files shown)


supakeen:
Why did this PR end up being a whole bunch of test changes as well? The size change looks good but I'm not sure if all the test stuff is OK?

thozza:
> Why did this PR end up being a whole bunch of test changes as well? The size change looks good but I'm not sure if all the test stuff is OK?

The PR targets `rhel-8.10` branch, so that we can make a "patch" release of `images`.

As for the tests, they need more "love"...

achilleas-k:
Had to bump the google repo snapshots because the files for the ones we had configured are gone, though the snapshot date `20240101` appears in the the snapshot list.

I also added repo configs for CS9 to the `Schutzfile` to include the CRB repo, which we need for `gpgme-devel`.

supakeen (dismissed):

supakeen (approved):

supakeen (approved):

PR #1598, opened 2025-06-15
distro: handle `centos-9` and similar versions better

mvo5:
This commit adds a special `versionStringForVerCmp()` that s a special version string for our version
compare that will assume that any version with no minor is automatically higher than any compare with a minor version.

The rational is that "centos-9" is always higher than any "rhel-9.X" version for our version compare (centos is always "rolling").

TODO: this should become an explicit chose in "distro.yaml" but until we have everything converted to generic.Distro accessing the properites from an image type is very hard so we start here.

Note that this automatically fixes some issues in our current image defintions:
1. we added "system-reinstall-bootc" to the cloud images but this change did not propagate to the centos-9 cloud image
2. We exlucde "microcode_ctl" in vhd for rhel >= 9.6 but this never propagated to centos-9
3. we remove "dnsmasq" from edge-commit for rhel < 9.6 but this never propagated to centos-9


pkg/distro/defs/loader_test.go
```diff
@@ -533,10 +533,14 @@ image_types:
       hostname: "foo"
       locale: "C.UTF-8"
       timezone: "DefaultTZ"
+      default_kernel_name: kernel
       condition:
         version_less_than:
           "2":
             timezone: "OverrideTZ"
+          # test-distro is version "1" (no minor) so considered "1" is > "1.4"
```
supakeen, line 541: I had to read this twice before understanding what it meant, but it is indeed correct.

supakeen (approved):
I'll expect that once this gets applied to rhel-10 that almalinux_kitten-10 will also work this way since it also lacks a number?

thozza (approved):
LGTM, thanks.

supakeen:
Needs checksum collision fixes.

mvo5:
> Needs checksum collision fixes.

Thanks! Fixed the conflict and pushed again.

supakeen (approved):

PR #1601, opened 2025-06-17
distro/rhel-10: new `vagrant-libvirt` image type (HMS-6116)

supakeen:
This image type was previously updated and improved for Fedora. I wanted to hold out for the full YAML changes but have now decided to land these faster.

Image was manually tested, might need refinement in followup(s) it is currently similar to the QCOW2 image type.

### DIFF
--- a/pkg/distro/defs/rhel-10/distro.yaml
+++ b/pkg/distro/defs/rhel-10/distro.yaml
@@ -554,7 +554,7 @@ image_types:
                   - "subscription-manager-cockpit"
 
   qcow2: &qcow2
-    image_config:
+    image_config: &qcow2_image_config
       default_target: "multi-user.target"
       kernel_options: ["console=tty0", "console=ttyS0,115200n8", "no_timer_check"]
       condition:
@@ -571,7 +571,8 @@ image_types:
       <<: *default_partition_tables
     package_sets:
       os:
-        - include:
+        - &qcow2_pkgset
+          include:
             - "@core"
             - "chrony"
             - "cloud-init"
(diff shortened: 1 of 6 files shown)


pkg/distro/defs/rhel-10/distro.yaml
```diff
@@ -634,6 +635,29 @@ image_types:
...
+            ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIN1YdxBpNlzxDqfJyw/QKow1F+wvG9hXGoqiysfJOn5Y vagrant insecure public key
+          # yamllint enable rule:line-length
+      files:
+        - path: "/etc/sudoers.d/vagrant"
+          user: "root"
+          group: "root"
+          mode: 440
+          data: |
+            vagrant ALL=(ALL) NOPASSWD: ALL
+    partition_table:
```
mvo5, line 655: One tiny suggestion to avoid having to add this:
```yaml
diff --git a/pkg/distro/defs/rhel-10/distro.yaml b/pkg/distro/defs/rhel-10/distro.yaml
index 4f8c149f8..a44b01d99 100644
--- a/pkg/distro/defs/rhel-10/distro.yaml
+++ b/pkg/distro/defs/rhel-10/distro.yaml
@@ -636,6 +636,7 @@ image_types:
                   - "subscription-manager-cockpit"
 
   "vagrant-libvirt":
+    <<: *qcow2
     image_config:
       <<: *qcow2_image_config
       users:
@@ -652,11 +653,6 @@ image_types:
           mode: 440
           data: |
             vagrant ALL=(ALL) NOPASSWD: ALL
-    partition_table:
-      <<: *default_partition_tables
-    package_sets:
-      os:
-        - *qcow2_pkgset
 
   oci: *qcow2
```
this will help with the full-yaml as it will "inherit" all the pipeline then too (once we move to full YAML for rhel-10).



(could be a followup of course too, I don't mind).

mvo5 (commented):
Looks good, one tiny suggestion inline.

pkg/distro/defs/rhel-10/distro.yaml
supakeen: Ah, I *think* this is what I wanted to do but got confused if the merging would happen correctly; let me try.

supakeen: Perfect, works fine and no checksums changed :)

pkg/distro/defs/rhel-10/distro.yaml
```diff
@@ -571,7 +571,8 @@ image_types:
       <<: *default_partition_tables
     package_sets:
       os:
-        - include:
+        - &qcow2_pkgset
+          include:
```
thozza, line 575: This change does not seem to be needed, because `qcow2_pkgset` is not used anywhere AFAICT.

thozza (commented):
CI is failing, and there is one unneeded change; otherwise, LGTM.

pkg/distro/defs/rhel-10/distro.yaml
supakeen: Thanks! That was a leftover from a previous approach, I've removed it.

thozza (approved):

PR #1604, opened 2025-06-19
isos: provide 32-bit EFI firmware

supakeen:
For (at least) Fedora it is wanted to be able to boot 64-bit systems that have 32-bit EFI firmware; see [1].

[1]: https://fedoraproject.org/wiki/Changes/32BitUefiSupport

---

See also:

1. https://pagure.io/fedora-kiwi-descriptions/pull-request/193
2. https://github.com/OSInside/kiwi/issues/2822

This probably needs careful QA.

Thanks to https://github.com/Conan-Kudo for the report.


Conan-Kudo:
Why not just do the things if and only if the EFI binaries are installed in the installroot? Distro-specific stuff should be unnecessary here.

supakeen:
> Why not just do the things if and only if the EFI binaries are installed in the installroot? Distro-specific stuff should be unnecessary here.

I don't know, that just sounds like another way of doing distro specific stuff except you'd do it by listing a package in a distros build root instead of an option. Do you normally get these packages into your build root through a comps group?

Does the contents of the `efiboot.img` in the above comment look right to you?

Conan-Kudo:
The contents look right to me. 👍🏾

Conan-Kudo:
> I don't know, that just sounds like another way of doing distro specific stuff except you'd do it by listing a package in a distros build root instead of an option. Do you normally get these packages into your build root through a comps group?

I don't think so. What I mean is that since osbuild supports multiple distributions, the correct logic is to only copy over things if they exist, and just ignore it if they don't. That way it doesn't matter how your package manifest needs to be structured.

lzap (approved):

Conan-Kudo (approved):

bcl:
Just noticed this. FYI I've got a fix for this that uses the platforms yaml instead of hard-coded go. It's part of my netinst PR which is still in progress.

bcl:
Actually, ignore me for now :) I've just realized that none of the RHEL yaml has been updated so my platform change needs to wait for a bit more progress. Carry on :)

bcl:
Note that the latest build of shim has dropped the ia32 support completely. See https://github.com/weldr/lorax/pull/1489

PR #1607, opened 2025-06-23
Many: delete repositories for EOL distributions (HMS-7044)

thozza:
- Delete F40
- Delete odd RHEL-8.Y releases. Only even releases are supported (via EUS, which is a different story, but this is a good start).
- Delete odd RHEL-9.Y releases that are not the latest GA YStream or in development. For the older even RHEL-9.Y releases, the situation is the same as for RHEL-8 (as noted in the EUS).


supakeen (approved):

thozza:
I forgot to regenerate the `.gitlab-ci.yml` after deleting the repositories. This is now fixed.

achilleas-k (commented):
Yes, thank you!!  This is going to have a lot of nice side-effects: Smaller CI cache, fewer flakes from runner exhaustion, a lot less CI work in general.

Needs a test fix though.

achilleas-k:
Should we also drop version conditions and `IsRHEL()` checks from the RHEL 8 image definitions?

thozza:
> There's a unit test that relies on RHEL 9.1 being in the distro registry.

Thanks, I missed that. It is fixed now.

supakeen (dismissed):

ondrejbudai (dismissed):

supakeen (approved):

mvo5 (approved):

supakeen (approved):

PR #1608, opened 2025-06-23
distro: allow AND and easier merging in conditions (HMS-8593)

mvo5:
This commit tweaks the condition handling in the YAMLs so that we can make merging easier on the YAML level.

This means that each condition now needs a unique description but that is probably good anyway to describe the reason for the condition.

The rhel-9 azure image type needs a logical AND for its image config. It checks for example that the arch is x86_64 and the rhel version >= 9.6.

This extends `when:` to allow multiple keys and when multiple keys are used they are put together as AND:

```yaml
image_types:
  "workstation-live-installer":
    ...
    image_config:
      locale: "en_US.UTF-8"
      iso_rootfs_type: "squashfs"
      conditions:
        "f40 and below uses ext4 based iso rootfs":
          when:
            version_less_than: "41"
            distro_name: "fedora"
          merge:
            iso_rootfs_type: "squashfs-ext4"
```
With that we can express all our intend nicely and we can also express what a "when" is doing, i.e. for some (like package sets) it will append, for others (image-types) it will merge and for installer-config it will override. All of this is now easier to read via:
```yaml
when: <expr>
override: ...
```

Alternative version to https://github.com/osbuild/images/pull/1600

### DIFF
--- a/pkg/distro/defs/fedora/distro.yaml
+++ b/pkg/distro/defs/fedora/distro.yaml
@@ -167,15 +167,18 @@
     additional_dracut_modules:
       - "net-lib"
     squashfs_rootfs: true
-    condition:
-      version_less_than:
-        "42":
-          # config is fully replaced
+    conditions:
+      "f41 uses ifcfg in dract but already a squashfs rootfs":
+        when:
+          version_equal: "41"
+        override:
           additional_dracut_modules:
             - "ifcfg"
           squashfs_rootfs: true
-        "41":
-          # config is fully replaced
+      "f40 and lower uses ifcfg in dracut and no squashfs rootfs":
+        when:
+          version_less_than: "41"
+        override:
           additional_dracut_modules:
             - "ifcfg"
           squashfs_rootfs: false
@@ -195,9 +198,11 @@
         - "redboot-auto-reboot"
         - "redboot-task-runner"
       kernel_options: *ostree_deployment_kernel_options
-      condition:
-        version_less_than:
-          "42":
+      conditions:
+        "f41 and below used zezere and parsec":
+          when:
+            version_less_than: "42"
+          merge:
             enabled_services:
               - "NetworkManager.service"
               - "firewalld.service"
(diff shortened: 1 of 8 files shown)


lzap:
Now that I am looking on it, these are not boolean conditions but rather statements. I wonder if it should be maybe (probably with the wrong indentation but you get the idea):

```yaml
image_types:
  "workstation-live-installer":
    ...
    image_config:
      locale: "en_US.UTF-8"
      iso_rootfs_type: "squashfs"
      conditions:
        - when:
            version_less_than: "41"
            distro_name: "fedora"
          merge:
            iso_rootfs_type: "squashfs-ext4"
```

mvo5:
[..]
> I do not see much value of using dictionary keys as comments because YAML itself support comments and this allows us to easily implement "OR" condition if we want to. I mean, there is a way forward even in your proposal but this looks more readable to me.

Please see the commit description , the reason is that this allows us to merge multiple conditions with pure YAML, e.g.
```yaml
image_types:
  "workstation...":
    ...
    image_config:
      conditions:
        <<: *conditions_from_other_image_type
        "f40 and below uses ext4 based iso rootfs":
       ...
```
(real example in https://github.com/osbuild/images/compare/main...mvo5:yaml-imageconfig-rhel8-starlark?expand=1#diff-9513fd81fd761032ce3eb6ae5901dde595f0aba41f041ba7a33d7dad0c0e58deR253)

Because there is no (pure) yaml merging of sequences it has to be a map.

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -167,15 +167,18 @@
     additional_dracut_modules:
       - "net-lib"
     squashfs_rootfs: true
-    condition:
-      version_less_than:
-        "42":
-          # config is fully replaced
+    conditions:
+      "f41 uses ifcfg in dract but already a squashfs rootfs":
```
bcl, line 171: 'but already' is confusing. I'd word it like "f41 dracut requires ifcfg, rootfs is squashfs"

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -167,15 +167,18 @@
...
+        when:
+          version_equal: "41"
+        override:
           additional_dracut_modules:
             - "ifcfg"
           squashfs_rootfs: true
-        "41":
-          # config is fully replaced
+      "f40 and lower uses ifcfg in dracut and no squashfs rootfs":
```
bcl, line 178: 'no squashfs rootfs' is misleading. 'squashfs+ext4 rootfs' is really what is created.

pkg/distro/defs/fedora/distro.yaml
```diff
@@ -822,26 +827,39 @@ image_types:
...
-              "43":
+            "f42 and below uses basesystem":
+              when:
+                version_less_than: "43"
+              append:
                 include:
                   - "basesystem"
-            version_greater_or_equal:
-              "41":
+            "f41+ needs bootupd":
```
bcl, line 853: Maybe consistently say above and below instead of using below and +?
eg. "f41 and above need bootupd"

bcl (approved):
Overall I like this much better than the other proposal. Just had a few comments.

pkg/distro/defs/fedora/distro.yaml
mvo5: I will address those in a followup

PR #1612, opened 2025-06-24
distro/rhel: introduce new image type: Azure SAP Apps

achilleas-k:
The SAP Applications variant of the Azure SAP image type should not include the rhel-system-roles-sap and tuned-profiles-sap-hana packages. Introduce a new image type that's identical to the azure-sap-rhui only without the two packages and call it azure-sapapps-rhui.

At the package set level, the `sap_pkgset` has been split into two sets now: `sap_base_pkgset` and `sap_extras_pkgset`.  The extras set includes the two packages that have been removed from the SAP Applications variant.

HMS-8738


mvo5 (dismissed):

supakeen (approved):

achilleas-k:
Forgot to remove the tuned profile from RHEL 10.  Fixed now.

supakeen (approved):

pkg/distro/defs/rhel-10/distro.yaml
```diff
@@ -94,14 +92,16 @@
...
+      profiles: ["sap-hana"]
+
+  sap_base_pkgset: &sap_base_pkgset
     include:
       # RHBZ#2076763
       - "@Server"
-      # SAP System Roles
-      # https:#access.redhat.com/sites/default/files/attachments/rhel_system_roles_for_sap_1.pdf
       - "ansible-core"
-      - "rhel-system-roles-sap"
```
thozza, line 104: Nitpick: `ansible-core` is there IMHO only due to the RHEL system roles, which are Ansible roles. Not sure how useful it is for the SAP Apps image...

thozza (approved):
LGTM. I've added one nitpick comment, but the PR is good as is and the comment can be addressed later if needed.

pkg/distro/defs/rhel-10/distro.yaml
achilleas-k: I realised this from the grouping under the comment, but decided to do the bare minimum change in the end.  Since you brought it up too, I think I'm going to open a follow-up PR and move `ansible-core` to the extras as well.

PR #1614, opened 2025-06-25
rhel9: enable latest azure changes on centos-9 too

mvo5:
This commit tweaks the conditions in defaultAzureKernelOptions and defaultAzureImageConfig so that when we check if a version is greater than 9.6 we also check if its centos. Centos versions are always just "9" so any greater-than version compare will yield false but for centos we want the same behavior as on the latest rhel.

This is not ideal for almakitten but this problem fixes itself once we move to image-config in YAML (c.f. https://github.com/osbuild/images/pull/1616) where the minor verison is automatically taken into account in the version compare (i.e. 9-stream is > 9.x).


achilleas-k (approved):
Nice, thanks!

PR #1620, opened 2025-06-25
distro/rhel9: xz compress azure-cvm image type [HMS-8587]

achilleas-k:
All other Azure image types are xz compressed because the disk size is very big and mostly empty.  Let's compress the CVM image for convenience and consistency.


supakeen (approved):
For WSL we had a long discussion on if changing the filename consists of breaking the API. Let's skip that here :>

achilleas-k:
> For WSL we had a long discussion on if changing the filename consists of breaking the API. Let's skip that here :>

yeah the CVM image just got merged into composer a few days ago.  I don't think we even made a release with it yet.

thozza (approved):

PR #1624, opened 2025-06-27
distro/rhel: move ansible-core to sap_extras_pkgset

achilleas-k:
Move ansible-core to the SAP extras package set.  This removes ansible-core from the SAP Apps Azure image type.  The package is included in the SAP images for the rhel-system-roles-sap package, so when we moved the system roles package out of the SAP apps image type in #1612, we should have moved ansible-core too.


thozza:
@achilleas-k needs rebasing

achilleas-k:
Done!

thozza (approved):
Perfect!

supakeen (approved):

achilleas-k:
Ok, ready now.

bcl (approved):
Change looks ok to me. But I wonder why rhel-system-roles-sap doesn't depend on ansible-core. Is it always used with it like the document describes, or is it optional?

PR #1633, opened 2025-06-30
many: implement gzip

supakeen:
`osbuild` has an `org.osbuild.gzip` stage that we never exposed in `images`. This is a small PR to add the relevant bits so it can be used; it also exposes it in `image/disk.go` even though nothing uses it yet.

Followup PRs will generalize the compression between `disk.go`, `ostree_disk.go`, and `archive.go`. After that we will likely generalize even further to a `compress.go` pipeline.

---

Note: Level is not directly exposed right now, do we want to? We don't have it available in the `xz` and `zstd` stages at all; the `gzip` stage however *does* have the option in its osbuild stage.

### DIFF
--- a/pkg/image/disk.go
+++ b/pkg/image/disk.go
@@ -131,6 +131,10 @@ func (img *DiskImage) InstantiateManifest(m *manifest.Manifest,
 		zstdPipeline := manifest.NewZstd(buildPipeline, imagePipeline)
 		zstdPipeline.SetFilename(img.Filename)
 		return zstdPipeline.Export(), nil
+	case "gzip":
+		gzipPipeline := manifest.NewGzip(buildPipeline, imagePipeline)
+		gzipPipeline.SetFilename(img.Filename)
+		return gzipPipeline.Export(), nil
 	case "":
 		// don't compress, but make sure the pipeline's filename is set
 		imagePipeline.SetFilename(img.Filename)


pkg/osbuild/gzip_stage.go
```diff
@@ -0,0 +1,41 @@
+package osbuild
+
+type GzipStageOptions struct {
+	// Filename for gz archive
+	Filename string `json:"filename"`
+	Level    *int   `json:"level,omitempty"`
```
mvo5, line 6: AFAICT this is not used or tested, maybe we can remove it and add it when we need it (it seems xz, zstd also do not expose the compression level?). Unless of course we will use it soon.

pkg/osbuild/gzip_stage_test.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func TestNewGzipStage(t *testing.T) {
+	inputFilename := "image.raw"
+	filename := "image.raw.gz"
+	pipeline := "os"
+
+	expectedStage := &Stage{
+		Type:    "org.osbuild.gzip",
+		Options: NewGzipStageOptions(filename),
```
mvo5, line 27: Using Options: &GzipStageOptions{Filename: filename}, here feels slightly better (its a bit of tradeoff of course). The reason is that if we (accidentally) change NewGzipStageOptions to e.g. switch to a default compression level of say "5" with this test setup no test would fail but with the explicit GzipStageOptions{} we would get an error.

mvo5 (commented):
Thank you! Very nice! Two tiny suggestions inline

pkg/osbuild/gzip_stage_test.go
supakeen: Mmm, this was purely for consistency with the others; I think we'll probably end up dropping the compression level.

pkg/osbuild/gzip_stage.go
mvo5: If we don't use it and don't have plans to use it I would be (strongly) in favor of YAGNI and just not do it (for now).

pkg/osbuild/gzip_stage_test.go
mvo5: No need to change anything here, I can make a note to myself to do a followup if needs changes in more places. But I still wanted to mention it because it highlights a general principle: testing the result of  NewGzipStageOptions() with the result of NewGzipStageOptions() in the test will test less than using the &GzipStageOptions{} directly here (unless of course we already check elsewhere that NewGzipStageOptions() has the expected result). But please ignore for now, this is also very theoretical because the NewGzipOptions() thing is so small.

mvo5 (dismissed):

pkg/manifest/gzip.go
```diff
@@ -0,0 +1,55 @@
+package manifest
+
+import (
+	"github.com/osbuild/images/pkg/artifact"
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+// The Gzip pipeline compresses a raw image file using zstd.
```
bcl, line 8: typo - 'file using gzip' :)

pkg/osbuild/gzip_stage.go
```diff
@@ -3,7 +3,6 @@ package osbuild
 type GzipStageOptions struct {
 	// Filename for gz archive
 	Filename string `json:"filename"`
-	Level    *int   `json:"level,omitempty"`
```
bcl: Looks like this was sqashed into the wrong commit

bcl (dismissed):
Looks good, except for those two minor things.

pkg/manifest/gzip.go
supakeen: Done.

pkg/manifest/gzip_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+func TestGzipSerialize(t *testing.T) {
+	mani := manifest.New()
+	runner := &runner.Linux{}
+	build := manifest.NewBuild(&mani, runner, nil, nil)
+
+	// setup
+	rawImage := manifest.NewRawImage(build, nil)
+	gzipPipeline := manifest.NewGzip(build, rawImage)
+	gzipPipeline.SetFilename("filename.gz")
```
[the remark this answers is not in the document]
supakeen: It's probably because setting the filename isn't always required (like when you directly consume it into the next pipeline it doesn't matter).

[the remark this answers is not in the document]
thozza: Sure, but that's another thing I don't like - pipelines not being explicit about the filenames they produce and the code making assumptions about filenames that are hardcoded in the parent pipeline... Moving away from this practice is a prerequisite if we ever end up allowing one to combine pipeline generators in YAML, not just in the code.

mvo5 (approved):

PR #1634, opened 2025-07-01
many: make compression more generic

supakeen:
After a chat with @mvo5 yesterday about there being a lot of duplication in regards to compression and even mismatches between implementations of the switch statements I've done the *first* followup to #1633.

This adds a single function that generates the wanted compression pipeline based on `img.Compression`. We could, in the future, perhaps go even further by making a `compress` pipeline instead; but I feel like this is a nice first step.

---

Draft because it depends on #1633. If you go commit-by-commit there are two 'step-back' commits where I unify the switch-statements *before* moving them all into a single function.

Only the WSL checksums change here, `img.Compression` isn't used much but it was being used in the tar pipelines; which have now changed from passing a stage option to `org.osbuild.tar` to having an additional pipeline to compress them.


pkg/image/image.go
```diff
@@ -27,3 +28,19 @@ func NewBase(name string) Base {
...
+		return manifest.NewXZ(buildPipeline, inputPipeline)
+	case "zstd":
+		return manifest.NewZstd(buildPipeline, inputPipeline)
+	case "gzip":
+		return manifest.NewGzip(buildPipeline, inputPipeline)
+	case "":
+		return inputPipeline
+	default:
+		// panic on unknown strings
+		panic(fmt.Sprintf("unsupported compression type %q", compression))
```
mvo5, line 44: (super nitpick) make this a `fmt.Errorf()` would prepare us already for a future with less panics :)

supakeen:
Rebased since #1635 also changed WSL manifests.

bcl (approved):
Looks good, great job going backwards before going forward!

PR #1636, opened 2025-07-01
distro: remove some hardcoded values from generic/images.go

mvo5:
distro: drop hardcoded presets in `iot-commit`

The `iotCommitImage` function in the generic images.go
code hardcodes some systemd presets.

This will become a problem when we move to rhel-8,9
to the generic distro. Here their "edge" images do
not use these presets. So this commit moves them into
the YAML image config instead of hardcoding them.

---

distro: drop hardcoded dracut modules from
 `iotInstallerImage`

The current generic `images.go:iotInstallerImage()` has
some hardcoded anaconda modules encoded. This will be
problematic when we switch other distros (like rhel-{8,9})
to the generic distro. Those are not using them.

So instead of hardcoding them make them part of the
`InstallConfig` and add them to the fedora distro.

### DIFF
--- a/pkg/distro/defs/fedora/distro.yaml
+++ b/pkg/distro/defs/fedora/distro.yaml
@@ -766,6 +766,14 @@ image_types:
           config:
             add_dracutmodules:
               - "fips"
+      # see https://github.com/ostreedev/ostree/issues/2840
+      presets:
+        - name: "ignition-firstboot-complete.service"
+          state: "enable"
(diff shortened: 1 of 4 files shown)


bcl (approved):
Looks good!

mvo5:
Sorry, had to force push a trivial whitespace fix (extra space in the yaml tag)

thozza (approved):

supakeen (approved):

PR #1639, opened 2025-07-02
distro: exclude container relative packages in ami

liangxiao1:
JIRA: https://issues.redhat.com/browse/RHEL-96793

Other type of images do not have container packages installed or exclude in distro files. But RHEL-10.0 and RHEL-9.6 updates AMIs have more packages than GA version which is not expected change. So exclude them in this change.

### DIFF
--- a/pkg/distro/defs/rhel-10/distro.yaml
+++ b/pkg/distro/defs/rhel-10/distro.yaml
@@ -1121,6 +1121,23 @@ image_types:
             - "dracut-config-rescue"
             # RHBZ#2075815
             - "qemu-guest-agent"
+            # RHEL-96793
+            - "conmon"
+            - "containers-common"
+            - "containers-common-extra"
+            - "container-selinux"
+            - "criu"
+            - "criu-libs"
+            - "crun"
+            - "libnet"
+            - "libnftnl"
+            - "netavark"
+            - "nftables"
+            - "passt"
+            - "passt-selinux"
+            - "podman"
+            - "protobuf-c"
+            - "shadow-utils-subid"
           conditions:
             <<: *conditions_pkgsets_insights_client_on_rhel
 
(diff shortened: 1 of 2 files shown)


supakeen:
> > It's likely that `podman` became a dependency of one of the selected packages. Do we know which one?
> 
> Exactly. Before we modify the package set, let's conduct some detective work and figure out what changed. Based on my brief research, I don't think we changed anything in the package set on our end, and there does not appear to be any change in the CS or RHEL comps groups that would add podman as a dependency to the `@core` group.
> 
> As Simon mentioned, the most probable change is that some package added `podman` as a weak dependency, which are by default installed for the image base package set.

Looking at the first (likely) culprit: `system-reinstall-bootc`:

```# dnf rq --recommends system-reinstall-bootc
Updating Subscription Management repositories.
subscription-manager is operating in container mode.
Last metadata expiration check: 0:00:04 ago on Wed Jul  2 08:13:26 2025.
podman
```

thozza:
I'd say that the appearance of these packages is not really "unexpected", since they are dependencies of other packages. The only difference is that RHEL 9.6/10.0 pungi configs add `"system-reinstall-bootc"` to the image as a customization, which doesn't pull in weak dependencies. However, that was AFAIK just a "quick fix", until the package was added to the image base package set. So, while I understand the sentiment that 9.6/10.0 update images contain more packages than the GA ones, due to the packages being added to the image base package set, this is not necessarily an issue requiring us to exclude any of them.

thozza:
> Thanks, @thozza agree that it is not necessary to be excluded. I am going to close this pr and copy your comments to close the jira ticket too. If any customer raise such concern about AMI, we can discuss whether exclude some of them again.

Agreed. FWIW, thank you for doing the initial work.

PR #1642, opened 2025-07-02
[RFC] image: drop `InstallWeakDeps` from image.DiskImage

mvo5:
This commit drops `InstallWeakDeps` from `image.DiskImage`. We already define this in `manifest.OSCustomizations` and in `distro.ImageConfig` and this var is no longer used (and actually confusing given that we have these two other places).

(adding RFC as their might be something I'm missing but I don't see what)


bcl (approved):
Looks good

thozza (approved):

PR #1641, opened 2025-07-02
manifest: handle content template name with spaces

xbhouse:
Fixes rhc command to support registering to content templates with or without spaces in the template name

Jira: [HMS-8657](https://issues.redhat.com/browse/HMS-8657)

PR #1646, opened 2025-07-03
manifest: use shutil.Quote() when building command strings

mvo5:
[I think this is ready now, the only open question is if/how this part of the code is currently (end-to-end) tested before I move this out of draft]

Followup for https://github.com/osbuild/images/pull/1641
as that got me thinking:

When building the strings for the subscription systemd
unit we need to be a bit more careful [0] with the quoting
because e.g. template names could contain strings like
```
Robert"); DROP TABLE
```
which would build into:
```
rhcConnect += fmt.Sprintf(" --content-template=\"%s\"", subscriptionOptions.TemplateName)
-> --content-template="Robert"); DROP TABLE"
```
which of course would mean that the command can no longer
run.

Note that this is only a (slightly theoretical) robustness issue,
the command injection is not a problem as the API already
allows adding arbitrary users to the image and this systemd
job runs on firstboot only. Still worth fixing if strange
names are used.

[0] AFAICT we don't do input validation for e.g. template_name,
c.f.
https://github.com/osbuild/osbuild-composer/blob/v144/internal/cloudapi/v2/openapi.v2.yml#L2295

---

internal: add `common.ShlexQuote()`

This commit adds a `common.ShlexQuote()` helper that can be
used to quote shell commands in the same way as pythons
`shlex.quote()`

### DIFF
--- a/pkg/manifest/os_test.go
+++ b/pkg/manifest/os_test.go
@@ -86,7 +86,7 @@ func TestSubscriptionManagerCommands(t *testing.T) {
 	}
 	pipeline := os.Serialize()
 	CheckSystemdStageOptions(t, pipeline.Stages, []string{
-		"/usr/sbin/subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY} --serverurl 'subscription.rhsm.redhat.com' --baseurl 'http://cdn.redhat.com/'",
+		`/usr/sbin/subscription-manager register --org="${ORG_ID}" --activationkey="${ACTIVATION_KEY}" --serverurl 'subscription.rhsm.redhat.com' --baseurl 'http://cdn.redhat.com/'`,
 	})
 }
 
@@ -101,7 +101,7 @@ func TestSubscriptionManagerInsightsCommands(t *testing.T) {
 	}
 	pipeline := os.Serialize()
 	CheckSystemdStageOptions(t, pipeline.Stages, []string{
-		"/usr/sbin/subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY} --serverurl 'subscription.rhsm.redhat.com' --baseurl 'http://cdn.redhat.com/'",
+		`/usr/sbin/subscription-manager register --org="${ORG_ID}" --activationkey="${ACTIVATION_KEY}" --serverurl 'subscription.rhsm.redhat.com' --baseurl 'http://cdn.redhat.com/'`,
 		"/usr/bin/insights-client --register",
 	})
 }
@@ -119,7 +119,7 @@ func TestRhcInsightsCommands(t *testing.T) {
 	os.OSCustomizations.PermissiveRHC = common.ToPtr(true)
 	pipeline := os.Serialize()
 	CheckSystemdStageOptions(t, pipeline.Stages, []string{
-		"/usr/bin/rhc connect --organization=${ORG_ID} --activation-key=${ACTIVATION_KEY} --server 'subscription.rhsm.redhat.com'",
+		`/usr/bin/rhc connect --organization="${ORG_ID}" --activation-key="${ACTIVATION_KEY}" --server 'subscription.rhsm.redhat.com'`,
 		"/usr/sbin/semanage permissive --add rhcd_t",
 	})
 }
--- a/pkg/manifest/subscription.go
+++ b/pkg/manifest/subscription.go
@@ -127,7 +127,7 @@ func subscriptionService(
 	if subscriptionOptions.Rhc {
 		var curlToAssociateSystem string
 		// Use rhc for registration instead of subscription manager
-		rhcConnect := fmt.Sprintf("/usr/bin/rhc connect --organization=${ORG_ID} --activation-key=${ACTIVATION_KEY} --server %s", shutil.Quote(subscriptionOptions.ServerUrl))
+		rhcConnect := fmt.Sprintf(`/usr/bin/rhc connect --organization="${ORG_ID}" --activation-key="${ACTIVATION_KEY}" --server %s`, shutil.Quote(subscriptionOptions.ServerUrl))
 		if subscriptionOptions.TemplateUUID != "" {
 			curlToAssociateSystem = getCurlToAssociateSystem(subscriptionOptions)
 		} else if subscriptionOptions.TemplateName != "" {
@@ -152,7 +152,7 @@ func subscriptionService(
 			files = append(files, icFile)
 		}
 	} else {
-		commands = []string{fmt.Sprintf("/usr/sbin/subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY} --serverurl %s --baseurl %s", shutil.Quote(subscriptionOptions.ServerUrl), shutil.Quote(subscriptionOptions.BaseUrl))}
+		commands = []string{fmt.Sprintf(`/usr/sbin/subscription-manager register --org="${ORG_ID}" --activationkey="${ACTIVATION_KEY}" --serverurl %s --baseurl %s`, shutil.Quote(subscriptionOptions.ServerUrl), shutil.Quote(subscriptionOptions.BaseUrl))}
 
 		// Insights is optional when using subscription-manager
 		if subscriptionOptions.Insights {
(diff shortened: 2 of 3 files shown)


achilleas-k (dismissed):
Draft approval.
LGTM :)

pkg/manifest/os_test.go
```diff
@@ -84,7 +86,7 @@ func TestSubscriptionManagerCommands(t *testing.T) {
 	}
 	pipeline := os.Serialize()
 	CheckSystemdStageOptions(t, pipeline.Stages, []string{
-		"/usr/sbin/subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY} --serverurl subscription.rhsm.redhat.com --baseurl http://cdn.redhat.com/",
+		"/usr/sbin/subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY} --serverurl 'subscription.rhsm.redhat.com' --baseurl 'http://cdn.redhat.com/'",
```
achilleas-k, line 89: Should we be quoting the env vars (`"${ORG_ID}"`) for the same reason?

achilleas-k (approved):

internal/common/shlex_quote.go
```diff
@@ -0,0 +1,13 @@
+package common
+
+import (
+	"strings"
+)
+
+// ShlexQuote implements a variant of pythons shlex.quote()
+// in go (c.f. https://github.com/python/cpython/blob/3.14/Lib/shlex.py#L320)
```
supakeen, line 8: Let's be clear here what the variation is; we're not checking the table of 'safe characters' and that's the only deviation right?

supakeen (changes requested):
Actually no, I had a hunch and found a second issue. When using `ShlexQuote` we should *not* double quote around its returned value (this is happening in quite a bit of places right now); see the difference.

Let's remove wrapping quotes everywhere?

```
user@muja ~ € echo '$(id)'      
$(id)
user@muja ~ € echo "'$(id)'"
'uid=1000(user) gid=1000(user) groups=1000(user),10(wheel),18(dialout),135(mock) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023'
```

mvo5:
> The approach here is fine 'enough' (I've played with it a bit) but like all escaping its use is conditional. Be aware that it's still possible with this style of quoting to pass certain arguments to commands that might not be expected. This is dependent on the command being given arguments but (for example) `ssh` takes a `-oLocalCommand=...` (if enabled), so do various other tools.

Can you describe via an example how ShlexQuote() would misbehave for LocalCommand? Is it because the argument is passed as a single argv[n] to the other side and then the shell tokenizes it there? Would `-oLocalCommand=''single argv''`  work in this case, i.e. will ssh honor the inner `'` ?

supakeen:
> > The approach here is fine 'enough' (I've played with it a bit) but like all escaping its use is conditional. Be aware that it's still possible with this style of quoting to pass certain arguments to commands that might not be expected. This is dependent on the command being given arguments but (for example) `ssh` takes a `-oLocalCommand=...` (if enabled), so do various other tools.
> 
> Can you describe via an example how ShlexQuote() would misbehave for LocalCommand? Is it because the argument is passed as a single argv[n] to the other side and then the shell tokenizes it there? Would `-oLocalCommand=''single argv''` work in this case, i.e. will ssh honor the inner `'` ?

You can't really do anything about this; it's up to the command how it parses its args (after it receives them from the shell). Just nice to note that some programs might do their own argument parsing and thus be vulnerable to (some form of) injection, even when its args were correctly shell-quoted :)

internal/common/shlex_quote.go
mvo5: I can expand this, the idea is really to just drop the optimization in the python code to skip qutoting if only "safe" chars are used. We could do that too but I don't (currently?) see why.

lzap:
Maybe a wild idea, but I am not feeling comfortable relying on escaping and then perhaps feeling more secure skipping proper input sanitization. Wild idea: could this escaping code issue a warning into logs or somewhat warn users and us about escape sequence being applied so then we can go ahead and properly sanitize inputs?

achilleas-k:
> could this escaping code issue a warning into logs or somewhat warn users and us about escape sequence being applied so then we can go ahead and properly sanitize inputs?

What's the benefit of doing this over what this PR is doing?

supakeen:
I'm sorry but sanitisation is not the answer to this sort of thing for me.

You can't sanitize because you don't yet know the context a value is being used it at that point in time; you can escape when you know the context a value is being used in. 

What's safe in a shell context is different from what's safe in a HTML context and while they might partially overlap they are also mutually exclusive. You also can't sanitize for the very small subset of 'safe in all possible contexts'.

Sanitization happens non-local which means the programmer has to look up the stack to see if a value is safe for the particular context they're going to be using it in. Having an explicit `shutil.Quote` at the place where the danger is, is nice.

Doing both adds nothing, escaping is enough.

lzap:
> You can't sanitize because you don't yet know the context

I absolutely can. We have a function that takes UUID as **string** and slaps that into a shell script. This is just plain wrong, no matter if you sanitize down the stack or not. How can I tell if data gets or do not get escaped down the stack? We might add a feature and forget to add escaping too. Can you say for sure 100% of all stages are escaped after this PR is merged?

I am not suggesting to get rid of sanitization, again, I see it as a last line of defense. What I am advocating for is defensive code that takes care of sanitization within its context. In the example above (registration), if native `uuid.UUID` type is not an option at least the function could simply parse UUID and error out if parsing fails. This type of sanitization is valid no matter if you send it to shell script or render onto HTML page.

The point of my comment was to help us to find those places in the codebase where extra caution is good by issuing a warning into logs. This is literally few lines of code, compare length of both strings before and after and that's it. At the same time, in doubtful with how useful this would be in practice. Thus the "wild idea" remark, this was merely a brainstorming.

> Doing both adds nothing, escaping is enough.

Defensive programming does not cost too much, in the example I mentioned above it is literally a UUID parse call with error. The locality is a valid argument for me, one can cover this with unit tests and the code does not depend on other functions. Not all input can be sanitized this way of course, is what you are trying to say I think.

supakeen:
>  How can I tell if data gets or do not get escaped down the stack? We might add a feature and forget to add escaping too. Can you say for sure 100% of all stages are escaped after this PR is merged?

It should be, if it is forgotten this is a bug. We'll still have unescaped things after this PR merges either now or in the future.
 
> Not all input can be sanitized this way of course, is what you are trying to say I think.

Yes, you didn't mention UUIDs in your original post so I thought you were also speaking in the general case.

---

As an aside, I think you mean validating the input, instead of sanitization? The latter (to me) means more generally that input is transformed to remove any harmful parts of it while the former means 'hey this isnt what I expected to get' :)

lzap:
> I think you mean validating the input, instead of sanitization?

Yeah that is what I meant. Sorry for the confusion.

I hope UBP will help with this quite a bit, this is a typical example that can be validated via JSON Schema because this will be a common constraint across all image types and OSes.

I will say, one example of such "sanitization" (pardon me, validation) would be unit names, I want them to be configurable for the firstboot and my initial draft validates them to alphanum plus underscore/dash with the idea that this is enough and safely can be used in filenames. This could be also subject of escaping in `SystemdStage`, not shell escaping but OS filename escaping perhaps.

supakeen:
> > I think you mean validating the input, instead of sanitization?
> 
> Yeah that is what I meant. Sorry for the confusion.
> 
> I hope UBP will help with this quite a bit, this is a typical example that can be validated via JSON Schema because this will be a common constraint across all image types and OSes.
> 
> I will say, one example of such "sanitization" (pardon me, validation) would be unit names, I want them to be configurable for the firstboot and my initial draft validates them to alphanum plus underscore/dash with the idea that this is enough and safely can be used in filenames. This could be also subject of escaping in `SystemdStage`, not shell escaping but OS filename escaping perhaps.

Yea, we can be more helpful there. Many `osbuild` stages already have ideas about what valid inputs are and it'd be nice if we could share those with any blueprint options that are directly wired to `osbuild` stage options?

achilleas-k (dismissed):

supakeen (dismissed):

achilleas-k:
Unit tests fail when rebased on main because of the tests added in de6472eaece52a19424773de37b5f3281d2f2d8b.

Going to click rebase here so we get the tests (and failures) in the PR itself.

supakeen (approved):

achilleas-k (approved):

PR #1653, opened 2025-07-09
distro: add test for unsupported partition mode "raw"

mvo5:
distro: add test for unsupported partition mode "raw"

This commit adds a missing test for the constraint that the
`iot-*` image types do not support `disk.RawPartitioningMode`.

---

distro: prefix all fedora tests with TestFedora

As more and more distros move into the generic distro code we
need to label the tests better. This commit prefixes all fedora
distro tests with `TestFedora`.


thozza (approved):
LGTM, thanks.

achilleas-k (approved):
Thanks!

PR #1661, opened 2025-07-14
distro: add support for `supported_partitioning_modes`

mvo5:
This commit adds support for supported_partitioning_modes via YAML so that the RHEL-8 constraints for "edge-raw-image" and "edge-simplified-installer" can be expressed.

Note that the rhel code is using `unsupportedPartitionModes` but the downside of that approach is that every time a new mode is added there is a risk of forgeting to update the list of unsupported modes (this happend e.g. when we added btrfs partitioning and the unsupported partition modes list was not updated for rhel8).

We also add the new `supported_partitioning_types` mode into the fedora YAML distro definitions and drop the custom code to validate/mutate the partitioning for rpm_ostree from the go code, the discussion from
https://github.com/osbuild/images/pull/1654 is that we can drop the mutation entirely and just rely on the check. Technically this is a change in behavior, but arguably this is fixing a bug because without this change we one could specify a btrfs partitioning mode for IoT and get AutoLVM instead.

This is covered by the TestFedoraDistro_PartioningModeConstraints test.

Alternative version to https://github.com/osbuild/images/pull/1654 as suggested by @thozza (thanks!)

### DIFF
--- a/pkg/distro/defs/fedora/distro.yaml
+++ b/pkg/distro/defs/fedora/distro.yaml
@@ -38,6 +38,13 @@
       - "rw"
       - "coreos.no_persist_ip"
 
+  rpm_ostree_imgtype_common: &rpm_ostree_imgtype_common
+    rpm_ostree: true
+    supported_partitioning_modes:
+      - ""  # empty string means default partitioning mode
+      - "auto-lvm"
+      - "lvm"
+
   environments:
     kvm: &kvm_env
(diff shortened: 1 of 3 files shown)


ondrejbudai (commented):
I like this! :)

achilleas-k (approved):

supakeen (approved):

PR #1670, opened 2025-07-17
manifest-gen: add `-buildconfig-allow-unknown` flag

mvo5:
This commit adds support for `-buildconfig-allow-unknown` that
allows to turn off stict unmarshal mode in json for the configs
in test/ - this is useful in `gen-manifests-diff` where we always
run the "old" gen-manifests against the new branch. If the new
code contains unknown (to the old code) fields it errors. With
this new addition we can run the "old" gen-manifest with
`-buildconfig-allow-unknown` and the new one without.

This allows us to diff manifests even if the build config changes,
like in PR#1656 where `distro.ImageOptions` grows a new
`Bootc` option (similar to OSTree).

---

buildconfig: add option to allow unknown fields

This commit adds an option to allow unknown fields when reading the
build config. This is important for "gen-manifests-diff" when it
needs to read configs that contain potentially new fields.

### DIFF
--- a/internal/buildconfig/buildconfig.go
+++ b/internal/buildconfig/buildconfig.go
@@ -16,7 +16,15 @@ type BuildConfig struct {
 	Depends   interface{}          `json:"depends,omitempty"` // ignored
 }
 
-func New(path string) (*BuildConfig, error) {
+type Options struct {
+	AllowUnknownFields bool
+}
+
+func New(path string, opts *Options) (*BuildConfig, error) {
+	if opts == nil {
+		opts = &Options{}
+	}
+
 	fp, err := os.Open(path)
 	if err != nil {
 		return nil, err
@@ -24,11 +32,13 @@ func New(path string) (*BuildConfig, error) {
 	defer fp.Close()
 
 	dec := json.NewDecoder(fp)
-	dec.DisallowUnknownFields()
+	if !opts.AllowUnknownFields {
+		dec.DisallowUnknownFields()
+	}
 	var conf BuildConfig
 
 	if err := dec.Decode(&conf); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot decode build config: %w", err)
 	}
 	if dec.More() {
 		return nil, fmt.Errorf("multiple configuration objects or extra data found in %q", path)
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -51,7 +51,7 @@ func run() error {
 	}
 
 	distroFac := distrofactory.NewDefault()
-	config, err := buildconfig.New(configFile)
+	config, err := buildconfig.New(configFile, nil)
 	if err != nil {
 		return err
 	}
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -168,7 +168,7 @@ func loadConfigMap(configPath string) *BuildConfigs {
 			cfgDir := filepath.Dir(configPath)
 			path = filepath.Join(cfgDir, path)
 		}
-		config, err := buildconfig.New(path)
+		config, err := buildconfig.New(path, nil)
 		if err != nil {
 			panic(err)
 		}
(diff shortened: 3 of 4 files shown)


internal/buildconfig/buildconfig.go
```diff
@@ -16,19 +16,29 @@ type BuildConfig struct {
...
-func New(path string) (*BuildConfig, error) {
+type Options struct {
+	AllowUnknownFields bool
+}
+
+func New(path string, opts *Options) (*BuildConfig, error) {
+	if opts == nil {
+		opts = &Options{}
+	}
+
```
lzap, line 27: I would consider passing options by value. This is a common pattern and while I prefer the "options pattern" the most, sometimes it is way too much code for a small option group like this. I feel readability of value passing is much better:

```go
New("path", nil)

// vs
New("path", Options{})
```

In the first case, I very do not like the possibility of growing such functions into things like `New("", nil, nil, nil, nil)` that is unreadable. I value little bit more verbose case which is super easy to read, understand, also editing it is faster when you need to add or flip an option.

Now, my personal feeling was that in Go standard library passing by value is prevalent but I asked AI to do me some stats and while it told me it cannot do full statistics it expressed an opinion passing by pointers is more standard there. However, it made a really good point: there is another pattern in the library and it calls it "configurable object":

```
server := &http.Server{
    Addr:         ":8080",
    ReadTimeout:  10 * time.Second,
    // ... other options
}
server.ListenAndServe()
```

A second recommendation was even more striking for me, why not to use just a regular public accessor? Example again from the standard library:

```
writer := csv.NewWriter(os.Stdout)
writer.Comma = ';'
writer.UseCRLF = true
```

I like this a lot, why to make things more complex than they need to be? Since this is a reference, it does not guarantee thread safety only passing by value would do this. Sure, the `New` method would have need to be separated into separate one that performs the work.

Well, at this point I usually delete my comment and carry on because I think I trapped myself. But let me add it anyways, I think pointer is fine and AI found many examples from the standard library. But personally, I am not so happy about passing pointers into methods.

mvo5: Thanks for your thoughts on this.

> I would consider passing options by value. This is a common pattern and while I prefer the "options pattern" the most, sometimes it is way too much code for a small option group like this. I feel readability of value passing is much better:
> 
> ```go
> New("path", nil)
> 
> // vs
> New("path", Options{})
> ```
>
> In the first case, I very do not like the possibility of growing such functions into things like `New("", nil, nil, nil, nil)` that is unreadable. I value little bit more verbose case which is super easy to read, understand, also editing it is faster when you need to add or flip an option.

Well, given that we own the code we can control how `New(opts)` grows. I.e. instead of `New("", nil, nil, nil)` we just add the (hypothetical) new things just to options and it will always be "New(required, optionalOptions)`. In many ways the `&Options{}` pattern is a (non-optimal) way to have kwargs in go. Fwiw, I have seen this pattern in many codebases.

[..] 
> A second recommendation was even more striking for me, why not to use just a regular public accessor? Example again from the standard library:
> 
> ```
> writer := csv.NewWriter(os.Stdout)
> writer.Comma = ';'
> writer.UseCRLF = true
> ```
> 
> I like this a lot, why to make things more complex than they need to be? Since this is a reference, it does not guarantee thread safety only passing by value would do this. Sure, the `New` method would have need to be separated but it is a small change with big impact in the big picture.

I think this works well for the example above, but not so well for the use case in the PR. Here we already do the reading in New() so it would be something like:
```
loader := blueprintload.NewLoader()`
loader.AllowUnknownFields = true
loader.Load()
``` 
seems seems a bit verbose given that it was just "blueprintload.New(path, Options{AllowUnknownFields: true}" before.
 
> Well, at this point I usually delete my comment and carry on because I think I trapped myself. But let me add it anyways, I think pointer is fine and AI found many examples from the standard library. But personally, I am not so happy about passing pointers into methods.

Idk, it seems fine to pass pointerer with the kwargs like struct above as long as the method is prepared. As I said above, I have seen this in many go codebases. The most simple alternative would be a boolean argument in  blueoprintload.New()path, allowUnknown). But the issue with that is that bools in function signatures are problematic, what does "blueprintoad.New(path, false)"  mean? Its harder to read than the kwargs style from this PR IMHO.

supakeen (approved):
I ran into this issue while debugging the package lists between 0.128 and the current main so functionally this is great.

Thanks!

PR #1673, opened 2025-07-18
customizations/subscription: conditionally enable semanage call

croissanne:
Only set rhcd_t to permissive on rhel 9 and 10. The selinux module is
missing on rhel 8.

### DIFF
--- a/pkg/customizations/subscription/subscription.go
+++ b/pkg/customizations/subscription/subscription.go
@@ -15,6 +15,7 @@ type ImageOptions struct {
 	BaseUrl       string `json:"base_url"`
 	Insights      bool   `json:"insights"`
 	Rhc           bool   `json:"rhc"`
+	PermissiveRHC bool   `json:"permissive_rhc"`
 	Proxy         string `json:"proxy"`
 	TemplateName  string `json:"template_name"`
 	TemplateUUID  string `json:"template_uuid"`
--- a/pkg/distro/rhel/images.go
+++ b/pkg/distro/rhel/images.go
@@ -3,6 +3,7 @@ package rhel
 import (
 	"fmt"
 	"math/rand"
+	"slices"
 
 	"github.com/osbuild/images/internal/workload"
 	"github.com/osbuild/images/pkg/blueprint"
@@ -245,6 +246,8 @@ func osCustomizations(
 		if options.Subscription.Proxy != "" {
 			osc.InsightsClientConfig = &osbuild.InsightsClientConfigStageOptions{Config: osbuild.InsightsClientConfig{Proxy: options.Subscription.Proxy}}
 		}
+		// selinux policy for rhc is set in registration service on rhel 9 and 10
+		options.Subscription.PermissiveRHC = slices.Contains([]string{"9", "10"}, t.Arch().Distro().Releasever())
 	} else {
 		subscriptionStatus = subscription.RHSMConfigNoSubscription
 	}
--- a/pkg/manifest/os_test.go
+++ b/pkg/manifest/os_test.go
@@ -113,6 +113,7 @@ func TestRhcInsightsCommands(t *testing.T) {
 		BaseUrl:       "http://cdn.redhat.com/",
 		Insights:      false,
 		Rhc:           true,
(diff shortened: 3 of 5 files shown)


supakeen:
I still don't feel this *has* to be passed through the options; why can't it be part of the `RHSMConfig` already contained in `ImageConfig`? The options to me feel like they only set the subscription status; not *how* we set it? If it doesn't fit in the `ImageConfig.RHSMConfig` (I see that that gets merged with blueprint) then a single new property in `ImageConfig` would suffice too?

Why does it have to be an option? I really thought those were for specific builds, not for things that always affect a certain image type.

pkg/customizations/subscription/subscription.go
```diff
@@ -15,6 +15,7 @@ type ImageOptions struct {
 	BaseUrl       string `json:"base_url"`
 	Insights      bool   `json:"insights"`
 	Rhc           bool   `json:"rhc"`
+	PermissiveRHC bool   `json:"permissive_rhc"`
 	Proxy         string `json:"proxy"`
 	TemplateName  string `json:"template_name"`
 	TemplateUUID  string `json:"template_uuid"`
```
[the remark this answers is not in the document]
croissanne: We shouldn't turn off selinux as a whole though. And I don't think the selinux.config stage can currently configure permissions per process type?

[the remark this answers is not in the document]
lzap: I was not proposing that, I was merely suggesting to move selinux workarounds at one place. Okay then.

lzap (dismissed):

supakeen (commented):
Am I missing something? This *should've* changed image checksums right? Since we now have different stage options to the systemd stage?

Right now it seems that this PR changes nothing (to the images being created) while its intent is to write a different service file on RHEL 8.

internal/common/pointers.go
```diff
@@ -3,3 +3,10 @@ package common
 func ToPtr[T any](x T) *T {
 	return &x
 }
+
+func FromPtr[T any](ref *T) (value T) {
```
mvo5, line 7: (not for this PR but I want to raise awareness as my idea is not really getting traction): a maybe interesting approach for this could be taken from https://github.com/osbuild/images/pull/860 - i.e. instead of wrapping pointers (with the risk of still dereferencing them and causing panics) we could move into a world of optional datatypes that encapsulate the pointers entirely so that a panic is never possible. Anyway, this totally fine in this context of course, just trying to raise awareness :)

mvo5, line 7: Fwiw, we have a very similar function in https://github.com/osbuild/blueprint-schema/blob/main/pkg/ptr/any.go#L22 but its called `ValueOrEmpty` and has a docstring. I kinda like it slightly better as from FromPtr() its not clear what happens with nil pointers.

mvo5 (commented):
Thank you, code looks fine, I have one naming suggestion (and/or we could import the existing code for this but its probably weird).

One meta-question, the commit message is a bit short on why we need to set rhc permissive and what prompts us to do it now, am I missing something here or could we add maybe a little bit more context why this is needed (and why we are doing it now, rhel9 is out since a while)? Happy to approve if its time critical but if not maybe we could add append a bit more background?

internal/common/pointers.go
croissanne: How rustful :)

croissanne: oohhhh, i missed that one.

croissanne:
> Thank you, code looks fine, I have one naming suggestion (and/or we could import the existing code for this but its probably weird).
> 
> One meta-question, the commit message is a bit short on why we need to set rhc permissive and what prompts us to do it now, am I missing something here or could we add maybe a little bit more context why this is needed (and why we are doing it now, rhel9 is out since a while)? Happy to approve if its time critical but if not maybe we could add append a bit more background?

added a bit more context in the commit message. lmk if more is needed.

mvo5 (dismissed):
Thank you! Love the new commit message! It answers all my questions :)

mvo5 (approved):

PR #1684, opened 2025-07-23
test/manifests/all-customizations: don't embed local file via URI

thozza:
The all-customizations (and empty) config is used for manifest tests in osbuild. Previously, the all-customizations config was specifying a file customization referencing a local file from the host using URI. This turned out to be problematic, because all manifests in images CI are generated and built on Fedora host, thus embedding /etc/os-release from Fedora. However, in osbuild, manifests are generated and built on the same distribution as the manifest's distribution. As a result, the manifest checksums differed for the all-customizations manifests on all distros, except for the Fedora version used by images CI.

Initially, the plan was to workaround the issue by mocking the local file. However, that would not work, because the manifest is actually being built as part of the CI pipeline. The local file is turned into CURL source in the manifest, with its URI pointing to a temporary path that would not exist on the host at the time of image build.

Therefore, to solve the issue with manifest tests in osbuild, do not use the file customization with URI, pointing to a local file, in the all-customizations config.

The URI scenario is moved to a dedicated config, so it is still boot-tested in images CI, but not used in osbuild for manifest tests.

I also kept the enhancements that were part of the initial version of this PR, as I believe they are still useful.


mvo5 (dismissed):
Thank you - out of curiosity, why is this draft? What is missing?

bcl (dismissed):
Looks good to me. Thanks!

achilleas-k (approved):
Thank you!!

PR #1687, opened 2025-07-24
distro: drop legacy loaders and update tests

mvo5:
This commit drops the legacy YAML loading functions and updates the test to use the {Distro,ImageType}YAML getters instead.


thozza (approved):
LGTM

PR #1688, opened 2025-07-24
distro: set `distro.ID` in `DistroYAML` by loader

mvo5:
distro: set `distro.ID` in `DistroYAML` by loader

By setting the parsed distro.ID into DistroYAML we can
simplify a lot of the cases where we currently use
`nameVer string` for the various helpers. By using the
id we avoid re-parsing the name and potentially raising
error. This will simplify the code in the comming commits.

---

distro: add/use `DistroYAML.ID` consistently

We use(d) `distroNameVer` a lot when accessing YAML
data because that is what we had when transitioning
from fedora/rhel distro abstraction to the generic
distro. But now that everything is a generic distro
we can actually use `distro.ID` everywhere which
simplifies a lot of code and avoids (re)parsing
the `distroNameVer`.

### DIFF
--- a/pkg/distro/defs/loader_test.go
+++ b/pkg/distro/defs/loader_test.go
@@ -119,8 +119,7 @@ image_types:
 `
 	it := makeTestImageType(t, fakePkgsSetYaml)
 
-	pkgSet, err := it.PackageSets("test-distro-1", "x86_64")
-	assert.NoError(t, err)
+	pkgSet := it.PackageSets(distro.ID{Name: "test-distro", MajorVersion: 1}, "x86_64")
 	assert.Equal(t, map[string]rpmmd.PackageSet{
 		"os": {
 			Include: []string{"from-condition-inc2", "inc1"},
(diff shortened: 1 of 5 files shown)


achilleas-k (dismissed):
LGTM, but the linter is unhappy.

supakeen (approved):
The YAML linter seems to be complaining about a file you didn't touch.

pkg/distro/defs/loader_test.go
```diff
@@ -841,9 +841,9 @@ func TestDistrosLoadingExact(t *testing.T) {
 	restore := defs.MockDataFS(baseDir)
 	defer restore()
 
-	distro, err := defs.NewDistroYAML("fedora-43")
+	dist, err := defs.NewDistroYAML("fedora-43")
 	require.NoError(t, err)
-	assert.Equal(t, &defs.DistroYAML{
+	assert.Equal(t, dist, &defs.DistroYAML{
```
thozza, line 839: Nitpick: technically, the `dist` should be the third argument to `Equal()`, because the second is the "expected" value and the third is the "actual" value.

thozza (approved):
Amazing! Thanks for figuring this out.

pkg/distro/defs/loader_test.go
mvo5: Nice catch! Thank you, I will fix this in a followup (usually I get this right but sometimes I forget :)

PR #1690, opened 2025-07-24
rhel{9,10}: centos WSL refinement (HMS-8922)

supakeen:
After a short chat with CentOS folks (see the [linked issue](https://issues.redhat.com/browse/HMS-8922)) they'd like to:

- Use wsl-setup.
- Drop cloud-init.
- Drop subscription-manager.
- Use centos-* named packages.

This PR makes those changes only for CentOS; all other RHEL-likes (RHEL, AlmaLinux, AlmaLinux Kitten) keep the RHEL-style WSL images with cloud-init and subscription-manager until we're ready to migrate RHEL as well.


supakeen:
CI failing on `wsl-setup` not being available in CentOS 10; need to update snapshots.

mvo5 (dismissed):
Thank you

pkg/distro/defs/rhel-10/distro.yaml
```diff
@@ -282,6 +282,16 @@
       - "mdadm"
       - "nss-softokn"
 
+  wsl_config: &wsl_config
+    config:
+      boot_systemd: true
+    distribution_config: &wsl_distribution_config
+      shortcut:
+        enabled: true
+        icon: /usr/share/pixmaps/fedora-logo.ico
```
thozza, line 291: I know that this is a copy&paste, but referencing the Fedora logo in the RHEL definition does not look right.

thozza (approved):
LGTM in general.

The WSL image type should install some sensible RPM package group, instead of listing all packages explicitly. Comps groups already solve the problem of installing the `*-release` and `*-logos`, as well as `subscription-manager`, on CS versus RHEL. We could get rid of the conditions in our image definitions.

pkg/distro/defs/rhel-10/distro.yaml
supakeen: It is correct; all EL's put the logo in that path (weirdly so).

supakeen:
Note that this PR will need to be force-pushed for even newer snapshots :(

supakeen:
Converting to draft until snapshots work, I'll also be applying this to CentOS 9 as well.

supakeen:
Ok, this PR is now based on #1694 and adds a commit for a package that was missed in that PR (`-logos`). All commits have been rewritten to apply to both 9 and 10.

Rebasing and undrafting once #1694 lands in `main`.

croissanne (approved):
this lgtm thanks!

thozza (approved):

PR #1692, opened 2025-07-25
distro: rename "distro.yaml" and move under "./data"

mvo5:
[draft as it needs needs https://github.com/osbuild/images/pull/1687]

Another tiny cleanup/RFC, happy about naming suggestions
or different ideas about the location (see below):

---

distro: rename "distro.yaml" -> "imagetypes.yaml"

This commit renames all our {fedora,rhel-{7,8,9}}/distro.yaml
files into `imagetypes.yaml`. Since the `distro.yaml` name was
picked our YAML based image defintions have evolved quite a
bit and now we also have the `distros.yaml` that describes
the individual distros and the `distro.yaml` (singular) is
really more about the image types than it is about a distro.

So lets rename it.

---

distrodefs: move distro defifinitions under `./data`

The distro definitions YAML is currently living under
`pkg/distro/defs`. Now that all our distros are converted
to YAML lets move them into our `./data` directory as
this seems a more appropriate place than next to the
go code.


supakeen (approved):
Thank you; I've wanted this :)

achilleas-k (dismissed):
LGTM

We're gonna have merge conflicts though, but that's okay :)

achilleas-k (approved):

PR #1696, opened 2025-07-25
[RFC/wild-idea] distro: drop {Build,Payload}Pipelines from imagetypes

mvo5:
[draft as this is obviously very controversial and also currently osbuild-composer uses the pipeline names to collect rpm metadata, however AFAICT composer could also just iterate over all pipelines and use "build" to identify the build pipeline name]

This started out from me investigating if we can inferre `[Build,Payload}Pipelines` and `Exports` when documenting the new YAML based image types. I think for `Exports` that is still something worth perusing but for 

All image types currently define the build and payload pipelines. But the build pipelines are the same (`["build"]`) for all images and the only place that uses it is the sbom writer so we can as well just use the static name there.

The payload pipelines are not used at all in the code AFAICT and osbuild just needs the export to know what pipelines it needs to run.

The extra complication is that the connection between the pipeline names and the code is quite loose and the pipelines are actually fully determined by the combination of the image function, platform and compression and could (probably) also be infered. But it AFAICT we don't need them at all so YAGNI.

### DIFF
--- a/pkg/distro/defs/fedora/distro.yaml
+++ b/pkg/distro/defs/fedora/distro.yaml
@@ -549,8 +549,6 @@ image_types:
     bootable: true
     default_size: 5_368_709_120  # 5 * datasizes.GibiByte
     image_func: "disk"
-    build_pipelines: ["build"]
-    payload_pipelines: ["os", "image", "vagrant", "archive"]
     exports: ["archive"]
     required_partition_sizes: *default_required_dir_sizes
     image_config: &image_config_vagrant
@@ -611,8 +609,6 @@ image_types:
     bootable: true
     default_size: 5_368_709_120  # 5 * datasizes.GibiByte
     image_func: "disk"
-    build_pipelines: ["build"]
-    payload_pipelines: ["os", "image", "qcow2"]
     exports: ["qcow2"]
     required_partition_sizes: *default_required_dir_sizes
     image_config: &image_config_qcow2
@@ -653,7 +649,6 @@ image_types:
     name_aliases: ["ami"]
     filename: "image.raw"
     mime_type: "application/octet-stream"
-    payload_pipelines: ["os", "image"]
     exports: ["image"]
     environment: *ec2_env
     platforms:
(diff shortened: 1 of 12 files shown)


achilleas-k:
> composer could also just iterate over all pipelines and use "build" to identify the build pipeline name

I don't like this for two reasons:
- It makes he pipeline name significant, which we've avoided in many cases until now.
- We had a quick chat with @supakeen recently about `osbuild --inspect`ing manifests to generate stage IDs.  This changes the names of pipelines to UUIDs.  Now, in the plan we talked about, that "inspected" manifest wouldn't be used in the build itself, but there's no reason it couldn't, and that would break our pipeline name assumption.  We might not ever do this, or anything like this, but I don't feel comfortable with the idea that renaming a pipeline would break some functionality that relies on specific strings.

I'm not a fan of listing pipeline names in each image type either (though it was my idea to begin with, iirc).  But I think there's a solution that doesn't involve magic strings/names nor listing everything.  The build and payload pipeline names are needed here:
https://github.com/osbuild/osbuild-composer/blob/e3255a578d26cad9e83e294706caaee5009e2207/internal/cloudapi/v2/server.go#L258-L264

They're attached to the osbuild job as metadata so the reader of the job result can parse the osbuild output and separate the pipelines by name, and they're taken directly from the image type that defines them statically.
What I think could work (should work, even), is if the pipeline generators themselves mark each pipeline accordingly.  So if the Pipeline interface has a method called `Type()` or `Role()` that returns `"build"` or `"payload"`, or anything else, (or alternatively, `IsBuild() bool`), then the (pre-serialised) manifest itself could report these dynamically:
```go
// pkg/manifest/manifest.go

func (m *Manifest) BuildPipelines() []string {
	var buildNames []string
	for _, pipeline := range m.pipelines {
		if pipeline.IsBuild() {
			buildNames = append(buildNames, pipeline.Name())
		}
	}
	return buildNames
}
```

or similar.

mvo5:
[...]
> They're attached to the osbuild job as metadata so the reader of the job result can parse the osbuild output and separate the pipelines by name, and they're taken directly from the image type that defines them statically. What I think could work (should work, even), is if the pipeline generators themselves mark each pipeline accordingly. So if the Pipeline interface has a method called `Type()` or `Role()` that returns `"build"` or `"payload"`, or anything else, (or alternatively, `IsBuild() bool`), then the (pre-serialised) manifest itself could report these dynamically:
[...]

I like this! I opened https://github.com/osbuild/images/pull/1732 that (hopefully) implements this suggestion. I will close this one. I also want to persue the exports idea but its (to me) less pressing currently.

PR #1701, opened 2025-07-28
distro/generic: test checkOptions() functions

achilleas-k:
Add tests for the checkOptions() image type function.  This is done in
preparation for a new way of declaring supported and required
customizations and options alongside the image definitions.

The tests themselves contain notes highlighting incorrect behaviour.  I
am leaving these unfixed intentionally.  Corrections to the config
validation will happen alongside the new definitions.

---

This commit was spun-off from and related to #1216.
It is meant to serve as a reference for the current behaviour of the image-type-based config validation (however incorrect), so that changes in the functionality of the validator get reflected in the testing as well.

The tests are not meant to cover each individual image type.  My goal was to cover each error condition, so for cases where the same rules apply to multiple image types, only a single test case is included.

### DIFF
--- a/pkg/distro/generic/export_test.go
+++ b/pkg/distro/generic/export_test.go
@@ -25,3 +25,7 @@ type ImageType = imageType
 func (t *imageType) GetDefaultImageConfig() *distro.ImageConfig {
 	return t.getDefaultImageConfig()
 }
+
+func ImageTypeCheckOptions(it *imageType, bp *blueprint.Blueprint, options distro.ImageOptions) ([]string, error) {
+	return it.checkOptions(bp, options)
+}
--- a/pkg/distro/generic/options_test.go
+++ b/pkg/distro/generic/options_test.go
@@ -1,0 +1,1613 @@
+package generic_test
+
+import (
+	"testing"
+
+	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/osbuild/images/pkg/distro"
+	"github.com/osbuild/images/pkg/distro/generic"
+	"github.com/osbuild/images/pkg/ostree"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCheckOptionsFedora(t *testing.T) {
+	type testCase struct {
+		distro  string
+		it      string
+		bp      blueprint.Blueprint
+		options distro.ImageOptions
+		expErr  string
+	}
+
+	testCases := map[string]testCase{
+		"f42/ami-ok": {
+			distro:  "fedora-42",
+			it:      "server-ami",
+			bp:      blueprint.Blueprint{},
+			options: distro.ImageOptions{},
+			expErr:  "",
+		},
+		"f42/ami-installer-error": {
+			distro: "fedora-42",
+			it:     "server-ami",
+			bp: blueprint.Blueprint{
+				Customizations: &blueprint.Customizations{
+					Installer: &blueprint.InstallerCustomization{
+						Unattended: true,
+					},
+				},
+			},
+			expErr: "installer customizations are not supported for \"server-ami\"",
+		},
+		"f42/ami-ostree-error": {
+			distro: "fedora-42",
+			it:     "server-ami",
+			options: distro.ImageOptions{
+				OSTree: &ostree.ImageOptions{
+					URL: "https://example.org/repo",
+				},
+			},
+			expErr: "OSTree is not supported for \"server-ami\"",
+		},
+		"f42/ostree-installer-requires-ostree-url": {
+			distro: "fedora-42",
+			it:     "iot-installer",
+			expErr: "boot ISO image type \"iot-installer\" requires specifying a URL from which to retrieve the OSTree commit",
+		},
+		"f42/ostree-disk-supported": {
+			distro: "fedora-42",
+			it:     "iot-qcow2",
+			bp: blueprint.Blueprint{
+				Customizations: &blueprint.Customizations{
+					User:  []blueprint.UserCustomization{{Name: "root"}},
+					Group: []blueprint.GroupCustomization{{Name: "admins"}},
+					Files: []blueprint.FileCustomization{{
+						Path: "/etc/osbuild/stamp",
+						Data: "Created by osbuild",
+					}},
+					Directories: []blueprint.DirectoryCustomization{{
+						Path: "/etc/osbuild",
+					}},
+					Services: &blueprint.ServicesCustomization{
+						Disabled: []string{"sshd.service"},
+					},
+					FIPS: common.ToPtr(true),
+				},
+			},
+			// NOTE: this should also require an ostree URL
+		},
+		"f42/ostree-disk-not-supported": {
+			distro: "fedora-42",
+			it:     "iot-qcow2",
+			bp: blueprint.Blueprint{
+				Customizations: &blueprint.Customizations{
+					User:  []blueprint.UserCustomization{{Name: "root"}},
+					Group: []blueprint.GroupCustomization{{Name: "admins"}},
+					Files: []blueprint.FileCustomization{{
+						Path: "/etc/osbuild/stamp",
+						Data: "Created by osbuild",
+					}},
+					Directories: []blueprint.DirectoryCustomization{{
+						Path: "/etc/osbuild",
+					}},
+					Services: &blueprint.ServicesCustomization{
+						Disabled: []string{"sshd.service"},
+					},
+					FIPS: common.ToPtr(true),
+					Kernel: &blueprint.KernelCustomization{
+						Name: "kernel-rt",
+					},
+				},
+			},
+			options: distro.ImageOptions{
+				OSTree: &ostree.ImageOptions{
+					URL: "https://example.org/repo",
+				},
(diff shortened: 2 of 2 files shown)


bcl (dismissed):
Looks like a good idea to me!

pkg/distro/generic/export_test.go
```diff
@@ -25,3 +25,11 @@ type ImageType = imageType
 func (t *imageType) GetDefaultImageConfig() *distro.ImageConfig {
 	return t.getDefaultImageConfig()
 }
+
+var (
+	CheckOptionsFedora = checkOptionsFedora
```
mvo5, line 30: I wonder if we could tweak this a little bit, while this is super targeted we will not check the checkOptionsCommon() which is run from imageType.checkOptions(). So *maybe* we could just run imageType.checkOptions() here and then we could cover the ostree checks too. The over nice advantage is that when we refactor and move the allowed options into the image type no code in this testfile will have to change, we can use exactly this corpus and validate our refactor is working. The downside of course is that it is slightly less targeted, i.e. with the existing test we know *exactly* we run the rhel9 check, if the code in imgtypes is refactored and now runs the rhel10 checks we will not know. OTOH that might be a good thing as we refactor. Anyway, just my musing, below a diff that works for me and now also errors in case that you added a note that it should error:
```go
diff --git a/pkg/distro/generic/checkoptions_test.go b/pkg/distro/ge
neric/checkoptions_test.go
index 48fba97f5..9c46f3e4f 100644
--- a/pkg/distro/generic/checkoptions_test.go
+++ b/pkg/distro/generic/checkoptions_test.go
@@ -515,7 +515,7 @@ func TestCheckOptionsFedora(t *testing.T) {
 			assert.NoError(err)
 			genit, ok := it.(*generic.ImageType) // chec
kOptions() functions require generic.ImageType
 			assert.True(ok, "image type %q for distro %q
 does not appear to be valid", tc.it, fedora.Name())
-			_, err = generic.CheckOptionsFedora(genit, &
tc.bp, tc.options)
+			_, err = genit.CheckOptions(&tc.bp, tc.optio
ns)
 			if tc.expErr == "" {
 				assert.NoError(err)
 			} else {
@@ -544,6 +544,15 @@ func TestCheckOptionsRhel8(t *testing.T) {
 	assert.NoError(t, err)
 
 	testCases := map[string]testCase{
+		"common-ostree-check": {
+			it: "ami",
+			options: distro.ImageOptions{
+				OSTree: &ostree.ImageOptions{
+					ImageRef: "foo",
+				},
+			},
+			expErr: `OSTree is not supported for "ami"`,
+		},
 		"ami-ok": {
 			it:      "ami",
 			bp:      blueprint.Blueprint{},
@@ -935,7 +944,7 @@ func TestCheckOptionsRhel8(t *testing.T) {
 			assert.NoError(err)
 			genit, ok := it.(*generic.ImageType) // chec
kOptions() functions require generic.ImageType
 			assert.True(ok, "image type %q for distro %q
 does not appear to be valid", tc.it, rhel8.Name())
-			_, err = generic.CheckOptionsRhel8(genit, &t
c.bp, tc.options)
+			_, err = genit.CheckOptions(&tc.bp, tc.optio
ns)
 			if tc.expErr == "" {
 				assert.NoError(err)
 			} else {
@@ -1504,7 +1513,7 @@ func TestCheckOptionsRhel9(t *testing.T) {
 			assert.NoError(err)
 			genit, ok := it.(*generic.ImageType) // chec
kOptions() functions require generic.ImageType
 			assert.True(ok, "image type %q for distro %q
 does not appear to be valid", tc.it, rhel9.Name())
-			_, err = generic.CheckOptionsRhel9(genit, &t
c.bp, tc.options)
+			_, err = genit.CheckOptions(&tc.bp, tc.optio
ns)
 			if tc.expErr == "" {
 				assert.NoError(err)
 			} else {
@@ -1592,7 +1601,7 @@ func TestCheckOptionsRhel10(t *testing.T) {
 			assert.NoError(err)
 			genit, ok := it.(*generic.ImageType) // chec
kOptions() functions require generic.ImageType
 			assert.True(ok, "image type %q for distro %q
 does not appear to be valid", tc.it, rhel10.Name())
-			_, err = generic.CheckOptionsRhel10(genit, &
tc.bp, tc.options)
+			_, err = genit.CheckOptions(&tc.bp, tc.optio
ns)
 			if tc.expErr == "" {
 				assert.NoError(err)
 			} else {
@@ -1660,7 +1669,7 @@ func TestCheckOptionsRhel7(t *testing.T) {
 			assert.NoError(err)
 			genit, ok := it.(*generic.ImageType) // chec
kOptions() functions require generic.ImageType
 			assert.True(ok, "image type %q for distro %q
 does not appear to be valid", tc.it, rhel7.Name())
-			_, err = generic.CheckOptionsRhel7(genit, &t
c.bp, tc.options)
+			_, err = genit.CheckOptions(&tc.bp, tc.optio
ns)
 			if tc.expErr == "" {
 				assert.NoError(err)
 			} else {
diff --git a/pkg/distro/generic/export_test.go b/pkg/distro/generic/
export_test.go
index cb39db9a3..e4c593e2c 100644
--- a/pkg/distro/generic/export_test.go
+++ b/pkg/distro/generic/export_test.go
@@ -26,10 +26,6 @@ func (t *imageType) GetDefaultImageConfig() *dist
ro.ImageConfig {
 	return t.getDefaultImageConfig()
 }
 
-var (
-	CheckOptionsFedora = checkOptionsFedora
-	CheckOptionsRhel7  = checkOptionsRhel7
-	CheckOptionsRhel8  = checkOptionsRhel8
-	CheckOptionsRhel9  = checkOptionsRhel9
-	CheckOptionsRhel10 = checkOptionsRhel10
-)
+func (t *imageType) CheckOptions(bp *blueprint.Blueprint, options d
istro.ImageOptions) ([]string, error) {
+	return t.checkOptions(bp, options)
+}

```

pkg/distro/generic/checkoptions_test.go
```diff
@@ -0,0 +1,1671 @@
+package generic_test
```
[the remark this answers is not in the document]
achilleas-k: It's a 1000+ SLOC file.  I wanted to make it clear it's only testing one family of functions and not everything in a file.

THAT SAID, I wasn't aware at the time that `options.go` only contains those check options functions.

pkg/distro/generic/options_test.go
```diff
@@ -0,0 +1,1671 @@
...
+			it: "vhd",
+			bp: blueprint.Blueprint{
+				Customizations: &blueprint.Customizations{
+					OpenSCAP: &blueprint.OpenSCAPCustomization{
+						ProfileID: "",
+					},
+				},
+			},
+			expErr: "OpenSCAP profile cannot be empty",
+		},
```
lzap, line 497: This is something that this validation should not care about and will be done by UBP schema. Can't hurt tho, also UBP is not yet implemented.

pkg/distro/generic/options_test.go
```diff
@@ -0,0 +1,1671 @@
...
+		// NOTE: the following tests verify the current behaviour of the
+		// function, but the behaviour itself is wrong
+		"ostree-disk-requires-ostree-url": {
+			it:     "iot-qcow2",
+			expErr: "", // NOTE: it should require a URL
+		},
+		"ostree-disk2-requires-ostree-url": {
+			it:     "iot-raw-xz",
+			expErr: "", // NOTE: it should require a URL
+		},
```
lzap, line 508: I am not sure I understand, is this a leftover perhaps?

achilleas-k: Ooooh, ok.  Sorry, I misunderstood something in your earlier comment.  I think I completely missed the part about `checkOptionsCommon()`.  I'll fix this all up tomorrow.  Thanks!

achilleas-k:
Merged them all into one test against the `imageType.checkOptions()` function.

pkg/distro/generic/options_test.go
```diff
@@ -0,0 +1,1613 @@
...
+			// partitions on RHEL 8. Let's ignore that for now and return to it
+			// when we redo the validation.
+			imageTypes, err := fedora.GetArch("x86_64")
+			assert.NoError(err)
+			it, err := imageTypes.GetImageType(tc.it)
+			assert.NoError(err)
+
+			genit, ok := it.(*generic.ImageType) // checkOptions() function is defined on generic.ImageType
+			assert.True(ok, "image type %q for distro %q does not appear to be valid", tc.it, fedora.Name())
+			_, err = generic.ImageTypeCheckOptions(genit, &tc.bp, tc.options)
```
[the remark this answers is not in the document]
achilleas-k: The only warning that gets generated now is when selecting FIPS on a non-FIPS-enabled host.  The way the check works makes it hard to mock right now.  I suppose we could skip it when we detect the test is running on a host with FIPS enabled.

achilleas-k:
> I'm wondering if there would be any value in testing also CentOS Stream `checkOptions()`. I don't know if there are any differences compared to the respective RHEL counterpart.

There is no mention of `centos` in the `checkOptions` functions, so I don't see a reason to check it explicitly.  The plan is to drop the distro-specific functions anyway.

PR #1702, opened 2025-07-29
distro: fix issues with yaml distro definitions and enable yaml checks

achilleas-k:
**distro/rhel9: remove duplicate anchors from yaml definitions**

I've been using `yq` a lot to get a flattened view of our yaml
definitions, with all references resolved.  While `yamllint` doesn't
seem to mind duplicated anchors, `yq` fails when they are found.

Duplicate anchors create ambiguity.  Remove them.

- platform_override_conditions_ami: never referenced, so it is
  unnecessary.
- default_partition_table_part_boot_payload: defined three times with
  the same body.  Convert the two extra anchors to references.

---

**distro/rhel8: remove duplicate anchors from yaml definitions**

platform_override_conditions_ami is never referenced, so it is
unnecessary.

---

**distro/defs: fix up issues in distros.yaml**

Fix issues caught by yamllint on distros.yaml:
- Document start marker (---).
- Comment indentation level.

---

**github: run yamllint and yq on distro definitions**

Check our yaml distro definitions for errors and warnings.

yamllint is run with `--strict` to convert warnings to errors.  Let's be
as clean as possible.
yq is also used because it will fail when a file has multiple anchors
with the same name.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -239,3 +239,28 @@ jobs:
         run: |
           python3 -m pylint --version
           python3 -m pylint $(grep -l "/usr/bin/env python3" -r test/scripts) test/scripts/*.py
+
+  yaml-checks:
+    name: "YAML checks"
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Install checkers
+        run: |
+          sudo apt update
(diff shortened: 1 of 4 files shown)


mvo5 (approved):
Thank you!

achilleas-k:
> Thanks; since this is a YAML cleanup PR I just noticed that in Fedora there's an (unused) `default_kernel_optons`; is that something you want to drop here as well?

oh good catch.  This PR is queued to merge now, so let's do it separately.

PR #1704, opened 2025-07-29
distro/rhel9: set default_kernel to kernel-uki-virt

achilleas-k:
When unset, the value defaults to kernel-core, which is wrong.


mvo5 (commented):
Out of curiosity, why did this not show up in the manifest diffs? Is there something missing somewhere? Or was it just overlooked?

mvo5:
> > Out of curiosity, why did this not show up in the manifest diffs? Is there something missing somewhere? Or was it just overlooked?
> 
> What do you mean? Which manifest diffs?
> 
> EDIT: If you mean changes to manifests during the transition from code to yaml, I think the CVM image definitions weren't merged until after the yaml transition.

Yeah, sorry for not being clear. What I mean is if something during the code->yaml transition went wrong that was somehow not caught but you answered my question :) Thanks!

PR #1707, opened 2025-07-29
rhel{9,10}: azure for non-RHEL (HMS-8949)

supakeen:
We only had an `azure-rhui` image type that builds importable disk images for Azure. This image type was only enabled on RHEL.

In this PR I introduce a new image type (which is the same as `azure-rhui`) except under the `azure` name. This image type is available in all based-on-RHEL-but-not-RHELs (CentOS, Alma Linux, Alma Linux Kitten) where the `rhui` name doesn't make much sense.

Separately this PR drops `system-reinstall-bootc` from CentOS (and other non-RHEL) as requested by CentOS maintainer(s).


achilleas-k (approved):
Thanks.  I'm approving but this does make me consider if our image type naming is confusing.  In the beginning there was just `vhd`.  Then came all the `azure-rhui` variants for the internal builds and marketplace.  Now, on CentOS, we have `vhd` and `azure`, where `azure` matches the marketplace images and `vhd` is a simpler (no LVM) Azure-compatible image, basically.

The image descriptions on osbuild.org certainly help, but let's come back to this and reconsider our naming scheme.

mvo5 (dismissed):

ondrejbudai:
I'm deeply concerned about the `system-reinstall-bootc` change. I would love if CentOS and RHEL images are as similar as possible because of the whole upstream-downstream relationship. If there are no technical reasons to drop it, it should stay there.

supakeen:
> I'm deeply concerned about the `system-reinstall-bootc` change. I would love if CentOS and RHEL images are as similar as possible because of the whole upstream-downstream relationship. If there are no technical reasons to drop it, it should stay there.

It seems like an issue should be filed in the CentOS project to request its inclusion. It's currently not in their Lorax-produced images. So technically this isn't a *change*, it'd be a change to start including it.

supakeen:
> Looks like this is going to get popped off the merge queue because of conflicts :(

Ah, I'll rebase after then.

supakeen:
@achilleas-k / @mvo5 this has been rebased, could you re-ack?

PR #1709, opened 2025-07-30
cloud/azure: add unit tests (HMS-8954)

croissanne:
- Move azure from upload to cloud
- mock clients
- unit tests


achilleas-k (approved):
LGTM

The `Client` struct holding interfaces rather than factories that initialise the internal client every time makes sense to me.

PR #1656, opened 2025-07-11
images: provide the ability to generate bootc disk images (HMS-8844)

mvo5:
[draft as this needs https://github.com/osbuild/images/pull/1655 and lacks a bunch of tests]

This branch provides a way to  create an imageType from a bootc reference. 

With that the API user can do:
```go
imgType, err := distro.ImageFromBootc(bootcRef, imgTypeName, archStr)
...
```
and that can then be passed to manifestgen.

/jira-epic HMS-8839

JIRA: [HMS-8844](https://issues.redhat.com/browse/HMS-8844)

PR #1713, opened 2025-07-31
distro: add support for filesystem from default_fs_type (HMS-8844)

mvo5:
[splitout of https://github.com/osbuild/images/pull/1656]

When we move to bootc we need a way to declare that certain
partitions use the distribution default filesystem as the
partition table itself is generic. This commit adds support
in the YAML to omit the filesystem type, i.e. to just write:
```yaml
    partition_table:
      test_arch:
        partitions:
          - payload_type: filesystem
            payload:
              mountpoint: "/"
```
which will then translated at YAML load time to the default
filesystem declared by the distribution. This allows us to
write a generic YAML based partition table in images that
is the same as the partition table in bootc-image-builder.

/jira-epic HMS-8844

### DIFF
--- a/pkg/distro/defs/loader.go
+++ b/pkg/distro/defs/loader.go
@@ -240,6 +240,10 @@ func NewDistroYAML(nameVer string) (*DistroYAML, error) {
 			if err := v.runTemplates(foundDistro); err != nil {
 				return nil, err
 			}
+			if err := v.setupDefaultFS(foundDistro.DefaultFSType.String()); err != nil {
+				return nil, err
+			}
+
 			foundDistro.imageTypes[name] = v
 		}
 	}
@@ -459,6 +463,44 @@ func (it *ImageTypeYAML) runTemplates(distro *DistroYAML) error {
 			}
 		}
 	}
+	return nil
+}
+
+func (it *ImageTypeYAML) setupDefaultFS(distroDefaultFS string) error {
+	subs := func(pts map[string]*disk.PartitionTable) error {
+		for _, pt := range pts {
+			err := pt.ForEachMountable(func(mnt disk.Mountable, _ []disk.Entity) error {
+				elem, ok := mnt.(*disk.Filesystem)
+				if !ok {
+					return nil
+				}
+				if elem.Type == "" {
+					if distroDefaultFS == "" {
+						return fmt.Errorf("mount %q requires a default filesystem for the distribution but none set", mnt.GetMountpoint())
+					}
+					elem.Type = distroDefaultFS
+				}
+				return nil
+			})
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	// we need to update both the partition tables and all
+	// partition tables overrides
+	if err := subs(it.PartitionTables); err != nil {
+		return err
+	}
+	if it.PartitionTablesOverrides != nil {
+		for _, cond := range it.PartitionTablesOverrides.Conditions {
+			if err := subs(cond.Override); err != nil {
+				return err
+			}
+		}
+	}
+
 	return nil
 }
 
--- a/pkg/distro/defs/loader_test.go
+++ b/pkg/distro/defs/loader_test.go
@@ -315,6 +315,127 @@ image_types:
 	}, partTable)
 }
 
+func TestDefsPartitionTableFilesystemDistroDefault(t *testing.T) {
+	fakeDistrosYaml := `
+distros:
+  - name: test-distro-1
+    defs_path: test-distro
+    default_fs_type: ext4
+`
+	fakeImageTypesYaml := `
+image_types:
+  test_type:
+    filename: test.img
+    platforms:
+      - arch: x86_64
+    partition_table:
+      test_arch:
+        partitions:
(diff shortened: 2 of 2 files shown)


achilleas-k:
Adding the distro default fs type to the distro config is great, we should have done this ages ago. 
 I'm not sure about the magic value `distro-default` in the yaml configs though.  Somehow to me it feels more natural to just use an empty value, implicitly falling back to the default.  I think `distro-default` feels to me like we're inching close to doing variables in our yaml configs.

achilleas-k (dismissed):
Splendid.  Ship it!

supakeen (dismissed):

supakeen:
This got removed from the merge queue due to (I think) the recent changes to `distro.ID`.

achilleas-k:
```diff
diff --git a/pkg/distro/defs/loader_test.go b/pkg/distro/defs/loader_test.go
index 0b5229e57..28faf92d3 100644
--- a/pkg/distro/defs/loader_test.go
+++ b/pkg/distro/defs/loader_test.go
@@ -339,12 +339,12 @@ image_types:
 	baseDir := makeFakeDistrosYAML(t, fakeDistrosYaml, fakeImageTypesYaml)
 	restore := defs.MockDataFS(baseDir)
 	defer restore()
-	distro, err := defs.NewDistroYAML("test-distro-1")
+	td, err := defs.NewDistroYAML("test-distro-1")
 	require.NoError(t, err)
-	it := distro.ImageTypes()["test_type"]
+	it := td.ImageTypes()["test_type"]
 	require.NotNil(t, it)
 
-	partTable, err := it.PartitionTable("test-distro-1", "test_arch")
+	partTable, err := it.PartitionTable(distro.ID{Name: "test-distro", MajorVersion: 1}, "test_arch")
 	require.NoError(t, err)
 	assert.Equal(t, &disk.PartitionTable{
 		Partitions: []disk.Partition{
@@ -390,12 +390,12 @@ image_types:
 	baseDir := makeFakeDistrosYAML(t, fakeDistrosYaml, fakeImageTypesYaml)
 	restore := defs.MockDataFS(baseDir)
 	defer restore()
-	distro, err := defs.NewDistroYAML("test-distro-1")
+	td, err := defs.NewDistroYAML("test-distro-1")
 	require.NoError(t, err)
-	it := distro.ImageTypes()["test_type"]
+	it := td.ImageTypes()["test_type"]
 	require.NotNil(t, it)
 
-	partTable, err := it.PartitionTable("test-distro-1", "x86_64")
+	partTable, err := it.PartitionTable(distro.ID{Name: "test-distro", MajorVersion: 1}, "x86_64")
 	require.NoError(t, err)
 	assert.Equal(t, &disk.PartitionTable{
 		Partitions: []disk.Partition{
```

achilleas-k:
Since Michael's out this week, I rebased and fixed this so we can get it merged.

pkg/distro/defs/loader.go
```diff
@@ -462,6 +466,44 @@ func (it *ImageTypeYAML) runTemplates(distro *DistroYAML) error {
...
+	if err := subs(it.PartitionTables); err != nil {
+		return err
+	}
+	if it.PartitionTablesOverrides != nil {
+		for _, cond := range it.PartitionTablesOverrides.Conditions {
+			if err := subs(cond.Override); err != nil {
+				return err
+			}
+		}
+	}
```
thozza, line 502: This also has a side effect that if the partition table explicitly specifies the FS type for a mountpoint, which is then overridden by the condition that does not define it, it would use the default FS type.

This is expected, given the implementation. However, it is not tested by a unit test.

What I mean is
```go
func TestDefsPartitionTableFilesystemPartTableOverrideDefault(t *testing.T) {
	fakeDistrosYaml := `
distros:
  - name: test-distro-1
    defs_path: test-distro
    default_fs_type: ext4
`
	fakeImageTypesYaml := `
image_types:
  test_type:
    filename: test.img
    platforms:
      - arch: x86_64
    partition_table:
      x86_64:
        partitions:
          - payload_type: filesystem
            payload:
              type: xfs
              mountpoint: "/"
    partition_tables_override:
      conditions:
        "test condition":
          when:
            distro_name: test-distro
          override:
            x86_64:
              partitions:
                - payload_type: filesystem
                  payload:
                    # note that no "type: <fstype>" is set here
                    mountpoint: "/"
`
	baseDir := makeFakeDistrosYAML(t, fakeDistrosYaml, fakeImageTypesYaml)
	restore := defs.MockDataFS(baseDir)
	defer restore()
	td, err := defs.NewDistroYAML("test-distro-1")
	require.NoError(t, err)
	it := td.ImageTypes()["test_type"]
	require.NotNil(t, it)

	partTable, err := it.PartitionTable(distro.ID{Name: "test-distro", MajorVersion: 1}, "x86_64")
	require.NoError(t, err)
	assert.Equal(t, &disk.PartitionTable{
		Partitions: []disk.Partition{
			{
				Payload: &disk.Filesystem{
					Type:       "ext4",
					Mountpoint: "/",
				},
			},
		},
	}, partTable)
}
```

thozza (approved):
LGTM

PR #1714, opened 2025-07-31
disk: pass default fs to FS customizations (MS-8990)

mvo5:
[needed for https://github.com/osbuild/images/pull/1656]

Tradtionally we did not pass the filesystem with the filesystem customizations which means that we always created "xfs". This is problematic on bootc systems that require us to use the container default filesystem and arguably also on fedora that declares its default fs to be "ext4" in our image definitions.

So this commit changes the behavior so that the defaultFs is passed when creating custom mountpoints. This changes nothing on rhel/centos but does change the behavior on fedora - which should be fine as this is a bugfix.

/jira-epic HMS-8844

JIRA: [HMS-8990](https://issues.redhat.com/browse/HMS-8990)

### DIFF
--- a/pkg/disk/btrfs.go
+++ b/pkg/disk/btrfs.go
@@ -61,6 +61,10 @@ func (b *Btrfs) GetChild(n uint) Entity {
 	return &b.Subvolumes[n]
 }
 func (b *Btrfs) CreateMountpoint(mountpoint, defaultFs string, size uint64) (Entity, error) {
+	if defaultFs != "btrfs" {
+		return nil, fmt.Errorf("only btrfs mountpoints are supported with btrfs subvolumes not %q", defaultFs)
+	}
+
 	name := mountpoint
 	if name == "/" {
 		name = "root"
--- a/pkg/disk/lvm.go
+++ b/pkg/disk/lvm.go
@@ -76,6 +76,10 @@ func (vg *LVMVolumeGroup) GetChild(n uint) Entity {
 }
 
 func (vg *LVMVolumeGroup) CreateMountpoint(mountpoint, defaultFs string, size uint64) (Entity, error) {
+	if defaultFs == "btrfs" {
+		return nil, fmt.Errorf("btrfs under lvm is not supported")
+	}
+
 	filesystem := Filesystem{
 		Type:         defaultFs,
 		Mountpoint:   mountpoint,
--- a/pkg/disk/disk_test.go
+++ b/pkg/disk/disk_test.go
@@ -194,6 +194,9 @@ func TestCreatePartitionTable(t *testing.T) {
 			if ptName == "luks+lvm" {
 				ptMode = partition.AutoLVMPartitioningMode
 			}
+			if ptName == "btrfs" {
+				ptMode = partition.BtrfsPartitioningMode
+			}
 			mpt, err := disk.NewPartitionTable(&pt, bp, uint64(13*MiB), ptMode, arch.ARCH_PPC64LE, nil, "", rng)
 			require.NoError(t, err, "Partition table generation failed: PT %q BP %q (%s)", ptName, bpName, err)
 			assert.NotNil(mpt, "Partition table generation failed: PT %q BP %q (nil partition table)", ptName, bpName)
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -117,6 +117,7 @@ func NewPartitionTable(basePT *PartitionTable, mountpoints []blueprint.Filesyste
 		ensureLVM = len(newMountpoints) > 0
 	case partition.BtrfsPartitioningMode:
 		ensureBtrfs = true
+		defaultFs = "btrfs"
 	default:
 		return nil, fmt.Errorf("unsupported partitioning mode %q", mode)
 	}


pkg/disk/lvm.go
```diff
@@ -73,10 +75,13 @@ func (vg *LVMVolumeGroup) GetChild(n uint) Entity {
 	return &vg.LogicalVolumes[n]
 }
 
-func (vg *LVMVolumeGroup) CreateMountpoint(mountpoint string, size uint64) (Entity, error) {
+func (vg *LVMVolumeGroup) CreateMountpoint(mountpoint, defaultFs string, size uint64) (Entity, error) {
+	if defaultFs == "" {
+		defaultFs = "xfs"
+	}
```
thozza, line 81: It is not ideal that this exact code is duplicated in all structures that implement the `MountpointCreator` interface, which creates a small risk of going out of sync.

I don't have a good idea how to solve this, other than trying to implement this fallback logic at the lowest possible level.

thozza (approved):
Thanks.

In general, this looks OK, I just don't like that the "default" behavior is implemented in two places.

In addition, this made me realize that we should consider defining `FSType` as an enum or at least as constants, instead of using string literals everywhere.

pkg/disk/lvm.go
achilleas-k: Or maybe it should just be an error?  It's a breaking API change for the `disk` package, but we make no guarantees about stability there.

I'd even go further and do some sanity checks here.  Like `Btrfs{}.CreateMountpoint("/path", "xfs", 10)` and `LVMVolumeGroup{}.CreateMountpoint("/path", "btrfs", 10)` should return errors I think.
Likewise, `T.CreateMountpoint("/path", "", 10)` should probably be an error for all types `T` except `Btrfs`.

mvo5: I like the suggestion, I looked into it but it raises a bunch of question. I will try harder and also write about some the issue I encountered here in a bit. For now I went with a tweak to the existing code that avoids the duplication that thozza mentioned above (i.e. only a single place to select the defaultFS).

mvo5: I like the suggestion, I started looking into it but adding logic into the individual CreateMountpoint() helpers raises some questions:

btrfs its both a filesystem and a partition mode. So when asking a fedora image (with defaultFs ext4) to create an extra "/stuff" mountpoint in btrfs partitioning mode we will call "CreateFilesystem()" once early to ensure we have a /boot partition. Here we need to call CreatePartition with default fs: ext4. Then we add the customization for "/stuff" which also calls CreateMountpoint() - because the mountpoint is under an btrfs root it automatically becomes btrfs. To follow the suggestion here we would have to set a different `defaultFS` for the CreateMountpoint(). So it would become something like:
```go
diff --git a/pkg/disk/btrfs.go b/pkg/disk/btrfs.go
index 2b360b39c..ed4eeb0a5 100644
--- a/pkg/disk/btrfs.go
+++ b/pkg/disk/btrfs.go
@@ -61,6 +61,10 @@ func (b *Btrfs) GetChild(n uint) Entity {
        return &b.Subvolumes[n]
 }
 func (b *Btrfs) CreateMountpoint(mountpoint, defaultFs string, size uint64) (Entity, error) {
+       if defaultFs != "btrfs" {
+               return nil, fmt.Errorf("internal error: CreateMountpoint for btrfs called on %q with fs: %q", mountpoint, defaultFs)
+       }
+
        name := mountpoint
        if name == "/" {
                name = "root"
index fd285b653..97bd884e3 100644
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -152,6 +152,12 @@ func NewPartitionTable(basePT *PartitionTable, mountpoints []blueprint.Filesyste
                }
        }
 
+       if ensureBtrfs {
+               // after we added thigs like /boot we now switch to
+               // a new defaultFs for "btrfs"
+               defaultFs = "btrfs"
+       }
+
        // second pass: deal with new mountpoints and newly created ones, after switching to
        // the LVM layout, if requested, which might introduce new mount points, i.e. `/boot`
        _, err := newPT.applyCustomization(newMountpoints, defaultFs, true)

```
but that would also not work because NewPartitionTable() can already run on a btrfs partition table and in this case even the first run already would have to have a different fs.

The other thing I'm contemplating about this is code organization. If we add code in LVMVolumeGroup{}.CreateMountpoint() that checks for btrfs then now the lvm code knows about btrfs, something it ideally would not have to know about (this one I'm probably overthinking).

Anyway, happy to explore this further but I'm not sure we really need it now given that `defaultFs` right now is more a hint.

achilleas-k: > If we add code in LVMVolumeGroup{}.CreateMountpoint() that checks for btrfs then now the lvm code knows about btrfs, something it ideally would not have to know about (this one I'm probably overthinking).

But it doesn't need to know about btrfs does it?  It only needs to know the list of filesystems it supports (xfs, ext4, swap, and maybe vfat).  Anything else is invalid.

mvo5: I updated the PR to do the suggested error checking now.

pkg/disk/btrfs.go
```diff
@@ -61,6 +61,10 @@ func (b *Btrfs) GetChild(n uint) Entity {
 	return &b.Subvolumes[n]
 }
 func (b *Btrfs) CreateMountpoint(mountpoint, defaultFs string, size uint64) (Entity, error) {
+	if defaultFs != "btrfs" {
+		return nil, fmt.Errorf("only btrfs mountpoints are supported with btrfs subvolums not %q", defaultFs)
```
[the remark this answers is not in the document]
mvo5: Thanks! Fixed.

PR #1715, opened 2025-08-01
osbuild: inspect wrapper (HMS-8973)

supakeen:
Allow for calling `osbuild --inspect`, returning a deserialized manifest which has `ID`'s filled in stages and a helper method on the pipeline to get its last stage ID; which is the same as the pipeline ID.

### DIFF
--- a/pkg/osbuild/osbuild-exec.go
+++ b/pkg/osbuild/osbuild-exec.go
@@ -144,3 +144,16 @@ func OSBuildVersion() (string, error) {
 	version = strings.TrimSpace(version)
 	return version, nil
 }
+
+// OSBuildInspect converts a manifest to an inspected manifest.
+func OSBuildInspect(manifest []byte) ([]byte, error) {
+	cmd := exec.Command("osbuild", "--inspect")
+	cmd.Stdin = bytes.NewBuffer(manifest)
+
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+
+	return out, nil
+}
--- a/pkg/osbuild/osbuild.go
+++ b/pkg/osbuild/osbuild.go
@@ -2,6 +2,10 @@
 // OSBuild (schema v2) types.
 package osbuild
 
+import (
+	"fmt"
+)
+
 const (
 	// should be "^\\/(?!\\.\\.)((?!\\/\\.\\.\\/).)+$" but Go doesn't support lookaheads
 	// therefore we have to instead check for the invalid cases, which is much simpler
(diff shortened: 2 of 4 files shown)


pkg/osbuild/osbuild-exec.go
```diff
@@ -144,3 +144,37 @@ func OSBuildVersion() (string, error) {
 	version = strings.TrimSpace(version)
 	return version, nil
 }
+
+// OSBuildInspect converts a manifest to an inspected manifest.
+func OSBuildInspect(manifest []byte) ([]byte, error) {
```
mvo5, line 149: Go can help with a lot of this, I think:
```go
func OSBuildInspect(manifest []byte) ([]byte, error) {
	cmd := exec.Command("osbuild", "--inspect")
	cmd.Stdin = bytes.NewBuffer(manifest)
	output, err := cmd.Output()
	if err != nil {
		return nil, err
	}
	return output, nil
}
```
is enough - I did not test though but I'm 80% confident

pkg/osbuild/osbuild-exec.go
```diff
@@ -144,3 +144,16 @@ func OSBuildVersion() (string, error) {
...
+	cmd := exec.Command("osbuild", "--inspect")
+	cmd.Stdin = bytes.NewBuffer(manifest)
+
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+
+	return out, nil
+}
```
lzap, line 159: This swallows any possible stderr output.

mvo5 (commented):
Thanks for this. I added a few comments inline for your consideration. A bit of a meta-question. The PR descriptions describes "what" the PR is doing (allowing to run osbuild-inspect) but it would be nice to expand a bit of the "why" part, i.e. what this will be used for, how its useful etc. I'm currently not sure what the use-case is just from reading it.

achilleas-k (approved):
Cheers!

lzap (approved):

pkg/osbuild/osbuild.go
```diff
@@ -51,3 +56,31 @@ func (p *Pipeline) AddStages(stages ...*Stage) {
 		p.AddStage(stage)
 	}
 }
+
+// Take some bytes and deserialize them into a Manifest; mostly used to take
+// an inspected manifest
+func NewManifestFromBytes(data []byte) (*Manifest, error) {
```
mvo5, line 62: I tried this for a real manifest but it did not go very well, it dies with `json: cannot unmarshal object into Go struct field Stage.pipelines.stages.inputs of type osbuild.Inputs` and looking at our osbuild inputs it seems we need to add quite a bit of custom unmarshaling in this area. What is the idea here? Will we expand the osbuild package so that it can unmarshal things properly back? it would be nice for tests as well to be able to transform the json back into more managable structs.

PR #1716, opened 2025-08-01
Add crc image types as image aliases (HMS-8730)

kingsleyzissou:
Since we support aliases for image definitions, we can add the image
types used in `image-builder-crc` to the alias of the corresponding
image type in the distro definitions.

### DIFF
--- a/pkg/imagefilter/filter.go
+++ b/pkg/imagefilter/filter.go
@@ -77,6 +77,17 @@ type filter struct {
 	terms []term
 }
 
+func containsAlias(term term, aliases []string) bool {
+	result := false
+	for _, alias := range aliases {
+		if term.pattern.Match(alias) {
+			result = true
+			break
+		}
+	}
+	return result
+}
+
 // Matches returns true if the given (distro,arch,imgType) tuple matches
 // the filter expressions
 func (fl filter) Matches(distro distro.Distro, arch distro.Arch, imgType distro.ImageType) bool {
@@ -89,13 +100,17 @@ func (fl filter) Matches(distro distro.Distro, arch distro.Arch, imgType distro.
 			m1 := term.pattern.Match(distro.Name())
 			m2 := term.pattern.Match(arch.Name())
 			m3 := term.pattern.Match(imgType.Name())
-			m = m && (m1 || m2 || m3)
+			m4 := containsAlias(term, imgType.Aliases())
+			m = m && (m1 || m2 || m3 || m4)
 		case prefixDistro:
 			m = m && term.pattern.Match(distro.Name())
 		case prefixArch:
 			m = m && term.pattern.Match(arch.Name())
 		case prefixType:
-			m = m && term.pattern.Match(imgType.Name())
+			// Check the main image type name
+			m1 := term.pattern.Match(imgType.Name())
+			m2 := containsAlias(term, imgType.Aliases())
+			m = m && (m1 || m2)
 			// mostly here to show how flexible this is
 		case prefixBootmode:
 			m = m && term.pattern.Match(imgType.BootMode().String())
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -608,7 +608,7 @@ image_types:
         image_format: "vagrant_virtualbox"
 
   "server-qcow2": &server_qcow2
-    name_aliases: ["qcow2"]
+    name_aliases: ["qcow2", "guest-image"]
     filename: "disk.qcow2"
     mime_type: "application/x-qemu-disk"
     environment: *kvm_env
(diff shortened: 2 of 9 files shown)


pkg/distro/defs/rhel-10/distro.yaml
```diff
@@ -837,6 +838,9 @@ image_types:
 
   "vagrant-libvirt": &vagrant_libvirt
     <<: *qcow2
+    # we have to reset the aliases otherwise this type
+    # will inherit the name aliases causing a conflict
+    name_aliases: []
```
kingsleyzissou, line 843: I was unsure whether to do this or maybe redefine a new base type? Will leave this open for comments & suggestions

mvo5:
Thank you \o/ - does that mean we can consolidate the aliases all in images and drop the custom naming in -crc :) I think it would be great if we had only a single place for th "source of truth"!

pkg/distro/defs/rhel-9/distro.yaml
```diff
@@ -2073,7 +2086,9 @@ image_types:
         - *anaconda_pkgset
 
   gce:
-    name_aliases: ["gce-rhui"]
+    # this image type is set to `gcp` in image-builder-crc
+    # & `osbuild-composer`, so set the alias here
```
thozza, line 2090: This comment, IMHO,  does not provide much value, since the situation is exactly the same with other added aliases above, which don't have this comment.

pkg/distro/defs/loader.go
```diff
@@ -68,6 +68,7 @@ type DistroYAML struct {
 	// for {{.Major}}, {{.Minor}} which will be expanded by the
 	// upper layers.
 	Name             string            `yaml:"name"`
+	Aliases          []string          `yaml:"name_aliases"`
```
thozza, line 71: As far as I can tell, this is not necessary, because it adds `name_aliases` for the distribution, not the image type.

At this point, we should not be handling aliases for distributions in `images`, since this also affects repo config loading. Moreover, distro aliases are, for now, intentionally handled on the higher levels - osbuild-composer. So if we were to bring that feature to `images`, so that, i.e., ibcli can use it, we should bring it from osbuild-composer, together with the unit and functional tests that we have for it.

pkg/distro/generic/distro.go
```diff
@@ -115,6 +113,10 @@ func (d *distribution) Name() string {
 	return d.DistroYAML.Name
 }
 
+func (d *distribution) Aliases() []string {
+	return d.DistroYAML.Aliases
+}
```
thozza, line 118: `Aliases()` on the distro level should be removed, since it is not used or required at this point as far as I can tell.

pkg/imagefilter/filter.go
```diff
@@ -89,13 +89,34 @@ func (fl filter) Matches(distro distro.Distro, arch distro.Arch, imgType distro.
...
+			// Also check all aliases
+			if !m1 {
+				for _, alias := range imgType.Aliases() {
+					if term.pattern.Match(alias) {
+						m1 = true
+						break
+					}
+				}
+			}
+			m = m && m1
```
thozza, line 119: It would be great to extend unit tests to cover matching aliases as well.

thozza (changes requested):
Thank you. Please see my inline comments.

pkg/distro/generic/distro.go
```diff
@@ -30,9 +30,7 @@ const (
 	blueprintPkgsKey = "blueprint"
 )
 
-var (
-	ErrDistroNotFound = errors.New("distribution not found")
-)
+var ErrDistroNotFound = errors.New("distribution not found")
```
thozza: Nitpick: this change could be dropped.

pkg/imagefilter/filter.go
```diff
@@ -89,13 +89,34 @@ func (fl filter) Matches(distro distro.Distro, arch distro.Arch, imgType distro.
...
+			// Also check image type aliases
+			m4 := false
+			if !m3 {
+				for _, alias := range imgType.Aliases() {
+					if term.pattern.Match(alias) {
+						m4 = true
+						break
+					}
+				}
+			}
```
thozza, line 101: This is already duplicated twice in this file, and aliases are checked also elsewhere, AFAIK. Would it make sense to factor this out to a dedicated function that checks the aliases? Or make it an imageType method?

kingsleyzissou:
@thozza I accidentally reverted some of the changes this morning. But think I may have got them all back now

pkg/imagefilter/filter.go
kingsleyzissou: I wonder if we could just have a dedicated function here for now and then if we need it in other places we could move it to an imageType method?

thozza: Sure. The way aliases work in `architecture.GetImageType()` is quite different compared to what the `filter` package does.

thozza (commented):
Looks good from my point of view. However, you had previously added some unit tests for the `filter` package's aliases, but these are now gone. It would be great to add them back.

thozza (approved):
Thanks. LGTM.

PR #1717, opened 2025-08-01
repositories: fedora-44

supakeen:
Introduce Fedora 44 (pointing to rawhide) early. Rawhide will become Fedora-44 on the 12th of August. Let's have it available earlier so we don't run into issues when Koji tries to build Fedora-44.

Note that the repositories here are (kinda) bogus but the distro `Fedora-44` *must* exist on that date as it's when `rawhide` builds will start reporting as it.


lzap (approved):

mvo5:
Actually it might be worth also editing pkg/distro/defs/distros.yaml and add it there explicitly as "prefix: true"

PR #1718, opened 2025-08-01
distro: fix openscap version check for rhel8

mvo5:
This commit partly reverts `e0182a9f93` which changed some ostree validation but also changed in the rhel8 checkOptions() code:
```
pkg/distro/rhel8/imagetype.go:

-		if common.VersionLessThan(t.arch.distro.osVersion, "8.7") {
+		if t.arch.distro.osVersion == "9.0" {
```
(c.f. https://github.com/osbuild/images/pull/91/files#diff-041430a8039876cfa501c6018f198eb7e7bc03fe206e5422112d43241f8b22b2L377)

It looks like a copy/paste mistake, it also is here for a while so maybe its not important. Stumbled over this while reviewing the (excellent) checkOptions tests PR#1701.

### DIFF
--- a/pkg/distro/generic/options.go
+++ b/pkg/distro/generic/options.go
@@ -407,7 +407,8 @@ func checkOptionsRhel8(t *imageType, bp *blueprint.Blueprint, options distro.Ima
 	}
 
 	if osc := customizations.GetOpenSCAP(); osc != nil {
-		if t.Arch().Distro().OsVersion() == "9.0" {
+		// only add support for RHEL 8.7 and above.
+		if common.VersionLessThan(t.Arch().Distro().OsVersion(), "8.7") {
 			return warnings, fmt.Errorf("OpenSCAP unsupported os version: %s", t.Arch().Distro().OsVersion())
 		}
 		if !oscap.IsProfileAllowed(osc.ProfileID, t.arch.distro.DistroYAML.OscapProfilesAllowList) {


lzap (approved):

achilleas-k:
This is now failing to generate manifests because we had oscap customizations set for configs with distro < RHEL 8.7.  I think the check was there because we only added oscap support in 8.7, but I don't know if it means it's not supported at all in previous versions (@kingsleyzissou?).

kingsleyzissou:
> This is now failing to generate manifests because we had oscap customizations set for configs with distro < RHEL 8.7. I think the check was there because we only added oscap support in 8.7, but I don't know if it means it's not supported at all in previous versions (@kingsleyzissou?).

Oh ya that's exactly it the check was there because that's when the feature was added initially. And yeah it wouldn't be supported at all, I don't think there is an osbuild stage for it in older versions than 8.7

achilleas-k:
> I don't think there is an osbuild stage for it in older versions than 8.7

But you can build a RHEL 8.4 image on RHEL 9.6.  But I wonder if the customization relies on an oscap package that was never released in 8.4.

kingsleyzissou:
> > I don't think there is an osbuild stage for it in older versions than 8.7
> 
> But you can build a RHEL 8.4 or 8.6 image on RHEL 9.6. But I wonder if the customization relies on an oscap package that was never released before 8.7.

Oh sorry, I misunderstood this. Ya in that case if you're building from a newer distro you should be fine. I think the package will be there... it's just that they made some optimizations in later releases (like checking if the environment is an osbuild build root).

thozza (approved):

mvo5:
This got merged indirectly via https://github.com/osbuild/images/pull/1872/commits and specially https://github.com/osbuild/images/pull/1872/commits/4a8284323a213e534214aba99addefc5efc24f6b

PR #1721, opened 2025-08-04
rhel10: system-reinstall-bootc back in c10s, new in qcow2 (HMS-8991)

supakeen:
Re-introduce system-reinstall-bootc in c10s after some discussion. Also introduce it in the qcow2-based image types in general as I think we want this available everywhere?


mvo5 (approved):
Fine with me

supakeen:
> Love the centos/RHEL unification, thank you! ❤️
> 
> Out of curiosity, can you diff the qcow2 depsolved package set before/after the change? I'm just curious how many packages this innocent change brings in. Overall, I'm for it, consistency is good.

Sure, I can take a look.

PR #1727, opened 2025-08-05
blueprint: installer kernel args (RHEL-107495)

supakeen:
Allows for the customization of kernel arguments as used by installer environments. Anaconda takes certain actions based on the kernel arguments supplied such as configuring the installed system with them.

We've had the ability to customize these for disk images for a while now. We also need them for installers.


supakeen:
Changed this to apply kernel arguments straight from `kernel.append` instead of a new blueprint, ready for review :)

bcl (approved):
Looks good

thozza (approved):
LGTM

PR #1728, opened 2025-08-05
pkg/awscloud: various cleanups (HMS-8869)

thozza:
This PR does a few cleanups for the `awscloud` package, which I noticed while porting it to AWS SDK v2.

Most notably, I changed all function argument types to be values, instead of pointers, in case the code didn't expect it to be `nil` and thus the value was not optional. Additionally, I modified the `AWS.Register()` method to accept `platform.BootMode` for the boot mode and `arch.Arch` for the image architecture.

### DIFF
--- a/pkg/cloud/awscloud/awscloud.go
+++ b/pkg/cloud/awscloud/awscloud.go
@@ -682,14 +682,14 @@ func (a *AWS) GetInstanceAddress(instanceID string) (string, error) {
 }
 
 // DeleteEC2Image deletes the specified image and its associated snapshot
-func (a *AWS) DeleteEC2Image(imageID, snapshotID *string) error {
+func (a *AWS) DeleteEC2Image(imageID, snapshotID string) error {
 	var retErr error
 
 	// firstly, deregister the image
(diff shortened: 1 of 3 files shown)


pkg/cloud/awscloud/awscloud.go
```diff
@@ -417,18 +417,13 @@ func (a *AWS) ShareImage(ami string, snapshotIDs, userIDs []string) error {
...
 	olog.Println("[AWS] 💿 Sharing ec2 AMI")
 	var launchPerms []ec2types.LaunchPermission
-	for _, id := range uIds {
+
+	for idx := range userIDs {
 		launchPerms = append(launchPerms, ec2types.LaunchPermission{
-			UserId: id,
+			UserId: aws.String(userIDs[idx]),
 		})
 	}
```
lzap, line 428: Yes, on top of that iteration over index is also faster. Nice.

lzap (approved):
Oh yes, more of these.

_Every time you kill a 👉🏻 a new 🐈 is born!_

achilleas-k (approved):
A real pointer massacre.

LGTM

PR #1732, opened 2025-08-06
manifest,distro: move pipeline role out of imagetypes

mvo5:
This is another attempt to move the `{Build,Payload}Pipeline` out of the image type YAML.

It follows the suggestion from @achilleas-k (thanks!) in https://github.com/osbuild/images/pull/1696#issuecomment-3156169592 and move the pipeline role into the pipeline itself. Then it drops the `{Build,Payload}Pipelines` methods from the image types and from the YAML and use the equivalent `manifest.Build,Payload}Pipelines`.

### DIFF
--- a/pkg/distro/defs/loader.go
+++ b/pkg/distro/defs/loader.go
@@ -371,8 +371,6 @@ type ImageTypeYAML struct {
 	DefaultSize uint64 `yaml:"default_size"`
 	// the image func name: disk,container,live-installer,...
 	Image                  string            `yaml:"image_func"`
-	BuildPipelines         []string          `yaml:"build_pipelines"`
-	PayloadPipelines       []string          `yaml:"payload_pipelines"`
 	Exports                []string          `yaml:"exports"`
 	RequiredPartitionSizes map[string]uint64 `yaml:"required_partition_sizes"`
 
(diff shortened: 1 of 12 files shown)


pkg/distro/defs/loader.go
```diff
@@ -370,8 +370,8 @@ type ImageTypeYAML struct {
 	DefaultSize uint64 `yaml:"default_size"`
 	// the image func name: disk,container,live-installer,...
 	Image                  string            `yaml:"image_func"`
-	BuildPipelines         []string          `yaml:"build_pipelines"`
-	PayloadPipelines       []string          `yaml:"payload_pipelines"`
+	XXXUnused              []string          `yaml:"build_pipelines"`
+	XXXUnused2             []string          `yaml:"payload_pipelines"`
```
achilleas-k, line 374: :thinking: ?

achilleas-k (commented):
What a great idea :sunglasses: 

Second commit message could use an update though.

pkg/distro/defs/loader.go
mvo5: *cough* quick draft *cough* :) fixed!

thozza (approved):
Thanks. While I'm not a fan of hardcoding the pipeline role in the pipeline itself, this is a reasonable incremental change. Moreover, it makes a lot of sense not to specify the pipelines in the image definitions YAML

PR #1734, opened 2025-08-07
manifest: extract `Export` from pipeline into its own interface

mvo5:
We currently have `Export` in the basic piepline interface and then a default implementation that panic()s when called. This commit removes the Export and the panic and uses the type system via a new `ExportingPipeline` to ensure type safety without the panic.


supakeen (approved):

bcl (approved):
Nice!

PR #1735, opened 2025-08-07
many: drop `Exports()` from imagetypes and use the manifest

mvo5:
[draft as it needs consensus first and also needs updates to ibcli/composer, based on the excellent suggestion from @achilleas-k - there is also the question how this will fit with bootc/bib, see https://github.com/osbuild/images/pull/1736 which has exports for all of raw/qcow2/etc currently, this side also needs tweaks probably a selector in the images function]

This PR removes the `distro.ImageType.Exports()` method and replaces it with the `manifest.Exports()`.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -124,7 +124,8 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("[ERROR] manifest generator creation failed: %w", err)
 	}
-	if err := mg.Generate(config.Blueprint, distribution, imgType, arch, &config.Options); err != nil {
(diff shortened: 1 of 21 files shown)


achilleas-k:
Related: #1738

supakeen:
I'd still like this so I'm reopening.

supakeen:
> > I'd still like this so I'm reopening.
> 
> 
> 
> Are you going to take it over?  I'd throw it onto my pile, but I don't know if I'll be able to get to it soon.

Yea somewhere after FOSDEM.

PR #1737, opened 2025-08-08
many: get anaconda modules from `installer_config` (HMS-9015)

supakeen:
When customizations for Anaconda modules were introduced we started using a default module state mapping. Recently I adjusted this mapping to include the runtime module, as without this module certain kickstart verbs and scripts no longer get ran starting with Fedora 42.

Doing this unconditionally means that the module also got enabled on RHEL 8. On RHEL 8 we don't use the `activatable_modules` configuration option of Anaconda as it didn't exist, thus this module is now absolutely required and without it Anaconda won't start.

This led to non-booting Anaconda installers on RHEL-8.

This PR moves the default module states into the `installer_config` defined on image types. We might want to revisit later to see if we want to change the values used here but for now I want to only introduce changes to the RHEL-8 image types.

---

The last commit fixes the linked issue, all the other commits are there to make it possible :)


pkg/osbuild/anaconda_stage.go
```diff
@@ -76,17 +63,15 @@ func filterEnabledModules(moduleStates map[string]bool) []string {
...
 
 	return &AnacondaStageOptions{
 		KickstartModules: filterEnabledModules(states),
 	}
 }
 
 func NewAnacondaStageOptions(enableModules, disableModules []string) *AnacondaStageOptions {
-	states := defaultModuleStates()
-	setModuleStates(states, enableModules, disableModules)
+	states := moduleStates(enableModules, disableModules)
```
achilleas-k, line 74: Without `defaultModuleStates()`, isn't this whole function equivalent to
```go
func NewAnacondaStageOptions(enableModules []string) *AnacondaStageOptions {
	slices.Sort(enableModules)
	return &AnacondaStageOptions{
		ActivatableModules: slices.Compact(enableModules),
	}
}

```
?

Before, we needed to iterate through to apply user customizations to change the default states, but now it's all set outside the function already.

achilleas-k: Now that I think about it, it looks like this PR drops the ability for the blueprint to disable a module that's enabled by default.

achilleas-k (changes requested):

achilleas-k (approved):
Thanks.  LGTM!

PR #1739, opened 2025-08-08
platform: drop hardcoded platforms and rename PlatformConf

mvo5:
many: drop the hardcoded platforms

The hardcoded platforms like platform.AArch64 are not used anymore,
all platform information comes from YAML. So drop them from the
code and replace them in the tests with their dynamic equivalent.

This also removes the platform.X86_64 from a bunch of tests and
replaces that too. We *could* have left a compatibility X86
platform but given that we don't actually use it outside tests
and that everything comes from yaml it seems more dangerous
to keep than to remove as it might lead to expectation that
are just not true (i.e. platform.X86 is not used at all
anymore).

---

platform: rename PlatformConf -> platform.Data

The platform.PlatformConf is not a great name, platform.Data
is (a bit) better. Ideas welcome. Ideal would be
platform.Platform but that name is taken already by the
interface. We could consider droping the interface now that
we have only a single platform data struct but that is
probably something for a followup.


bcl:
Minor complaint :) Having Platform in the name is redundant since it is part of the platform module. Makes things more verbose than they need to be.

achilleas-k (approved):
Nice cleanup.  Thanks!

supakeen:
@achilleas-k seems it is unmergeable due to some API changes :)

supakeen (approved):
Don't particularly like `platform.Data` but also don't have a better idea.

PR #1742, opened 2025-08-11
pkg/awscloud: make additional methods usable for osbuild-composer

thozza:
Modify the `DeleteEC2Image()` and `TerminateInstancesEC2()` methods, so that they are easier to use by osbuild-composer's service maintenance binary. This will allow to de-duplicate these functions in osbuild-composer and use the implementation from `osbuild/images`.

/jira-epic HMS-8802


achilleas-k (approved):
Very nice.  Thank you.

LGTM

PR #1744, opened 2025-08-11
Fix creation of ESP when it's defined in the blueprint [HMS-9032]

### LINKED ISSUE #1726: addPlainPartition() does not check for /boot/efi
https://github.com/osbuild/images/blob/6a13edd5a7e1ed71a09abbe2e112ff98fb2410a4/pkg/disk/partition_table.go#L1422-L1434

achilleas-k:
This PR fixes several issues with the creation of the ESP when it is defined in the blueprint.

1. Fixes creating duplicate ESPs.  The function that checked if the ESP was defined in the blueprint only checked for partitions with type "plain".  However an empty type is also valid (and is equivalent to "plain").
2. Fixes the automatic selection of the partition type (GUID or DOS ID) when the mountpoint is `/boot/efi` and the type is not defined.
3. Adds the standard fstab mount options for the custom ESP.

Fixes #1726.


bcl (approved):
Looks good!

PR #1745, opened 2025-08-11
Set the default menu option for grub2 isos

bcl:
This switches the default menu from 0 to 1 on iso image types.


achilleas-k (dismissed):
Thank you!  LGTM

pkg/platform/platform.go
```diff
@@ -137,13 +137,15 @@ type Platform interface {
 	GetBootFiles() [][2]string
 	GetBootloader() Bootloader
 	GetFIPSMenu() bool
+	GetDefaultMenu() int
```
thozza, line 140: I'm not really sure if the default boot menu should be set / customized via the `Platform` definition :thinking: Having it in the `ImageConfig` and specified per image type may make more sense.

thozza (commented):
Thank you.

I'm not convinced that setting/customizing the boot menu via the `Platform` is the best place to do it. It seems to be a result of the fact that ISOs have their platform defined, so it was somewhat convenient. However, in my opinion, it would make more sense logically in the `ImageConfig` or `InstallerConfig`.

I'm wondering, what was your reasoning for picking the `Platform`?

supakeen (changes requested):
Putting this on request changes so it's no longer green and no one accidentally merges.

bcl:
> I'm wondering, what was your reasoning for picking the `Platform`?

Probably because that's where I put the fips menu bool. Should that be moved as well?

supakeen (approved):
Thanks, fits better in installer config. I have a small remark on the timeout and default integers but since I'll wire it up in a followup it doesn't matter yet here :)

thozza (dismissed):
LGTM, but there are conflicts :crying_cat_face:

achilleas-k (approved):

thozza (approved):

PR #1748, opened 2025-08-12
pkg/upload: copy over Pulp upload implementation from osbuild-composer (HMS-9043)

thozza:
SSIA

Additionally, change the code not to call `logrus`, but `olog` instead.

/jira-epic HMS-8802

JIRA: [HMS-9043](https://issues.redhat.com/browse/HMS-9043)


achilleas-k (approved):
I'm not entirely sure we still use this client, but let's do the move anyway.  We can clean it up later if we don't need it.

PR #1750, opened 2025-08-12
azure: add support for booting a VM & azure boot tests (HMS-9168)

croissanne:
ssia


croissanne:
Ok interesting, it seems the 8.4 repos got a little update, because building a 8.4 vhd with the current snapshots results in the vm fialing to boot on azure.

PR #1752, opened 2025-08-13
cmd: delete unused binaries

thozza:
- `cmd/osbuild-mock-openid-provider` - The binary is used for testing in the osbuild-composer repository and it is a leftover from the time when images repo was forked off the composer repository.
- `cmd/osbuild-composer-image-definitions` - The binary is not used anywhere and nowadays we have much nicer ways to
list supported combinations of distro / arch / image type. Moreover, there is cmd/list-images, which does the same thing and it actually used by tests.
- `cmd/osbuild-package-sets` - This binary is not used anywhere and given that the package sets are now in YAML, it most probably lost its original purpose.


lzap:
There is a `README` in the `cmd/` that explains what each binary does, can you clean it up too?

kingsleyzissou (approved):

lzap:
Yeah sorry it was in composer.

PR #1757, opened 2025-08-14
installer(s): refactoring, simplification, deduplication

### LINKED ISSUE #1756: Consolidate ISO options from `ImageConfig` to `InstallerConfig`
Some ISO-specific options are still in the `ImageConfig` and some of them directly collide with the options in the `InstallerConfig`. The `ImageConfig.ISORootFsType` and `InstallerConfig.SquashfsRootfs` are good examples, where both affect the same attribute of how the ISO is built.

Ideally, all `ISO*` options from the `ImageConfig` should be moved to `InstallerConfig` and in the case of `ISORootFSType`, also merged with the existing `InstallerConfig` option that does almost the same thing.

https://github.com/osbuild/images/blob/0b1224086249a94fe95280ae8f8e6bcdf149c86f/pkg/distro/image_config.go#L138-L143

supakeen:
This PR will likely be draft for a while but feel free to put your other annoyances/papercuts with the installer things in here as comments. Please not that certain commits duplicate things and later commits deduplicate them; I'm doing that to keep commits to a single purpose so they don't move a section around *and* deduplicate at the same time.

Related: #1756.

---

This PR will, at times, probably contain things that don't make sense; that's because I'm trying things out :)

There are (so far) 3 distinct phases to this PR, the first two commits resolved #1756. Then there's a batch of commits that clean up the YAML as I went through it. Then there's the introduction of `InstallerCustomizations` which is similar to `OSCustomizations`, except for installer.

The introduction of `InstallerCustomizations` hopefully leads to more deduplication in the image funcs down the road.

### DIFF
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -1574,7 +1574,21 @@ image_types:
       - "bootiso"
     exports: ["bootiso"]
     required_partition_sizes: *default_required_dir_sizes
-    installer_config: *default_installer_config
+    installer_config: 
+      <<: *default_installer_config
+      # for some reason the live-installer never had or never took into account
+      # the additional dracut modules. This might be a bug but for now we reset
+      # them to empty to avoid changing the live-installer definition
+      additional_dracut_modules: []
+      # also reset the condition that adds dracut modules
+      conditions:
+        <<: *default_installer_config_conditions
+        "f41 dracut requires ifcfg, rootfs is squashfs":
+          when:
+            version_equal: "41"
+          shallow_merge:
+            additional_dracut_modules: []
+
     image_config:
       locale: "en_US.UTF-8"
     package_sets:
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -318,6 +318,56 @@ func osCustomizations(t *imageType, osPackageSet rpmmd.PackageSet, options distr
 	return osc, nil
 }
 
+func installerCustomizations(t *imageType, c *blueprint.Customizations) (manifest.InstallerCustomizations, error) {
+	isc := manifest.InstallerCustomizations{}
+	isc.FIPS = c.GetFIPS()
+	isc.UseLegacyAnacondaConfig = t.ImageTypeYAML.UseLegacyAnacondaConfig
+
+	installerConfig, err := t.getDefaultInstallerConfig()
+	if err != nil {
+		return isc, err
+	}
+
+	if installerConfig != nil {
+		isc.EnabledAnacondaModules = append(isc.EnabledAnacondaModules, installerConfig.EnabledAnacondaModules...)
+		isc.AdditionalDracutModules = append(isc.AdditionalDracutModules, installerConfig.AdditionalDracutModules...)
+		isc.AdditionalDrivers = append(isc.AdditionalDrivers, installerConfig.AdditionalDrivers...)
+
+		if menu := installerConfig.DefaultMenu; menu != nil {
+			isc.DefaultMenu = *menu
+		}
+
+		if isoroot := installerConfig.ISORootfsType; isoroot != nil {
+			isc.ISORootfsType = *isoroot
+		}
+
+		if isoboot := installerConfig.ISOBootType; isoboot != nil {
+			isc.ISOBoot = *isoboot
+		}
+
+		// Put the kickstart file in the root of the iso, some image
+		// types (like rhel10/image-installer) put them there, some
+		// others (like fedora/image-installer) do not and because
+		// its not uniform we need to make it configurable.
+		// XXX: unify with rhel-11 ? or rhel-10.x?
+		if rootkickstart := installerConfig.ISORootKickstart; rootkickstart != nil {
+			isc.ISORootKickstart = *rootkickstart
+		}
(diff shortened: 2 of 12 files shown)


pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -17,16 +17,16 @@ import (
...
+type ISORootfsType uint64
 
 // These constants are used by the ISO images to control the style of the root filesystem
 const ( // Rootfs type enum
-	SquashfsExt4Rootfs RootfsType = iota // Create an EXT4 rootfs compressed by Squashfs
-	SquashfsRootfs                       // Create a plain squashfs rootfs
-	ErofsRootfs                          // Create a plain erofs rootfs
+	SquashfsExt4Rootfs ISORootfsType = iota // Create an EXT4 rootfs compressed by Squashfs
+	SquashfsRootfs                          // Create a plain squashfs rootfs
+	ErofsRootfs                             // Create a plain erofs rootfs
```
achilleas-k, line 26: The enum values could be renamed here too for the same reason.  In other words, `SquashfsExt4ISORootfs` is a better name than `SquashfsExt4Rootfs` (ignoring the length).

It's fine to leave this for a follow-up though.

pkg/distro/generic/images.go
```diff
@@ -579,34 +604,30 @@ func imageInstallerImage(workload workload.Workload,
...
-		img.AdditionalDrivers = append(img.AdditionalDrivers, installerConfig.AdditionalDrivers...)
-		if img.Kickstart.Unattended {
-			img.AdditionalKernelOpts = append(img.AdditionalKernelOpts, installerConfig.KickstartUnattendedExtraKernelOpts...)
-		}
+	// TODO: left over bits
+	img.InstallerCustomizations.EnabledAnacondaModules = append(img.InstallerCustomizations.EnabledAnacondaModules, anaconda.ModuleUsers)
+
+	if img.Kickstart.Unattended {
+		img.InstallerCustomizations.AdditionalKernelOpts = append(img.InstallerCustomizations.AdditionalKernelOpts, installerConfig.KickstartUnattendedExtraKernelOpts...)
+	}
```
achilleas-k: What does the comment mean here (left over bits)?  A more clarifying TODO or an explanation of what's wrong or out of place would be nice here.

pkg/distro/generic/images.go
```diff
@@ -799,40 +806,16 @@ func iotInstallerImage(workload workload.Workload,
...
-
-		if isoroot := installerConfig.ISORootfsType; isoroot != nil {
-			img.RootfsType = *isoroot
-		}
-
-		if isoboot := installerConfig.ISOBootType; isoboot != nil {
-			img.ISOBoot = *isoboot
-		}
-	}
+	// XXX left over bits
```
achilleas-k: Same here.

data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -1574,7 +1574,20 @@ image_types:
       - "bootiso"
     exports: ["bootiso"]
     required_partition_sizes: *default_required_dir_sizes
-    installer_config: *default_installer_config
+    installer_config: 
+      <<: *default_installer_config
+      # for some reason the live installer didn't have the additional dracut modules, this
+      # might be a bug, reset them to empty?
```
achilleas-k, line 1580: I'd remove the implied uncertainty and spell it out clearer here.  Like
```
For some reason, the live-installer never had or never took into account the
additional dracut modules. This might be a bug but, for now, we reset them to
empty to avoid changing the live-installer definition.
```

Or something to that effect.

pkg/distro/generic/images.go
```diff
@@ -345,6 +345,15 @@ func installerCustomizations(t *imageType, c *blueprint.Customizations) (manifes
 		if isoboot := installerConfig.ISOBootType; isoboot != nil {
 			isc.ISOBoot = *isoboot
 		}
+
+		// Put the kickstart file in the root of the iso, some image
+		// types (like rhel10/image-installer) put them there, some
+		// others (like fedora/image-installer) do not and because
+		// its not uniform we need to make it configurable.
+		// XXX: unify with rhel-11 ? or rhel-10.x?
```
achilleas-k, line 353: Let's turn this into an issue/task so we don't forget about it

achilleas-k (approved):
This is great.  Thanks for all the work!

A couple of minor notes, but no blockers.

LGTM

pkg/distro/generic/images.go
supakeen: Left over bits are things that I feel could move into the `installerCustomization` and `InstallerCustomizations` directly, maybe in the future (mostly `img.Kickstart` things). I'll adjust the comment.

pkg/distro/generic/images.go
supakeen: Left over bits are things that I feel could move into the `installerCustomization` and `InstallerCustomizations` directly, maybe in the future (mostly `img.Kickstart` things). I'll adjust the comment.

pkg/distro/generic/images.go
supakeen: Sure.

supakeen: https://github.com/osbuild/images/issues/1761

PR #1758, opened 2025-08-14
Partially align Azure CVM package list with other Azure cloud images [HMS-9011]

achilleas-k:
This PR adds missing packages from the Azure CVM image type that are considered important for core or common use cases.

This does not completely align the CVM image type with the rest of the Azure image types, but it adds the most important packages that were identified to be missing.  Further refinement might be needed later.


bcl (approved):
Ack

supakeen (approved):

PR #1768, opened 2025-08-18
distrodefs/fedora: add supported customizations for net installer

achilleas-k:
The net installer supports only FIPS and the locale.

https://github.com/osbuild/images/pull/1216#issuecomment-3189972728

~~Draft: Based on #1767~~


mvo5 (approved):
This is so simple we could merge it in one go but fine to keep separate.

supakeen:
It should also support kernel argument customization after #1786 lands.

achilleas-k:
> One of the tests need updating now :)

Ooops.  Fixed!

PR #1769, opened 2025-08-19
many: drop `ISORootKickstart`

supakeen:
Since a while we've had a boolean option that determines if the kickstart file we write for certain image types ends up in the root of an ISO or in the `/usr/share/anaconda/interactive-defaults.ks` path.

This changes things only for the `minimal-installer` in Fedora as all other installer adjacent image types had this set to `true`.

I've tested the `minimal-installer` to still function.

Historically we introduced the `minimal-installer` (then called `image-installer`) to test the Anaconda webui. At that point in time the webui didn't support any kickstarts excepts the one in `interactive-defaults.ks`.

Right now it only leads to confusion. I'll likely re-introduce the ability to modify the kickstart path(s) through the installer configs in the image definitions at a later point in time if we need it instead of having a boolean toggle.

---

Note that with this change the `InteractiveDefaults` name no longer really makes any sense; I might add a commit or two to rename it once I think about better names but I'll likely do that in a follow-up as it's all quite intertwined with eachother.

Note that the `minimal-installer` image type for which checksums change isn't one that we ship or publish anywhere.


bcl (approved):
Looks good!

pkg/image/installer_image_test.go
```diff
@@ -251,12 +250,9 @@ func TestTarInstallerUnsetKSOptions(t *testing.T) {
...
 
 	mfs := instantiateAndSerialize(t, img, mockPackageSets(), nil, nil)
-	// the tar installer doesn't set a custom kickstart path unless the
-	// unattended option is enabled, so the inst.ks option isn't set and
-	// interactive-defaults.ks is used
+
 	assert.Contains(t, mfs, fmt.Sprintf(`"inst.stage2=hd:LABEL=%s"`, isolabel))
-	assert.Contains(t, mfs, fmt.Sprintf("%q", osbuild.KickstartPathInteractiveDefaults))
-	assert.NotContains(t, mfs, "osbuild.ks") // no mention of the default (custom) value
+	assert.Contains(t, mfs, fmt.Sprintf(`"inst.ks=hd:LABEL=%s:/osbuild.ks"`, isolabel))
```
thozza, line 255: Nitpick: Shouldn't the unit test consider `img.Kickstart.Path` value? :thinking:

thozza (approved):
LGTM.

You could have also deleted https://github.com/osbuild/images/blob/1c5cba87c0dbbece9fb1232b73aa85c28329a5d9/pkg/osbuild/kickstart_stage.go#L13

PR #1770, opened 2025-08-19
many: drop the workload.Workload type entirely

mvo5:
Note that this is only a first step, we can do more here
(and probably should) but for now this is a really
low-hanging fruit. It ties into the whole discussion
of unifying workload/oscustomization/imageconfig
and seems like a step in the right direction to me.

But I'm open for discussion of course, maybe it
should be part of a bigger refactor, idk.

---

many: drop the workload.Workload type entirely

Workloads are no longer used in the images library and there
is no appetite to use them. So this commit removes the last
use of them: the workload.Custom{} that was used to pass
image type configuration to the images functions.

The workload.Custom is replaces with the regular
manifest.OSCustomizations that already contains (almost)
anything that is needed. So instead of having two concepts
to configure the image we consolidate it to one.

---

many: rename `workload` usage to `imgTypeCustomizations`

Now that the workload concept is removed and everything the
initial image type configuration is done via OSCustomization
we need to rename "workload" as it will be confusing without
the historic context (and arguably it was always a bit
confusing).

The new name needs to capture that the OSCustomizations
applied here come from the initial imagetype. I picked
`imgTypeCustomizations` as it seems to capture best
what this is doing. But its long and I'm open for more
ideas, I initially used `seed` as its short but its
probably too abstract and it captures the concept poorly.

Its also unclear how long-lived this name will be,
ideally it all will be merged into a single OSCustomizations
variable inside the various images but that needs to
happen in followups or the churn is too big.


supakeen (approved):
Sure, though I feel this could have dropped the idiom entirely and just apply them to `img.OSCustomizations` directly. Let's do that as a follow up since you already mentioned it.

mvo5:
> Sure, though I feel this could have dropped the idiom entirely and just apply them to `img.OSCustomizations` directly. Let's do that as a follow up since you already mentioned it.

I agree, I would like to merge it into a single thing. One reason why its not currently is https://github.com/osbuild/images/blob/main/pkg/manifest/os.go#L338 - the workload packages are resolved in their own transaction so we need something new in OSCustomizations (ImageTypePackages?). I will look into it.

supakeen (approved):

thozza:
> I agree, I would like to merge it into a single thing. One reason why its not currently is https://github.com/osbuild/images/blob/main/pkg/manifest/os.go#L338 - the workload packages are resolved in their own transaction so we need something new in OSCustomizations (ImageTypePackages?). I will look into it.

These are now effectively the "custom packages" requested by the user in the Blueprint. These are not related to the Image Type. Something along the lines of "custom packages", "additional packages", "user packages", or "blueprint packages" would, in my opinion, be more descriptive of what the package set represents.

PR #1771, opened 2025-08-19
test/scripts: new quickceck script for CI image builds

achilleas-k:
This PR adds a test script that is meant to be run in a GitHub action and determine if the GitLab pipelines should be triggered.  This will significantly speed up PR pipelines when it is obvious that no rebuilds are needed.  It's not wired up to anything right now.  Adding it to the existing GitHub workflows ended up in some yak shaving of the interactions between the gitlab-helper and trigger-gitlab.  I din't want to forget about this in the meantime.

---

Add a new test scripts that helps decide if any manifests will need to be rebuilt when comparing two branches.  It determines this by checking for diffs in paths that will, if changed, cause changes in the generated manifests or builds configuration.

These checks are quicker than generating all real manifests and checking against the build cache, but are only accurate because of our current test setup.  If the test setup changes to depend on anything outside the paths that are currently checked, the script will need to be updated.


bcl:
Should `data/dependencies/osbuild` also be checked for a change? It may already be covered by checking for manifest checksum changes but thought I'd ask :)

achilleas-k:
Test builds aren't triggered if the osbuild dependency version is changed.  They do if the commit ID in the Schutzfile is though.  The commit ID of the osbuild version used is part of the path in the images CI cache.

See below
https://github.com/osbuild/images/blob/1c5cba87c0dbbece9fb1232b73aa85c28329a5d9/test/scripts/imgtestlib.py#L170-L186

bcl (approved):
Ack, faster is better :)

PR #1773, opened 2025-08-19
Add the netinst image type to RHEL 8, 9, and 10

bcl:
See commit messages :)


supakeen (approved):
This is all pretty mechanical, I'm assuming the firmware changes here for Edge are OK; as they were for IoT previously.

bcl:
> This is all pretty mechanical, I'm assuming the firmware changes here for Edge are OK; as they were for IoT previously.

The changes are to keep it from changing the manifest, since this is RHEL and has been in production, and they aren't actually installer images.

All the packages and firmware were taken from the current RHEL builds, so should be exactly the same as the current production boot.iso

I'm sure there is room to consolidate things here, but we should start with something that matches. I think anything that builds an image with anaconda should be using the same package and firmware set, but that can be done next.

supakeen:
> > This is all pretty mechanical, I'm assuming the firmware changes here for Edge are OK; as they were for IoT previously.
> 
> The changes are to keep it from changing the manifest, since this is RHEL and has been in production, and they aren't actually installer images.
> 
> All the packages and firmware were taken from the current RHEL builds, so should be exactly the same as the current production boot.iso
> 
> I'm sure there is room to consolidate things here, but we should start with something that matches. I think anything that builds an image with anaconda should be using the same package and firmware set, but that can be done next.

Same, in that case would you want to also special case `image-installer`; since those manifest checksums are also changing at the moment.

mvo5 (approved):
Codewise this looks fine, no test is failing and given that this is a new image type the risk of regressions is low (except the firemware pkg changes for platform but those are from the official isos so…) so "ship it"!

supakeen (approved):

PR #1775, opened 2025-08-20
2xlarge DO NOT MERGE

croissanne:


supakeen:
So, slightly under 3 hours (175 minutes 11 seconds, queued for 12 seconds) that doesn't seem enough of an improvement to warrant switching to larger instance types and indicates that we likely spend *most* of our time waiting for AWS/provisioning instances as opposed to actually building images.

PR #1776, opened 2025-08-21
[RFC] kernel arguments options support  for all installers

mvo5:
When debugging a test failure of centos9 iso install I want to use the cool new feature by Simon to add kernel arguments to the installer (c.f. https://github.com/osbuild/images/pull/1727/files) test in bib via a blueprint kernel append.

I noticed then that its currently implemented for a subset of our installers, so I created this quick draft to see what it would take and it seems (beside the duplication) straightforward. Is there a reason not to support the kargs for the missing installers? If not I'm happy to flesh this one out more, it needs tests (obviously) but I did not want to write them before there is agreement that is what we want.

[edit/reminder: the AanacondaContainerInstaller needs a coresponding change in bib to fully work]

### DIFF
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -541,6 +541,12 @@ func liveInstallerImage(imgTypeCustomizations manifest.OSCustomizations,
 
 	img := image.NewAnacondaLiveInstaller()
 
+	var err error
+	img.OSCustomizations, err = osCustomizations(t, packageSets[osPkgsKey], options, containers, bp.Customizations)
+	if err != nil {
+		return nil, err
+	}
+
 	img.Platform = t.platform
 	img.ImgTypeCustomizations = imgTypeCustomizations
 	img.ExtraBasePackages = packageSets[installerPkgsKey]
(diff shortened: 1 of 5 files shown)


supakeen:
I have nothing against applying kernel arguments for all installer types (they're arguably the most valuable for the network installer) however; I didn't do it in the initial PR because it predated the installer customizations *and* I didn't like putting `OSCustomizations` in all installer types.

My idea at the time was to (after implementing `InstallerCustomizations`) copy this from `OSCustomizations` to `InstallerCustomizations` in `installerCustomizations` instead but I'm still unsure on that.

mvo5:
Fwiw, I'm fine moving this to InstallerCustomizations, I'm not sure yet what the final structure should be yet

mvo5:
I'm closing this one, I think it is/was a bit misguided (unless people tell me differently). I opened https://github.com/osbuild/images/pull/1786 instead. 

The difference is that we already have a way in the InstallerCustomizations to add kernel options so reusing that seems the most sensible approach.

PR #1778, opened 2025-08-21
rhel: fix openscap profile allowlists (HMS-9095)

croissanne:
Updated based on
https://github.com/osbuild/image-builder-crc/tree/8311bea5e2fc331ac27fc79c50349a381a006696/distributions, which uses the oscap tooling to fetch a list of supported profiles per distro.

### DIFF
--- a/data/distrodefs/distros.yaml
+++ b/data/distrodefs/distros.yaml
@@ -103,7 +103,7 @@ distros:
           - azure
     # rhel & centos share the same list of allowed profiles so a
     # single allow list can be used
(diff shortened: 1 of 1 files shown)


mvo5 (approved):
Thank you!

croissanne:
> Approving but I don't have the backstory if these are now actually the correct allow lists for each major?

they should be. the list i linked to is generated from https://github.com/osbuild/image-builder-crc/blob/8311bea5e2fc331ac27fc79c50349a381a006696/cmd/oscap/main.go ; which calls oscap to list profiles of  all the datastreams.

PR #1780, opened 2025-08-21
distro,image: drop `imgTypeCustomizations`

mvo5:
In https://github.com/osbuild/images/pull/1770 we did the
first step to remove the workload and replaced them with
the (badly named) `imgTypeCustomizations`. This commit
now removes this too and folds it into the existing
OSCustomizations for clarity and to have one less thing
to worry about.

The payload repositories are now passed to the image
functions and get added to the OSCustomizations there
as a new `PayloadRepos`. They will be used when resolving
blueprint packages.


supakeen (approved):
Looks good to me, let's have @achilleas-k / @thozza also look but this fits in the process.

PR #1781, opened 2025-08-22
Basic support for aboot support in bootc-image-builder.

alexlarsson:
This adds required features for making bootc-image-builder support aboot images. 

The two features are:

Add support for Osinfo to find the kernel dir (/usr/lib/moduler/*) in the target container, and check if it contains an aboot.img. This will be used by bootc-image-builder to trigger the "aboot mode" for the image. This is similar to what was done for `bootc install` in https://github.com/bootc-dev/bootc/pull/1532.

Support adding a "disk.Raw" payload, that specifies a file from an earlier pipeline and writes it to a device. This uses the new org.ostree.write-device stage (added in v154).

The intent is that bootc-image-builder will detect that there is an aboot.img in the target image, and then find the correct partition in the partition table and add a disk.Raw  payload to the raw boot partition to initialize it from the file. The corresponding bc-i-b code is available here: https://github.com/alexlarsson/bootc-image-builder/tree/aboot-support for reading, but will later be a separate PR.

Note: The bc-i-b branch above doesn't build against this PR, because bc-i-b doesn't build against osbuild-images v0.178 due to API changes, but I tested it against v0.177, and it worked.

### DIFF
--- a/pkg/blueprint/disk_customizations.go
+++ b/pkg/blueprint/disk_customizations.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"slices"
 	"strings"
+	"unicode/utf16"
 
 	"github.com/google/uuid"
 	"github.com/osbuild/images/pkg/datasizes"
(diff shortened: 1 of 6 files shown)


alexlarsson:
Sorry, this was the wrong source branch, created https://github.com/osbuild/images/pull/1782 with the real one

PR #1785, opened 2025-08-24
feat: add CustomLoraxTemplatePath support to InstallerCustomizations

brycevertera:
## Summary
Add support for custom lorax template paths in anaconda installer manifests. This enables bootc-image-builder and other consumers to specify custom lorax templates for ISO builds via the `InstallerCustomizations.CustomLoraxTemplatePath` field.

## Changes
- Add `CustomLoraxTemplatePath string` field to [InstallerCustomizations] struct
- Update lorax template selection logic in [anaconda_installer.go] to prioritize custom paths
- Extend [getLoraxTemplatePath()] to handle custom template override
- Add comprehensive unit tests in [anaconda_installer_lorax_test.go]

## API Changes
```go
type InstallerCustomizations struct {
    // ... existing fields
    CustomLoraxTemplatePath string `json:"custom_lorax_template_path,omitempty"`
    // ... existing fields
}

### DIFF
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -368,6 +368,10 @@ func installerCustomizations(t *imageType, c *blueprint.Customizations) (manifes
 		if isoboot := installerConfig.ISOBootType; isoboot != nil {
 			isc.ISOBoot = *isoboot
 		}
+
+		if customPath := installerConfig.CustomLoraxTemplatePath; customPath != nil {
+			isc.CustomLoraxTemplatePath = *customPath
+		}
 	}
 
 	installerCust, err := c.GetInstaller()
--- a/pkg/distro/installer_config.go
+++ b/pkg/distro/installer_config.go
@@ -26,6 +26,9 @@ type InstallerConfig struct {
 
 	// BootType defines what type of bootloader is used for the iso
 	ISOBootType *manifest.ISOBootType `yaml:"iso_boot_type,omitempty"`
+
+	// CustomLoraxTemplatePath allows overriding the default Lorax template path
+	CustomLoraxTemplatePath *string `yaml:"custom_lorax_template_path,omitempty"`
 }
 
 // InheritFrom inherits unset values from the provided parent configuration and
--- a/pkg/manifest/anaconda_installer.go
+++ b/pkg/manifest/anaconda_installer.go
@@ -299,7 +299,11 @@ func (p *AnacondaInstaller) payloadStages() []*osbuild.Stage {
 	}
 
 	LoraxPath := "99-generic/runtime-postinstall.tmpl"
-	if p.InstallerCustomizations.UseRHELLoraxTemplates {
+
+	// Check for custom lorax template path first (highest priority)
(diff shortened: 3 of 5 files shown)


brycevertera:
> @brycevertera Thanks! Just a few notes before I take a look at the code:
> 
>     1. Please get rid of the merge commit and rebase instead.
> 
>     2. Your `test` commits seem to start with `est` instead :)
> 
> 
> I'm also wondering what your use case is for executing custom Lorax templates, especially in this way where the custom path overwrites `runtime-postinstall`. I'd have (maybe) expected this to be a (list?) of additional Lorax scripts to run?

Will rebase :)

Oops, will squash

Use case is for custom templates. This started because Rocky Linux does not have the Lorax templates for RHEL, so when it detects it's a RHEL-based OS it automatically wants to use the RHEL specific folder for the Lorax scripts, whereas Rocky Linux only has 80-rocky and generic. Instead of making exception specific to Rocky, allow override to let the user use whatever they want if they so choose.

supakeen (approved):
Thank you, functionally this is fine and does what it says on the title.

---

However, the current PR is 'not enough' to wire this up all the way. Currently there is no way of setting this path either through the YAML definitions.

To do this there are two more (hopefully small) steps to take:

1. Add a field to: https://github.com/osbuild/images/blob/main/pkg/distro/installer_config.go which is the struct used to deserialize YAML (a `*string` with `omitempty` probably).
2. Copy the field into `InstallerCustomizations` here: https://github.com/osbuild/images/blob/main/pkg/distro/generic/images.go#L366

I don't think we want to expose this field in blueprints so that'd be where it stops.

If you were planning to do that as a follow-up I personally would prefer that step to be in this PR so it's one functional chunk.

With the above done we can ack and merge, unless others have concerns.

---

Now for something else that you don't *have* to do, but might since the test case is doing a lot of testing surrounding the order of precedence between boolean and custom path:

Why not drop `UseRHELLoraxTemplates` entirely? We could rename `CustomLoraxTemplatePath` to `LoraxTemplatePath`. Always set it in the YAML installer configs and get rid of the boolean toggle entirely.

supakeen (changes requested):
Sorry, the previous comment accidentally approved. I did request some changes which are to wire it up in `distro/installer_config.go` and `distro/generic/images.go`

brycevertera:
> @bcl Note that I suggest to that in a follow-up that we can do ourselves since it's a bit of our thing :)
> 
> @brycevertera could you run `./tools/prepare-source.sh`, it will run the formatters that the CI is complaining about.

Ah, yes!

supakeen:
I don't think this PR is going to go all the way and end up in blueprints and I do feel it's useful to have this in installer config (which also gets read into installer customizations).

@brycevertera I see some tests are failing can we fix those so we can get this in?

I'd like to iterate on this after it lands (I'll probably drop the `UseRHELLoraxTemplates` boolean like I said earlier, but that doesn't need to be part of this PR).

supakeen:
@brycevertera I'll close this since #1886 has been merged, please re-open (or a different one if you want to build on top of it) if that isn't sufficient yet.

PR #1789, opened 2025-08-26
image,manifest: move installer metadata into InstallerCustomiaztions

mvo5:
Some pretty mechanical refactor to avoid some duplication. Originally I planned multiple PRs but maybe this is straightforward enough for a single PR(?). In any case, happy to split it up if desired.

---

distro: move variant for live installer into YAML

We have a lot of duplicated extraction to get the installer product, isoLabel etc. This commit consolidates it into the InstallerCustomizations.

---

image,manifest: pass InstallerCustomizations in NewAnaconda*

By passing the InstallerCustomiations into the various
image.NewAnaconda*() function we can simplify the code by
only passing a single struct (that is part of the anaconda
structs anyway).

This also means we can use the product/preview etc values
from the customizations and don't need to carry an extra
copy. So this is dropped too.

### DIFF
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -1634,6 +1634,7 @@ image_types:
     boot_iso: true
     image_func: "live_installer"
     iso_label: "Workstation"
+    variant: "Workstation"
     exports: ["bootiso"]
     required_partition_sizes: *default_required_dir_sizes
     installer_config:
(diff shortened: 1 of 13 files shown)


supakeen (dismissed):
Might be better suited under some form of Anaconda-specific customizations; but let's start with this.

thozza (approved):
LGTM

supakeen (approved):

PR #1790, opened 2025-08-26
image: add platform/filename to image.New*()

mvo5:
image: add platform/filename to image.New*()

We require a platform and a filename for all images we create
so instead of setting it as the first thing after New*() make
it required. This ensures that its not missed.

---

image: make platform/filename part of image.Base and unexport

Every image we use has a platform and a filename so make it
part of `image.Base` instead of duplicating it everywhere.

### DIFF
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -447,8 +447,7 @@ func diskImage(t *imageType,
 	containers []container.SourceSpec,
 	rng *rand.Rand) (image.ImageKind, error) {
 
-	img := image.NewDiskImage()
-	img.Platform = t.platform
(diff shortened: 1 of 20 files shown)


supakeen:
@mvo5 There's a `wip` commit in here; was that your intent?

mvo5:
> @mvo5 There's a `wip` commit in here; was that your intent?

Ups, silly me! Thanks for noticing, I had fixed it locally but then did not force push :/ Fixed now.

supakeen (approved):
I like it. Bit worried how this will (eventually) mesh with multi-files-per-'image' but we can cross that bridge then.

PR #1792, opened 2025-08-27
policy: drop `/boot/efi` from mountpoints (HMS-9210)

supakeen:
Customizing `/boot/efi` works fine in general. It is necessary to build certain image types into Raspberry Pi 3/4 compatible MBR partition tables through customizations.


achilleas-k:
> the second commit message could be `test: add /boot/efi to...` the current "modify" sounds a bit (to me) like it already exists

I read it as "modify the partition" in the sense that it is being customized by the blueprint.
I see your point though, but wont block on it.

achilleas-k (approved):

thozza (approved):

achilleas-k:
> > > Is there some way I can see serial output from these test machines?
> > 
> > 
> > You might be able to catch the boot test while it's happening, but that would be annoying. You're better doing it manually I think. Though I don't know how to get console output in EC2. Shouldn't be too hard, right?!
> 
> Depends on ones familiarity level with AWS which for me is about below zero. I'll see if I can figure _something_ out tomorrow.

Happy to help figure it out.  I should probably learn a thing or two at some point myself.

lzap (dismissed):
Surprised to see `/boot/efi` in the denied list, isn't this by definition a separate partition so it must be always available as a mount point? I think I might be missing the point of the list. But LGT:

thozza:
> Surprised to see `/boot/efi` in the denied list, isn't this by definition a separate partition so it must be always available as a mount point? I think I might be missing the point of the list. But LGT:

Previously, the user could not customize it, which is the purpose of the list. It is a policy of which mountpoints can / can't be customized via BP.

lzap (approved):

bcl (approved):
Looks good to me, but schutzbot is angry. Restarted the failed tests just in case it was a glitch.

supakeen:
Yes, as noted in the previous comments this PR is actually failing to boot in AWS. I need to do some digging on the why but haven't had time yet :)

achilleas-k:
This is a problem with the vfat volume ID.  When the esp isn't customised, we create it with a [fixed ID](https://github.com/osbuild/images/blob/d805717430483632f7db58ed98ab404b9b4f9a09/pkg/disk/disk.go#L106).  This works out fine.
However, when it is customised, we create a [random one](https://github.com/osbuild/images/blob/d805717430483632f7db58ed98ab404b9b4f9a09/pkg/disk/disk.go#L455-L464) that has two issues:
- No dash separator between the two components.
- Lowercase letters.

This is not an issue for the mkfs stage.  The `mkfs.vfat` command requires an ID without a dash, but [we remove it when generating the stage options](https://github.com/osbuild/images/blob/d805717430483632f7db58ed98ab404b9b4f9a09/pkg/osbuild/mkfs_fat_stage.go#L40).  The command also doesn't care about the case of the characters.

However, this becomes an issue when generating `/etc/fstab`, since that requires both uppercase letters and the dash-separator, in order to match the node path created under `/dev/disk/by-uuid`.

Investigation steps:

1. Generate a manifest with a custom esp
```json
{
  "name": "esp",
  "blueprint": {
    "name": "esp",
    "customizations": {
      "disk": {
        "partitions": [
          {
            "type": "plain",
            "mountpoint": "/boot/efi",
            "fs_type": "vfat",
            "part_uuid": "7B7795E7",
            "minsize": "500 MiB"
          }
        ]
      }
    }
  }
}
```
```
› go run ./cmd/gen-manifests --metadata=false --distros fedora-42 --types server-qcow2 --arches x86_64 --output manifests --config ./esp.json
```

2. Build it
```
sudo osbuild --cache-max-size unlimited --output-directory ./builds --checkpoint "*" --export qcow2 ./manifests/fedora_42-x86_64-server_qcow2-esp.json
```

3. Boot in UEFI mode and observe timeout waiting for `/dev/disk/by-uuid/<some vfat volid>`

4. Modify the manifest: In the `org.osbuild.fstab` stage, change the `uuid` field of the `/boot/efi` filesystem to be all uppercase and with a dash `-` in the middle (separating the ID into two 4-character blocks).

5. Repeat steps 2 and 3.  It should boot successfully.


Also, we need to allow setting the volume ID for vfat filesystems in the blueprint: https://github.com/osbuild/blueprint/issues/31

achilleas-k:
I'll fix the blueprint validation later today.

supakeen:
> That way we wont care where the volid came from and what its original form was like.

But, we can also generate volid's in the correct form in `NewVolIDFromRand` and then keep the rest the same; this seems like the smaller change?

With a follow-up to allow specifying (and validating that it's correct) in blueprint?

achilleas-k:
> I'll fix the blueprint validation later today.

Turns out that's not a bug.  My mistake.

mvo5:
> > > That way we wont care where the volid came from and what its original form was like.
> > 
> > 
> > But, we can also generate volid's in the correct form in `NewVolIDFromRand` and then keep the rest the same; this seems like the smaller change?
> > With a follow-up to allow specifying (and validating that it's correct) in blueprint?
> 
> We could. My thinking was that with a formatting fix in the stage options generator we wont have to hunt down every source of this ID (automatic/internal, blueprint[1](#user-content-fn-1-375b390eb0d0c27c3eeb07e82a9138aa), yaml, using the library and creating a `disk.PartitionTable` in code). As long as the ID is valid in one of the two forms, both stages will generate the correct options.
> ## Footnotes
> 
>     1. though I just remembered that we don't support setting **filesystem** UUIDs in blueprints, but we might at some point. [↩](#user-content-fnref-1-375b390eb0d0c27c3eeb07e82a9138aa)

I did a quick experiment to just tweak the volid generateion in https://github.com/osbuild/images/compare/main...mvo5:volid?expand=1 (if we go with this approach it would need a tiny bit of work to pass the rng)

supakeen:
> > > > That way we wont care where the volid came from and what its original form was like.
> > > 
> > > 
> > > But, we can also generate volid's in the correct form in `NewVolIDFromRand` and then keep the rest the same; this seems like the smaller change?
> > > With a follow-up to allow specifying (and validating that it's correct) in blueprint?
> > 
> > 
> > We could. My thinking was that with a formatting fix in the stage options generator we wont have to hunt down every source of this ID (automatic/internal, blueprint[1](#user-content-fn-1-375b390eb0d0c27c3eeb07e82a9138aa), yaml, using the library and creating a `disk.PartitionTable` in code). As long as the ID is valid in one of the two forms, both stages will generate the correct options.
> > ## Footnotes
> > ```
> > 1. though I just remembered that we don't support setting **filesystem** UUIDs in blueprints, but we might at some point. [↩](#user-content-fnref-1-375b390eb0d0c27c3eeb07e82a9138aa)
> > ```
> 
> I did a quick experiment to just tweak the volid generateion in https://github.com/osbuild/images/compare/main...mvo5:volid?expand=1 (if we go with this approach it would need a tiny bit of work to pass the rng)

It should be hexadecimal, I have the following here:

```diff
diff --git a/pkg/disk/disk.go b/pkg/disk/disk.go
index c35246c15..a5f72af45 100644
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -453,14 +453,16 @@ func newRandomUUIDFromReader(r io.Reader) (uuid.UUID, error) {
 }
 
 // NewVolIDFromRand creates a random 32 bit hex string to use as a volume ID
-// for FAT filesystems.
+// for FAT filesystems. It ensures there's a dash in the middle and the string
+// is uppercased.
 func NewVolIDFromRand(r *rand.Rand) string {
        volid := make([]byte, 4)
        len, _ := r.Read(volid)
        if len != 4 {
                panic("expected four random bytes")
        }
-       return hex.EncodeToString(volid)
+       asHex = hex.EncodeToString(volid).ToUpper()
+       return fmt.Sprintf("%s-%s", asHex[:2], asHex[2:])
 }
```

But let's not all do this work; who's going to do it and in what way?

supakeen:
Draft for a bit, I need to figure out the normalization and checksum changes of the last commit.

Issue #1566, opened 2025-05-27
Rethink the manifest checksum check


achilleas-k:
The manifest checksum check has become a bit annoying.  It's not required, so a failing check doesn't block a PR, but it very often creates merge conflicts, requiring extra work to rebase PRs that should would otherwise allow automatic rebasing.  I have a couple of ideas for how to get around these issues while also keeping the information they provide.

## Background

**Skip this section if you know what the manifest checksums are and their purpose.**

The checksums and the CI test were introduced in #1380.  For anyone who's not familiar with the need for the checks, but also as a reminder for everyone else, the purpose of the checksums is to signal to patch authors and reviewers that a certain commit has an effect on one or more generated manifests.  This can be very useful in two scenarios:
1. A contributor is authoring a commit that refactors some part of manifest generation to make the code easier to read, but should be functionally identical.  The commit or PR should have no changes in the checksum file, because manifests should not be affected by the change.  The CI test verifies that the checksums are valid.
2. A contributor is authoring a commit that changes specific image types, or test configurations.  Each commit that affects a manifest should be accompanied by a change in the checksums file and should be limited to just the configurations it has affected, i.e. there were no unintended side effects that changed unrelated images or configs.

In the past, we used to have full manifests checked into the repository for this same purpose.  Commits that changed manifests showed the exact change in the generated manifest itself for reviewing and tracing.  However, the people who remember those times do not have fond memories of it.  I (I think it was me) proposed checksums as a way to have a signal that something changed without needing big manifests checked into the repository and also big, hard to read patches that eventually get ignored.

### 1. Replace manifest checksums file and check with a CI check and GitHub action

I think with a bit of engineering we can automate the notification that manifests have changed.  This is also something we tried in the past but got rid of because it was a bit useless.  The old solution was simply a comment on the PR by Schutzbot that mentioned that manifests changed, without any extra information.  I think a solution that might work is something that integrates with GitHub's annotations, or a comment that lists every commit in a PR that affects a manifest along with a link to a diff of the actual manifest (or a folded codeblock).  This can be done in a way that blocks the PR and forces reviewing, so reviewers are encouraged to check each diff and dismiss it if it's valid.

Benefits:
- Requires no interaction by authors and no knowledge for external contributors.
- Enforces reviewing of changes in manifests with more information than the current system.

Drawbacks:
[...]

PR #1796, opened 2025-08-27
test: update manifest checksums

achilleas-k:
A bit of a rebase issue happened here.
PR #1750 updated the repository URLs (snapshots) for RHEL 8.4, modifying all manifests.  PR #1773 added a new image type (netinst), which introduced new manifests.

The netinst PR was merged a couple of days ago, adding the new manifests to main, but #1750 was based on an old commit and didn't pick those manifests up to change them.

We don't run the manifest checksum verification on the merge queue, nor do we require it for merging to begin with, so this created the inconsistency in main.

We knew this was a situation that could happen (see drawbacks at the end of this issue description #1566), but I was pretty sure we wouldn't run into it in reality.


mvo5:
Silly question - should we run the manifest checksums in the merge queue then to prevent this? Or am I misreading the chain of events?

achilleas-k:
> Silly question - should we run the manifest checksums in the merge queue then to prevent this? Or am I misreading the chain of events?

You are not misreading, no.  We could run them in the merge queue, but currently we don't require the checksums to be valid in the PR itself.  I kept them as non-required as a trial.  If we want to check them in the merge queue to prevent this, we will have to make them required, which is the more important change IMO.  Something to discuss.

PR #1797, opened 2025-08-27
test: ensure manifest from bib/images are identical

mvo5:
This is meant as a temporary measure to ensure that we
produce identical manifests from the bootc integration of
images. We need it because we have no manifest checksums
for bootc based images (yet).

So the idea is:
1. have this tool validate that the manifests are identical
2. create manifests checksums for bootc images (just like we
   do for regular images)
3. merge bib PR#1014 that switches disk generation to images
4. drop using the tool
5. add something like `test-integration-bib` that runs the bib tests with the latest images

Keep the tool in the backpocket because we will need it for
the ISO move to images too (c.f. https://github.com/osbuild/images/pull/1804)

### DIFF
--- a/tools/gen-bootc-diff
+++ b/tools/gen-bootc-diff
@@ -1,0 +1,121 @@
+#!/bin/bash
+#
+# This tool is meant as a temporary measure to ensure that we
+# produce identical manifests from the bootc integration of
+# images. We need it because we have no manifest checksums
+# for bootc based images (yet).
+
+set -e
+set -o pipefail
+
+(cd "$(dirname "$0")"/../cmd/gen-manifests && go build)
+
+# we only need to test a single type, all bootc disk manifests
+# contain all possible types by default
+TYPE=qcow2
+
+BPDIR=$(mktemp -d)
+trap 'rm -rf -- "$BPDIR"' EXIT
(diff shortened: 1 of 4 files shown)


pkg/distro/distro.go
```diff
@@ -135,10 +135,16 @@ type ImageType interface {
 	Manifest(bp *blueprint.Blueprint, options ImageOptions, repos []rpmmd.RepoConfig, seed *int64) (*manifest.Manifest, []string, error)
 }
 
+type BootcRef struct {
```
mvo5, line 138: XXX: double check if this is still needed now that we have the bootc distro

mvo5: It is not, it will be removed in the next push

tools/gen-bootc-diff
```diff
@@ -92,7 +92,9 @@ do
 		ROOTFS="ext4"
 	    fi
 
-	    sudo go run github.com/osbuild/bootc-image-builder/bib/cmd/bootc-image-builder@main \
+	    sudo go run \
+		 -buildvcs=false \
```
achilleas-k, line 96: Why is `-buildvcs=false` needed here (and above)?

thozza (approved):
LGTM

Issue #1798, opened 2025-08-27
Fedora 43 and Rawhide minimal disk images boot to systemd-firstboot, initial-setup does not run


AdamWill:
Between Fedora-Rawhide-20250414.n.0 and Fedora-Rawhide-20250416.n.0 (yeah, April, sorry I didn't notice sooner!), the aarch64 minimal disk images for Fedora stopped working right. These are built with osbuild. The similar Server disk image built with Kiwi does not have the same issue.

The observed symptom is that when you boot the image you get systemd-firstboot, not initial-setup as intended. In openQA testing this means we get a blank screen, because systemd-firstboot apparently only runs on the serial terminal, not tty1. If you force-disable systemd-firstboot with `systemd.firstboot=off`, the system boots to a login prompt, not to initial-setup.

On further investigation, things are pretty weird. If we list out the contents of `/etc/systemd/system/multi-user.target.wants` on the booted system (using the systemd debug shell), we find:
```
NetworkManager.service
audit-rules.service
auditd.service
chronyd.service
flatpak-add-fedora-repos.service
remote-cryptsetup.target
remote-fs.target
sshd.service
sssd.service
systemd-homed.service
systemd-oomd.service
```
but if we look at the same directory *in the raw disk image itself*, its contents are different:
```
auditd.service
audit-rules.service
boot-efi.mount
boot.mount
chronyd.service
flatpak-add-fedora-repos.service
initial-setup.service
-.mount
NetworkManager.service
remote-fs.target
sshd.service
sssd.service
```
@supakeen confirmed that if he boots with `rd.break` and checks the contents of `/sysroot/etc/systemd/system/multi-user.target.wants`, it's the same as the disk image - i.e. it contains `initial-setup.service`. Once he exits the rd.break environment and proceeds with boot, it flips over to the other contents, somehow.

We do not understand what's going on here. Maybe @keszybz or @poettering has some idea why the contents of `/etc` would apparently change during boot like this?

supakeen:
Closing this as downstream updates have been submitted. Thanks for the report and the help debugging!

PR #1799, opened 2025-08-27
fedora: do not mark images as firstboot (HMS-9222)

supakeen:
When we write `uninitialized` into `/etc/machine-id` it triggers `systemd`'s first boot capabilities. This was wanted behavior at the time but leads to interplay with other bits that are unexpected.

Specifically in the `minimal-raw` images we are using mount units and have an `rw` root filesystem. This, in combination with the system marked as booting for the first time triggers `systemd`'s first boot full preset behavior [1].

This then means `systemd` overwrites all relevant configuration in `/etc` with the presets. These presets do *not* include our mount units, nor the `initial-setup` that this image normally uses.

An alternative is to drop mount units, write `/etc/fstab` and kernel argument as `ro` again. This prevents `systemd` from performing its reset (probably accidentally). However, in Fedora 43 the `setup` package [2] was also changed so we now have a (correct) `!unprovisioned` password hash for the `root` password. If this is present `systemd-firstboot` will prompt for it; and we end up with duplicate provisioning systems.

[1]: https://src.fedoraproject.org/rpms/systemd/blob/rawhide/f/systemd.spec#_839
[2]: https://src.fedoraproject.org/rpms/setup/c/7ced36d60b67c9e74f7951123225200597e3d2fa

---

Reported by @AdamWill in https://github.com/osbuild/images/issues/1798, downstream tracker bug: https://bugzilla.redhat.com/show_bug.cgi?id=2391379


supakeen:
Note that I made the smallest possible change here. It's probably a good idea to turn this off globally even if this behavior only manifests in combination with a read/write sysroot.

achilleas-k:
Thanks for the (late night) investigation on this!

PR #1800, opened 2025-08-28
distro: ensure correct bootc image type Filename() output

mvo5:
[edit: this will unblock https://github.com/osbuild/image-builder-cli/pull/245]

This fixes a bug in the new bootc based imagetypes. The filename was not returned with the correct extension. The reason is that the the file extension is hardcoded in pkg/image/bootc_disk.go, we have no way to access this information the level when the image type is generated.

When the bootc imagetypes move into YAML the problem hopefully solves itself as will then set the filename in the YAML just like we do for the other image types (this will need changes to bootc_disk.go then too).

So this commit needs to duplicate it for now. No
explicit test as this is indirectly tested via ibcli and it will go away when moving into yaml based bootc definitions (see https://github.com/osbuild/images/pull/1656 for an outline what that might look like).

(fwiw with https://github.com/osbuild/images/pull/1039 this wouldn't be an issue for ibcli).

### DIFF
--- a/pkg/distro/bootc/bootc.go
+++ b/pkg/distro/bootc/bootc.go
@@ -56,6 +56,8 @@ type BootcImageType struct {
 
 	name   string
 	export string
+	// file extension
+	ext string
 }
 
 func (d *BootcDistro) SetBuildContainer(imgref string) (err error) {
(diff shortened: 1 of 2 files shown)


ondrejbudai (approved):
Looks, reasonable, thanks!

PR #1806, opened 2025-08-28
Fix handling of root user in kickstart stage [RHEL-4644]

achilleas-k:
The root user should not be handled using the standard `user` command in kickstart.  If the root user is added to a blueprint for an ISO, it only supports two options: ssh key and password.  This commit changes the kickstart stage options generator to:
1. use the rootpw command to set the root password when specified, and
2. return an error if unsupported options are defined for the root user.

See also osbuild/osbuild#2178 for some details about the current behaviour.


pkg/osbuild/kickstart_stage.go
```diff
@@ -149,6 +149,70 @@ func NewKickstartStageOptions(
 		users = usersOptions.Users
 	}
 
+	var rootpw *RootPasswordOptions
+	for name, user := range users {
+		// User options for the root user have no effect in kickstart. In other
```
mvo5, line 154: I wonder if extracting this into a helper would be slightly easier to read, then the helper name could convey that this is exclusively about adjusting/validating the root user. That would have helped me I think, from a casual reading it was not obvious to me that the loop is only dealing with root and the "break" at the end also made me pause and wonder (but of course if its all about root it makes perfect sense).

mvo5 (commented):
Looks good, thanks for the tests. One tiny suggestion inline about using a helper though that I think would make it slightly nicer/easier to read.

pkg/osbuild/kickstart_stage.go
achilleas-k: Indeed, it's much more readable as a function.

mvo5 (approved):
Thank you, love it!

thozza (approved):
LGTM

PR #1808, opened 2025-08-29
github: remove matrix value from test names

achilleas-k:
Remove the matrix value from the names of tests so we can mark the whole step as required.  This should save us from having to update the required tests in repository settings every time we add or remove a Fedora version.


mvo5 (approved):

thozza (approved):
Thanks! :heart:

PR #1817, opened 2025-09-02
Add support to set partition GPT attributes in yaml format

alexlarsson:
This is going to be necessary for automotive to set up aboot partitions.

### DIFF
--- a/pkg/disk/partition.go
+++ b/pkg/disk/partition.go
@@ -3,6 +3,7 @@ package disk
 import (
 	"encoding/json"
 	"fmt"
+	"slices"
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/datasizes"
@@ -27,6 +28,9 @@ type Partition struct {
 
 	// If nil, the partition is raw; It doesn't contain a payload.
 	Payload PayloadEntity `json:"payload,omitempty" yaml:"payload,omitempty"`
+
+	// Partition GPT attribute flags to set
+	Attrs []uint `json:"attrs,omitempty" yaml:"attrs,omitempty"`
 }
 
 func (p *Partition) Clone() Entity {
@@ -41,6 +45,7 @@ func (p *Partition) Clone() Entity {
 		Bootable: p.Bootable,
 		UUID:     p.UUID,
 		Label:    p.Label,
+		Attrs:    slices.Clone(p.Attrs),
 	}
 
 	if p.Payload != nil {
--- a/pkg/disk/partition_test.go
+++ b/pkg/disk/partition_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"go/types"
+	"reflect"
 	"testing"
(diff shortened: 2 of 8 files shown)


pkg/disk/partition.go
```diff
@@ -47,6 +50,11 @@ func (p *Partition) Clone() Entity {
 		partition.Payload = p.Payload.Clone().(PayloadEntity)
 	}
 
+	if p.Attrs != nil {
```
mvo5, line 53: (nitpick) with the (relatively) new slices package this could be slightly more compact:
```go
diff --git a/pkg/disk/partition.go b/pkg/disk/partition.go
index 3c9f785bd..e5ebd141c 100644
--- a/pkg/disk/partition.go
+++ b/pkg/disk/partition.go
@@ -3,6 +3,7 @@ package disk
 import (
        "encoding/json"
        "fmt"
+       "slices"
 
        "github.com/osbuild/images/internal/common"
        "github.com/osbuild/images/pkg/datasizes"
@@ -49,11 +50,7 @@ func (p *Partition) Clone() Entity {
        if p.Payload != nil {
                partition.Payload = p.Payload.Clone().(PayloadEntity)
        }
-
-       if p.Attrs != nil {
-               partition.Attrs = make([]uint, len(p.Attrs))
-               _ = copy(partition.Attrs, p.Attrs)
-       }
+       partition.Attrs = slices.Clone(p.Attrs)
 
        return partition
 }
```

pkg/disk/partition.go
```diff
@@ -47,6 +50,11 @@ func (p *Partition) Clone() Entity {
...
+		partition.Attrs = make([]uint, len(p.Attrs))
+		_ = copy(partition.Attrs, p.Attrs)
+	}
```
mvo5, line 56: Its probably a bit silly but I wonder if having a tiny test:
```go
func TestPartitionClone(t *testing.T) {
	p1 := &disk.Partition{
		Start:    123,
		Size:     456,
		Type:     "0x83",
		Bootable: true,
		UUID:     "7a23f22e-3da3-4b3c-a1b7-cc3ebb995307",
		Label:    "part-label",
		Payload:  &disk.Raw{SourcePath: "/source/path"},
		Attrs:    []uint{7, 8, 9},
	}
	p2 := p1.Clone().(*disk.Partition)
	assert.Equal(t, p1, p2)
	// payload/attrs got cloned as well
	assert.False(t, reflect.ValueOf(p1.Payload).Pointer() == reflect.ValueOf(p2.Payload).Pointer())
	assert.False(t, reflect.ValueOf(p1.Attrs).Pointer() == reflect.ValueOf(p2.Attrs).Pointer())
}
```
would make sense?

mvo5 (approved):
Thanks for this! Tiny suggestion inline but could be followups (or ignored)

alexlarsson:
I tested this (cherry-picked to another branch) and did a full build. It works fine there.

achilleas-k (approved):

PR #1821, opened 2025-09-02
many: add bootc based manifest checksums

mvo5:
[draft as this needs cleanup in the commits, some tests are missing etc]

This PR adds manifest checksums for the bootc image types. This is useful to get manifest diffs and regression tests for bootc based manifest generation. Its slightly different from the regular distro based tests because we do not want to use real containers for speed reasons. So instead there is a new test/bootc-fake-containers.yaml that fakes the inputs we normally read from the container itself.


mvo5:
The ibcli test failure will fix itself once https://github.com/osbuild/image-builder-cli/pull/294 is merged in ibcli. Its not specific to this PR but to the new images library.

achilleas-k (approved):
This turned out pretty good.  Thanks for all the work.

LGTM

PR #1822, opened 2025-09-02
distro: Fix build container use in bootc distro

alexlarsson:
This fixes two things:
 1) If SetBuildContainer() is not called, buildImgRef is nil
 2) If it is called, buildImgRef is properly set to the value passed.

Without this, when using a build container, I got errors about a missing "target" pipeline when using https://github.com/osbuild/bootc-image-builder/pull/1014

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -513,10 +513,22 @@ func main() {
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


alexlarsson:
Hmm, seems like something is using buildImgRef here causing a ci failure.

PR #1825, opened 2025-09-03
Add support for geometry option in vfat mkfs

alexlarsson:
This adds support for the geometry option in the filesystem payload of a partition, and adds support in the actual fat stage to use it.

### DIFF
--- a/pkg/osbuild/mkfs_stage.go
+++ b/pkg/osbuild/mkfs_stage.go
@@ -37,6 +37,8 @@ func GenFsStages(pt *disk.PartitionTable, filename string, soucePipeline string)
 		case *disk.Filesystem:
 			stageDevices := getDevicesForFsStage(path, filename)
 
+			// Make a copy so we can mark the ones we handled
+			mkfsOptions := e.MkfsOptions
 			switch e.GetFSType() {
 			case "xfs":
 				options := &MkfsXfsStageOptions{
@@ -49,11 +51,12 @@ func GenFsStages(pt *disk.PartitionTable, filename string, soucePipeline string)
 					VolID: strings.ReplaceAll(e.UUID, "-", ""),
 					Label: e.Label,
 				}
-				if e.MkfsOptions.Geometry != nil {
+				if mkfsOptions.Geometry != nil {
 					options.Geometry = &MkfsFATStageGeometryOptions{
 						Heads:           e.MkfsOptions.Geometry.Heads,
 						SectorsPerTrack: e.MkfsOptions.Geometry.SectorsPerTrack,
 					}
+					mkfsOptions.Geometry = nil // Handled
 				}
 
 				stages = append(stages, NewMkfsFATStage(options, stageDevices))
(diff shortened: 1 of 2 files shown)


pkg/disk/filesystem.go
```diff
@@ -9,47 +9,106 @@ import (
...
+func (o MkfsOptionVerity) MarshalJSON() ([]byte, error) {
+	return json.Marshal("verity")
 }
 
-// String converts MkfsOption into a human readable string
-func (option MkfsOption) String() string {
-	return getMkfsOptionMapping()[int(option)]
+type MkfsOptionGeometry struct {
+	Heads           int `json:"heads" yaml:"heads"`
+	SectorsPerTrack int `json:"sectors-per-track" yaml:"sectors-per-track"`
```
alexlarsson, line 28: Hmm, this should maybe be sectors_per_track. The stage uses sectors-per-track, but we should probably match the other payload fields which seem to use underscore.

pkg/disk/filesystem.go
```diff
@@ -9,47 +9,106 @@ import (
 	"github.com/google/uuid"
 )
 
-type MkfsOption int
+type MkfsOption interface {
```
mvo5, line 12: A maybe silly question - if we would just do:
```go
type MkfsOptions struct {
	Verity bool `json:"verity,omitempty" yaml:"verity,omitempty"`

	Heads           int `json:"heads,omitempty" yaml:"heads,omitempty"`
	SectorsPerTrack int `json:"sectors-per-track,omitempty" yaml:"sectors-per-track,omitempty"`
}
```
here instead of going the interface approach, what would be loose? It would be simpler to read and more inline with the rest of the disk code which mostly uses similar (simple) structs. But maybe I'm missing something?

(full diff with minor test failure here: 
[pr-1825-struct-only-approach.diff.txt](https://github.com/user-attachments/files/22119987/pr-1825-struct-only-approach.diff.txt)

pkg/disk/filesystem.go
```diff
@@ -9,47 +9,106 @@ import (
...
 
-// UnmarshalJSON converts a JSON string into an MkfsOption
-func (option *MkfsOption) UnmarshalJSON(data []byte) error {
-	val, err := unmarshalHelper(data, getMkfsOptionMapping())
-	if err != nil {
-		return err
+type optionFactory func() MkfsOption
+
+var registry = map[string]optionFactory{
+	"verity":   func() MkfsOption { return &MkfsOptionVerity{} },
```
mvo5, line 51: Ideally we would (intendant of the implementation via interface or flat struct) some sort of validation that checks that the given filesystem type supports the options so that we error early if someone tries to enable "verity" on vfat by accident.

mvo5 (commented):
Thanks, this looks interesting! One small question/suggestion about maybe considering a slightly simpler approach for the marshal/unmarshal (see inline diff).

pkg/disk/filesystem.go
alexlarsson: I'd still like a Geometry struct to show that these are related and you must either specify none or both. But for sure this is nicer. It will change how the json is serialized, but i don't think anything is tied to the existing format, right?

alexlarsson: done

pkg/disk/filesystem.go
alexlarsson: done

mvo5 (dismissed):
Thanks! I like this! I added small suggestion about adding tests but we can always do a followup for them.

pkg/osbuild/mkfs_stage.go
```diff
@@ -50,20 +51,37 @@ func GenFsStages(pt *disk.PartitionTable, filename string) []*Stage {
...
 					options.Verity = common.ToPtr(true)
+					mkfsOptions.Verity = false // Handled
 				}
 
 				stages = append(stages, NewMkfsExt4Stage(options, stageDevices))
 			default:
 				panic(fmt.Sprintf("unknown fs type: %s", e.GetFSType()))
 			}
+
+			if mkfsOptions.Geometry != nil {
```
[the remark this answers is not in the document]
alexlarsson: added

pkg/osbuild/mkfs_stage.go
```diff
@@ -50,20 +51,37 @@ func GenFsStages(pt *disk.PartitionTable, filename string) []*Stage {
 					VolID: strings.ReplaceAll(e.UUID, "-", ""),
 					Label: e.Label,
 				}
+				if mkfsOptions.Geometry != nil {
```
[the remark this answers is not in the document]
alexlarsson: added

mvo5 (approved):

alexlarsson:
@achilleas-k could you take a look too?

achilleas-k (approved):
TY!  LGTM

PR #1829, opened 2025-09-03
Support raw payloads in yaml partition info

alexlarsson:
Primarily, this removes the SourcePipeline option in the Raw payload type, and instead always uses whatever is a "natural" source for the type of image being built. This makes a lot more sense, because the pipeline names is an internally generated thing and the user shouldn't need to know what to use here.

We also add Raw handling to Partition.partitionTableFeatures() to avoid this causing a panic.

### DIFF
--- a/internal/testdisk/partition.go
+++ b/internal/testdisk/partition.go
@@ -400,8 +400,7 @@ func MakeFakePartitionTable(mntPoints ...string) *disk.PartitionTable {
 			payload = swap
 		case "raw":
 			payload = &disk.Raw{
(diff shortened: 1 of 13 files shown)


alexlarsson:
@achilleas-k 

>  In particular, I don't think there's much sense in supporting this for the installer pipelines (AnacondaInstallerISOTree, CoreOSISOTree) for now. I would prefer it if raw partitions in that case become errors instead. If we ever need them there we can add them.

I'm not sure how to best do this, as I'm not very experienced in the non-bootc side of things.

PR #1831, opened 2025-09-04
distrodefs/fedora: support enabled_modules for image types with packages

achilleas-k:
Add support for enabled_modules wherever package installation is supported.

This was missed in the initial implementation of the option validation. It was caught during the update in osbuild-composer.

### DIFF
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -760,6 +760,7 @@
       - "packages"
       - "modules"
       - "groups"
+      - "enabled_modules"
       - "containers"
       - "minimal"
       - "customizations.cacerts"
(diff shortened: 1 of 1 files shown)


mgold1234 (approved):
/lgtm

thozza (approved):

achilleas-k:
> I am unsure about ostree, specifically ostree commit. That is a tar, is Anaconda involved at all? I guess I am wrong.

These are package modules (modularity).  We called them `enabled_modules` because the top-level `modules` field was taken (and basically equivalent to `packages`—long story).  So anywhere that packages are supported, we also support modules.  The ostree commit is in fact the problem here.  We had tests in osbuild-composer where we enable and install modules in a Fedora iot-commit and they were failing now because I didn't mark them as supported.

I'm working on some dynamic testing that should catch problems like this in the future.

PR #1832, opened 2025-09-04
RHEL-10: use GCP guest tools for GCE image type (HMS-9232)

thozza:
SSIA

During the work, I also realized that the `google-cloud-sdk` repository that was in our repo configs was never used for any content, and also not added as a custom repo config to the GCE image. Therefore, let's delete it.


achilleas-k (approved):
Very nice :rocket: 

LGTM

mvo5 (approved):

PR #1833, opened 2025-09-04
tools: delete image-info

thozza:
We consolidated the image-info implementation into a single place (osbuild repo) some time ago. This copy was somehow missed when that happened, but it should not exist. Let's delete it.


lzap (approved):
The the winner of Patch of the Week is... 🏆 

`+0 −2,883`

PR #1837, opened 2025-09-05
manifestgen,imagefilter: simplify the API

mvo5:
manifestgen: return manifest via Generator.Generate()

Simplify the way to use manifestgen to just return the
osbuild manifest bytes from `manifestgen.Generator.Generate()`
instead of the previous (indirect) way via `Options.Output`.

The previous way way was a bit too indirect and cumbersome
and this approach is more inline with what the existing
Serialize() is doing.

This will require updates in bib/ibcli.

---

manifestgen: simplify Generate() to only take imageType

We used to require to pass distro,arch as well but that
is entirely unnecessary (and actually arguably dangerous)
as the image type that is already passed has references
to its architecture and distro already.

---

imagefilter: simplify result and just return imageType

The imagefilter was returning the tuple (distro,arch,imgtype)
but that is unnecessary. The image type already has references
to the matching arch and distro. So just drop that.

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -2,7 +2,6 @@
 package main
 
 import (
-	"bytes"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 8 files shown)


achilleas-k (dismissed):
Nice simplifications.  LGTM!

thozza (dismissed):
Makes sense, thanks!

supakeen (approved):

thozza (approved):

supakeen:
Tests have started failing after rebase.

PR #1841, opened 2025-09-07
manifest/os: prepend dracut stage

supakeen:
Dracut configuration files need to be written *before* dracut generates an initrd (which happens in the RPM installation). Alternatively we *could* regenerate dracut after the config stage is written.

Without moving the dracut configuration stages before the RPM stages initrd's generated as part of the build don't include any of the dracut configuration. Only subsequent calls to dracut (for example, when the booted system updates its kernel) have them applied.

---

It is my idea that our `dracut_conf` bits never *actually* worked (even before they were refactored into YAML). I'd like for people to confirm building an image before/after this commit and to verify that the expected modules end up inside the initrd because I *must*'ve missed something.

This changes a lot of manifests due to moving a stage around. It is also possible that some builds start failing as dracut tries to copy binaries into the initrd, it will (likely) fail if those binaries do not exist.


bcl:
I have comments, but they are in #1853 :)

PR #1848, opened 2025-09-08
bootc: add integration test for disk.yaml

mvo5:
[draft: as I'm not super happy yet but useful for discussion]
[based on https://github.com/osbuild/images/pull/1845]

---

bootc,manifestest: use *json.RawMessage instead of
 map[string]

This is an alternative idea to the previous commit, using
json.RawMessage allows us to use the "native" structs at the
expense of a bit more verbosity inside the test itself.

Thanks to Thozza and Achilleas for their suggestions her here!

---

bootc: convert TestInteratedBuildDiskYAML to use new
 manifestest

This commit converts the TestInteratedBuildDiskYAML to use the
new helpers from manifesttest. A bit better then checking for
the raw json but still a bit unsatificatory because of the
disconnect between the osbuild types and the manifesttest types.

---

manifesttest: add NewManifestFromBytes() and Pipeline/Stage
 etc

This commit is a step towards a more realistic manifesttest. It
will unmarshal the json into a similar structure as osbuild.Manifest
would do. Ideally this would not be needed and osbuild would do
it all via `osbuild.NewManifestFromBytes`. However because we
use interfaces in osbuild.Stage the unmarshal is not symetric
to the marshal and we cannot unmarshal back without writing a
bunch of code in a custom osbuild.Stage.Unmarshal handler.
Doing that just for testing seems a bit overkill. Its a bit
sad though because it means more duplication and drift.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -101,6 +101,9 @@ jobs:
       - name: Run unit tests for bib container
         run: sudo go test ./pkg/bib/container/... --fail-if-podman-missing
 
+      - name: Run tests for bootc container
+        run: sudo go test ./pkg/distro/bootc/...
+
   unit-tests-cs:
     strategy:
       matrix:
--- a/pkg/distro/bootc/bootctest/bootctest.go
+++ b/pkg/distro/bootc/bootctest/bootctest.go
@@ -1,0 +1,118 @@
+package bootctest
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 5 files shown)


achilleas-k (commented):
I think I understand why you're not entirely happy with this, but the general approach is good.  The last commit seems to clean things up a bit, but the previous version wasn't bad either.

thozza (approved):
This looks great. I think that it is a reasonable compromise that does the job. :clap:

mvo5:
This is very stange - the commits for this have landed in main (e.g. https://github.com/osbuild/images/commit/ec1e12cb5611cf41b74bc7200ca3e010a6607572) but for some reason GH closed this PR with unmerged commits?!?

PR #1850, opened 2025-09-09
pkg/rpmmd: simplify package search in depsolved package list and `panic()` less (HMS-9310)

thozza:
Simplify the `rpmmd` implementation and reduce code that will need to be ported as part of the `rpmmd` data structures consolidation by deleting the `GetVerStrFromPackageSpecList*()` functions. Let's use `rpmmd.GetPackage()` everywhere and let the caller do whatever it needs with the returned `rpmmd.PackageSpec` instance.

I also used this opportunity to reduce the amount of `panic()` calls.

/jira-epic HMS-8910

JIRA: [HMS-9310](https://issues.redhat.com/browse/HMS-9310)

### DIFF
--- a/pkg/manifest/anaconda_installer.go
+++ b/pkg/manifest/anaconda_installer.go
@@ -190,8 +190,11 @@ func (p *AnacondaInstaller) serializeStart(inputs Inputs) error {
 	}
 	p.packageSpecs = inputs.Depsolved.Packages
 	if p.kernelName != "" {
-		// XXX: return error instead of panic
-		p.kernelVer = rpmmd.GetVerStrFromPackageSpecListPanic(p.packageSpecs, p.kernelName)
+		kernelPkg, err := rpmmd.GetPackage(p.packageSpecs, p.kernelName)
+		if err != nil {
+			return fmt.Errorf("AnacondaInstaller: failed to find kernel package %q in the depsolved packages", p.kernelName)
+		}
+		p.kernelVer = kernelPkg.GetEVRA()
 	}
 	p.repos = append(p.repos, inputs.Depsolved.Repos...)
 	return nil
--- a/pkg/manifest/anaconda_installer.go
+++ b/pkg/manifest/anaconda_installer.go
@@ -1,6 +1,7 @@
 package manifest
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
@@ -183,15 +184,17 @@ func (p *AnacondaInstaller) getPackageSpecs() []rpmmd.PackageSpec {
 	return p.packageSpecs
 }
 
-func (p *AnacondaInstaller) serializeStart(inputs Inputs) {
+func (p *AnacondaInstaller) serializeStart(inputs Inputs) error {
 	if len(p.packageSpecs) > 0 {
-		panic("double call to serializeStart()")
+		return errors.New("AnacondaInstaller: double call to serializeStart()")
 	}
 	p.packageSpecs = inputs.Depsolved.Packages
 	if p.kernelName != "" {
+		// XXX: return error instead of panic
 		p.kernelVer = rpmmd.GetVerStrFromPackageSpecListPanic(p.packageSpecs, p.kernelName)
 	}
 	p.repos = append(p.repos, inputs.Depsolved.Repos...)
+	return nil
 }
 
 func (p *AnacondaInstaller) serializeEnd() {
--- a/pkg/manifest/anaconda_installer.go
+++ b/pkg/manifest/anaconda_installer.go
@@ -192,7 +192,7 @@ func (p *AnacondaInstaller) serializeStart(inputs Inputs) error {
 	if p.kernelName != "" {
 		kernelPkg, err := rpmmd.GetPackage(p.packageSpecs, p.kernelName)
 		if err != nil {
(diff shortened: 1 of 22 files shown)


lzap (dismissed):
That is a lot of `panic` kills, well done. Some linter complaints but looks good to me.

pkg/manifest/empty.go
```diff
@@ -65,16 +67,17 @@ func (p *ContentTest) getOSTreeCommits() []ostree.CommitSpec {
 	return p.commitSpecs
 }
 
-func (p *ContentTest) serializeStart(inputs Inputs) {
+func (p *ContentTest) serializeStart(inputs Inputs) error {
 	if p.serializing {
-		panic("double call to serializeStart()")
+		return fmt.Errorf("double call to serializeStart()")
```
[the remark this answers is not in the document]
thozza: In principle, I agree with using the `panic()` here, and I've been advocating for using it in the past in justified cases. However, with the function returning `error`, it felt silly to keep the `panic()` around. I have no strong opinion, I can make it `panic()` again if others feel the same way.

The use of `fmt.Errorf()` was a mechanical mistake on my side while making the changes; I'll fix it.

As for the custom `error` type, I don't see any benefit in adding it for now. It would be just additional code, given that we do not have any use case to distinguish different error types in this case.

pkg/manifest/pipeline.go
```diff
@@ -174,7 +174,9 @@ func NewBase(name string, build Build) Base {
 
 // serializeStart must be called exactly once before each call
 // to serialize().
-func (p Base) serializeStart(inputs Inputs) {
+func (p Base) serializeStart(inputs Inputs) error {
+	// XXX: we could do the "double call" check and other common mistakes here
```
mvo5, line 178: Have a consistent way of doing this here would actually be nice. I ran into issues with the current "indirect" way of tracking via things like containerSpecs and its also from a casual reading view not obvious that the set/unset of eg. containerSpecs has this side effect so being explicit here might be a good idea.

mvo5 (approved):
I like it! One suggestion about the more details in the new errors but not a blocker

pkg/manifest/pipeline.go
thozza: I had the same thought.

My initial idea would be to add an explicit `bool` variable in the `Base` to track this. And maybe doing the checks in the `Base.serializeStart()` and then all structs that embed it would call it as the first thing in their `serializeStart()` implementation.

pkg/manifest/anaconda_installer.go
```diff
@@ -183,15 +184,17 @@ func (p *AnacondaInstaller) getPackageSpecs() []rpmmd.PackageSpec {
 	return p.packageSpecs
 }
 
-func (p *AnacondaInstaller) serializeStart(inputs Inputs) {
+func (p *AnacondaInstaller) serializeStart(inputs Inputs) error {
 	if len(p.packageSpecs) > 0 {
-		panic("double call to serializeStart()")
+		return errors.New("double call to serializeStart()")
```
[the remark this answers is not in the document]
thozza: Fair comment. I'll think about a good way to achieve this.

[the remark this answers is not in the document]
mvo5: Yeah, its a bit annoying and busy work. I never looked too deep but I wonder if somehow we could make the API in a way that makes this less of an issue.

achilleas-k (approved):

pkg/rpmmd/repository.go
```diff
@@ -246,23 +246,6 @@ func GetPackage(pkgs []PackageSpec, packageName string) (PackageSpec, error) {
 	return PackageSpec{}, fmt.Errorf("package %q not found in the PackageSpec list", packageName)
 }
 
-func GetVerStrFromPackageSpecList(pkgs []PackageSpec, packageName string) (string, error) {
```
bcl, line 249: Why not leave this one and use it? All the calls to GetPackage are now checking err and doing GetEVRA when there is no error.

pkg/manifest/anaconda_installer.go
```diff
@@ -183,15 +184,20 @@ func (p *AnacondaInstaller) getPackageSpecs() []rpmmd.PackageSpec {
...
 	if len(p.packageSpecs) > 0 {
-		panic("double call to serializeStart()")
+		return errors.New("AnacondaInstaller: double call to serializeStart()")
 	}
 	p.packageSpecs = inputs.Depsolved.Packages
 	if p.kernelName != "" {
-		p.kernelVer = rpmmd.GetVerStrFromPackageSpecListPanic(p.packageSpecs, p.kernelName)
+		kernelPkg, err := rpmmd.GetPackage(p.packageSpecs, p.kernelName)
+		if err != nil {
+			return fmt.Errorf("AnacondaInstaller: failed to find kernel package %q in the depsolved packages", p.kernelName)
```
bcl, line 195: GetPackage (or GetVerStrFromPackageSpecList) already return a useful error message. This should probably just wrap it with the added context so that it isn't repeating it.

mvo5 (commented):
This looks fine, thanks! One tiny suggestion about what looks like a change that belongs into a earlier commit though.

bcl (approved):
Looks good!

PR #1851, opened 2025-09-09
osbuild: add Duration field to Status and parse it

lzap:
osbuild: remove JSON control character from tests

The control character is being trimmed by the parser, linters do not
like this and I was confused when working with the test data too. This
patch removes it, but there is one test case that actually sends the
character into the parser to ensure it works properly.

---

osbuild: add Duration field to Status and parse it

Osbuild now emits duration of each stage/module in its status messages.
This patch adds Duration field to Status struct and parses it from
the JSON input. The field is of type time.Duration and is set to 0
if the duration is not present in the JSON (for example in older
osbuild versions).

---

osbuild: remove broken message fields

Osbuild does not always provide message fields in progress updates, this
was parsed as "Stage " and "Pipeline " which does not make sense.

---

I am not sure about the last commit, can drop if I am missing something. @mvo5 

https://issues.redhat.com/browse/HMS-9311


pkg/osbuild/monitor_test.go
```diff
@@ -14,9 +14,9 @@ import (
 	"github.com/osbuild/images/pkg/osbuild"
 )
 
-const osbuildMonitorLines_curl = `{"message": "source/org.osbuild.curl (org.osbuild.curl): Downloaded https://rpmrepo.osbuild.org/v2/mirror/public/f39/f39-x86_64-fedora-20231109/Packages/k/kpartx-0.9.5-2.fc39.x86_64.rpm\n", "context": {"origin": "org.osbuild", "pipeline": {"name": "source org.osbuild.curl", "id": "598849389c35f93efe2412446f5ca6919434417b9bcea040ea5f9203de81db2c", "stage": {}}, "id": "7355d3857aa5c7b3a0c476c13d4b242a625fe190f2e7796df2335f3a34429db3"}, "progress": {"name": "pipelines/sources", "total": 4, "done": 0}, "timestamp": 1731589338.8252223}
```
mvo5, line 17: Removing the control character makes the test less realistic, this is especially problematic when we do multiline test. The other issue is that its no longer trivial to just copy and paste output from osbuild to generate test-cases. I would not do it and instead add a comment with context (like "\x1e" is included because this follows the json-seq rfc or something).

The fact that our jsonseq parser is navie and just strips is the control chars instead of properly checking is actually an issue (not a big one) but if we fix it we would then be in a weird situation because the tests would no longer work as they are not realistic anymore.

pkg/osbuild/monitor_test.go
```diff
@@ -72,13 +72,14 @@ func TestScannerSimple(t *testing.T) {
...
+const osbuildMontiorLines_subprogress = `{"message": "Starting module org.osbuild.rpm", "context": {"origin": "osbuild.monitor", "pipeline": {"name": "build", "id": "32e87da44d9a519e89770723a33b7ecdd4ab85b872ae6ab8aaa94bdef9a275c7", "stage": {"name": "org.osbuild.rpm", "id": "bf00d0e1e216ffb796de06a1a7e9bb947d5a357f3f18ffea41a5611ee3ee0eac"}}, "id": "04c5aad63ba70bc39df10ad208cff66a108e44458e44eea41b305aee7a533877"}, "progress": {"name": "pipelines/sources", "total": 4, "done": 1, "progress": {"name": "pipeline: build", "total": 8, "done": 2, "progress": {"name": "sub-sub-progress", "total": 16, "done": 4}}}, "timestamp": 1731600115.148399}
 `
 
 func TestScannerSubprogress(t *testing.T) {
 	ts1 := 1731600115.14839 * 1000
 
-	r := bytes.NewBufferString(osbuildMontiorLines_subprogress)
+	// test with the control character
+	r := bytes.NewBufferString("\x1e" + osbuildMontiorLines_subprogress)
```
mvo5, line 81: For multi-line strings this will give us a very strange half-json-seq and half-not

pkg/osbuild/monitor.go
```diff
@@ -123,22 +123,28 @@ func (sr *StatusScanner) Status() (*Status, error) {
...
-			Total:   status.Progress.Total,
-			Message: fmt.Sprintf("Pipeline %s", pipelineName),
+			Done:  status.Progress.Done,
+			Total: status.Progress.Total,
 		},
 		Timestamp: ts,
 		Duration:  time.Duration(status.Duration * float64(time.Second)),
 	}
 
+	if pipelineName != "" {
```
mvo5, line 133: A test case for this change wold be nice. In what situations does osbuild not give us a pipeline and/or a stage name? What does the progress struct look like in this case? Would be great to make sure that we provide still soemthing vaguely nice looking to the user and/or fix osbuild (if it needs fixing).

pkg/osbuild/monitor_test.go
lzap: The problem is, this character can only be escaped in regular quotes, it is not visible which is a good argument against it.

lzap:
Implemented a helper that prefix JSONSeq control characters automatically so it does not need to be in the codebase.

As was requested, added an additional test for empty input which creates some problems with Timestamp parsing which I solved.

pkg/osbuild/monitor.go
mvo5: Aha, nice. Thanks for this extra context!

achilleas-k:
> We could just load the test data from an external file and then even go:embed it for easier access

I like this and had the same thought while reviewing.

bcl (approved):
Looks ok to me, assuming @mvo5's issues are resolved.

lzap:
I believe so, thank you very much. CLI followup: https://github.com/osbuild/image-builder-cli/pull/303

PR #1853, opened 2025-09-09
os: Run dracut when config files are added to image

bcl:
Previously the initrd was only recreated if fips mode was set, and partitions were created. But it also should be run when dracut configuration files are created using OSCustomizations.DracutConf. In addition, these dracut config files need to be created before the dracut stage runs.

With these changes the dracut config stages are created before org.osbuild.dracut is run, and dracut is run when fips is enabled or when configuration files have been added.

This changes quite a few manifests, fixing unnoticed issues with their initrd not including changes as expected because dracut was either not being run, or was running too soon.

eg. ec2 image should include sgdisk and blacklist the amdgpu and nouveau modules, but this was not actually being applied.


bcl:
ha! So, I like my solution better. Because it is mine ;) 
But also because your solution isn't complete -- in many of those cases dracut still isn't running because it is only added if fips has been selected and there is a /boot partition. If you take a look at the full manifests you will see there are dracut.conf changes but nothing adds the dracut run.
It also 'feels' a bit odd to be prepending stages when everything else goes in the order it is called, makes it harder for me to think about it.

supakeen:
> ha! So, I like my solution better. Because it is mine ;) 
> 
> But also because your solution isn't complete -- in many of those cases dracut still isn't running because it is only added if fips has been selected and there is a /boot partition. If you take a look at the full manifests you will see there are dracut.conf changes but nothing adds the dracut run.
> 
> It also 'feels' a bit odd to be prepending stages when everything else goes in the order it is called, makes it harder for me to think about it.

Hah, I like mine better (also because it's mine) because AFAIK dracut already runs during RPM installation and this way we avoid a second run (it can be quite slow).

I don't like prepending that much either but we also do it with the kernel cmdline for the same reason that it needs to be there before RPMs get installed.

achilleas-k:
I kinda like the explicitness of running dracut ourselves.  The slowness can be mitigated by [disabling dracut during the rpm stage](https://github.com/osbuild/osbuild/blob/60130ca19628c3ed5b4d67b1a76f812a67aca370/stages/org.osbuild.rpm.meta.json#L34-L36).  On the other hand, letting it run during rpm installation feels more "natural", so to speak, and the way we disable dracut in the rpm stage feels a little bit hacky.

Not sure where I lean on this one.

bcl:
Hm, I didn't realize it was running during rpm install. Which means there is also another issue, when we run the org.osbuild.dracut stage it has some defaults, like no early microcode and reproducible flag which are not passed when rpm runs it (compare the output of lsinitrd):

```
Image: ./initramfs-5.14.0-605.el9.x86_64.img: 84M
========================================================================
Early CPIO image
========================================================================
drwxr-xr-x   3 root     root            0 Aug 18 07:15 .
-rw-r--r--   1 root     root            2 Aug 18 07:15 early_cpio
drwxr-xr-x   3 root     root            0 Aug 18 07:15 kernel
drwxr-xr-x   3 root     root            0 Aug 18 07:15 kernel/x86
drwxr-xr-x   2 root     root            0 Aug 18 07:15 kernel/x86/microcode
-rw-r--r--   1 root     root       176162 Aug 18 07:15 kernel/x86/microcode/AuthenticAMD.bin
-rw-r--r--   1 root     root     20648960 Aug 18 07:15 kernel/x86/microcode/GenuineIntel.bin
========================================================================
Version: dracut-057-102.git20250818.el9

Arguments:  -f
```

With my changes:

```
Image: initramfs-5.14.0-605.el9.x86_64.img: 65M
========================================================================
Version: dracut-057-102.git20250818.el9

Arguments:  --no-hostonly --kver '5.14.0-605.el9.x86_64' --force -v --show-modules --no-early-microcode --reproducible
```

I think it's better to run it explicitly, with the argument we want. But apparently we've never done that so now we are in the situation of larger changes and more unknowns :/

I'm also worried that there may be cases when it gets run without `--no-hostonly` This can also be set by including the dracut-config-generic package. But if that's not in the package list it's going to build the initrd for the build host/build environment which may not be generic enough to always boot.

PR #1856, opened 2025-09-11
fedora & rhel-8,9,10: drop unused no_timer_check kernel parameter

lzap:
This parameter was needed in the past (circa 2014-2016) to work around issues
on older hardware, but it has not been necessary since the RHEL 7 era.

As explained in arch/x86/kernel/apic/io_apic.c, its original purpose was to
handle buggy SMP boards with incorrect mptable information about the timer IRQ.
On modern hypervisors like KVM, VMware, and Hyper-V, the kernel's timer
handling is robust, and the behavior already mimics the effect of this
parameter, making it redundant.

There is no reason to add this legacy parameter on RHEL 8 or later, and Fedora
19 and later. Removing it aligns with best practices by eliminating outdated
configuration, which helps prevent future confusion.

---

https://issues.redhat.com/browse/RHEL-11974


achilleas-k:
> [..]
>
> > If this was for me, I do not understand at all as I am not familiar with `images` workflow yet. I am shooting in the dark with this one trying to repeat what I have learned during reviews.
>
> More a question for the super experienced people, tbh I still often am not sure about policy details like this myself so I just want to make it explicit and hope that someone like @achilleas-k knows the answer :)

Taking the bug report at face value, this should not have any functional effect.  If that's the case, then I think it should be fine to drop this everywhere.

Since it came up though, I did try to find out more about this.
I tracked down the addition of the option in cloud images to [two RHBZs](https://bugzilla.redhat.com/buglist.cgi?columnlist=product%2Ccomponent%2Cassigned_to%2Cstatus%2Csummary%2Clast_change_time%2Cseverity%2Cpriority&list_id=13614067&order=status%2C%20priority%2C%20assigned_to%2C%20id%2C%20&query_format=advanced&short_desc=no_timer_check&short_desc_type=allwordssubstr): This one is for RHEL 8 rhel-guest-image (what we call the `qcow2` image): https://bugzilla.redhat.com/show_bug.cgi?id=1144155.  The other is for RHEL 6 and internal only.
The option works around an issue that only occurs in nested virtualisation (specifically KVM) scenarios.

I also found suggestions that this is [enabled by default when running on KVM](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/configuring_and_managing_windows_virtual_machines/optimizing-windows-virtual-machines), which I think is what the current bug/request is suggesting with the pasted function docstring.

It seems to me it's safe to drop the explicit option in images as far back as RHEL 8.

lzap:
Ok, rebased to resolve conflicts in manifests.

mvo5 (dismissed):

bcl (approved):
Looks good!

thozza:
@lzap this needs a rebase

thozza (dismissed):

supakeen (dismissed):

bcl (approved):
Ack Ack

PR #1858, opened 2025-09-13
Support uploading to libvirt

FrostyX:
I tried implementing this via `github.com/libvirt/libvirt-go` but it seems that it doesn't support uploading a file over SSH and it would have to be workarounded by copying the file via scp, and then uploaded to libvirt localy.

So I used virsh instead.

### DIFF
--- a/pkg/cloud/libvirt/libvirt.go
+++ b/pkg/cloud/libvirt/libvirt.go
@@ -1,0 +1,94 @@
+package libvirt
+
+import (
+	"fmt"
+	"github.com/osbuild/images/pkg/cloud"
+	"io"
+	"os"
+	"os/exec"
+)
+
+type libvirtUploader struct {
+	connection string
+	pool       string
+	volume     string
+}
+
+func NewUploader(connection string, pool string, volume string) (cloud.Uploader, error) {
+	return &libvirtUploader{
+		connection: connection,
+		pool:       pool,
+		volume:     volume,
+	}, nil
+}
+
+func (lu *libvirtUploader) CanShowProgress() bool {
+	return false
+}
+
+func (lu *libvirtUploader) Check(status io.Writer) error {
+	cmd := exec.Command("virsh", "help")
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("Can't execute virsh:\n%v", err)
+	}
+	return nil
+}
+
+func (lu *libvirtUploader) UploadAndRegister(r io.Reader, f *os.File, status io.Writer) (err error) {
+	fmt.Fprintf(status, "Uploading to libvirt...\n")
+
+	err = lu.Check(io.Discard)
+	if err != nil {
+		return err
+	}
+	err = lu.VolCreate()
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(status, "Starting to upload %s\n", f.Name())
+	fmt.Fprintf(status, "Progressbar is not supported.\n")
+	fmt.Fprintf(status, "This can take a while...\n")
+
+	err = lu.VolUpload(f)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(status, "File %s uploaded to %s\n", f.Name(), lu.connection)
+	return nil
+}
+
+func (lu *libvirtUploader) VolCreate() (err error) {
+	cmd := exec.Command(
+		"virsh",
+		"--connect", lu.connection,
+		"vol-create-as",
+		"--pool", lu.pool,
+		lu.volume,
+		"1M",
+	)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("virsh vol-create-as failed: %v", err)
+	}
+	return nil
+}
+
+func (lu *libvirtUploader) VolUpload(f *os.File) (err error) {
+	cmd := exec.Command(
+		"virsh",
+		"--connect", lu.connection,
+		"vol-upload",
+		"--pool", lu.pool,
+		lu.volume,
+		"--sparse",
+		"--file", f.Name(),
+	)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("virsh vol-upload failed: %v", err)
+	}
+	return nil
+}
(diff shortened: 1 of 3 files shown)


FrostyX:
This is the first Go code that I ever wrote, so it is entirely possible it sucks. Please let me know what should I improve. I tried to model the code after `pkg/awscloud`, so that there is at least some consistency.

mvo5:
It seems there are reasonably modern libvirt go bindings in https://gitlab.com/libvirt/libvirt-go-module/-/commits/master?ref_type=HEADS - could this be an option to avoid the external deps?

pkg/cloud/uploader.go
```diff
@@ -20,5 +21,8 @@ type Uploader interface {
 	// the reader and write status message to the given
 	// status writer.
 	// To implement progress a proxy reader can be used.
-	UploadAndRegister(f io.Reader, status io.Writer) error
+	UploadAndRegister(r io.Reader, f *os.File, status io.Writer) error
+
+	// Can this uploader show a progressbar while uploading the image?
+	CanShowProgress() bool
```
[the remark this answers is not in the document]
FrostyX: I use it in this PR https://github.com/osbuild/image-builder-cli/pull/300
I should have linked it, sorry.

FrostyX:
Thank you all for the review and feedback. I will try to reimplement this using `libvirt-go-module` and get back to you :-)

mvo5:
> I reimplemented the uploader to not rely on `virsh` anymore. Please take a look.

\o/ thanks!
 
> There is this ugly glitch that the final message appears before the progress bar is finished.
> 
> ```
> Uploading to libvirt...
> 1.62 GiB / 1.68 GiB [------------------------------------------------------------>__] 96.24% 1.18 GiB p/s
> File /home/jkadlcik/git/copr-image-builder/output/qcow2/disk.qcow2 uploaded to qemu:///system
> 1.68 GiB / 1.68 GiB [--------------------------------------------------------------] 100.00% 1.34 GiB p/s
> ```
> 
> But as I understand it, it's kinda related to our previous discussion about whether to pass `io.Reader` or `*os.File` when calling `UploadAndRegister`, so I am keeping the glitch in, and I would like to ask what to do next.

I suspect the glitch will be fixed with https://github.com/osbuild/image-builder-cli/pull/303/commits/d551763d807505d78f299ea32a85fb1c5356af27 which is part pf #303

FrostyX:
> I suspect the glitch will be fixed with https://github.com/osbuild/image-builder-cli/commit/d551763d807505d78f299ea32a85fb1c5356af27 which is part pf https://github.com/osbuild/images/pull/303

I rebased both this PR and https://github.com/osbuild/image-builder-cli/pull/300 against `main` but it seems it didn't help.

pkg/cloud/libvirt/libvirt.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (lu *libvirtUploader) Check(status io.Writer) error {
+	return nil
+}
+
+func (lu *libvirtUploader) UploadAndRegister(r io.Reader, f *os.File, status io.Writer) (err error) {
+	fmt.Fprintf(status, "Uploading to libvirt...\n")
+
+	// Connect to libvirt
```
mvo5, line 33: (nitpick) I'm not a fan of this kind of "what" comments (i.e. comments that explain what is happening in the next few lines). Its pretty clear from the code already and I feel like its not really adding much, so my suggestion would be to drop this (and the ones below that just explain the "what"). Note that I'm a huge fan of "why" comments, i.e. if needed, why something is happening, but here it seems everything is pretty straightforward :)

pkg/cloud/libvirt/libvirt.go
```diff
@@ -0,0 +1,118 @@
...
+	conn, err := lv.NewConnect(lu.connection)
+	if err != nil {
+		return fmt.Errorf("Failed to connect to libvirt: %v", err)
```
mvo5, line 36: Please use `%w` when wrapping errors, that ensures that the upper layers of go can use things like `errors.Is` to unwrap the underlying error type (here and below).

pkg/cloud/libvirt/libvirt.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (lu *libvirtUploader) VolumeXML(name string, size int64) string {
+	return fmt.Sprintf(`
+<volume>
+  <name>%s</name>
+  <capacity unit="bytes">%d</capacity>
+  <target>
+	<format type="qcow2"/>
+  </target>
+</volume>`, name, size)
```
mvo5, line 79: (nitpitck) we may want to use `xml.EscapeText([]byte(name)` here just to be on the safe side, so that a name like `my-silly-name</invaid-xml>` does not break anything. OTOH it will throw an errror to the user so I might overthinking it :)

pkg/cloud/libvirt/libvirt.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (lu *libvirtUploader) Upload(conn *lv.Connect, vol *lv.StorageVol, r io.Reader, size int64) (err error) {
+	// Initialize the upload stream
```
mvo5, line 83: (same comment about "what" coments as above)

pkg/cloud/libvirt/libvirt.go
```diff
@@ -0,0 +1,118 @@
...
+	// Stream the content using a 64KB buffer
+	buf := make([]byte, 64*1024)
+	for {
+		n, err := r.Read(buf)
+		if n > 0 {
+			if _, sendErr := stream.Send(buf[:n]); sendErr != nil {
+				return fmt.Errorf("Failed to stream the buffer: %v", sendErr)
+			}
+		}
+		if err != nil {
```
mvo5, line 104: (nitpick) Please move the error handling up, right under `n, err := ..`. This is the idiomatic way in go (your version is fine of course but using the idiomatic way makes reading easier/quicker.

mvo5 (dismissed):
Thanks a lot A few small suggestion about comments/error handling inline. This looks good otherwise.

Question for the more experienced members of the team - I assume the new libvirt-dev dependency (that we probably also need in the .github/workflow/tests.yml and container builds etc) is fine(?). 

We will probably need a stub implementation when build without cgo but I can take care of this.

supakeen (approved):
Since I already wrote code in this PR myself my review might not hold much value. I agree that we should escape the XML but that can be done as a follow-up; otherwise this is OK.

PR #1865, opened 2025-09-15
tools: run `-fake-bootc=true` in old manifest-gen too

mvo5:
Now that the fake bootc manifest generation has landed we can also add this flag when we run gen-manifests from the main repository.


supakeen (approved):

bcl (approved):
Ack!

PR #1869, opened 2025-09-16
tools: error in gen-manifest-checksums on failure

mvo5:
If `gen-manifests` fails because of e.g. unsupported/invalid configuration error instead of just continuing and leaving missing manifest checksums behind.


achilleas-k (approved):
Huh.  I noticed this a few days ago and thought I fixed it 😅 

TY!! <3

lzap (approved):

PR #1873, opened 2025-09-16
Add new pxe-tar image type (HMS-9284)

bcl:
This adds a new image type that consists of a kernel, inird, and compressed rootfs suitable for PXE booting systems. The tar also contains a README and an example grub.cfg


pkg/manifest/pxetree.go
```diff
@@ -11,27 +11,34 @@ import (
 
 type PXETree struct {
 	Base
-
-	// TODO
-	// - Add compression control (squashfs/erofs, etc.)
+	RootfsCompression string
+	RootfsType        ISORootfsType
```
lzap, line 16: Would be super useful to be able to pass level of compression, if that is something osbuild can support. For testing, I would probably set level 1 instead of the default 5 (or 9?) which would speed up build creationg.

But this is completely random, probably material for a different PR.

lzap (commented):
Oh I started a review, I meant to just drop random comments.

data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -2254,3 +2279,23 @@ image_types:
       - "distro"
       - "customizations.fips"
       - "customizations.locale"
+
+  "pxe-tar":
```
supakeen, line 2283: Let's call it `pxe-tar-xz` to be consistent with the other image types in this file that are compressed.

I understand this doesn't have a variant so keeping that out of the name is probably fine for now.

PR #1874, opened 2025-09-17
distro: `hidden` image types (HMS-9375)

supakeen:
Some of our image types are 'internal use only', in `osbuild-composer` we explicitly exclude them [1].

Instead of this information living in the executables this PR introduces the ability to have it be determined by the library (if `imagefilter` is being used to list images). We can still allow an `--all` option in, for example, `image-builder`.

Note that this PR doesn't yet mark any images as hidden; I'd like to keep that part of the discussion separate from the implementation.

[1]: https://github.com/osbuild/osbuild-composer/blob/main/cmd/osbuild-composer/config.go#L110-L117

### DIFF
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -1378,6 +1378,7 @@ image_types:
   "iot-bootable-container":
     <<: *rpm_ostree_imgtype_common
     filename: "iot-bootable-container.tar"
+    hidden: true
     mime_type: "application/x-tar"
     image_func: "bootable_container"
     exports: ["ostree-encapsulate"]
--- a/data/distrodefs/rhel-10/imagetypes.yaml
+++ b/data/distrodefs/rhel-10/imagetypes.yaml
@@ -1187,6 +1187,7 @@ image_types:
 
   "azure-rhui": &azure_rhui
     <<: *azure
+    hidden: true
     supported_blueprint_options: *supported_options_disk
 
   "azure-sap-rhui":
@@ -1484,6 +1485,7 @@ image_types:
   # RHEL internal-only x86_64 EC2 image type
   ec2: &ec2
     <<: *ami
+    hidden: true
     # we have to reset the aliases otherwise this type
     # will inherit the name aliases causing a conflict
     name_aliases: []
--- a/data/distrodefs/rhel-7/imagetypes.yaml
+++ b/data/distrodefs/rhel-7/imagetypes.yaml
@@ -282,6 +282,7 @@ image_config:
 
 image_types:
   "azure-rhui":
+    hidden: true
     filename: "disk.vhd.xz"
     mime_type: "application/xz"
     image_func: "disk"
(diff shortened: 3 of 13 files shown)


supakeen:
Note, I'm figuring out the test case situation at the moment so I'll probably amend the last commit with it.

achilleas-k:
I'm not sure this is the right way to go.  In my view, the showing or hiding of image types is a "product" decision, something that's decided by the packager of a frontend.  For example, osbuild-composer hides RHUI image types when packaged in distros, but we enable them when deploying the service.  We separate these two use cases by API, but that's all going away.

It makes sense to have one central location for the list, but I don't know if it's something that belongs in the library next to the definitions themselves.  It's a deployment/packaging/installation-time configuration and I think it would be simpler if we treated it as such.

But maybe this does make sense.  I wont argue strongly against this if others think it's a good idea.  I guess it does make sense for a distro maintainer to mark certain image configurations as hidden from image type listings.  It'll be possible to override them anyway if necessary.

lzap (approved):

lzap:
Now I am reading @achilleas-k 's comment. Hmmm.

The context is important, what is not something a CLI user would be building could make sense for a Lightspeed user. I wonder how many contexts there are. Maybe instead of a simple "visibility" boolean, this could be a slice of strings that define use cases like `cli`, `lightspeed` or `rhui`. Then one could list images from any context at any point.

supakeen:
> Now I am reading @achilleas-k 's comment. Hmmm.
> 
> The context is important, what is not something a CLI user would be building could make sense for a Lightspeed user. I wonder how many contexts there are. Maybe instead of a simple "visibility" boolean, this could be a slice of strings that define use cases like `cli`, `lightspeed` or `rhui`. Then one could list images from any context at any point.

We sadly didn't get to discuss this during the team meeting but the alternative is that we do implement these `FilterOptions` and allow passing in a list of image types to exclude; each executable that uses `imagefilter` can then decide what it wants to leave out.

supakeen:
Converting to draft until we can have a chat about which direction to pick.

mvo5:
> I'm not sure this is the right way to go. In my view, the showing or hiding of image types is a "product" decision, something that's decided by the packager of a frontend. For example, osbuild-composer hides RHUI image types when packaged in distros, but we enable them when deploying the service. We separate these two use cases by API, but that's all going away.
> 
> It makes sense to have one central location for the list, but I don't know if it's something that belongs in the library next to the definitions themselves. It's a deployment/packaging/installation-time configuration and I think it would be simpler if we treated it as such.
[..]

Its an interesting use-case. When I was chatting with Simon about this before we talked about the possiblity of instad of a boolean have `tags: [internal, service-visible]` etc. And then `imagefiler.Filter("tag:!internal")` or `imagefilter.Filter("tag:service-visible")` could be used. It would still be a central location but more flexible.

Long term I think we need to sovle the "imagedefinitions.d" problem anyway, i.e. a way to override/amend existing imagetypes, that could then be used by the downstreams to tweak the tags (add/remove/change).

supakeen:
> > Just to double check - do you feel the same about the "tags" idea above, ie do you still feel this is not flexible enough/too central ?
> 
> If I'm reading that correctly the library still holds the tags so changing them still means a library change. I don't see that being a significant difference, it would still need to be rebuilt to change the tags.

With the idea to make the YAML definitions distribution owned and external to the library (e.g. similar to kickstarts), would the concept of either a hidden attribute; or tags, make more sense?

PR #1876, opened 2025-09-17
add boot test for WSL images (x86_64) (HMS-9293)

croissanne:
Boots a windows vm from a snapshot on azure, copies the wsl image, runs the batch script which imports the wsl image and runs the test script.

### DIFF
--- a/test/scripts/boot-image
+++ b/test/scripts/boot-image
@@ -5,10 +5,12 @@ import json
 import os
 import uuid
 from tempfile import TemporaryDirectory
+import shutil
 
 import imgtestlib as testlib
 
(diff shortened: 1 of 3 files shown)


achilleas-k (dismissed):
Thank you.  This is great!
LGTM

supakeen (approved):
Very cool. I'm a bit vague on if and how we would like to test the OOBE (we're skipping it since we're using `-d`) but maybe that's not up to us.

croissanne:
> Very cool. I'm a bit vague on if and how we would like to test the OOBE (we're skipping it since we're using `-d`) but maybe that's not up to us.

Not against it, but really not looking forward to doing that interactively in the windows command prompt. Maybe powershell would work better idk.

PR #1877, opened 2025-09-17
centos: allow the `ec2` image type (HMS-9326)

supakeen:
CentOS has indicated that they want to have `xz`-compressed AMI's. We already have the `ec2` image type for this. Let's remove it from RHEL-only images so it becomes available.


supakeen:
> The change looks fine, the backstory makes me wonder if we should rename "ec2" to "ami-xz" or something though (and keep the previous name as alias), wdyt? Still approving as this is kinda orthogonal

We can and probably should, in a follow-up. Renaming image type base names is always a bit hectic if I recall from previous times I did this (in Fedora).

lzap (approved):

PR #1878, opened 2025-09-17
tools: conflict-less manifest checksums

lzap:
The current approach is overwriting manifest checksum files which causes
git conflicts. This patch appends last X characters of the checksum to the
filename to avoid conflicts.

While this doubles amount of git changes for each manifest update, it
prevents merge conflicts and is easier to handle. It also does not
increase the overall size of the repository because file overwrite is a
new blob for git anyway.

---

I did not explain in the commit message but I guess this is clear - file creation and deletion is a conflict-less operation in git, unless file with the same name is changed which is very unlikely with shasum in the filename.

Curious what you think about this. Did not care to change the validation part yet as I am really not sure what are the chances of getting this in...

Edit: Ah if I squash the two commits I guess it should pass right away.


mvo5:
This is interesting. In the case of conflicts we now have slightly confusing behavior I believe (unless I misunderstand how this works):
- branch A would create conflicts, main has $prefix-ABCD now gets a new set of checksums files from A with $prefix-DCBA
- branch A gets merged, we now have chechums/$prefix-{ABCD,DCBA} in the repo
- the next (unrelated) run of ./gen-manifest-checksums.sh will delete the extra ones and it will show up in another unrelated PR

Not saying its better or worse, just if we want this changed behavior (which is also going to be a bit annoying) over the current conflicts (I don't have a strong opinion yet)?

achilleas-k:
> though it _is_ annoying to pull in people to re-review each time this happens

This definitely slows things down, I agree.  Would be nice to get away from this, though like you said, it's not too much trouble to fix.

I share Michael's concerns about creating inconsistencies after a clean rebase.  With our current checks, we would be able to add PRs to the merge queue that would then fail the check once the test started on the queue.  Which doesn't really solve anything in the end.

lzap:
Haven't realized that files would start to stack up. Well at least I learned the `--exec` trick, I was doing it interactively.

PR #1882, opened 2025-09-18
manifest: add error to Pipeline.serialize() 

mvo5:
This was suggested by Achilleas in https://github.com/osbuild/images/pull/1879#discussion_r2358519118 - its a start if its liked I can do serialzeEnd() too and go over the "in-between" panics in the various helpers (e.g. anaconda_installer still has some indirect panics).

---

manifest: add `error` to Pipeline.serialize()

This commit makes the Pipeline.serialize() add support for
error returns. This diff is purely automatic and does not
change any of the existing panics into errors.

---

manifest: convert panic() in serialize() to errors

This commit changes the panics() in serialize() to errors.

PR #1883, opened 2025-09-19
manifest: remove most panics()

mvo5:
With https://github.com/osbuild/images/pull/1882 we can not return errors in more places and need to panic less. This PR contains some followups to eliminate more panics. With it merged we go from 95 panics() in pkg/manifest to just 20 (most in serializeEnd, the rest in things like `getBuildPackages()` which can be caused by unsupported architecutures, see https://github.com/osbuild/images/blob/v0.193.0/pkg/manifest/anaconda_installer.go#L127)


lzap (approved):

bcl (approved):
Looks good!

PR #1884, opened 2025-09-19
disk: sort partitions in a partition table by their start position [RHEL-83078]

achilleas-k:
The relayout() function until now would change the order of partitions to put the root partition, volume group, or btrfs volume at the end, but only did so by changing the start sector of partitions.  The order in the array, which affects the creation order and determines the partition number on the table, was never affected.  While this is generally valid, it was surprising that the layout order did not match the partition numbering.

This commit changes the relayout() function to always sort partitions in the table by their start position.


bcl (approved):
Looks good to me. Even if someone depends on the wrong numbering this is an improvement for all future users. Having partition numbers not match order is confusing :)

bcl (approved):
Ack!

PR #1885, opened 2025-09-19
osbuild: add pipeline name to monitor status

lzap:
In order to print reasonable duration statistics, we need to know which pipeline is currently running. The osbuild monitor JSON status messages now include the pipeline name in the context object. Without this, one could not tell if the slowest stage was RPM in os or build pipeline.

It also adds a testdata that is a complete and unmodified osbuild of a Fedora 42 raw image.

---

https://github.com/osbuild/osbuild/issues/2196

HMS-9311


pkg/osbuild/monitor_test.go
```diff
@@ -155,6 +159,27 @@ func TestScannerDuration(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	for {
+		st, err := scanner.Status()
+		assert.NoError(t, err)
+		if st == nil {
+			break
+		}
+
+		assert.NotEmpty(t, st.Pipeline)
```
mvo5, line 179: I would prefer if the test would be a bit more specific, maybe we can collect all pipeline names and do a `assert.Equal(t, []string{"build", "sources", ...)` here, wdyt?

Issue #1578, opened 2025-06-03
Enable Customization of partition table to set start offset of ESP


ajayramaswamy:
I am using imagebuilder cli to make a sd card image for my FriendlyElec Nano Pi R5s. The rock chip RK3568 needs the uboot image to be embedded after the partition table and before the ESP,  I would like to specify that the esp starts at 16M but right now the start offset is hard coded to 8M

ref
https://opensource.rock-chips.com/wiki_Boot_option

https://github.com/osbuild/images/blob/8a654cdb09d3bb2421a1a4aeeb733872fade2970/pkg/distro/defs/fedora/distro.yaml#L400

mvo5:
How exactly would this work? How would the uboot be installed after the partition table? This is a manual step for you currently, is that right?

If so, I wonder if we should add the ability to have "binary payloads" for partitions so that during the build of the image a file can be put into the right place on the disk image. This combined the ability for you to create your own image type via YAML (similar to e.g. https://github.com/osbuild/images/blob/main/pkg/distro/defs/fedora/distro.yaml#L846) would also sove this, right? (but I understand the blueprint offset would be a quicker path to success for you probably?).

ajayramaswamy:
right now the way to do this is 
1. first generate the Fedora-42-minimal-aarch64.img.xz with imagebuilder
2. write that to sdcard
3. then add the uboot to the sdcard by dd
`dd bs=32k seek=1 if=u-boot-rockchip.bin of=/dev/sdcard
`

right now I have to mount sdcard backup esp delete esp create with new offset and then add uboot and restore esp, this is necessary since the uboot-rockchip.bin ia around 10MB and growing after every release

in case this rfe is accepeted and implemented then in the blueprint.toml I would set the offset of the ESP to 16M and after that embed the uboot with 
`dd bs=32k seek=1 if=u-boot-rockchip.bin of=Fedora-42-minimal-aarch64.img conv=notrunc,fsync
`

If you can think of a way of adding the dd parts to image builder then it would be great.

Thanks

supakeen:
This issue was linked in proposed Fedora blocker bug: https://bugzilla.redhat.com/show_bug.cgi?id=2396309 I'll take a look at this today/tomorrow on if we should either bump the offset by default *or* allow control over the offset through some means.

ajayramaswamy:
works for me now!

PR #1888, opened 2025-09-22
fedora: double aarch64 partition table offset

supakeen:
Move the partition table offset 16 MiB from the start of the disk instead of the previous 8 MiB. For some devices (notably, some RockChip) devices it seems that 8 MiB is not enough to fit firmware [1].

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2396309

---

Opening this as a draft PR for #1578, pending confirmation that this works as a fix and is acceptable. I'll likely do a follow-up PR to allow customizing `start_offset` from the disk customizations in blueprints as well which gives more control to the user though that might warrant some (more) discussion.


achilleas-k:
> Fwiw, I like this approach better (at least for now)

Compared to what?

PR #1899, opened 2025-09-24
test/configs: add jq to all test configs that require it

achilleas-k:
Add jq to the package list of each configuration that might need it.  jq is now required by test/scripts/check-host-config.sh script in order to parse each test config and verify the customizations.

The requirements are determined by the following criteria:
- The config is used in an image type that is booted in CI (AWS, Azure, and WSL).
- The config doesn't already install jq or something that includes it (for example, the core group on RHEL 9+ installs jq).

For a lot of image types, the empty-rhel config was replaced by the jq-only-rhel config.  The empty-rhel config is now used for the net installer.

A new config was added for RHEL 7 because it doesn't have jq available and can't be included with the new jq-only-rhel config group.

Verified by generating all depsolved manifests and calling:

    grep "jq.*\.rpm" *ami* *ec2* *azure* *vhd* -L

---

PR includes some additional cleanups and formatting fixes.

PR #1900, opened 2025-09-24
Convert the test config map to a config list

achilleas-k:
DRAFT: includes #1899

This PR converts the config-map.json file to a list.

The purpose of this change is to have more flexibility with configs that
need to apply to multiple filter sets.  For example, with the existing
config map, it is not possible to apply the same config to the
server-qcow2 file for Fedora  and at the same time all image
types for CentOS Stream 9, because the following is not possible:

```json
{
    "configs/example.json": {
        "distros": [
            "fedora-*"
        ],
        "image-types": [
            "server-qcow2"
        ]
    },
    "configs/example.json": {
        "distros": [
            "centos-9"
        ]
    }
}
```

By converting the config map to a list, the above configuration is now
possible as follows:

```json
[
    {
        "path": "configs/example.json",
        "filters": {
            "distros": [
                "fedora-*"
            ],
            "image-types": [
                "server-qcow2"
            ]
        }
    },
    {
        "path": "configs/example.json",
        "filters": {
            "distros": [
                "centos-9"
            ]
        }
    }
]
```


thozza (approved):
This makes a lot of sense. Also, nice cleanup 😉

achilleas-k:
This will probably conflict with some of the other open PRs and it's a pretty low priority right now.  Not gonna set it to draft or anything, just people know.

croissanne (dismissed):

PR #1901, opened 2025-09-25
defs: typo in RHEL-7 UBI URL

supakeen:
Don't even know how I spotted this but I doubt there's actually a space in the URL?


mvo5 (approved):
Thanks! no idea how that ended there :( I wonder if we should generate cross-arch manifests to test for this - oh well. Thanks for spotting/fixing it!

P.S. Its still mind-boggling to me that this works:
```
$   podman run -it --rm registry.access.redhat.com/ubi7/ubi:latest 
# 
```
and I get a fully patched system that dates back to 2014… good job RH!

PR #1904, opened 2025-09-25
distro: turn all errors from distro.ValidateConfig() into warnings (HMS-9444)

achilleas-k:
The stricter blueprint validation is causing a lot of issues in the service and frontend where customizations that were previously silently ignored are causing errors.  This is causing issues because some error cases aren't being reported to users and it appears that the service is simply not responding to a build request.

Relax the blueprint validation errors for now, turning them into warnings.  We should turn them back into errors when consumers of the library have adapted to handle the new error conditions.


mvo5 (dismissed):
Very nice! Thank you!

supakeen (approved):

PR #1906, opened 2025-09-25
[WIP] many: add support for bootc container based anaconda ISOs

mvo5:
[draft as this is still a bit messy (but not too terrible anymore) and misses tests]

This PR adds support for a new `bootc-instalelr` image type that is a bootc container based anaconda ISOs. Instead of building our iso from rpm packages and just putting the bootc container into the rpm generated ISO this spike builds the ISO from a bootc container.

This requires that the container has the right content of course (i.e. anaconda). We would move some of the responsibility to the user (or to a tool that helps the user and hides the complexity here like bootc-iso-imagectl or something) to ensure that.

See also https://github.com/osbuild/bootc-image-builder/pull/1053


supakeen (approved):
I've amended https://github.com/osbuild/images/pull/1906/commits/972d95f538d0292f2880b536e06950aa8d9a5142 to be compatible with the changes in #1949.

Otherwise while I have my concerns in general about shipping Anaconda as a container let's go forward with this. Code is OK.

thozza (approved):

achilleas-k (approved):

PR #1907, opened 2025-09-26
pkg/rpmmd: unify RPM package structure implementations (HMS-8911)

thozza:
`rpmmd` package used to contain multiple data structures representing an RPM package, with a varying number of package metadata. All `depsolvednf.Solver` commands were using a different `rpmmd` struct in/as their return value.

This PR unifies all structs in `rpmmd`, which represent an RPM package metadata, into a single struct called `rpmmd.Package`. The struct definition is based on DNF4 and DNF5 code and covers all properties relevant for generating manifests and any future artifacts, such as SBOM documents.

`depsolvednf.Solver` API implementation for calling `osbuild-depsolve-dnf` is decoupled from the `rpmmd` structures. All `depsolvednf.Solver` commands (depsolve, package search, repo dump) return value now uses the same `rpmmd` struct to represent an RPM package. For now, the scope of set metadata in the `rpmmd.Package` struct is the same as with the original data structures. The next step is to enhance `osbuild-depsolve-dnf` to return complete RPM package metadata in its responses.


pkg/rpmmd/package.go
```diff
@@ -0,0 +1,172 @@
...
+func (p *Package) GetEVRA() string {
+	if p.Epoch == 0 {
+		return fmt.Sprintf("%s-%s.%s", p.Version, p.Release, p.Arch)
+	}
+	return fmt.Sprintf("%d:%s-%s.%s", p.Epoch, p.Version, p.Release, p.Arch)
+}
+
+type PackageList []Package
+
+func (pl *PackageList) GetPackage(packageName string) (Package, error) {
```
mvo5, line 141: (super nitpick) Maybe just `Package()` (c.f. https://go.dev/doc/effective_go#Getters) but we don't really follow this in many places so feel free to ignore.

pkg/rpmmd/package.go
```diff
@@ -0,0 +1,172 @@
...
+	if pl == nil || len(*pl) == 0 {
```
mvo5, line 142: It seems fine to me to ask for a package in an empty list and get a "NotFound" (slices.Index() also does not differentiate this), i.e. I would just remove this check (unless there is a use-case I'm missing of course).

pkg/rpmmd/package.go
mvo5, line 141: Given that a slice is already a (fat) pointer, could we just make this a `func (pl PackageList) GetPackages...` ? That avoids the need to check for nil in here as range over a nil list is fine, something like:
```go
func (pl PackageList) Package(packageName string) (Package, error) {
	for _, pkg := range pl {
		if pkg.Name == packageName {
			return pkg, nil
		}
	}

	return Package{}, fmt.Errorf("package %q not found in the Package list", packageName)
}
```

mvo5 (commented):
Some very early ideas/questions inline (or, nitpicks?) - I already love that we have much nicer names now.

pkg/rpmmd/package.go
thozza: Thanks for pointing this out, I'll adjust the getter, as well as other getters for the `rpmmd.Package` struct. Shorter names and following Go idioms is a good thing :wink:

PR #1909, opened 2025-09-29
pkg/distro/test_distro: ensure Manifest.*Pipeline() works as expected

thozza:
In the past, each image type has been defining a static list of its build and payload pipelines. This is now handled dynamically on the `Manifest` type level by iterating over all pipelines and picking the ones based on their type.

This however didn't work properly for the test_distro image types, which were returning a `Manifest` instance with two pipelines (build and os), but the build pipeline was not a proper `manifest.Build` type. Thus the Manifest.BuildPipelines() returned an empty list, while Manifest.PayloadPipelines() returned both pipelines.

To fix this, a new manifest.ContentTestBuild pipeline is added, which provides the means to create a testing build pipeline implementing the `manifest.Build` interface. Note that the manifest.ContentTest can't implement the `manifest.Build` interface, because then even non-build pipelines would be treated as build pipelines by
Manifest.BuildPipelines() method.

Modify the manifest.NewContentTest() to require a build pipeline, which is the case for all other pipeline implementations that are not used only for testing.

This change is needed in order to not break osbuild-composer API unit tests, which use the distro.TestDistro. Previously the API there was setting fallback values for pipeline names, which was a compatibility layer from 4 years ago when we migrated from v1 to v2 osbuild manifests. Removing the unneeded compatibility layer in osbuild-composer revealed the limitation of distro.TestDistro. See osbuild-composer PR#4875 [0] for more information.

[0] https://github.com/osbuild/osbuild-composer/pull/4875

### DIFF
--- a/pkg/distro/test_distro/distro.go
+++ b/pkg/distro/test_distro/distro.go
@@ -18,9 +18,6 @@ import (
 const (
 	// package set names
 
-	// build package set name
-	buildPkgsKey = "build"
-
 	// main/common os image package set name
 	osPkgsKey = "os"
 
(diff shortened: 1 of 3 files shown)


achilleas-k (approved):
Neat.  Thank you.
LGTM

PR #1912, opened 2025-09-29
github: use PR user.login instead of actor to autoapprove dependabot PRs

achilleas-k:
Using github.actor to determine if a PR was created by dependabot and autoapprove it can be exploited using a "Confused Deputy" attack.

See https://github.com/osbuild/images/security/advisories/GHSA-v669-m9gr-3mgp


lzap (approved):
Interesting.

thozza:
> Using github.actor to determine if a PR was created by dependabot and autoapprove it can be exploited using a "Confused Deputy" attack.
> 
> The automerge action is not useful anyway because we always need a second approval, which requires human review.

The second approval by a human is the reason why I thought that this attack is quite theoretical... :innocent:

achilleas-k:
> Note that the GHSA link is only visible to contributors of this project with the appropriate rights, for example: not me. Otherwise fine.

Oh, I wasn't aware.  Good to know.  I'll see if I can make it (or some part of it) public.

PR #1913, opened 2025-09-30
manifest: add firstboot support (HMS-9187)

lzap:
Add native support for firstboot and registration with support for arbitrary ordering.

_The format of the blueprint was rewritten in the middle of the PR, merged into `blueprint`._

---

Replaces: https://github.com/osbuild/images/pull/1705

Fixes: https://redhat.atlassian.net/browse/HMS-9187


pkg/manifest/firstboot.go
```diff
@@ -0,0 +1,168 @@
...
+	"io/fs"
+	"strings"
+
+	"github.com/osbuild/images/internal/common"
+	"github.com/osbuild/images/pkg/customizations/firstboot"
+	"github.com/osbuild/images/pkg/customizations/fsnode"
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+// checkName prevents path traversal
```
supakeen, line 15: While this is certainly a side-effect of this function it's also overly restrictive for preventing path traversal (`systemd-firstboot-первый-старт.service` is a perfectly valid filename). This seems better suited to live in the blueprint library?

In `images` we'd then take the `basename` (to be defensive and have some locality) and then throw it through `systemd-escape` when we put it in the stage (since `systemd` has its own ideas for escaping). We already have a helper for the last bit in `internal/common/helpers.go` (`MountUnitNameFor`) that could be duplicated (or expanded) to also handle service names?

pkg/manifest/firstboot.go
```diff
@@ -0,0 +1,168 @@
...
+		if !(('a' <= r && r <= 'z') || ('A' <= r && r <= 'Z') || ('0' <= r && r <= '9') || r == '-' || r == '_') {
+			return fmt.Errorf("name can only contain alphanumeric characters, dashes, and underscores")
+		}
+	}
+
+	return nil
+}
+
+var tmplFirstbootAAP = `#!/usr/bin/bash
+curl -s -i --data "host_config_key={{ .HostConfigKey }}" {{ .URL -}}
```
supakeen, line 31: And if we are being defensive then we might as well escape this too so it's obvious and local. We have `Quote` in `pkg/util/shlex_quote.go`.

pkg/manifest/firstboot.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	var certs []string
+	var files []*fsnode.File
+	var executables []string
+
+	var ci int
+	eachCustom := func(opt firstboot.CustomFirstbootOptions) error {
+		// keep the naming convention consistent with the existing "osbuild-first-boot"
+		name := fmt.Sprintf("osbuild-first-%s", opt.Name)
```
supakeen, line 61: This is where we'd use `systemd-escape` (see earlier comment).

pkg/manifest/firstboot.go
```diff
@@ -0,0 +1,168 @@
...
+		if err != nil {
+			return fmt.Errorf("error creating firstboot file node: %w", err)
+		}
+		files = append(files, f)
+
+		// prepare data for the systemd unit
+		if opt.IgnoreFailure {
+			exec = "-" + exec
+		}
+		executables = append(executables, exec)
```
supakeen, line 79: This bit is repeated in each `each*` function, lift it to a helper?

pkg/manifest/firstboot.go
```diff
@@ -0,0 +1,168 @@
...
+	var result strings.Builder
+	err = tmpl.Execute(&result, data)
+	if err != nil {
+		return "", fmt.Errorf("error rendering firstboot unit: %w", err)
+	}
+
+	return result.String(), nil
+}
+
+func firstbootFileNodes(fbo *firstboot.FirstbootOptions) ([]string, []*fsnode.File, *osbuild.SystemdUnitCreateStageOptions, error) {
```
supakeen, line 49: It does a bit more than just file nodes right?

pkg/manifest/firstboot.go
```diff
@@ -0,0 +1,168 @@
...
+		if opt.IgnoreFailure {
+			exec = "-" + exec
+		}
+		executables = append(executables, exec)
+
+		return nil
+	}
+
+	// iterate over all firstboot scripts and create the necessary structures
+	fbo.Each(eachCustom, eachSatellite, eachAAP)
```
supakeen, line 140: Check error?

pkg/manifest/os.go
```diff
@@ -658,6 +661,24 @@ func (p *OS) serialize() (osbuild.Pipeline, error) {
 		pipeline = prependStage(pipeline, osbuild.NewDracutConfStage(dracutConfConfig))
 	}
 
+	fbCerts, fbFiles, unit, err := firstbootFileNodes(p.OSCustomizations.Firstboot)
+	if err != nil {
+		panic(err.Error())
```
supakeen, line 666: Just `err`?

pkg/manifest/os.go
```diff
@@ -658,6 +661,24 @@ func (p *OS) serialize() (osbuild.Pipeline, error) {
 		pipeline = prependStage(pipeline, osbuild.NewDracutConfStage(dracutConfConfig))
 	}
 
+	fbCerts, fbFiles, unit, err := firstbootFileNodes(p.OSCustomizations.Firstboot)
```
supakeen, line 664: Bit of a nitpick, but why not call `unit` -> `fbUnit`?

pkg/customizations/firstboot/firstboot.go
```diff
@@ -0,0 +1,163 @@
...
+				return err
+			}
+		case AAPFirstbootOptions:
+			if err := cbAAP(s); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown firstboot option type: %T", script)
+		}
+	}
```
supakeen, line 161: I'm not sure about this idiom of having an interface to put everything into a single array then taking them out again. It feels a bit that you don't really gain anything since you only use `.Each` to access this? I'll have to think more when I'm more awake.

supakeen (commented):
Did a quick read through, there might be other or more things but this is what jumped out with my morning coffee.

Two things:

1. You forgot to add `Firstboot` to `OSCustomizations`.
2. The panic thing commit can be a separate PR since it's the first thing I saw when opening this PR and I thought 'wait this isnt firstboot' ;)

pkg/manifest/firstboot.go
lzap: I think using `systemd-escape` is an overkill, but you are right that there is no reason to be too restrictive. I will just use `filepath.Base(userInput)` to do the same job.

pkg/manifest/firstboot.go
lzap: It is really on the edge of 3-4 lines where I deem it acceptable and it is very local stuff. I think perhaps inner function might be ok let me try.

pkg/manifest/firstboot.go
lzap: Yeah refactoring caused this.

pkg/customizations/firstboot/firstboot.go
lzap: I do not like it too, but keep in mind the BP allows you to order all firstboot scripts, including AAP and Satellite, so there has to be a common structure holding it. Unless I want to store indices and implement some kind of custom iteration, that sounds uglier than this.

What is a bit weird is the `Each` iterator, I have to agree. I am not sure about it, removing it will not get rid of the type assertion tho.

avitova (commented):
I tried very hard to find something, but this actually looks very nice. No objections.

achilleas-k (commented):
I think the duplication of the Blueprint structures in `pkg/customizations/firstboot/` is a bit unnecessary.  The `pkg/customizations/` types are meant to be convenient internal representations of options that serve as intermediate types between blueprint customizations and osbuild stages.  Sometimes, it's convenient to have something that very closely resembles the blueprint itself (like with users and groups), because that makes the most sense.  In other cases, it might resemble the stage options closer, or be completely independent.

I think in this case, the Blueprint structures aren't convenient internally.  They're designed to be convenient for user input, but the union types make it awkward to handle.  So copying them from one set of union types to another isn't really giving us any benefit.

Looking at the implementations of the `each` functions, it looks like for every firstboot customization we basically generate an executable file and an exec line for the systemd unit (with the optional `-` prefix).  Also certs for AAP and Satellite.  It seems to me that these things would be perfect as an intermediate representation of the firstboot customizations.  So what I'm imagining is the following:

1. `pkg/customizations/firstboot/` defines a type, `Script` that's basically
```go
type Script struct {
    Filename      string
    Contents      string
    IgnoreFailure bool
    Certs         []string
}
```

2. The `FirstBootOptionsFromBP()`, instead of copying the BP structs to identical ones inside images, implements analogues to the `each` functions (the ones that are currently in `pkg/manifest/`) that take `blueprint.FirstBootCustomization` and return `[]firstboot.Script`.

3. In `pkg/manifest/`, a function takes `[]firstboot.Script` and produces a set of `[]fsnode.File`, CA certs, and systemd unit create stage options (equivalent to what `parse()` does now, but I imagine with a simpler implementation).


This way, it would be a bit easier to (for example) define a firstboot script internally for something we want to define statically in an image type.  Consider:
```go
regFirstboot := firstboot.SatelliteFirstbootOptions{
    FirstbootCommonOptions: firstboot.FirstbootCommonOptions{
        Name:          "satellite",
        IgnoreFailure: true,
    },
    CACerts: []string{"cert1", "cert2"},
    Command: "#!/usr/bin/bash\ncurl https://sat.example.com/register",
}

certs, files, unit, err := parse(&firstboot.FirstbootOptions{
  Scripts: []firstboot.FirstbootOption{
      regFirstBoot,
  }
})
```

vs

```go
regFirstboot := firstboot.Script{
    Filename: "satellite",
    Command: "#!/usr/bin/bash\ncurl https://sat.example.com/register",
    IgnoreFailure: true
    CACerts: []string{"cert1", "cert2"},
}

files, certs, unit, err := genFirstbootComponents([]firstboot.Script{regFirstBoot})
```
(minor difference, but more readable IMO).

achilleas-k (changes requested):
Thank you.  This looks a lot nicer now.  Some minor changes requested below.

pkg/customizations/firstboot/firstboot.go
```diff
@@ -0,0 +1,187 @@
...
+func (AAPFirstbootOptions) isFirstbootOption()       {}
+
+var ErrFirstbootAlreadySet = errors.New("firstboot customization already set")
+
+var tmplFirstbootAAP = `#!/usr/bin/bash
+curl -i --data {{ .HostConfigKey }} {{ .URL }}
+`
+
+func renderFirstboot(tmplStr string, data any) (string, error) {
+	tmpl, err := template.New("firstboot-unit").Parse(tmplStr)
```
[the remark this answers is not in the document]
lzap: The idea we implemented in CRC was that I expected more of these integrations to be added and I wanted a common way of rendering these that is capable enough for longer scripts. Granted that only two lines is much easier to do using string formatting. Let me change that now.

supakeen (changes requested):
> I rebased, this is ready.

It's failing in CI with the following, which leads to the check host saying things were unsuccessful:

```
Dec 16 13:51:44 my-host (-first-.)[1092]: osbuild-custom-first-boot.service: Executing: /usr/local/bin/osbuild-first-.
Dec 16 13:51:45 my-host osbuild-first-.[1094]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Dec 16 13:51:45 my-host osbuild-first-.[1094]:                                  Dload  Upload   Total   Spent    Left  Speed
Dec 16 13:51:45 my-host osbuild-first-.[1094]: [158B blob data]
Dec 16 13:51:45 my-host osbuild-first-.[1094]: curl: (7) Failed to connect to localhost port 443 after 7 ms: Could not connect to server
Dec 16 13:51:45 my-host systemd[1]: osbuild-custom-first-boot.service: Child 1092 belongs to osbuild-custom-first-boot.service.
Dec 16 13:51:45 my-host systemd[1]: osbuild-custom-first-boot.service: Main process exited, code=exited, status=7/NOTRUNNING
Dec 16 13:51:45 my-host systemd[1]: osbuild-custom-first-boot.service: Failed with result 'exit-code'.
```

```
❌ Status for all failed units
× osbuild-custom-first-boot.service
     Loaded: loaded (/usr/lib/systemd/system/osbuild-custom-first-boot.service; enabled; preset: disabled)
     Active: failed (Result: exit-code) since Tue 2025-12-16 14:52:28 GMT; 30s ago
 Invocation: 4ec8bc4876f64bd7b9ae839f90c64818
    Process: 1435 ExecStartPre=/usr/bin/rm /var/local/.osbuild-custom-first-boot (code=exited, status=0/SUCCESS)
    Process: 1440 ExecStart=/usr/local/bin/osbuild-first-. (code=exited, status=7)
    Process: 1458 ExecStart=/usr/local/bin/osbuild-first-. (code=exited, status=7)
   Main PID: 1458 (code=exited, status=7)
   Mem peak: 2.3M
        CPU: 33ms
```

lzap:
> It's failing in CI with the following

Yeah already on it thanks. Feel free to drop code comments tho.

thozza (dismissed):
Thanks.

In general, the code looks good.

I've added some comments inline. While I have reservations regarding the ordering of the firstboot unit files, which I raised on multiple occasions, I can live with the code as is (Alhough I don't want to be the one who fixes this aspect in the future, because it will be bad to change any hard-coded ordering out of fear that **some** users may depend on it and we don't want to break them).

The blocking issues from my PoV:

- `handleFirstbootOptions()` should be moved to the `osbuild` package
- The first commit assigns to `osc.Firstboot`, which is added by the second commit.

lzap:
I know this was long, way too long for what it is doing. But I haven't gave up, please re-review @supakeen and @thozza and anyone else!

supakeen (commented):
Needs rebase.

lzap:
Rebased, can I get reviews in?

avitova (commented):
I am only unsure about the logic to create names for firstboot files, otherwise LGTM.

supakeen (commented):
Generally fine at this point. Anything still open (for example allowing users to define service ordering and dependencies themselves) can, IMHO, be follow-ups.

@avitova's remarks need to be addressed so not approving.

pkg/customizations/firstboot/firstboot.go
```diff
@@ -0,0 +1,167 @@
...
+		if inputName == "" || !filepath.IsLocal(inputName) ||
+			slices.Contains(alreadyUsed, inputName) || reservedRegexp.MatchString(inputName) {
+			ci++
+			return fmt.Sprintf("osbuild-first-%s-%d", prefix, ci)
+		}
+
+		// keep the naming convention consistent with the existing "osbuild-first-boot"
+		alreadyUsed = append(alreadyUsed, inputName)
+		return fmt.Sprintf("osbuild-first-%s", inputName)
+	}
```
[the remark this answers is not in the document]
lzap: How about this? I also added a test.

lzap:
Rebased.

supakeen (dismissed):

lzap:
Resolved conclicts so many reviews but why everyone dropped it @avitova @achilleas-k @thozza thanks

pkg/manifest/os.go
```diff
@@ -967,8 +987,10 @@ func (p *OS) serialize() (osbuild.Pipeline, error) {
 		}))
 	}
 
-	if len(p.OSCustomizations.CACerts) > 0 {
-		for _, cc := range p.OSCustomizations.CACerts {
+	allCACerts := append([]string{}, p.OSCustomizations.CACerts...)
+	allCACerts = append(allCACerts, fbCerts...)
+	if len(allCACerts) > 0 {
+		for _, cc := range allCACerts {
```
avitova, line 1045: I believe the condition is not needed, but no t a blocker

avitova (dismissed):
nice!

pkg/manifest/os.go
lzap: Note there is a line below the for block `pipeline.AddStage(osbuild.NewUpdateCATrustStage())` that must be in such if statement anyways, looks a little bit more consistent to have both stages being added in the if block.

avitova: oh, okay :+1: did not notice that

pkg/customizations/firstboot/firstboot.go
```diff
@@ -0,0 +1,179 @@
...
+// FirstbootOptionsFromBP converts a blueprint FirstbootCustomization to
+// FirstbootOptions. Validation is done in the blueprint package, so this function
+// assumes the input is valid, however, JSON unmarshalling errors are possible.
+func FirstbootOptionsFromBP(bpFirstboot blueprint.FirstbootCustomization) (*FirstbootOptions, error) {
+	fo := &FirstbootOptions{}
+	var satDone, aapDone bool
+	var ci int
+	var alreadyUsed []string
+
+	nameFunc := func(inputName, prefix string) string {
```
brlane-rht: I don't like the use of anonymous functions, it feels too much like javascript and makes it impossible to test the function to make sure it behaves as expected.
You're using it because you want ci and alreadyUsed to be available, but I think it would be cleaner and easier to test if you put those common values into a struct with a name generator function method.

brlane-rht (changes requested):
It would be a bit cleaner if you split out the new code into separate commits first, one for the new firstboot customization, one for the new osbuild stage, then hook everything up to the rest of the code. I also agree with the change to OSCustomizations, it should be squashed into the other commit -- the expectation is that OSCustomization has already been setup when the manifest serialize gets called.

This should also have a reference to a Jira ticket if there is one.

lzap:
Note for myself (I am busy atm):

* [x] Get rid of anon funcs
* [x] Add Before/After absolute ordering: https://github.com/osbuild/blueprint/pull/52
* [x] Add relative ordering (this was missing and is a bug)
* [x] Update blueprint in separate commit
* [x] Separate commit: new customization
* [x] Separate commit: new stage
* [x] Separate commit: hook it up
* [x] Ensure change to OSCustomizations is squashed into the other commit
* [x] Jira link

lzap:
Finally rebased, tested. It is now possible to specify ordering via before/after. By default, all scripts (aap/satellite/custom) run in parallel. I hope I fixed all comments it was a lot! Review carefully.

lzap:
I know this is a long PR and things were rewritten quite a bit, but I would appreciate re-reviews: @thozza @avitova @brlane-rht thanks!

pkg/customizations/firstboot/firstboot.go
```diff
@@ -0,0 +1,129 @@
...
+
+var reservedRegexp = regexp.MustCompile(`^(custom|satellite|aap)-\d+$`)
+
+// scriptNameGenerator assigns stable executable and unit names for firstboot scripts.
+type scriptNameGenerator struct {
+	counter     int
+	alreadyUsed []string
+}
+
+func (g *scriptNameGenerator) generate(inputName, prefix string) string {
```
brlane-rht: This feels overly complex, but maybe that's just me. But at the least could you add a docstring that explains how it is expected to act?

pkg/customizations/firstboot/name_generator_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestScriptNameGenerator(t *testing.T) {
+	var ng scriptNameGenerator
+
+	assert.Equal(t, "osbuild-first-my-script", ng.generate("my-script", "custom"))
```
brlane-rht: I think this could use some comments, eg. at first I thought it was wrong because it used `custom-2` instead of `custom-42`.

pkg/osbuild/firstboot.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/osbuild/images/pkg/customizations/firstboot"
+	"github.com/osbuild/images/pkg/customizations/fsnode"
+)
+
+var firstbootBaseAfter = []string{"network-online.target", "osbuild-first-boot.service"}
+
+func firstbootMarkerPath(unitFilename string) string {
+	// e.g. osbuild-first-setup.service -> /var/local/.osbuild-first-setup
+	base := strings.TrimSuffix(unitFilename, ".service")
+	return fmt.Sprintf("/var/local/.%s", base)
```
brlane-rht: Does this have to be in /var/local? Seems to me it would be nicer if it was in, say, /var/local/firstboot instead of cluttering up a directory used by other things.

pkg/osbuild/firstboot.go
```diff
@@ -0,0 +1,107 @@
...
+		unitFilename := script.Filename + ".service"
+		markerPath := firstbootMarkerPath(unitFilename)
+
+		f, err := fsnode.NewFile(markerPath, common.ToPtr(fs.FileMode(0770)), "root", "root", []byte{})
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("error creating firstboot marker node: %w", err)
+		}
+		files = append(files, f)
+
+		exec := fmt.Sprintf("/usr/local/bin/%s", script.Filename)
```
brlane-rht: Should we be using /usr/local/ for this? These are written at image creation time so I think they need to be someplace like /usr/libexec/firstboot maybe?

data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -1077,6 +1077,7 @@
       - "customizations.disk"
       - "customizations.files"
       - "customizations.filesystem"
+      - "customizations.firstboot"
```
brlane-rht, line 1080: It might be nice to list all the distro:imagetypes in the commit message since the git diff doesn't really show which ones it is allowed in.

brlane-rht:
Thanks for organizing things better. I'm not quite ready to ack it because of the /var/local and /usr/local usage, I don't think we should be touching things under those paths.

pkg/customizations/firstboot/firstboot.go
lzap: I think you did put a light on a smelly part - I have been refactoring this so many times so it grew to this monstrosity. I am going to vastly simplify this - just accept a name optionally, if its not provided do "custom-%d" and error out if the name is "aap" or "satellite" (reserved for the registration first boot scripts).

pkg/customizations/firstboot/name_generator_test.go
lzap: I vastly simplified this, "aap" or "satellite" or anything starting with "custom" is simply not allowed. Updated the tests. I actually got rid of the generator type, this is all done by the function I think it is reasonably short.

lzap:
I rebased and modified the patch set, thanks for review @bcl all your comments were incorporated. I appreciate the paths change, yeah, these are much better:

```
const (
	firstbootMarkerPath = "/var/local/osbuild-first-boot/.run"
	firstbootScriptDir  = "/usr/libexec/osbuild-first-boot"
)
```

FYI I am using this name because we already have a service named `osbuild-first-boot` so I am keeping the naming consistency.

Also simplified the name checker a bit, documented the function responsible for ensuring the naming consistency and conflicts, expanded unit tests.

Finally the allowed customization list was re-reviewed, what I am doing here is adding this customization to all image types which allows `customization.files` - I am assuming that this firstboot feature is nothing else than a fancy systemd unit file deployment. The commit message has a list of image types which were excluded.

Resolved all the import conflicts.

pkg/customizations/firstboot/firstboot.go
```diff
@@ -0,0 +1,142 @@
...
+	Certs         []string
+	After         []string
+	Before        []string
+}
+
+var ErrFirstbootAlreadySet = errors.New("firstboot customization already set")
+var ErrFirstbootNameTaken = errors.New("firstboot name already taken")
+
+func validateCustomName(name string) error {
+	if !filepath.IsLocal(name) {
```
achilleas-k: If I'm reading the `IsLocal()` docs correctly, this will allow `name` to be a path and not just a filename, right (e.g. `path/to/script.sh`).  Do we want that?

pkg/osbuild/firstboot.go
```diff
@@ -0,0 +1,111 @@
...
+		after := append([]string{}, firstbootBaseAfter...)
+		after = append(after, script.After...)
+		if prevUnit != "" {
+			after = append(after, prevUnit)
+		}
+		after = dedupeOrdered(after)
+
+		var execStartPre []string
+		if i == len(fbo.Scripts)-1 {
+			execStartPre = []string{"/usr/bin/rm " + firstbootMarkerPath}
```
achilleas-k, line 78: Any reason this is a `pre` command and not a `post`?  It seems strange that we have a marker that all services depend on but it's removed before running the last one.  It works, but the logic feels a bit off.

achilleas-k (changes requested):
LGTM generally.  Couple of inline comments, but I wont block on those (though I'm not sure about the `IsLocal()` one).  Requesting changes though because the last commit message includes lists of conflicting files, presumably from a rebase.

Sidenote: I was checking to see if we added this to the docs when we added the blueprint part (we didn't) and noticed that we have this page: https://osbuild.org/docs/user-guide/firstboot/
Which we should rewrite to use the new options now.

pkg/osbuild/firstboot.go
brlane-rht: If you use post, and the exec fails, it will run again on next boot which probably isn't something you want firstboot to do.

lzap: I tested this and honestly this is quite long PR I do not remember when/why I switched this from post to pre. But @brlane-rht has a point there, what we have [in our documentation](https://osbuild.org/docs/user-guide/firstboot/) is actually not right.

I created myself a Jira ticket to fix the docs tho: https://redhat.atlassian.net/browse/HMS-10883 and I fill fix it there as well.

lzap:
Rebased. It now uses basename to cut all the crap, it also removes all attempts to escape the path. Added some unit tests for that.

achilleas-k (dismissed):
Had a quick look at the changes since the last review and LGTM.

avitova (dismissed):
LGTM, let's get this in. You are a very patient person.

lzap:
Tests did fail because I moved (on request) the marker into `/var/local/osbuild-first-boot/.run` but the directory `/var/local/osbuild-first-boot` did not exist. I had to add an extra directory stage that creates the directory first. No other changes, please re-ack @avitova @achilleas-k 

Thanks @bcl for spotting the failure!

lzap:
Forgot to add the same for the new `/usr/libexec` dir, rebased.

brlane-rht (approved):
Looks good

supakeen (approved):

PR #1915, opened 2025-10-01
Custom AWS tags as slices instead of key=value strings

FrostyX:
See https://github.com/osbuild/images/pull/1903#issuecomment-3346755825

### DIFF
--- a/pkg/cloud/awscloud/awscloud.go
+++ b/pkg/cloud/awscloud/awscloud.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"strings"
 	"time"
 
 	"slices"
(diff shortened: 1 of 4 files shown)


thozza (dismissed):
Thanks, much appreciated :heart_eyes:

thozza (approved):

FrostyX:
No problem @thozza, it was a good suggestion.
I also updated my https://github.com/osbuild/image-builder-cli/pull/327 to use this new interface.

FrostyX:
Oh, lovely. Thank you for the review :-)

mvo5:
Heh, now I'm late to the party. I was wonding if instead of `[][2]string` we could do `struct KeyVal {Key, Val string}` which then would become `[]KeyVal`. I feel `[2]string` for a key/val is a bit too loose.

thozza:
> Heh, now I'm late to the party. I was wonding if instead of `[][2]string` we could do `struct KeyVal {Key, Val string}` which then would become `[]KeyVal`. I feel `[2]string` for a key/val is a bit too loose.

That would work for me. It felt like too much to define a new `struct` for this, but it is not the worst thing either...

PR #1918, opened 2025-10-01
fedora: Raise /boot to 2 GiB

Conan-Kudo:
With the increase in content in initramfs, a long-overdue raise of the default size for /boot is warranted.

Reference: https://pagure.io/fesco/issue/3482


supakeen:
Thanks @Conan-Kudo. You'll need to run `./tools/gen-manifest-checksums.sh` in your branch and amend it to your commit to make this pass CI or if you turn on letting me push to your branch I can do it for you.

In the FESco ticket it lists that this should probably only apply to x86 desktops but this PR changes it for everything?

supakeen:
@achilleas-k If you ack instead we can just merge directly when it comes out of draft :)

supakeen:
Ticket has been accepted by FESco, undrafting, merging.

supakeen:
> While out of scope for me, I would probably suggest this gets done for the RHEL definitions too as the conditions are exactly the same.

RHEL 10 ships without a separate `/boot` partition in disk images, but we can take a look for RHEL 8 and 9.

supakeen:
> As long as it's possible for a separate `/boot`, this probably needs to be accounted for.

#1930 :)

PR #1922, opened 2025-10-03
disk: drop static UUIDs in defs

supakeen:
We have long had a comment in our YAML definitions to drop the UUIDs and have them be autogenerated. This PR does so.


mvo5 (commented):
Removing this for fedora makes a lot of sense, I'm a bit hesitant for rhel, especially the older rhel releases.

supakeen:
> Removing this for fedora makes a lot of sense, I'm a bit hesitant for rhel, especially the older rhel releases.

What causes the hesitation?

bcl (approved):
Looks good!

PR #1923, opened 2025-10-03
allow boot mode selection through blueprint.DiskCustomization

martin-paulus:
When using the bootc-image-builder we found that when creating a bootc-based Fedora IoT image a 1 MiB bootable partition was inserted at the start of the disk image. For our Raspberry Pi-based project we desire a DOS MBR partition table where the first partition is large enough to hold the initial bootloader and the device tree overlays. Since this does not fit inside the 1 MiB partition, and the Raspberry Pi is unable to locate the boot loader and device tree overlays when they are located on the second partition of the disk, we designed a method for _optionally_ customizing the "boot mode" through the blueprint.

This pull request defines a way to customize the "boot mode" of a disk image through the blueprint. Another pull request is made to the blueprint repo[1] to parse the new option from the blueprint configuration.

[1]: https://github.com/osbuild/blueprint/pull/35

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -1304,6 +1304,12 @@ func NewCustomPartitionTable(customizations *blueprint.DiskCustomization, option
 	if customizations == nil {
 		customizations = &blueprint.DiskCustomization{}
 	}
+	if customizations.BootMode != "" {
+		bootMode, ok := platform.BootModeMap[customizations.BootMode]
+		if ok {
+			options.BootMode = bootMode
+		}
+	}
 
 	errPrefix := "error generating partition table:"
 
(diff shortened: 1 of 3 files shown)


mvo5:
[..]
> Personally I don't think we should then make the boot platform blueprint-customizable, but instead we should fix up bootc-image-builder to _not_ do this. Let's see what others think, cc @achilleas-k / @mvo5.

I agree here, it seems the only use-case to make this configurable right now is to fix the fact that we (unconditionally) add the (x86 specific) bios partition on aarch64 which is  a bug that Simons work in https://github.com/osbuild/images/pull/1927 fixes. 

So unless there is something I'm missing, I am in favor of merging the fix in #1927 and closing this PR (but please speak up if there is a different use-case that I missed or haven't considered).

martin-paulus:
> So unless there is something I'm missing, I am in favor of merging the fix in https://github.com/osbuild/images/pull/1927 and closing this PR (but please speak up if there is a different use-case that I missed or haven't considered).

Not to "speak up", but to confirm: my desired output is an  aarch64 image with a DOS partition table but without a BIOS boot partition. If you prefer to facilitate that otherwise, I will happily use your solution. My goal with the blueprint and images pull requests was merely to stop depending on local patches, and instead use the upstream builds.

PR #1926, opened 2025-10-06
Repos: Add definitions for AlmaLinux 9.7, 9.8, 10.1 and 10.2

eabdullin1:
Add AlmaLinux 9.7 and 9.8 repositories.
Add AlmaLinux 10.1 and 10.2 repositories.
Use vault.almalinux.org instead of repo.almalinux.org for AlmaLinux 9.5, 9.4 and 8.9.

For now, it’s expected that the repositories are not available.


achilleas-k (approved):
Thank you!

eabdullin1:
> I confirmed that the GPG keys match those available from https://repo.almalinux.org/almalinux/
> 
> But there are no 10.1, 10.2, 9.7 or 9.8 repos at those URLs.

It’s not available at the moment.  
Is it required to have these links working?  
If so, is it okay to use `mirrorlist` instead of `baseurl`? In that case, our mirror service will redirect to the latest stable version if the requested one is not available.

PR #1928, opened 2025-10-06
disk: make addPartitionsForBootMode() arch specific

mvo5:
When building a disk from a disk customization blueprint we will ensure that the right partitions for the boot mode of the image is added. This code currently checks the boot-mode of the image but for the aarch64/riscv64 architectures the added partitions are subtly wrong if the inputs (i.e. boot-mode) are wrong.

As a first step this commit makes addPartitionsForBootMode() look at the architecture and silently discard boot modes that do not make sense for the architecture. This is not ideal and we really should instead error if we encounter an invalid architecture+bootmode combination. But this will fix the issue outlined in
https://github.com/osbuild/images/pull/1923 where we create incorrect (x86) bios partitions on aarch64 bootc images (and ricv64 too).

Note that we may want to tweak/update some of the test-cases, we had explicit tests that we add a BIOS partition on aarch64 which are now dropped (and I suspect were never quite accurate(?))

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -1126,10 +1126,10 @@ func addPartitionsForBootMode(pt *PartitionTable, disk *blueprint.DiskCustomizat
 		return nil
 	}
 
-	// UEFI is not supported for PPC CPUs so we don't need to
-	// add an ESP partition there
 	switch architecture {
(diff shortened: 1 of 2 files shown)


supakeen (dismissed):

mvo5:
Sorry, had to force push because of conflict.

Issue #1920, opened 2025-10-02
Accept strings for `default_size`


supakeen:
We accept strings in many places for disk and partition sizes but `default_size` only accepts an int for bytes. Let's allow strings there too?

PR #1929, opened 2025-10-06
distro: add support for `default_size` to be a string

mvo5:
distro: add support for `default_size` to be a string

Make `default_size` a datasizes.Size type which means
it can be written as `size: 1234` or `size: 1 Kb` etc.

This also converts all existing image types to use
the simpler string format for the default_size.

Closes: https://github.com/osbuild/images/issues/1920

---

datasizes: add support for YAML unmarshal of datasizes

Add support to unmarshal YAML based datasizes just like we
support TOML and JSON. This is extremly simple and just
reuses the existing `common.UnmarshalYAMLviaJSON()`.

### DIFF
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -894,7 +894,7 @@ image_types:
     mime_type: "application/x-tar"
     environment: *kvm_env
     bootable: true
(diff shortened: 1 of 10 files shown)


supakeen (approved):
A lovely quality of life change. Thanks!

PR #1931, opened 2025-10-06
many: make `required_partition_sizes` take datasize.Size

mvo5:
[build on top of https://github.com/osbuild/images/pull/1929/commits]

This commit allows us to use nice strings in the
`required_partition_sizes` structs.

It might be a bit too much for the little gain but I'm opening it for feedback.

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -93,7 +93,7 @@ const DefaultBootPartitionSize = 1 * datasizes.GiB
 // containing the root filesystem is grown to fill any left over space on the
 // partition table. Logical Volumes are not grown to fill the space in the
 // Volume Group since they are trivial to grow on a live system.
-func NewPartitionTable(basePT *PartitionTable, mountpoints []blueprint.FilesystemCustomization, imageSize uint64, mode partition.PartitioningMode, architecture arch.Arch, requiredSizes map[string]uint64, defaultFs string, rng *rand.Rand) (*PartitionTable, error) {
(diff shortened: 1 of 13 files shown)


pkg/disk/partition_table.go
```diff
@@ -143,9 +143,9 @@ func NewPartitionTable(basePT *PartitionTable, mountpoints []blueprint.Filesyste
 
 	// If no separate requiredSizes are given then we use our defaults
 	if requiredSizes == nil {
-		requiredSizes = map[string]uint64{
-			"/":    1073741824,
-			"/usr": 2147483648,
+		requiredSizes = map[string]datasizes.Size{
+			"/":    datasizes.Size(1073741824),
+			"/usr": datasizes.Size(2147483648),
```
thozza, line 148: Nitpick: could we use the size constants here instead of the crazy number? :innocent:

thozza (commented):
I left out these sizes when submitting https://github.com/osbuild/images/pull/1579 due to reasons described there, but mainly because the YAMLification was still only partially done at that time, and I didn't want to introduce additional conflicts.

Given where we are now, I **insist** that we make this change and finish the transition to allow specifying any sizes as a string with units.

That said, this may be a good opportunity to consider unifying the implementation and replacing https://github.com/osbuild/images/blob/4a1679b4d4229450a1ebd88ff07f732c9d325a31/pkg/datasizes/parse.go#L59-L80 with `datasize.Size` everywhere, effectively eliminating `ParseSizeInJSONMapping()` .

pkg/disk/partition_table.go
mvo5: Sure!

thozza (approved):
LGTM

supakeen (approved):
Consistency is good :)

PR #1957, opened 2025-10-16
Delete `internal/worker`

thozza:
While working on something else, I noticed that we still keep the `internal/worker` in this repository, which is a leftover from the osbuild-composer split.

To enable the removal, refactor the `remotefile` package implementation and move it to the `pkg/` directory, since it has nothing to do with customizations. The `remotefile` resolver is not used anywhere, AFAIK, so I could also delete it, but I think we may actually use it for some functionality we discussed in the past (i.e., embedding custom files referenced by a remote URL in https://github.com/osbuild/images/blob/f21ffd7a58b2fd872f9d5788d1d62bfeafd2de8d/pkg/manifest/os.go#L1231).


thozza:
> Some error messages need to be updated in tests.

The issue is actually that the order in which resolver resolves the entries is not stable.

achilleas-k:
> > Some error messages need to be updated in tests.
> 
> The issue is actually that the order in which resolver resolves the entries is not stable.

Oh!  Problem.
Can we put the error messages in an array and test with `ElementsMatch()` against an array of expected ones?

lzap:
For the record, I am picking this PR up since I need this for https://github.com/osbuild/osbuild/pull/2326

PR #1958, opened 2025-10-17
Make bootupd metadata generation configurable per distro

pcdubs:
Add a configurable option for bootupd metadata generation that allows
distributions to enable it only where the bootupd package is available.

This introduces the bootupd_gen_metadata field in ImageConfig, which
controls whether bootupctl backend generate-update-metadata is run for
ostree commits. Distributions can set this conditionally based on
version to ensure bootupctl is only invoked when bootupd is present.


lzap (approved):

supakeen:
I'm wondering if this is OK to apply across the board like this. It's probably fine for Fedora; but does this work (for example) on RHEL 8 or do we need to make this into an image config option that's only applied where things are new enough?

supakeen (changes requested):
Looking at the CI this is indeed failing to work for RHEL-8, and RHEL-9:

```
subprocess.CalledProcessError: Command '['chroot', '/run/osbuild/tree', '/usr/bin/bootupctl', 'backend', 'generate-update-metadata', '/']' returned non-zero exit status 127.
```

It fails specifically because there's no `bootupctl` installed in RHEL-8, or RHEL-9 (is it available?).

```
chroot: failed to run command ‘/usr/bin/bootupctl’: No such file or directory
```

Depending on the outcome of the above there are two options:

1. Include `bootupd` on RHEL-8/RHEL-9 (if it exists and is of new enough version).
2. Make this an option in image config (or perhaps: on the distro) and only set it where it's supported (this would let us set it per-minor as well).

@pcdubs This would make the PR a bunch more involved; let me know if you want to handle it yourself or if you want me to force-push things into this PR for you.

say-paul:
> Ah, I see `rawhide` for Fedora now started failing in your PR due to no fault of your own:
> 
> ```
> Failed to enable unit: Unit greenboot-grub2-set-counter.service does not exist
> ```
> 
> If you want you can address it in this PR (a conditional on Fedora < 44) or I can PR it and you can rebase after?
> 
> Other than that this PR is good now :)

I have addressed this is in: https://github.com/osbuild/images/commit/59ac5d6977dabfd46db80fe81d95c33b154368e3
Seems to work for f43 but not rawhide.

supakeen:
> > Ah, I see `rawhide` for Fedora now started failing in your PR due to no fault of your own:
> > ```
> > Failed to enable unit: Unit greenboot-grub2-set-counter.service does not exist
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If you want you can address it in this PR (a conditional on Fedora < 44) or I can PR it and you can rebase after?
> > Other than that this PR is good now :)
> 
> I have addressed this is in: [59ac5d6](https://github.com/osbuild/images/commit/59ac5d6977dabfd46db80fe81d95c33b154368e3) ~Seems to work for f43 but not rawhide.~ I see for [F43](https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/11798741572) its still pulling in the old greenboot. I wonder why its doing so as new version is already out: https://packages.fedoraproject.org/pkgs/greenboot-rs/greenboot/

As for why it's not failing on Fedora 43:

It's pulling in the old version because at the date we made the snapshots that we use in our CI the old version was likely still in Fedora 43 (due to freeze, or such).

This is "expected" to start failing in Fedora 43 as well as soon as we update our snapshots and I have started a snapshot-update job.

---

From a quick scan I don't think `shallow_merge` is the right thing to use in this case or something finicky with the condition. I'll take a look at what can/should be used for this and PR it so we can rebase this PR on top of it.

Can you confirm if:

```
- "NetworkManager.service"
- "firewalld.service"
- "sshd.service"
- "greenboot-healthcheck"
```

Is the full set of services that should be enabled in Fedora >= 43?

PR #1960, opened 2025-10-20
fedora: /boot on btrfs for Fedora Cloud 44 (HMS-9737)

supakeen:
We enforce a partition for /boot even when there's already a subvolume with a mountpoint for /boot in the partition table. Let's change this requirement to only add a boot partition if there's a btrfs root but *no* boot.

Initially my focus is on supporting this in your YAML definitions hence some shortcuts are taken here; we assume that there must be a btrfs subvolume with a `/boot` mountpoint. This isn't a hard requirement (I could see using the `/` subvolume instead) but that will need a bit more finagling and discussion.

Also moves Fedora Cloud 44's `/boot` onto btrfs per [this](https://fedoraproject.org/wiki/Changes/BtrfsBootForCloud) change proposal.


Conan-Kudo:
I don't think there's a such thing as "nice code" for configuring bootloaders. At least not in my experience.

thozza (commented):
In general, this LGTM.

There is one corner case that could lead to a panic, so it would be good to address it.

PR #1961, opened 2025-10-20
gitlab: run ostree manifest generation and builds only when needed

achilleas-k:
With the way the gitlab-ci.yml was generated before, we've been running some useless jobs.  On CentOS 10, RHEL 10, and RHEL 7 distros, there are no ostree-based image types.  The 'generate-ostree-build-config' job would then run 'gen-manifests' for ostree-based image type names, produce no manifests, and end successfully with no builds queued for the next step.

This is wasteful.  Let's stop doing this.


thozza (approved):
😍

PR #1966, opened 2025-10-23
osbuild-exec: add buildlog and monitor support

croissanne:
This commit adds a new `BuildLog` option to the `OSBuildOptions` that
    can be used to generate a streamed buildlog (e.g. to a file or a
    websocket). This supports the `--with-buildlog` option in
    image-builder-cli.
    
With the buildlog option set, stderr is forwarded to stdout to make sure
    statements appear in order. If `BuildLogMu` is set, this mutex will be
    used for the synced writer which writes to both stdout and the
    buildlog. Callers can use the same mutex to write statements to the same
    buildlog.
    
The errorWriter option got removed, and callers can just use `Stdout`
    and `Stderr` to set the stdout and stderr of the command.

### DIFF
--- a/pkg/osbuild/osbuild-exec.go
+++ b/pkg/osbuild/osbuild-exec.go
@@ -43,8 +43,10 @@ type OSBuildOptions struct {
 	Stdout     io.Writer
 	Stderr     io.Writer
 
-	Monitor   MonitorType
-	MonitorFD uintptr
+	// If MonitorFD is set, a file (MonitorW) needs to be inherited by the osbuild process. The
+	// caller should make sure to close it afterwards.
+	Monitor     MonitorType
+	MonitorFile *os.File
 
 	JSONOutput bool
 
@@ -79,8 +81,10 @@ func NewOSBuildCmd(manifest []byte, optsPtr *OSBuildOptions) *exec.Cmd {
 	if opts.Monitor != "" {
 		cmd.Args = append(cmd.Args, fmt.Sprintf("--monitor=%s", opts.Monitor))
 	}
-	if opts.MonitorFD != 0 {
-		cmd.Args = append(cmd.Args, fmt.Sprintf("--monitor-fd=%d", opts.MonitorFD))
+
+	if opts.MonitorFile != nil {
+		cmd.Args = append(cmd.Args, "--monitor-fd=3")
+		cmd.ExtraFiles = []*os.File{opts.MonitorFile}
 	}
 
 	if opts.JSONOutput {
--- a/pkg/osbuild/osbuild-exec_test.go
+++ b/pkg/osbuild/osbuild-exec_test.go
@@ -51,6 +51,8 @@ func TestNewOSBuildCmdNilOptions(t *testing.T) {
 
 func TestNewOSBuildCmdFullOptions(t *testing.T) {
 	mf := []byte(`{"real": "manifest"}`)
+	_, wp, err := os.Pipe()
+	assert.NoError(t, err)
 	cmd := osbuild.NewOSBuildCmd(
 		mf,
 		&osbuild.OSBuildOptions{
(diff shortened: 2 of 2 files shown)


pkg/osbuild/osbuild-exec.go
```diff
@@ -27,9 +27,10 @@ const (
 var OSBuildCmd = "osbuild"
 
 type OSBuildOptions struct {
-	StoreDir  string
-	OutputDir string
-	ExtraEnv  []string
+	StoreDir    string
+	OutputDir   string
+	Checkpoints []string
+	ExtraEnv    []string
```
achilleas-k: Move exports here too?  I like the idea of all the osbuild args being bundled together and only having function args for things that aren't directly passed to the osbuild call (if we ever add any).

pkg/osbuild/osbuild-exec.go
```diff
@@ -179,3 +225,19 @@ func OSBuildInspect(manifest []byte) ([]byte, error) {
...
+func newSyncedWriter(mu *sync.Mutex, w io.Writer) io.Writer {
+	return &syncedWriter{mu: mu, w: w}
+}
+
+func (sw *syncedWriter) Write(p []byte) (n int, err error) {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+
+	return sw.w.Write(p)
+}
```
lzap: This writer decorator could be simplified to:

```go
type SyncedWriter struct {
	mu sync.Mutex
	w  io.Writer
}

func NewSyncedWriter(w io.Writer) *SyncedWriter {
	return &SyncedWriter{w: w}
}

func (sw *SyncedWriter) Write(p []byte) (n int, err error) {
	sw.mu.Lock()
	defer sw.mu.Unlock()

	return sw.w.Write(p)
}
```

I suggest to move this to common as this writer can be pretty useful elsewhere.

pkg/osbuild/osbuild-exec.go
```diff
@@ -24,6 +24,8 @@ const (
 	MonitorLog     = "LogMonitor"
 )
 
+var OSBuildCmd = "osbuild"
```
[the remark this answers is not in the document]
croissanne: Yea, the problem is that ibcli also uses it for testing, but maybe we should move all tests from there into here somehow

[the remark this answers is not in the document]
croissanne: let's merge https://github.com/osbuild/images/pull/1969 first, rebase and then use path

pkg/osbuild/osbuild-exec_test.go
```diff
@@ -1,19 +1,24 @@
...
+	"time"
 
 	"github.com/stretchr/testify/assert"
 
 	"github.com/osbuild/images/pkg/datasizes"
 	"github.com/osbuild/images/pkg/osbuild"
 )
 
-func makeFakeOsbuild(t *testing.T, content string) string {
+func makeFakeOSBuild(t *testing.T, content string) string {
```
mvo5: :heart:

pkg/osbuild/osbuild-exec_test.go
```diff
@@ -98,22 +103,81 @@ func TestNewOSBuildCmdFullOptions(t *testing.T) {
...
 		JSONOutput: true,
 	}
-	result, err := osbuild.RunOSBuild([]byte(`{"fake":"manifest"}`), nil, opts)
+	result, err := osbuild.RunOSBuild([]byte(`{"fake":"manifest"}`), opts)
 	assert.NoError(t, err)
 	assert.NotNil(t, result)
 	assert.True(t, result.Success)
 }
+
+func TestRunOSBuildBuildLog(t *testing.T) {
```
mvo5: :heart:

pkg/osbuild/osbuild-exec_test.go
```diff
@@ -98,22 +103,81 @@ func TestNewOSBuildCmdFullOptions(t *testing.T) {
...
+	assert.Equal(t, `osbuild-stdout-output
+osbuild-stderr-output
+`, buildLog.String())
+	assert.Equal(t, `osbuild-stdout-output
+osbuild-stderr-output
+`, stdout.String())
+	assert.Empty(t, stderr.Bytes())
+}
+
+func TestSyncWriter(t *testing.T) {
```
mvo5: :heart:

mvo5 (approved):
Thank you!

supakeen (approved):

PR #1969, opened 2025-10-27
osbuild: tweak how we can mock osbuild slightly

mvo5:
Tiny commit to tweak the way we allow to change the osbuild binary to run. This is a followup for PR#1965. The main motivation is to unexport `osbuild.OSBuildCmd` as it is global state and potentially racy, i.e. two threads doing:
```go
osbuild.OSBuildCmd = "foo"
osbuild.RunOSBuild()
```
and
```go
osbuild.OSBuildCmd = "bar"
osbuild.RunOSBuild()
```
would potentially race.

I know the risk is small (because this is mean only for testing and tests are run in sequence in the same pkg by default) but if it is ever used outside of tests it can becomes a source of subtle and hard to spot bugs.

The use-case to tweak osbuild.OSBuildCmd to test osbuild.RunOSBuild() from the outside one can just modify PATH to point to a different osbuild binary (we do something similar in e.g. ibcli, i.e. just modifying PATH).

### DIFF
--- a/pkg/osbuild/export_test.go
+++ b/pkg/osbuild/export_test.go
@@ -3,3 +3,11 @@ package osbuild
 type (
 	StatusJSON = statusJSON
 )
+
+func MockOsbuildCmd(s string) (restore func()) {
+	saved := osbuildCmd
+	osbuildCmd = s
+	return func() {
+		osbuildCmd = saved
+	}
+}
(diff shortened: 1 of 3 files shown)


achilleas-k:
Our AWS-based CI runners are not working today.  This PR is too small and don't affect those builds.  Admin merging.

PR #1970, opened 2025-10-27
deps: switch yaml libraries

supakeen:
During a recent chat it was brought up that `go.pkg.in/yaml.v3` is unmaintained. The YAML organization forked the unmaintained package and is now updating it. The `v3` we use here only receives security fixes while `v4` will eventually support new features and more of YAML.

Let's do the easy an uncontroversial swap first as this is a maintained drop-in replacement.

---

Note, this is *not* `goccy/go-yaml` which has additional features it is `yaml/go-yaml`.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -19,7 +19,7 @@ import (
 	"time"
 
 	"github.com/gobwas/glob"
-	"gopkg.in/yaml.v3"
+	"go.yaml.in/yaml/v3"
 
 	"github.com/osbuild/blueprint/pkg/blueprint"
 	"github.com/osbuild/images/internal/buildconfig"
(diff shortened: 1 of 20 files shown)


supakeen:
> Nice. Just for the record:
> 
>     * cli: direct dependency
> 
>     * crc: indirect
> 
>     * composer: indirect
> 
>     * blueprint-schema: indirect
> 
>     * blueprint: indirect
> 
> 
> Are you tracking this work in an issue @supakeen or was this just a random thing?

I created PRs for the ones that have direct deps on them, so https://github.com/osbuild/image-builder-cli/pull/354 and https://github.com/osbuild/bootc-image-builder/pull/1084 which means that afterwards *we* probably don't have any projects that cause indirect deps on it; though others of our 3rd party dependencies might still rely on it.

PR #1980, opened 2025-10-30
reporegistry: various cleanups

mvo5:
This started as a quick drive-by to tweak things I noticed while working
on the tests for https://github.com/osbuild/image-builder-cli/pull/360

It grew into more when I dove into the API of reporegitry and how we use.

I would like to do some followups (that will require change in composer) after this one:
1. AFAICT all API usage of rr.ReposByArchName() sets includeTagged=False so I would like to drop that argument
2. rr.DistroHasRepos() should either become a pure bool/error return not the full repos (only weldr uses it currently)or we just use ReposByArchName()
3. Renaming ReposByArchName to something like ReposByDistroArchExcludeTagged or something descriptive

All of the above of course is more controversial so happy to discuss

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -388,23 +388,6 @@ func save(ms manifest.OSBuildManifest, depsolved map[string]depsolvednf.Depsolve
 	return nil
 }
 
-func filterRepos(repos []rpmmd.RepoConfig, typeName string) []rpmmd.RepoConfig {
-	filtered := make([]rpmmd.RepoConfig, 0)
-	for _, repo := range repos {
-		if len(repo.ImageTypeTags) == 0 {
-			filtered = append(filtered, repo)
-		} else {
-			for _, tt := range repo.ImageTypeTags {
-				if tt == typeName {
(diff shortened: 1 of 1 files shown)


pkg/reporegistry/reporegistry.go
```diff
@@ -118,6 +120,12 @@ func (r *RepoRegistry) DistroHasRepos(distro, arch string) ([]rpmmd.RepoConfig,
 	return repos, nil
 }
 
+// (deprecated) DistroHasRepos is just here for compatbility with weldr which could
+// use `ReposByArchName(distro, arch, true)` instead.
+func (r *RepoRegistry) DistroHasRepos(distro, arch string) ([]rpmmd.RepoConfig, error) {
+	return r.reposByDistroArch(distro, arch)
+}
+
```
thozza: So why keep it and not fix the Weldr API to use `ReposByArchName()`?

mvo5: Excellent point, I should have linked tohttps://github.com/osbuild/osbuild-composer/pull/4910 - sorry for that. Once that is in I can drop this compat wrapper.

PR #1985, opened 2025-11-03
depsolvednf: add expected packages json file (HMS-9334)

lzap:
Extract test data into a separate file. Then use Go embed to load the file.

---

I tried to find more of these, the biggest structures we have are in `pkg/osbuild/common_test.go` but in these cases JSON actually looks worse than Go structs so I dropped the idea.


supakeen (approved):

lzap:
Might be surprising saying that as the author, but I am with @thozza with this one. I created the ticket because it was recommended by I think @achilleas-k or @mvo5 in other PR and it sounded like a great idea because in the PR we had a good use case (it was JSON from osbuild monitor output).

But here, the Go syntax is slightly better. And for the partitioning tables it is miles better I had a commit locally but I had to drop it it was so bad. So I will take this opportunity as the author and close this one. The last thing I want to do is to add extra work for others when editing this part of our codebase.

Damn, and I was this close filing a PR in `images`, getting reviews, passing tests without a single rebase! Well, I will have to wait for my next chance. :-) Thanks.

PR #1987, opened 2025-11-04
many: tar numeric owner (RHEL-102854)

supakeen:
```
osbuild/tar: allow numeric owner
    
In a recent osbuild PR [1] an option was added to allow using numeric
owners in tar archives. This is wanted because including user and group
names in metadata can lead to unwanted side-effects when unpacking on a
system that contains user/group names.
    
Allow the tar stage to set this option in `images` as well.
```

```
image/archive: set numeric owner by default
   
It is always a good idea to use numeric owner for the tar archive image
type. Let's do that.
```


pkg/image/archive.go
```diff
@@ -39,6 +40,7 @@ func (img *Archive) InstantiateManifest(m *manifest.Manifest,
 	osPipeline.OSVersion = img.OSVersion
 
 	tarPipeline := manifest.NewTar(buildPipeline, osPipeline, "archive")
+	tarPipeline.NumericOwner = common.ToPtr(true)
```
lzap: This is totally not Go, this should have been negative non-pointer. Something like: `SkipNumericOwner bool`

lzap (approved):
Small rant, looks good.

Edit: My second comment did not store, so putting it here:

_Having said that, I knew you are just following the suite. So I have no issues with that._

supakeen:
@lzap Thanks for the review; I apparently forgot to attach changed checksums so I've amended my last commit. Sadly this requires a re-review.

lzap (approved):

PR #1988, opened 2025-11-04
[RFC] bootc: add/switch to a non-running container by default

mvo5:
[draft as this needs discussion and tests and maybe a PR to bootc]

This commit switches our bootc introspection to not actually run the container but only mount it. We need this so that we can create osbuild manifests for bootc containers without having to worry about attacks. A container can potentially contain any code so its
risky to run it in our service.

We have the same problem when building (rpm) images where the maintainer scripts can also run arbitrary code and we run those in a secure instance. This is okay for image building but quite expensive for manifest generation so lets try to avoid it.

The one wringle here is that we can no longer run
```console
$ bootc install print-configuration
```
from the container. So this commit re-implements the configuration reading from bootc. This is obviously bad and we need to find a better solution.

One options is
```console
$ bootc install print-configuration --rootdir /path/to/container
```
so that we can run a trusted bootc to introspect the contaner config. The downside of this of course is that if our bootc is too old to understand new features from the container (this is the beauty of running it from the container).

Another option might be to embrace this and accept that we run arbitrary code and just trust that the podman isolation will protect us. We will still need to harden our implementation and ensure we run our podman container as a non-root user (not sure this happens in the service currently).

### DIFF
--- a/pkg/bib/container/nxcontainer.go
+++ b/pkg/bib/container/nxcontainer.go
@@ -1,0 +1,145 @@
+package container
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"slices"
+	"sort"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+)
+
+// Never add anything to this code that actually starts the the container
+// or execs code from it
+
+// NxContainer is a container non-executing container that runs *no* code
+// from the container and only creates/mounts it.
+type NxContainer struct {
+	id   string
+	root string
+	arch string
+}
+
+func NewNxContainer(ref string) (*NxContainer, error) {
+	output, err := exec.Command("podman", "create", ref).Output()
+	if err != nil {
+		if e, ok := err.(*exec.ExitError); ok {
+			return nil, fmt.Errorf("creating %s container failed: %w\nstderr:\n%s", ref, e, e.Stderr)
+		}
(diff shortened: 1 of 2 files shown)


achilleas-k:
Another option would be to standardise on the configuration paths and formats, make a spec that we can share between the two projects, so we can just read them from the containers as files instead of relying on a binary.  But I see how that can be harder to maintain in the long run.

pkg/bib/container/nxcontainer.go
```diff
@@ -0,0 +1,145 @@
...
+	if output, err := exec.Command("podman", "umount", c.id).CombinedOutput(); err != nil {
+		return fmt.Errorf("umount %s nxcontainer failed: %w\noutput:\n%s", c.id, err, output)
+	}
+
+	/* #nosec G204 */
+	if output, err := exec.Command("podman", "rm", c.id).CombinedOutput(); err != nil {
+		return fmt.Errorf("rm %s nxcontainer failed: %w\noutput:\n%s", c.id, err, output)
+	}
+	return nil
+}
```
lzap: I do not have local branch with ctags for this one so I cannot see myself but I am hoping this is called via `defer` to ensure it never gets skipped. We do not want to run out of resources on the server.

achilleas-k: It is deferred already yes.

achilleas-k (approved):
LGTM.  Typo comments below.

PR #1989, opened 2025-11-05
test: qemu based testing via bibs vm.py (local and in CI) for qcow2  [HMS-9572]

mvo5:
test: make runner image-type dependant

With the new qemu based testing of local images we want to make
sure that we pick gilab runners that support nested KVM. So this
commit adds a new `get_ci_runner_for(image_type)` helper that
allows to select specific runner for specific image types in
the Schutzfile. With that we can pick `rhos-01` which
supports kvm.

---

test: add support for local qemu testing via bibs vm.py

Initial support for running tests locally using boot-image
via the `vm.py` runner from bib for cloud-init supporting
images.

The root ssh key is added via cloud-init as this is what
our users are expecting (thanks to Ondrej for suggesting
this) and this way we do not need to modify the image itself.
This works for most images, for the others we need to discuss
(see below). This means we need "cloud-utils" installed.

We could also use test.thing but adding the polyfills is
actually not trivial and we need to modify the image. We could
also modify the image via guestfs and just add the root ssh
key directly, see mutate_qemu_image_for_login in [0]
for an example but the dowside of this is that we
modify the image (and its fiddly, we need to ensure
selinux is set correctly for anything we write via guestfs).

Currently only "qcow" is supported, cross arch boot is supported
for aarch64 (via the vmtest.vm.QEMU but it would be trivial
to extend this to more arches).

Note that because the QEMU test runner already supports cdrom
we can trivial test our installer with it too.

[0] https://github.com/osbuild/images/compare/main...mvo5:non-test-thing?expand=1

---

Schutzfile: increase RNG to force (re)tests

Ensure that we (re) test all images with the new qemu based image
boot tests. The easiest way to archive this it to just increase
the rngseed number.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,13 @@
 [tool.autopep8]
 max_line_length = 120
+
+[project]
+name = "images"
+version = "0.1.0"
+
+dependencies = [
+    "vmtest @ git+https://github.com/osbuild/bootc-image-builder.git",
+]
+
+[tool.setuptools]
+packages = []
--- a/test/scripts/boot-image
+++ b/test/scripts/boot-image
@@ -3,11 +3,16 @@ import argparse
 import contextlib
 import json
 import os
+import pathlib
 import uuid
-from tempfile import TemporaryDirectory
+import subprocess
+import shlex
 import shutil
+import textwrap
+from tempfile import TemporaryDirectory
 
 import imgtestlib as testlib
+from vmtest.vm import QEMU
 
 BASE_TEST_SCRIPT = "test/scripts/check-host-config.sh"
 WSL_TEST_SCRIPT = "test/scripts/wsl-entrypoint.bat"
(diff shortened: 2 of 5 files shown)


Schutzfile
```diff
@@ -4,7 +4,13 @@
     "bootc-image-builder": {
       "ref": "quay.io/centos-bootc/bootc-image-builder@sha256:9893e7209e5f449b86ababfd2ee02a58cca2e5990f77b06c3539227531fc8120"
     },
-    "gitlab-ci-runner": "aws/fedora-42"
+    "gitlab-ci-runner": "aws/fedora-42",
+    "gitlab-ci-runner-for": {
+      "x86_64": {
+        "qcow2": "rhos-01/fedora-42",
```
kingsleyzissou: ❤️

test/scripts/install-dependencies
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 
+# we need qemu-kvm here but libvirtt-devel pulls this in
```
lzap:
```suggestion
# we need qemu-kvm here but libvirt-devel pulls this in
```

test/scripts/boot-image
```diff
@@ -90,6 +95,106 @@ def ensure_uncompressed(filepath):
...
+    if not can_run:
+        print(f"WARNING: skipping {image_path} with {config_file}: {skip_reason}")
+        return
+    cmd = [BASE_TEST_SCRIPT, config_file]
+    with ensure_uncompressed(image_path) as raw_image_path:
+        with create_ssh_key(key_type="rsa") as (privkey_path, pubkey_path):
+            ssh_key = pathlib.Path(pubkey_path).read_text(encoding="utf8").strip()
+            # put cmd into iso as well via cloud-init
+            with make_cloud_init_iso(ssh_key) as cloud_init_iso:
+                with QEMU(raw_image_path, arch=arch, cdrom=cloud_init_iso) as vm:
```
supakeen: Let's save some indentation: `with A() as a, B() as b:`.

test/scripts/boot-image
```diff
@@ -90,6 +95,106 @@ def ensure_uncompressed(filepath):
...
+        meta_data.write_text('{"instance-id": "i-1234567890abcdef0"}')
+        iso_path = pathlib.Path(tmpdir) / "cloud-init.iso"
+        subprocess.check_call(
+            ["cloud-localds", os.fspath(iso_path), user_data.name, meta_data.name],
+            cwd=tmpdir,
+        )
+        yield iso_path
+
+
+def can_run_qemu_test(arch, image_path, config_file):
```
supakeen: Do we also need/want to test if `cloud-init` is actually in the image, since it's used to write keys and such?

mvo5:
> I am probably coming late or missed the conversation, but why exactly are we using `paramiko` vs simply shelling out to `ssh` / `scp` clients?

There is no deep reason, we should probably just change it, we need to work in this area anyway as there is a bug in paramiko with the privkey handling and it is probably simpler to just switch.

pyproject.toml
```diff
@@ -1,2 +1,13 @@
 [tool.autopep8]
 max_line_length = 120
+
+[project]
+name = "images"
+version = "0.1.0"
+
+dependencies = [
+    "vmtest @ git+https://github.com/osbuild/bootc-image-builder.git",
```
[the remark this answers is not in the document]
mvo5: Simon also raised concerns here, see https://github.com/osbuild/images/pull/1989#discussion_r2497594341 - I have no strong opinion here, but moving it into images *might* be an easiest?

PR #1992, opened 2025-11-06
data: update riscv repo URL for f42

mvo5:
The repository URL for f42 is not/no longer available on fedora.riscv.rocks and we need to switch to
riscv-koji.fedoraproject.org.

(we will need a similar update to f43 but that will also require adding the new keys as there is no rsicv64 in our repos for f43 yet)


lzap (approved):
URLs alive, all good.

PR #1995, opened 2025-11-06
test: add support to check for server-qcow2 (no cloud-init)

mvo5:
test: add support to check for server-qcow2 (no cloud-init)

This commit adds support for non-cloud init image types like
the `server-qcow2`. To test those we need to mutate the
image so that we can inject our test user and credetials
(note that in the future with test.thing and systemd
credentials passing this will be no longer necessary).

We also need a customization that contains the "osbuild"
user.

It can be run via:
```
./test/scripts/build-image fedora-43 server-qcow2 ./test/configs/all-customizations
./test/scripts/boot-image ./build/fedora_42-x86_64-server_qcow2
```

---

test: tweak all-customizations config to include a osbuild  user

Similar to the unattended-iso.json configs lets make sure we have
the `osbuild` user in all-customizations. This ensures that we
can inject credentials when needed.

This commit also changes the kernel commandline in all customizations
from debug to systemd.journal.forward_to_console to make debugging
boots easier.

### DIFF
--- a/test/scripts/boot-image
+++ b/test/scripts/boot-image
@@ -178,6 +178,10 @@ def ensure_can_run_qemu_test(arch, image_path, config_file):
                 raise CannotRunQemuTest("centos-9 will not start an install and waits on source selection")
             case name if name.startswith("rhel-9"):
                 raise CannotRunQemuTest("rhel-9 will not start an install and waits on source selection")
+    # server-qcow2 has no cloud-init so we need at least a osbuild user
+    if info.get("image-type") == "server-qcow2":
+        if not any(u["name"] == "osbuild" for u in customizations.get("user", [])):
+            raise CannotRunQemuTest("server-qcow2 needs a customization with the osbuild user")
 
 
 def qemu_cmd_scp_and_run(vm, cmd, privkey_path):
@@ -199,6 +203,10 @@ def boot_qemu(arch, image_path, config_file, keep_booted=False):
         uncompressed_image_path = cm.enter_context(ensure_uncompressed(image_path))
         (privkey_path, pubkey_path) = cm.enter_context(create_ssh_key())
         cloud_init_iso = cm.enter_context(make_cloud_init_iso(pubkey_path))
+        # ugly, find a better/declarative way for non-cloud init images
+        if "server_qcow2" in image_path:
+            print("WARNING: modifiyng the image to include osbuild ssh login")
+            mutate_qemu_image_for_login(image_path, pubkey_path)
         with QEMU(uncompressed_image_path, arch=arch, cdrom=cloud_init_iso) as vm:
             try:
                 qemu_cmd_scp_and_run(vm, cmd, privkey_path)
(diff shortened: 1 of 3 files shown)


test/configs/all-customizations.json
```diff
@@ -26,12 +26,15 @@
...
       },
       "user": [
         {
-          "name": "user1",
-          "key": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC61wMCjOSHwbVb4VfVyl5sn497qW4PsdQ7Ty7aD6wDNZ/QjjULkDV/yW5WjDlDQ7UqFH0Sr7vywjqDizUAqK7zM5FsUKsUXWHWwg/ehKg8j9xKcMv11AkFoUoujtfAujnKODkk58XSA9whPr7qcw3vPrmog680pnMSzf9LC7J6kXfs6lkoKfBh9VnlxusCrw2yg0qI1fHAZBLPx7mW6+me71QZsS6sVz8v8KXyrXsKTdnF50FjzHcK9HXDBtSJS5wA3fkcRYymJe0o6WMWNdgSRVpoSiWaHHmFgdMUJaYoCfhXzyl7LtNb3Q+Sveg+tJK7JaRXBLMUllOlJ6ll5Hod root@localhost"
+          "name": "osbuild",
+          "key": "ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBNCP83Bls0eAZnMzIOTDtRFnbVZVbebfshwpilYUQ8kwLNHCiKFcdu9KjDPe0BacACP0Mna83puuH/H92BF3jlc= github.com/osbuild/images",
+          "groups": [
+            "wheel"
+          ]
```
lzap: I currently cannot find this key shared via team BitWarden, can you add it?

test/scripts/boot-image
```diff
@@ -199,6 +203,10 @@ def boot_qemu(arch, image_path, config_file, keep_booted=False):
         uncompressed_image_path = cm.enter_context(ensure_uncompressed(image_path))
         (privkey_path, pubkey_path) = cm.enter_context(create_ssh_key())
         cloud_init_iso = cm.enter_context(make_cloud_init_iso(pubkey_path))
+        # XXX: fugly, find a better/declarative way for non-cloud init images
```
lzap:
```suggestion
        # XXX: ugly, find a better/declarative way for non-cloud init images
```

lzap (dismissed):

test/configs/all-customizations.json
mvo5: Excellent point, I revert this particular ssh key change.

test/scripts/boot-image
```diff
@@ -272,6 +280,55 @@ def boot_qemu_iso_no_unattended_support(arch, installer_iso_path, config_file):
...
+def mutate_qemu_image_for_login(filepath, pubkey_path):
+    """
+    Mutate the qemu image using libguestfs which uses qemu to boot a microVM
+    to allow root-login via the given pubkey
+    """
+    pubkey = pathlib.Path(pubkey_path).read_text(encoding="utf8")
+    # pylint: disable=import-outside-toplevel
+    import guestfs
+    g = guestfs.GuestFS()
+    try:
```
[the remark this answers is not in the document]
mvo5: Excellent idea, done.

test/scripts/boot-image
```diff
@@ -272,6 +280,63 @@ def boot_qemu_iso_no_unattended_support(arch, installer_iso_path, config_file):
...
+        # the customization has an osbuild user that can do sudo
+        uid = 1000
+        ssh_dir = "/home/osbuild/.ssh"
+        g.mkdir_p(ssh_dir)
+        g.chmod(0o700, ssh_dir)
+        g.chown(uid, uid, ssh_dir)
+        auth_keys_file = f"{ssh_dir}/authorized_keys"
+        g.write_append(auth_keys_file, pubkey.encode('utf-8'))
+        g.chmod(0o600, auth_keys_file)
+        g.chown(uid, uid, auth_keys_file)
```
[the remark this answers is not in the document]
mvo5: I updated the text to capture that while the customizations add an osbuild user we may not have access to the ssh key. We could of course force the use of a ssh key in the customizations that is available via the `CI_PRIV_SSH_KEY` env. The downside would be that running it locally requires more setup, this is why I have a slight preference for this but of course the upside is that this way we don't need to modify the image which is a very nice bonus.

If we want to go this way (which is fine with me) I will close this PR and we can add the server-qcow2 in a simpler way (i.e. by just ensuring that we have an osbuild user that can can do passwordless sudo. We will also have to ensure that passwordless sudo is possible, I don't think we support that outside of kickstarts right now but maybe I just missed the option :) If its missing it seems like a very nice feature to add.

As for why not just "root" - that was my initial idea, we can still just do it this way, the downsides:
a) less uniform, right now we always run "check-base-host" via the osbuild user so we would have to add conditions on vm.run() what user to pick and the behavior of the host check would be (slightly) different. Not a big deal of course
b) some systems do not allow root-login at all and having to debug when this is the case can be a slight headache (i.e. remembering that the failure might be because of that). by going via osbuild this is not a concern.
But of course happy to adjust as needed.

PR #1996, opened 2025-11-06
centos: `container` and `container-minimal` types (HMS-9661)

supakeen:
Additional image types for CentOS 9 and 10 that produce OCI archives. The definitions come directly from HMS-9661 [1] though I did a bit of deduplication.

Initially I'm only enabling these new types for non-RHEL (CentOS and Alma Linux) though we might want them in RHEL as well?

[1]: https://issues.redhat.com/browse/HMS-9661


data/distrodefs/rhel-10/imagetypes.yaml
```diff
@@ -821,6 +821,28 @@
...
+      - "distro"
+      - "packages"
+      - "modules"
+      - "groups"
+      - "enabled_modules"
+      - "minimal"
+      - "containers"
+      - "customizations.directories"
+      - "customizations.files"
+      - "customizations.firewall"
```
mvo5: Sorry, this is a bit unrelated and fine of course but I wonder if people configure firewalls in their containers. No need to change anything, I just stumbled upon it and was wondering :)

data/distrodefs/rhel-10/imagetypes.yaml
```diff
@@ -2373,3 +2394,87 @@ image_types:
...
             - xrdb
             - xz
+
+  container: &container
+    filename: "container.tar"
+    mime_type: "application/x-tar"
+    image_func: "container"
+    bootable: false
+    exports: ["container"]
+    required_partition_sizes: *default_required_dir_sizes
```
mvo5: (nitpick) AFAICT we don't need this as we have no partition table for the container(s)

PR #1998, opened 2025-11-06
many: lock root password for os pipeline

supakeen:
```    
    many: lock root password for os pipeline
    
    Some Fedora image types explicitly lock the root password. Let's plumb
    the same functionality in the OS pipeline as we have for OSTree
    deployments.
```

```
    many: sync up os and ostree-deployment
    
    Name properties that do the same the same and use the same helper in
    both places.
```

### DIFF
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -483,7 +483,7 @@ func ostreeDeploymentCustomizations(
 	}
 
 	if imageConfig.LockRootUser != nil {
-		deploymentConf.LockRoot = *imageConfig.LockRootUser
+		deploymentConf.LockRootUser = *imageConfig.LockRootUser
 	}
 
 	for _, fs := range c.GetFilesystems() {
--- a/pkg/manifest/ostree_deployment.go
+++ b/pkg/manifest/ostree_deployment.go
@@ -42,7 +42,7 @@ type OSTreeDeploymentCustomizations struct {
 
 	// Lock the root account in the deployment unless the user defined root
 	// user options in the build configuration.
-	LockRoot bool
+	LockRootUser bool
 
 	// What type of mount configuration should we create, systemd units, fstab
 	// or none
@@ -133,7 +133,7 @@ func (p *OSTreeDeployment) getBuildPackages(Distro) ([]string, error) {
 		"rpm-ostree",
 	}
 
-	if len(p.Users) > 0 || p.LockRoot {
+	if len(p.Users) > 0 || p.LockRootUser {
 		packages = append(packages, "shadow-utils")
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/distro/generic/images.go
```diff
@@ -83,6 +83,12 @@ func osCustomizations(t *imageType, osPackageSet rpmmd.PackageSet, options distr
 
 		osc.Users = users.UsersFromBP(c.GetUsers())
 		osc.Users = append(osc.Users, imageConfig.Users...)
+
+		// we also don't really want to lock the root account (as far
+		// as it exists) on installer ISOs
+		if imageConfig.LockRoot != nil {
+			osc.LockRoot = *imageConfig.LockRoot
+		}
```
lzap: Please use `common.ValueOrEmpty` for this, is much shorter.

lzap (dismissed):

pkg/distro/generic/images.go
supakeen: I didn't know it existed, it's not used at all in this file yet. I'll probably do a follow-up to rewrite the idiom everywhere as it is indeed nicer.

lzap: Tried my best with `grep` but there is probably more: https://github.com/osbuild/images/pull/2000

pkg/distro/image_config.go
```diff
@@ -89,7 +89,8 @@ type ImageConfig struct {
 
 	WSL *wsl.WSL `yaml:"wsl,omitempty"`
 
-	Users []users.User
+	Users    []users.User
+	LockRoot *bool `yaml:"lock_root,omitempty"`
```
mvo5: We have a `LockRootUser` option a few lines down (line 121 in this file) - how are the two related? And could we maybe reuse/unify the names (as they sound so similar)?

lzap (dismissed):

pkg/manifest/os.go
```diff
@@ -612,6 +612,15 @@ func (p *OS) serialize() (osbuild.Pipeline, error) {
 		pipeline.AddStage(osbuild.GenGroupsStage(p.OSCustomizations.Groups))
 	}
 
+	if p.OSCustomizations.LockRootUser {
+		usersOptions, err := osbuild.NewUsersStageOptionsForLockedRoot()
+		if err != nil {
+			return osbuild.Pipeline{}, err
+		}
+
+		pipeline.AddStage(osbuild.NewUsersStage(usersOptions))
```
mvo5: Without the error return in `osbuild.NewUsersStageOptionsForLockedRoot()` this could be a bit shorter:
```go
	if p.OSCustomizations.LockRootUser {
		pipeline.AddStage(osbuild.NewUsersStage(osbuild.NewUsersStageOptionsForLockedRoot())
   }
```

mvo5 (approved):

PR #2000, opened 2025-11-07
many: use common.ValueOrEmpty

lzap:
We introduced this helper some time ago, tried my best to find some instances where it can be used.

### DIFF
--- a/pkg/osbuild/grub2_legacy_stage.go
+++ b/pkg/osbuild/grub2_legacy_stage.go
@@ -148,9 +148,7 @@ func NewGrub2LegacyStageOptions(cfg *GRUB2Config,
 		},
 	}
 
-	if cfg != nil {
-		stageOptions.Config.GRUB2Config = *cfg
-	}
+	stageOptions.Config.GRUB2Config = common.ValueOrEmpty(cfg)
 
 	// NB: previously, the distributor was part of the GRUB2LegacyConfig struct and
 	// was always set. Now it is part of GRUB2Config, which could override it above.


pkg/distro/generic/images.go
```diff
@@ -121,9 +114,9 @@ func osCustomizations(t *imageType, osPackageSet rpmmd.PackageSet, options distr
 
 	language, keyboard := c.GetPrimaryLocale()
 	if language != nil {
-		osc.Language = *language
+		osc.Language = common.ValueOrEmpty(language)
```
mvo5: Slightly curious, this is of course fine but given that in the line above we did the `if language != nil` check, why not keep it as is with the "naked" dereference? Its fine if you prefer it for e.e. consistency, mostly curious

pkg/distro/generic/images.go
```diff
@@ -47,13 +48,9 @@ func osCustomizations(t *imageType, osPackageSet rpmmd.PackageSet, options distr
 		// an argument as fallback or make it not return the standard "kernel"
 		// when it's unset.
 		osc.KernelName = c.GetKernel().Name
-		if imageConfig.DefaultKernelName != nil {
-			osc.KernelName = *imageConfig.DefaultKernelName
-		}
+		osc.KernelName = common.ValueOrEmpty(imageConfig.DefaultKernelName)
```
thozza: We can't just simply use the `common.ValueOrEmpty()` everywhere, because it overrides the potentially previously set value with the type's default value. In this case, the kernel package name gets overridden with an empty string.

thozza (commented):
We can't just simply use the `common.ValueOrEmpty()` everywhere, because it overrides the potentially previously set value with the type's default value.

PR #2001, opened 2025-11-07
depsolvednf: Check for empty depsolve executable string

bcl:
Improve the error message when no osbuild-depsolve-dnf is installed. Without this check the error looks like:

error: error depsolving: running osbuild-depsolve-dnf failed:
starting  failed: exec: no command

Which is a bit unclear. With this extra check the error looks like:

error: error depsolving: running osbuild-depsolve-dnf failed:
osbuild-depsolve-dnf command undefined


thozza (approved):
Thanks.

Also, this just revealed places that need renaming `dnfjson` to `osbuild-depsolve-dnf` 😬

Issue #1864, opened 2025-09-15
Mocked source IDs for file customizations affected by repo definitions


achilleas-k:
The mocked checksums (IDs) for file customizations with URLs were affected by changes in repository definitions (adding a new repo).  This shouldn't happen.  It's not very important, as long as the checksum generation is stable, but it would be good to know what's going on.

-----

The checksum changes for other manifests seem inconsequential. Further changes work and I don't really have time to dive into the why. I'm including them in the commit.

_Originally posted by @supakeen in https://github.com/osbuild/images/issues/1863#issuecomment-3291286059_

achilleas-k:
https://github.com/osbuild/images/blob/49c3d90f745129ff6c660d52ef1aab12dc89f0b3/cmd/gen-manifests/main.go#L232-L245

This function writes a file in a tmpdir and replaces the URI with it, which makes the manifest generation stable and independent of the host.  The URI replacement however persists to the next blueprint that uses the same configuration file, because the URI modification happens on the object in the config map.  All instances of the same configuration file in the config map are references to the same underlying object.

PR #2004, opened 2025-11-10
fedora: f41 eol

supakeen:
We don't support building EOL distributions with our embedded definitions. Drop the Fedora 41 repositories and generated manifest checksums.

Note that checksums change for `file_customizations`. If I recall correctly that's because we do some things based on the filename for our pRNG.

This also removes all conditions in `fedora/imagetypes.yaml` that apply only to 'f42 and up' or 'below f42' (e.g, those that assume Fedora 41 to exist).

### DIFF
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -137,13 +137,6 @@
       - "xrdb"
       - "xz"
     conditions:
-      "f41 and below use tigervnc":
-        when:
-          version_less_than: "42"
-        append:
-          include:
-            - "tigervnc-server-minimal"
-            - "tigervnc-server-module"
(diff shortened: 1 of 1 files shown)


supakeen:
Would we want to drop all references to Fedora 41 in general? We can rename it in test cases and such.

achilleas-k:
> If I recall correctly that's because we do some things based on the filename for our pRNG.

#1864

achilleas-k:
> Would we want to drop all references to Fedora 41 in general? We can rename it in test cases and such.

If it's just test strings, I think it's okay to keep.

supakeen:
Undrafting as we want this in the next release of `image-builder` (which will happen after Fedora 41's EOL).

mvo5 (approved):

PR #2008, opened 2025-11-12
many: unattended network installer (HMS-9572, HMS-9769)

supakeen:
For @kingsleyzissou and @mvo5's work it's useful to be able to do unattended network installers. Since our installers are a bit ... convoluted I decided to implement this for them which ended up in a state I'm not entirely happy about. However, I also can't see a nicer way that wouldn't require pretty major refactoring so here we go :)

```
    customizations: `IsZero` for kickstarts
    
    For the implementation of kickstarts in the network installer we want to
    be able to know when a `kickstart.Options` is empty. This is so we can
    add the kickstart only if it actually contains values that we need to
    act upon instead of always (but with an empty kickstart file).
    
    Note that this also adjusts the setting of users and groups from the
    blueprint to use actual empty slices instead of slices of length 0. The
    former is considered empty, the latter is not.
```

```
    many: kickstart support for network installer
    
    Add support for kickstarts to the network installer. This works slightly
    differently from payload-type installers (image, bootc, ostree) because
    for the network installer we want *no* kickstart at all when no
    customizations have been passed. The others always have a kickstart
    because they always set some form of payload.
    
    Not entirely happy about this; but it does work and isn't overly weird?
```

### DIFF
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -2178,8 +2178,11 @@ image_types:
     supported_blueprint_options:
       - "distro"
       - "customizations.fips"
-      - "customizations.locale"
+      - "customizations.group"
       - "customizations.installer"
+      - "customizations.kernel.append"
+      - "customizations.locale"
+      - "customizations.user"
 
   "pxe-tar-xz":
     filename: "pxe.tar.xz"


internal/common/empty.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// compare a struct to its zero (empty) value. note that if a struct contains
+// maps or slices that an empty slice is different from a slice of length 0.
+// the latter is *not* an empty value and thus a struct won't compare empty
+func IsEmptyStruct(object interface{}) (bool, error) {
```
achilleas-k: Lets follow the conventional form for docstrings here with
```go
// <FunctionName> <does things>
```
so
```
// IsEmptyStruct compares a struct to its zero...
```

supakeen:
It seems Python linters are failing, but I didn't touch any Python in this PR so I think that can be ignored? :)

data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -2277,6 +2277,7 @@ image_types:
       - "distro"
       - "customizations.fips"
       - "customizations.locale"
+      - "customizations.installer"
```
mvo5: AIUI we support more than those [0] and we also need to update the other distros (rhel-*)?

[0] If not, having kernel.append would be very nice as it allows us to set console=ttyS0 so that we see what the installer is doing during the automatic test

mvo5: I did a quick experiment in https://github.com/osbuild/images/commit/144d6bcda88c77d9bc2ba222cbcfef687f15d362 (branch https://github.com/osbuild/images/compare/main...mvo5:qemu-runner-bib-netinst-real-test?expand=1)

kingsleyzissou: image

I'm getting the same and I think we explicitly need to set the user & group customisations in the yaml file otherwise it doesn't look like the get set. But ya, same issue with the source here

ETA: I built a fedora-42 iso (aarch) for what it's worth

supakeen:
> Is there some other use case for this, except testing? If not, I'm inclined to one of the inline comments about injecting the KS into the image or serving it via HTTP to the VM and using kernel options `inst.ks=http://<server>/path/to/kickstart.cfg` instead.
> 
> If there's another use case, then let's do it. It is acceptable. In general, I agree with other comments.

It is useful for the automated testing (though it could work around it). In my opinion though, do we have a reason why we don't want to support installer customizations on the network installer, I assume we want things like adding users and such working in the service as well? :)

pkg/customizations/kickstart/kickstart.go
```diff
@@ -76,6 +85,14 @@ func New(customizations *blueprint.Customizations) (*Options, error) {
 	return options, nil
 }
 
+func (options Options) IsZero() bool {
+	// we throw away the error value here, the only expected error is if the
+	// argument to IsEmptyStruct isn't a struct, and `Options` is always a
+	// struct
+	zero, _ := common.IsEmptyStruct(options)
```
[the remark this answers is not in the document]
supakeen: I've dropped the commit that introduced `common/empty.go` and put the reflect call directly into `kickstart.go`.

achilleas-k:
> do we have a reason why we don't want to support installer customizations on the network installer, I assume we want things like adding users and such working in the service as well?

I don't know.  Maybe we should answer this question first before deciding [what we want to do here](https://github.com/osbuild/images/pull/2008#pullrequestreview-3464417413).

My take is we should limit the scope to whatever was requested.  In this case, it's just testing.  The task even mentions that
> One idea discussed was using multiple external kickstart files to boot the same image multiple times to test different parts of anaconda.

which suggests we don't really need to be baking kickstarts into the ISO at all, but can do what Tomas suggested, kickstarts from a URL passed in through kernel args.  We could boot the same ISO multiple times with different external kickstarts.

It's always fun to add more features, but let's consider the support and test burden as well as the complexity.  Let's not add more than we need to.

mvo5:
[..]
> It's always fun to add more features, but let's consider the support and test burden as well as the complexity. Let's not add more than we need to.

Sure thing and sorry that I caused this ruckus - I will explore how to do this without anything special in the iso.

kingsleyzissou (approved):
The changes look all right to me, I'll let one of the regular maintainers have a final say though

PR #2011, opened 2025-11-12
manifest: add treeinfo stage (HMS-8574)

lzap:
Generated ISO files are missing .treeinfo file and thus it cannot be used easily with virt-install/virt-manager/cockpit to install VMs. This commit adds a new TreeinfoStage to osbuild that generates the .treeinfo file based on the parameters passed from the anaconda installer ISO manifest.

All fields supported by the new osbuild stage are defined, but one is currently skipped: the ShortName because we do not have it in distrodefs.

---

A new stage that is filled with all the information I could gather at this moment, the only exception is the short name which is either `F` or `RHEL` and I am not sure where exactly this is used (maybe @thozza knows) but `virt-install`, which is the reason why this missing feature was reported to us, [does not use it](https://github.com/virt-manager/virt-manager/blob/f9a01fbf96b35cb30473abfb96b731be9b8b6c9e/virtinst/install/urldetect.py#L51). It does, however, still parse the legacy "general" section so the manifest stage does copy these fields to make sure this works properly.

Osbuild v164 required. The osbuild stage that this uses is: https://github.com/osbuild/osbuild/pull/2200/files

I have built CentOS 10 and this is the result:

```
#!/bin/bash
MF=centos_10-x86_64-image_installer-unattended_iso.json
IBCACHE=/var/cache/image-builder
IBSTORE=$IBCACHE/store
IBEXPORT=$IBCACHE/export
IBMF=$IBCACHE/manifests

go run ./cmd/gen-manifests --distros centos-10 --arches x86_64 --types image-installer --config ./test/configs/unattended-iso.json --output $IBMF
jq .manifest $IBMF/$MF | sudo python3 -m osbuild --libdir /home/lzap/osbuild --store $IBSTORE --output-directory $IBEXPORT --export build --export os --export bootiso-tree -
```

Here is the generated `.treeinfo`.

```
[release]
name=CentOS Stream
version=10-stream

[tree]
arch=x86_64
build_timestamp=1763028879
platforms=x86_64

[checksums]
images/pxeboot/vmlinuz=sha256:60d7944e3029798bbb85161f39b9b5af421fe1ccb7fef4cd3eb150168df26fd8
images/pxeboot/initrd.img=sha256:a2e19856b5d2ada23adceceff709168039df25068bf22f063837eba30de4fc22
images/install.img=sha256:317cce0c229893b07c5e0e4753ea2b88e07981bef26905a7a138b91bc9a18339

[stage2]
mainimage=images/install.img

[images]
images-x86_64={'initrd': 'images/pxeboot/initrd.img', 'kernel': 'images/pxeboot/vmlinuz'}

[general]
arch=x86_64
family=CentOS Stream
name=CentOS Stream 10-stream
platforms=x86_64
timestamp=1763028879
version=10-stream

[header]
version=1.0
```

I see problems:

* Version should have been `10` and not `10-stream` I think
* The images section is not serialized correctly

Looks like a bug in `ostree` stage to me because the JSON looks ok:

```json
          {
            "type": "org.osbuild.treeinfo",
            "options": {
              "path": ".treeinfo",
              "treeinfo": {
                "release": {
                  "name": "CentOS Stream",
                  "version": "10-stream"
                },
                "tree": {
                  "arch": "x86_64",
                  "build_timestamp": 1763028879,
                  "platforms": [
                    "x86_64"
                  ]
                },
                "checksums": [
                  "images/pxeboot/vmlinuz",
                  "images/pxeboot/initrd.img",
                  "images/install.img"
                ],
                "stage2": {
                  "mainimage": "images/install.img"
                },
                "images": {
                  "images-x86_64": {
                    "initrd": "images/pxeboot/initrd.img",
                    "kernel": "images/pxeboot/vmlinuz"
                  }
                },
                "general": {
                  "arch": "x86_64",
                  "family": "CentOS Stream",
                  "name": "CentOS Stream 10-stream",
                  "platforms": [
                    "x86_64"
                  ],
                  "timestamp": 1763028879,
                  "version": "10-stream"
                }
              }
            }
          }
```


lzap:
Cannot find why this line is broken:

```
images-x86_64={'initrd': 'images/pxeboot/initrd.img', 'kernel': 'images/pxeboot/vmlinuz'}
```

Is it something in here, or in `osbuild` stage??? There is a test for that and it works. Hmmmm

lzap:
> Malformed version: 164

Hmmm isn't this related to my `data/dependencies/osbuild` change? My editor removed the newline I think or added one. But is this really relevant to my issue?

> Which suggests we need to bump the osbuild commit ID in Schutzfile.

Yeah I think I messed up the osbuild dependency update then. But still when I run this locally, I am getting that weirdly-formatted result. It builds fine, creates the file, but it has this messed up line.

supakeen:
I tested the ISOs that are being modified in this PR, these ISOs all work under `qemu` without this PR applied.

- `centos-10-network-installer` fails to install, no repositories available.
- `centos-10-image-installer` installs.
- `fedora-43-live-installer` fails to install, `org.fedoraproject.Anaconda.BootloaderInstallationError: gen_grub_cfgstub script failed` (unsure if this is due to `.treeinfo` being present).

Since this PR wants to add support for `virt-install` without passing arguments manually I also ran `virt-install` on these ISOs.

- `centos-10-network-installer`, fails to install, no repositories available.
- `centos-10-image-installer`, fails to install, no repositories available (this is weird, it shouldn't be looking at repositories).
- `fedora-43-live-installer`, fails to boot, emergency shell in dracut (maybe pointed at wrong initrd or kernel?).

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -254,12 +257,15 @@ func (p *AnacondaInstallerISOTree) NewSquashfsStage() (*osbuild.Stage, error) {
 
 	switch p.anacondaPipeline.Type {
 	case AnacondaInstallerTypePayload, AnacondaInstallerTypeNetinst:
+		p.Stage2Image = "images/install.img"
 		squashfsOptions = osbuild.SquashfsStageOptions{
-			Filename: "images/install.img",
+			Filename: p.Stage2Image,
 		}
```
thozza: Nitpick: the assignment now looks the same for all installer types, so it could be done just once after the `case` block.

pkg/osbuild/treeinfo_stage.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"encoding/json"
+
+	"github.com/osbuild/images/pkg/arch"
+)
+
+// treeinfoStageOptions represents the options of the treeinfo stage.
+// All paths are relative to .treeinfo path specified via Path field.
+// See https://github.com/osbuild/osbuild/blob/v165/stages/org.osbuild.treeinfo.meta.json
+type treeinfoStageOptions struct {
```
thozza: Nitpick: Not exporting this struct makes very little sense, given that all other structs are exported. But in fact, the `NewTreeinfoStage()` does not take any of these types as an argument, so none of them need to be exported. In any case, it would be good to be consistent.

PR #2012, opened 2025-11-12
data: add `supported_blueprint_options` to bootc-generic (HMS-9714)

mvo5:
This adds the `supported_blueprint_options` to the bootc based image types so that we can more accurately warn about unsupported options.

/jira HMS-9714


mvo5:
Fwiw, it works in https://github.com/osbuild/bootc-image-builder/pull/1138 (there is a integration tests there that validates it on stderr)

PR #2017, opened 2025-11-14
test: add network-installer boot test 

mvo5:
[draft as this needs/is just build on top of https://github.com/osbuild/images/pull/2008]

This is the unattended installer changes (pr#2008) with a tiny commit on top that essentially just enables the cdrom testing we added in pr#1991 for network-installer.

### DIFF
--- a/test/scripts/boot-image
+++ b/test/scripts/boot-image
@@ -156,6 +156,17 @@ def ensure_can_run_qemu_test(arch, image_path, config_file):
     customizations = config.get("blueprint", {}).get("customizations", {})
     if "openscap" in customizations:
         raise CannotRunQemuTest("openscap is not working right now")
+    # network-installer has a bunch of extra constraints
+    info_path = pathlib.Path(image_path).parent / "../info.json"
+    info = json.loads(info_path).read_text(encoding="utf8")
+    if info.get("image-type") in ["network-installer", "everything-network-installer"]:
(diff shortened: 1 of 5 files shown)


test/scripts/boot-image
```diff
@@ -156,6 +156,15 @@ def ensure_can_run_qemu_test(arch, image_path, config_file):
...
     if "openscap" in customizations:
         raise CannotRunQemuTest("openscap is not working right now")
+    # network-installer has a bunch of extra constraints
+    if config.get("image-type") in ["network-installer", "everything-network-installer"]:
+        # network install needs subscribed content but we cannot provide
+        # that currently
+        match config.get("distro"):
+            case name if name.startswith("rhel"):
+                raise CannotRunQemuTest("rhel network-installer needs subscription keys which we cannot provide yet")
+            case "centos-10":
```
mvo5: I strongly suspect we need to also skip centos-9 because afaict it will not automatically use a network repo and in our unattended kickstart file we do not include a "url" - this seems to be handled automatically (and correctly) in centos-10s anaconda.

mvo5: There is now also a `software selection error` in fedora 43 (c.f. https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/12121649535) which might (or might not be) a fluke

PR #2025, opened 2025-11-19
bootc: Select manifest based on YAML image_func

bcl:
Instead of using the name use the image_func field like we do for non-bootc images.

Disk types were already setting "bootc_disk", just not using it. Add "bootc_iso" and "bootc_legacy_iso".

### DIFF
--- a/data/distrodefs/bootc-generic/imagetypes.yaml
+++ b/data/distrodefs/bootc-generic/imagetypes.yaml
@@ -149,6 +149,7 @@ image_types:
     exports: ["bootiso"]
     filename: "installer.iso"
     boot_iso: true
+    image_func: "bootc_iso"
 
   # this image type is special (in many ways) and all is a bit ugly:
   # - we want to get rid of it (here for compat with bib)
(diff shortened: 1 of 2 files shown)


internal/test/boot_helpers.go
```diff
@@ -31,14 +31,19 @@ func run(c string, args ...string) ([]byte, []byte, error) {
...
+// SshRun runs the ssh command, retrying 3 times with a 10s delay if there is an error
 func SshRun(ip, user, key, hostsfile string, command ...string) error {
 	sshargs := []string{"-i", key, "-o", fmt.Sprintf("UserKnownHostsFile=%s", hostsfile), "-l", user, ip}
 	sshargs = append(sshargs, command...)
-	_, _, err := run("ssh", sshargs...)
-	if err != nil {
-		return err
+	var err error
+	for try := 0; try < 3; try++ {
+		_, _, err = run("ssh", sshargs...)
```
mvo5: I'm not sure we should rerun the command itself multiple times, I mean, unless we are sure its the ssh part that is failing not the command itself (which I don't think we can differentiate right now) - the reason is that the command itself maybe broken but re-running may fix it (or the command itself is flaky and this way we never learn). I mean, maybe it does not matter in practice but I would feel better if we only tried ssh with "true" or a similar "nil" command and then run the actual command only once - but maybe I'm overthinking it.

supakeen (approved):

PR #2028, opened 2025-11-21
test: import vmtest from bib

mvo5:
This imports the vmtest module into images (from bootc-image-builder) with the full history.

Now that we use it for qcow2 and cdrom testing it seems to make sense to have it here. However I'm open for other options of course, it could become its own repo for example (leaving it in bib is probably not good as we are currently gutting this repo). Once nice aspect about having it here is that it would allow us to run full image tests against any changes which probably helps a lot with the robustness.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,14 +1,20 @@
+# Note that this is pyproject file is here only for the vmtest utils.
+
 [tool.autopep8]
 max_line_length = 120
 
-[project]
-name = "images"
-version = "0.1.0"
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,14 +1,21 @@
...
+
 [project]
-name = "images"
+name = "vmtest"
 version = "0.1.0"
-
 dependencies = [
-    "vmtest @ git+https://github.com/osbuild/bootc-image-builder.git",
+  "boto3==1.33.13",
+  "paramiko==4.0.0",
```
kingsleyzissou: Do we still need this dependency now since we dropped paramiko for regular ssh & scp?

supakeen (approved):

PR #2035, opened 2025-11-24
defs/rhel8: exclude RHEL-specifics for non-RHEL

supakeen:
Non-RHEL distributions don't use RHUI image types, they should be excluded from non-RHEL.

### DIFF
--- a/data/distrodefs/rhel.yaml
+++ b/data/distrodefs/rhel.yaml
@@ -243,6 +243,15 @@ distros:
         ignore_image_types:
           - "azure-rhui"
           - "vhd"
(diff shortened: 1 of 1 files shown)


thozza:
> Aren't most of these already done through customizations in pungi (or other schedulers) nowadays on the 'base' image types for these clouds?

That's exactly my point. For those that are produced as `xz` archives by default, it probably still makes sense to exclude them. But in general, these `-sap`, `-ha`, `-eap` image types may be useful also outside of RHEL.

I would not block on it right now. Some required content may not even be available in these distros.

PR #2036, opened 2025-11-24
vmtest: make VM.run() more like subprocess.run() 

mvo5:
Now that vmtest is used more it should get some love. Here is the first step:

This makes vmtest.VM.run() behave more like subprocess.run(), i.e. it will raise an exception by the command fails and return a subprocess.CalledProgress on sucesss. The old code is really a bit cumbersome to use.


supakeen (approved):

thozza (approved):
Great stuff ❤️

PR #2040, opened 2025-11-25
support bootc verb in bootc-installer

ondrejbudai:
Anaconda has recently got a new verb for bootc container images, this commit adds support for it for bootc-installer.

This is currently opt-in via yaml configs, we need to decide how to enable it automatically.

The code is testable with this `Containerfile`:

```dockerfile
FROM quay.io/fedora/fedora-bootc:rawhide
RUN dnf --setopt=tsflags=noscripts install -y dnf5-plugins  && dnf copr enable -y @rhinstaller/Anaconda
RUN dnf install -y \
     anaconda \
     anaconda-install-env-deps \
     anaconda-dracut \
     dracut-config-generic \
     dracut-network \
     net-tools \
     squashfs-tools \
     grub2-efi-x64-cdboot \
     python3-mako \
     lorax-templates-* \
     biosdevname \
     prefixdevname \
     && dnf clean all
# shim-x64 is marked installed but the files are not in the expected
# place for https://github.com/osbuild/osbuild/blob/v160/stages/org.osbuild.grub2.iso#L91, see
# workaround via reinstall, we could add a config to the grub2.iso
# stage to allow a different prefix that then would be used by
# anaconda.
# once https://github.com/osbuild/osbuild/pull/2202 is merged we
# can update images/ to set the correct efi_src_dir and this can
# be removed
RUN dnf reinstall -y shim-x64
RUN mkdir -p /boot/efi && cp -av /usr/lib/efi/{shim,grub2}/*/EFI /boot/efi/
# lorax wants to create a symlink in /mnt which points to /var/mnt
# on bootc but /var/mnt does not exist on some images.
#
# If https://gitlab.com/fedora/bootc/base-images/-/merge_requests/294
# gets merged this will be no longer needed
RUN mkdir /var/mnt
```

and this command:

```console
# podman build -t fedora-installer:44 .
# podman pull quay.io/centos-bootc/centos-bootc:stream10
# image-builder build --bootc-ref localhost/fedora-installer:44 --bootc-installer-payload-ref quay.io/centos-bootc/centos-bootc:stream10 bootc-installer --default-fstype ext4
```

Note that you have to have osbuild with https://github.com/osbuild/osbuild/pull/2242, marking as a draft until it gets merged.

### DIFF
--- a/data/distrodefs/bootc-generic/imagetypes.yaml
+++ b/data/distrodefs/bootc-generic/imagetypes.yaml
@@ -163,6 +163,9 @@ image_types:
     filename: "installer.iso"
     boot_iso: true
     image_func: "bootc_legacy_iso"
+    # Uncomment the following lines to use the bootc verb instead of ostreecontainer
+    # installer_config:
+      # bootc_install_verb: bootc
 
   # XXX: ideally we would use name_aliases but the loader lib
   # does not not fully support this yet
--- a/pkg/distro/bootc/bootc.go
+++ b/pkg/distro/bootc/bootc.go
@@ -556,6 +556,11 @@ func (t *BootcImageType) manifestForISO(bp *blueprint.Blueprint, options distro.
 	// see https://github.com/osbuild/bootc-image-builder/issues/733
 	img.InstallerCustomizations.ISORootfsType = manifest.SquashfsRootfs
 
+	installerConfig := t.ImageTypeYAML.InstallerConfig(t.arch.distro.id, t.arch.Name())
+	if installerConfig != nil && installerConfig.BootcInstallVerb != nil {
+		img.BootcInstallVerb = *installerConfig.BootcInstallVerb
+	}
+
 	installRootfsType, err := disk.NewFSType(t.arch.distro.defaultFs)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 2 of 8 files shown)


bcl (approved):
Looks good to me, after fixing the conflict.

lzap (approved):

supakeen (commented):
This is fine to have but needs rebasing and linter appeasement.

A small question, if Anaconda supports the `bootc` verb depends on the Anaconda version but we do not have bootc definitions per-version (as we don't really know what we're building).

Should this then be left to the user or container build to determine instead of the YAML definitions which would apply to all? I can imagine it not specifically making sense for us to inject a kickstart in `bootc-installer` but that having to be done in the installer container instead?

data/distrodefs/bootc-generic/imagetypes.yaml
```diff
@@ -163,6 +163,9 @@ image_types:
     filename: "installer.iso"
     boot_iso: true
     image_func: "bootc_legacy_iso"
+    # Uncomment the following lines to use the bootc verb instead of ostreecontainer
+    # installer_config:
+      # bootc_install_verb: bootc
```
supakeen: This is under `anaconda-iso`, should it be under `bootc-installer` instead?

supakeen:
Not particularly relevant anymore due to `bootc-generic-iso`?

PR #2041, opened 2025-11-25
cloud/azure: optimise zero slice check when uploading page blob

croissanne:
`bytes.Equal` or `bytes.Compare` can be significantly faster (depending on the platform and hardware) than simply iterating over the slice.

---

With 1024 samples and a chunksize of 4k it's about 15-17 times faster on my laptop. With larger chunksizes (4MiB like in the example), the advantage seems to be diminished, but still about 8 times.

https://gist.github.com/croissanne/3c3a7bb41026cdd88526faf144be6db7

```shell
sanne@worky ~/workstuff/journal-extra/playground $ go run ./playground.go 
AllZeros: %s 2m44.204447552s
Comp: %s 20.391145901s
Equal: %s 19.218426318s
all zeros slower? true, bytes.Compare is about 8.052732708069499 times faster
```

### DIFF
--- a/pkg/cloud/azure/azurestorage.go
+++ b/pkg/cloud/azure/azurestorage.go
@@ -63,17 +63,6 @@ const DefaultUploadThreads = 16
 // See https://learn.microsoft.com/en-us/rest/api/storageservices/put-page
 const PageBlobMaxUploadPagesBytes = 4 * datasizes.MiB
 
-// allZerosSlice returns true if all values in the slice are equal to 0
-func allZerosSlice(slice []byte) bool {
-	for i := 0; i < len(slice); i++ {
-		if slice[i] != 0 {
-			return false
-		}
-	}
-
-	return true
-}
-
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):
Thanks!

ondrejbudai (dismissed):

pkg/cloud/azure/azurestorage.go
```diff
@@ -142,6 +131,7 @@ func (c StorageClient) UploadPageBlob(metadata BlobMetadata, fileName string, th
 	// Run the upload
 	run := true
 	var wg sync.WaitGroup
+	zeros := bytes.Repeat([]byte{0}, PageBlobMaxUploadPagesBytes)
```
thozza: Nitpick: this could be simply `zeros := make([]byte, PageBlobMaxUploadPagesBytes)`

thozza (commented):
Thanks for digging into the options to make the upload faster. I've added two suggestions to make the code simpler and slightly faster.

PR #2042, opened 2025-11-25
[RFC] osbuild: introduce PathChanger interface and use it

mvo5:
[this is a version of the relabel bootc work that is fine grained, there is a simpler version in ihttps://github.com/osbuild/images/pull/2053  that I like (a lot!) based on the idea by Alex outlined in **https://github.com/osbuild/images/pull/2037#issuecomment-3575925113** - ie. simplify this to just /var, /etc if any postStages are found - but if we want the finer grained one I I would prefer if the detailed knowledge what changed is part of the stage itself instead of something that the manifest package investigates]

It is useful for the higher layers to know what files are
touched by a specific stage. This is used in e.g. the
bootc disk where we want to target the relabeling.

So this commit introduce a new `PathChanger` interface
that can be implemented by the relevant stages and is
then used in RawBootcImage to relabel only the specific
paths that the stages touch.

Build on top of the excellent work by Alex in PR#2037.

### DIFF
--- a/pkg/osbuild/stage.go
+++ b/pkg/osbuild/stage.go
@@ -28,3 +28,7 @@ func (s *Stage) MountOSTree(osName, ref string, serial int) {
 	ostreeMount := NewOSTreeDeploymentMount(name, osName, ref, serial)
 	s.Mounts = append(s.Mounts, *ostreeMount)
 }
+
+type PathChanger interface {
+	PathsChanged() []string
+}
--- a/pkg/manifest/raw_bootc.go
+++ b/pkg/manifest/raw_bootc.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"slices"
-	"strings"
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/artifact"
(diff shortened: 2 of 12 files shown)


pkg/osbuild/stage.go
```diff
@@ -28,3 +28,7 @@ func (s *Stage) MountOSTree(osName, ref string, serial int) {
 	ostreeMount := NewOSTreeDeploymentMount(name, osName, ref, serial)
 	s.Mounts = append(s.Mounts, *ostreeMount)
 }
+
+type FileChanger interface {
+	FilesChanged() []string
+}
```
thozza: Nitpick: I had a slight problem with naming the functions using "files" in the original PR, because this is about paths (symlinks, directories, files) in general, not just files. Before, the code was limited to bootc, so it was not a big deal from my PoV. Since we are now generalizing the solution, which is great, let's use something more descriptive (maybe `PathsChanged()`)?

At the end of the day, I can live with the naming as is, just wanted to point this out.

mvo5: Yeah, I did this as a quick draft to show the concept, its incomplete, I need to implement the other stages like chmod/chown and iirc some more if we want to continue. I like your name better, I think `PathsChanged()` it is :) 

The alternative would be just he simple approach Alex outlined in https://github.com/osbuild/images/pull/2037#issuecomment-3575925113 ie just always relabeling /etc, /var if we have post stages. I really like the simplicity of the appraoch but also worry about the precision (long term). So I'm a bit on the fence about it and wanted to hear the others first :) But happy to properly flesh this out.

PR #2043, opened 2025-11-26
cmd: check-host-config rewritten in Go (HMS-9805)

lzap:
Yeah. Cozy winter evening happened.

I wrote the whole framework and implemented two checks and then asked AI to implement the rest. Only checkers in the `check` package were generated and reviewed by me, all the rest is my own code. Let me know how you like it. I intentionally did not make any changes to the behavior, this is pure refactoring which will open doors to more features like using YAML to find which tests are missing etc.

Also the output is just pure text, but it could be maybe `gotest` now or something that CI understands and can parse for nicer results. But one step after another, I would rather keep it simple for now.

Now, I really wanted this to be fast so everything runs in parallel, but logs from all goroutines are collected and presented in a nicer way sequentially so the output is not all messed up. I heavily use what is called context function mocking where the platform functions (Exec, Exists, Grep) can be overriden with a mock via special context value. The key type can be not exported making it impossible to mess around with the functions from other packages.

The hard part is now to integrate this in the CICD. We want to build the binaries for each PR and then upload them somewhere and take it from here. For now, I am settling up with simply running "go build" for each test and then copying the binaries over to the image the same way as we did for the script. If we cache GOCACHE maybe this might not be a bad idea at all.

You can safely run this locally, I have reviewed and tested everything:

```
$ go run ./cmd/check-host-config -config test/configs/all-customizations.json 
[main           ] Loading build config from test/configs/all-customizations.json
[files          ] Checking existence of file: /etc/systemd/system/custom.service
[files          ] Checking if file exists: /etc/systemd/system/custom.service
[files          ] File does not exist: /etc/systemd/system/custom.service
[hostname       ] Executing: hostname
[hostname       ] Command succeeded: hostname
[hostname       ] Comparing 'dev.home.lan' with expected hostname 'my-host'
[cacerts        ] Parsing CA cert 1
[cacerts        ] Extracting serial from CA cert 1: 27894af897dd2423607045716438a725f28a6d0b
[cacerts        ] Extracting CN from CA cert 1: Test CA for osbuild
[cacerts        ] Checking CA cert 1 anchor file serial '27894af897dd2423607045716438a725f28a6d0b'
[cacerts        ] Checking if file exists: /etc/pki/ca-trust/source/anchors/27894af897dd2423607045716438a725f28a6d0b.pem
[cacerts        ] File does not exist: /etc/pki/ca-trust/source/anchors/27894af897dd2423607045716438a725f28a6d0b.pem
[srv-disabled   ] Checking disabled service: bluetooth.service
[srv-disabled   ] Executing: systemctl is-enabled bluetooth.service
[srv-disabled   ] Command succeeded: systemctl
[srv-masked     ] Executing: systemctl list-unit-files --state=masked
[srv-masked     ] Command succeeded: systemctl
[srv-masked     ] Checking masked service: nfs-server
[users          ] Checking user: user1
[users          ] Executing: id user1
[users          ] Command failed: id (exit code: exit status 1)
[srv-enabled    ] Checking enabled service: sshd.service
[srv-enabled    ] Executing: systemctl is-enabled sshd.service
[srv-enabled    ] Command succeeded: systemctl
[srv-enabled    ] Service was enabled service=sshd.service state=enabled
[srv-enabled    ] Checking enabled service: custom.service
[srv-enabled    ] Executing: systemctl is-enabled custom.service
[srv-enabled    ] Command failed: systemctl (exit code: exit status 4)
[fw-srv-disabled] Checking disabled firewall service: telnet
[fw-srv-disabled] Executing: sudo firewall-cmd --query-service=telnet
[fw-srv-disabled] Command failed: sudo (exit code: exit status 1)
[fw-srv-disabled] Firewall service was disabled service=telnet state=no
[fw-srv-enabled ] Checking enabled firewall service: ftp
[fw-srv-enabled ] Executing: sudo firewall-cmd --query-service=ftp
[fw-srv-enabled ] Command failed: sudo (exit code: exit status 1)
[fw-ports       ] Checking enabled firewall port: 1337/udp
[fw-ports       ] Executing: sudo firewall-cmd --query-port=1337/udp
[fw-ports       ] Command failed: sudo (exit code: exit status 1)
Results:
✅ Firewall Services Disabled Check
⚠️  Hostname Check (warn: hostname does not match, got dev.home.lan expected my-host)
❌ Files Check (fail: file does not exist: /etc/systemd/system/custom.service)
❌ CA Certs Check (fail: file missing for cert 1 at /etc/pki/ca-trust/source/anchors/27894af897dd2423607045716438a725f28a6d0b.pem)
❌ Services Disabled Check (fail: service is not disabled: bluetooth.service state: enabled)
❌ Services Masked Check (fail: service is not masked: nfs-server)
❌ Users Check (fail: user does not exist: user1)
❌ Services Enabled Check (fail: service is not enabled: custom.service error: exit status 4)
❌ Firewall Services Enabled Check (fail: firewall service is not enabled: ftp error: exit status 1)
❌ Firewall Ports Check (fail: firewall port is not enabled: 1337:udp error: exit status 1)
```

The result codes indicate a test which failed as the first one, starting from 10 (1-9 are reserved). These goroutines can fail in random order, so it must not be used for any workarounds it is purely informative.


lzap:
Rebased, modularity fails but this is a bug (reported, see the issue above). Then OpenSCAP fails because the threshold was not reached for some cases. And finally bunch of AWS limits reached. Other than that, green.

I see some `jq-only` config, can I delete this one now? The `jq` should no longer be required.

lzap:
Breaking up commits from other packages does make sense, yeah. I was referring to breaking up things within one package. Let me do that, also, I think it make sense to put checks into separate commit at well.

cmd/check-host-config/wait.go
```diff
@@ -0,0 +1,111 @@
...
+	os.Unsetenv("DBUS_VERBOSE")
+
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
+	lastLogTime := time.Now()
+
+	// Set up signal handling for OS termination signals
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGTERM, syscall.SIGINT)
+	defer signal.Stop(sigChan)
```
supakeen: Would it be nicer to have this set up in `main()` instead and have an overarching context starting form there that we use everywhere (and create new contexts on) that gets cancelled on these signals?

lzap: Good observation, in fact, this is a less readable way of doing the same thing. I can utilize `NotifyContext` since context is being passed both to the wait and the main checking routines.

lzap:
> The last commit is a `!fixup`. Are you waiting for reviews before doing the rebase for that?

I was debugging a problem with OpenSCAP but I fixed it and merged it now as it works as expected.

Only one random test failure (SSH disconnect) other than that all green. Resolved conflicts and rebased.

lzap:
Started seeing the VCS git error after rebase, not sure which commit caused it but this error must stop. Here is my solution: https://github.com/osbuild/images/pull/2100

Will rebase this on top of that one.

cmd/check-host-config/wait.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/osbuild/images/cmd/check-host-config/mockos"
+)
+
+// runningWait emulates 'systemctl is-system-running --wait'
+// It blocks until the system reaches "running" or fails on other states.
+// It is requried for older versions of systemd that don't support the option (EL8)
```
thozza: Typo: `requried` -> `required`

thozza (commented):
Firstly, thank you for taking the time and rewriting the check in Go. It is something that we wanted for a long time, so it is great to see you taking on this task and making it a real thing. I like the `check.Check` interface and the check metadata; it is nice. The overall design is solid.

I truly and honestly appreciate your effort. It must have been a lot of work.

That said, this PR is huge. Given its size, I can only provide high-level feedback without dedicating significant time to a line-by-line review.

Here is my general feedback for now:

## Patch set organization

WRT splitting code changes into smaller commits... I don't care too much about how the code was implemented and if this reflects the "real" history. I am not convinced that it is important at all. What is important from my PoV is to split the change into small, atomic logical steps that keep the code working and that are reasonably easy to review and follow. This is not the case with this PR.

For example, adding the basics and then calling the Bash and Go version in parallel, while you'd replace one check at a time from the Bash script and move it to Go, would make it possible for me to also check that the check logic stayed the same. This is nearly impossible for me without extensive effort.

I know this means reworking the commits, which is additional effort on top of the implementation work you've already done. However, this would help me give you a better and a faster review.

## Parallelism

I get that you wanted the checks to run fast, but I'm wondering if running checks in parallel is actually necessary here. The original bash script ran sequentially and was fast enough. Most checks complete in milliseconds (just querying systemd/firewalld), and the "slow" ones like OpenSCAP and Modularity are often skipped anyway. The parallelism adds complexity with goroutines, channels, and WaitGroups, plus there's potential for race conditions when multiple checks hit systemd or firewalld simultaneously. Would it be simpler to just run them sequentially? I think so.

## Testing approach

I actually don't like the unit testing approach. I'm not convinced that it tests the actual code paths that are executed during an actual command run. The mockos package uses context-based function injection for mocking, which is a different approach from what we use elsewhere in the codebase. Our established pattern is the Mock*() (restore func()) style you can see in pkg/osbuild/export_test.go or pkg/bib/blueprintload/export_test.go. Using a consistent mocking approach across the project makes it easier for contributors to understand and maintain the tests.

From what I've read in the discussion and from what I understood from the changes, I think the context-based mocking was chosen specifically to support parallel test execution—you can't safely use global variable mocking when tests run concurrently. But if we switch to sequential execution (per the point above), we could use the standard Mock*() (restore func()) pattern instead, which would be more consistent with the rest of the codebase.

With the current mocking approach, the tests return early from ExecContext when a mock is present, so the actual command execution code path is never exercised. This means we're testing "does the check logic work given this assumed output" but not "does the real command get constructed correctly" or "does the real output parsing work." The tests are valuable, but they're more like logic tests than integration tests. It might be worth noting this limitation or considering how to get coverage of the real code paths.

lzap:
**Parallelism**

Parallelism is cheap in Go, the function that runs all checks is literally 40 lines and the printing loop is 30 lines. It was never the performance why I started it this way, it just feels natural to write software these days that is fast from the day one. I would like to hear from others about this one tho because the code that runs tests checks will unlikely need any changes in the future it will be the checks what we will be working on. I can give up this one since these are just host tests, but we are speaking about few lines here and this is clean Go.

**Mocking**

Granted, using context for mocking is weird and somewhat an anti-pattern, and I expected a pushback. I will say, I already expressed my concerns about the mentioned established pattern also known as monkey patching and I am really hesitant to do this in any new code. Inability to run tests in parallel is a major drawback and it is a good practice to treat testing code as runtime (production) code. Having said all of that, since the nature of this code (test helper/checks) it is not a big deal and I will refactor. However, I will likely introduce a special mocking helper with some extra guards.

**Unit tests vs integration tests**

I hate mocked unit tests with passion because one can have 100% green project that simply does not work. But I strongly believe that host config checks need properly mocked unit tests from its nature, it is a good fit here. We run this code in variety of environments, it is its purpose to be executed on different architectures and platforms. This is why it is very important to have unit test suite where there can be multiple mocked options - an output from command from RHEL, from modern Fedora, these can be pretty different as we know (e.g. DNF).

I do not want to turn those unit tests into integration tests (terminology can be blurry: from mocked to real tests) that will actually execute commands because the problem is the runtime environment. We cannot simply use the host system, that is different from laptop to laptop. Therefore I would need to come up with some VMs or containers to provide those environments. But we already have that - our images! I do not want to create a new testing framework just to execute the same code in our real tests. That is why I think, despite mocked unit tests are more often than not the wrong answer, it is actually a good fit here.

**Conclusion**

Although I do not find readability or "potential for race conditions" of 100 Go lines as an issue, I will drop this as well for straightforward imperative approach. I am going to replace context mocking with a monkey patching. I am not fan at all, I would rather pass an interface into the checks and actually move away from this in the whole project. But okay, let's see how that looks like. This checks two of your major concerns, but I need to stop here. Let me explain and elaborate on both mocking and patch organization because there is a reason why it looks this way.

I see mocked unit tests as a great complementary to the actual boot tests we do in real environments. The workflow allows me to quickly validate and find typos or obvious errors before pushing and running the whole suite. Any integration tests we would implement tho would never reach to the massive 1300 tests we have currently which possibly all could be boot tested at some point. Unit tests allow me to create mocks for various edge cases that appear as bugs after running the real tests easily.

Please realize that one run of all the boot tests is approximately 3-5 hours lately and to break down everything into "logical steps that keep the code working" is very impractical for dozen of checks we have today. I would be waiting for weeks in order to achieve that every commit actually works. That is why I intentionally have chosen the "big bang" approach, why I heavily relied on extensive mocking and unit testing to figure out the code first and then "debug" the problems and edge cases on the real boot tests. You can see in the history of this PR about 20-30 pushes, many test failures and three found bugs in the original shell script. I had to do it this way and debug _multiple host checks at the same time_ and yet it took over a month to get to the green state. If I was doing it one check after another, I would be still working on the third one I can tell you.

Having said that, I expected reviewers to open the [check-host-config.sh](https://github.com/osbuild/images/blob/main/test/scripts/check-host-config.sh) in a separate browser tab without telling them, this must have been painful to experience. Sorry about that, when I was reviewing the PR on github I sort of automatically did that but I should let you know. So rather than creating a fake history, and I had my reasons for not doing that, I would ask for patience with review here.

Of course, I am open to breaking the big commit that only adds new code into individual commits if that makes things easier on your side.

cmd/check-host-config/wait.go
```diff
@@ -0,0 +1,107 @@
...
+
+		if ctx.Err() != nil {
+			return fmt.Errorf("timeout before waiting for running state: %w", ctx.Err())
+		}
+
+		cmdStr := "systemctl is-system-running"
+		if logSometimes != nil {
+			logSometimes.Printf("Executing: %s\n", cmdStr)
+		}
+		stdout, stderr, err := check.Exec("systemctl", "is-system-running")
```
supakeen: Does `check.Exec` need to take the context? Otherwise I *think* if the command blocks on something we'll get stuck?

cmd/check-host-config/check/utils.go
```diff
@@ -0,0 +1,86 @@
...
+// OSRelease contains parsed fields from /etc/os-release
+type OSRelease struct {
+	ID           string
+	VersionID    string
+	Version      string
+	MajorVersion int // Extracted major version from VersionID (e.g., 9 from "9.0")
+}
+
+// ParseOSRelease reads and parses /etc/os-release file, extracting
+// the id, version_id, and version fields.
```
supakeen: I'm reasonably sure we have something that's at least similar in `images` itself, perhaps that can be re-used?

cmd/check-host-config/check/osfuncs.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+var ExecCommand func(name string, arg ...string) *exec.Cmd = exec.Command
+
+var Exec func(name string, arg ...string) ([]byte, []byte, error) = func(name string, arg ...string) ([]byte, []byte, error) {
```
supakeen: Let's take a context here so we can cancel stuck commands by using `exec.CommandContext`; potentially we also want to be a session leader to make that work in all cases, but if we don't hit those cases we can skip that for now.

supakeen (commented):
I've added some comments; it's a lot of code so I didn't go through all the checks.

I really want this so we can have less modifications to our images, and I prefer it much over the shell script (which has lead to plenty of issues in the past including silently ignoring stuff).

The approach you've taken here is that you overengineered it and pulled in a lot of stuff that's very forward-looking for problems that do not yet exist. This makes it hard to review and hard for me to suppress my YAGNI. In this case there's no point in blocking the PR on it because (to me) the benefits outweigh the extra maintenance and mental overhead.

Happy to approve after we either dismiss or implement my comments :)

achilleas-k:
Merge conflict from the recent change, but should be simple to resolve.

lzap:
> Merge conflict from the recent change, but should be simple to resolve.

Yeah I plan to further simplify the code, will do tomorrow. Also was thinking of introducing `testify` package which should save a lot of lines too.

cmd/check-host-config/wait.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		if ctx.Err() != nil {
+			return fmt.Errorf("timeout before waiting for running state: %w", ctx.Err())
+		}
+
+		cmdStr := "systemctl is-system-running"
+		if logSometimes != nil {
+			logSometimes.Printf("Executing: %s\n", cmdStr)
+		}
```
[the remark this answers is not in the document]
lzap: The problem is this spills way to many logs, because the mockable default `Exec` implementation performs logging of every single command. While this is pretty useful for the actual checks to keep things simple (no extra logging statements are necessary), it is counter productive here.

You know what, let me ditch this code, instead of 500ms let's just call `systemctl` every 15 seconds and leave logging as is. The default wait time is 15 minutes which is how many 220 logs in total in the worst case? I think CICD will survive 220 lines. Simplicity is what I am going for now.

cmd/check-host-config/check/utils.go
lzap: I have found `ReadOSReleaseFromTree` and I will use it instead, nice.

cmd/check-host-config/wait.go
```diff
@@ -0,0 +1,82 @@
...
+			return fmt.Errorf("system is at non-running state: %q", stdout)
+		}
+	}
+}
+
+// listBadUnits returns a space-separated string of systemd units that are
+// still in the activating state. It calls systemctl list-units to get the list.
+// This is only used in case of timeout to help with debugging.
+func listBadUnits() string {
+	stdout, _, _, err := check.Exec("systemctl", "list-units", "--state=activating,failed", "--plain", "--no-legend", "--no-pager")
```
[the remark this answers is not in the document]
lzap: Yeah that is RHEL9+ only.

cmd/check-host-config/check/files.go
```diff
@@ -0,0 +1,40 @@
...
+	// Note that this test only checks for the existance of the filesystem
+	// customizatons target path not the content. For the simple case when
+	// "data" is provided we could check but for the "uri" case we do not
+	// know the content as the file usually comes from the host.  The
+	// existing testing framework makes the content check difficult, so we
+	// settle for this for now. There is an alternative approach in
+	// https://github.com/osbuild/images/pull/1157/commits/7784f3dc6b435fa03951263e48ea7cfca84c2ebd
+	// that may eventually be considered that is more direct and runs
+	// runs locally but different from the existing paradigm so it
+	// needs further discussion.
```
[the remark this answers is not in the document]
lzap: Yes but I aim for no changes in this PR, except fixing obvious bugs. There is much more we can do now pretty easily.

supakeen (approved):

PR #2045, opened 2025-11-26
[RFC] test: add/use a pytest wrapper for the build/boot tests

mvo5:
[draft as this is potentially controversial]

This is a tiny wrapper around the build/boot tests scripts to run them via pytest. Why another layer of indirection you ask?

1. it makes it easy to enumerate all tests one can run via `pytest --collect-only` (708!)
2. it makes it easy to run selected tests: `pytest -k rhel-10.2`
3. it provides a uniform way to that is the same for all pytest projects
4. with a sufficiently small number of tests (via -k) we could do --parallel
5. it abstracts away from the underlying scripts, we can (potentially) change how we run this without changing the higher layers
6. (new) the (py)tests are split into `build_only` and `build_and_boot` testcases now, this makes it trivial to use `pytest --collect-only -k build_only` (or `-k build_and_boot) and see that we do 443 build tests and 265 boot tests.

Now that tests are easy to run locally it seems like a really nice feature.

[edit: Another thing I would like to do is make the tests that are skipped currently more visible, either via pytest.skip() when ensure_can_run_qemu_test() returns can exception or by having a declaratve skip-test list so that we can easily see what is currently skipped for what reason, right now the set of boot tests that we claim to do is a bit bigger than what we actually do (as we do not e.g. minimal-raw as it has no cloud-init]

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -229,7 +229,7 @@ jobs:
       - name: Testing imgtestlib and test scripts
         run: |
           pip install .
-          python3 -m pytest -v
+          python3 -m pytest -v -k "not images_integration"
 
   python-lint:
     name: "🐍 Lint (test scripts)"
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,0 +1,4 @@
+def pytest_configure(config):
+    config.addinivalue_line(
+        "markers", "images_integration"
+    )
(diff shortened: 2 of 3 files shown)


supakeen:
From the title I was kinda against this but after seeing this in action I feel it's nice. There's a little bug though:

```
FAILED test/test_build_integration.py::test_build_and_boot[rhel-10.0-x86_64-image-installer-jq-only] - NameError: name 'image_type' is not defined
```

PR #2048, opened 2025-11-27
test: enable openscap testing in qemu

mvo5:
We were skipping the opscap tests before when doing qemu based testing. But this is possible now with this PR. 

[edit: this skips openscap on fedora now because of the sssd bug, c.f. https://github.com/osbuild/images/issues/2049]


supakeen (dismissed):

kingsleyzissou:
> Conflict. I'm also unsure if (like FIPS) OpenSCAP is actually a supported thing in Fedora. It doesn't surprise me that it doesn't work.

There are 3 (maybe 4) profiles for oscap on fedora. I think we've just chosen the strictest one

mvo5:
This will need https://github.com/osbuild/images/pull/2051 first I guess (well, or a full variant of it that actually update rhel-10.1)

[edit. manually applied it, lets see if that helps]

thozza (dismissed):
LGTM, Thanks.

supakeen (dismissed):

kingsleyzissou (dismissed):

mvo5:
Sorry, had to force push because of conflicts

thozza (approved):

kingsleyzissou (approved):

PR #2054, opened 2025-12-01
fedora: write `/etc/sysconfig/kernel` (HMS-9736)

supakeen:
When the `/etc/sysconfig/kernel` file does not exist with the appropriate values then the default kernel does not get updated when a new kernel is installed with DNF.

Normally this file is written by either Kiwi [1] or Anaconda [2] hence we should write it as well. We already wrote this for RHEL-likes.

[1]: https://pagure.io/fedora-kiwi-descriptions/blob/rawhide/f/root/etc/sysconfig/kernel
[2]: https://github.com/rhinstaller/anaconda/blob/8ea01bacf57df26dcc14d736218bc02bbb07f766/pyanaconda/modules/storage/bootloader/utils.py#L148


mvo5 (dismissed):
Thank you, lets go with it as it fixes a real issue and kiwi is doing the same (anaconda is smarter though and tries to match the kernel to the package name).

thozza (dismissed):
ACK, thanks!

mvo5 (approved):

thozza (approved):

PR #2057, opened 2025-12-02
test: set the correct tag for openstack runners

ondrejbudai:
schutzbot uses tags for limiting the number of concurrent PRs on different platforms. This is needed because we have different quotas on AWS vs. OpenStack. See
https://github.com/osbuild/gitlab-ci-config/blob/main/config.toml.tpl

This commits tags the rhos01 runners with the proper terraform/openstack tag which should prevent hitting the openstack quota which is way lower than the aws one.


thozza (dismissed):
Thank you.

ondrejbudai:
fixed pylint

PR #2059, opened 2025-12-02
drop: `minimal` (HMS-9738)

supakeen:
We have had experimental support for `minimal` images. These images only install the packages defined by their architecture and platform.

This was added when we were experimenting with blueprint-defined distributions. Nowadays we offer the functionality to do so through our YAML definitions.

---

See also: https://github.com/osbuild/blueprint/pull/39


mvo5 (dismissed):
I'm in favor of removing this as well

test/config-list.json
```diff
@@ -299,28 +299,6 @@
...
-    "filters": {
-      "distros": [
-        "fedora*"
-      ],
-      "image-types": [
-        "server-qcow2"
-      ]
-    }
-  },
   {
```
lzap: Yes, smaller test matrix. I am hoping for more.

lzap (dismissed):
Absolutely, can we assume now that all images have ssh for testing?

PR #2060, opened 2025-12-03
rhel{9,10}: drop `initscripts-rename-device` (HMS-9515)

supakeen:
`initscript-rename-device` gets pulled into images as an indirect dependency of a package in the core group. This package adds `/usr/lib/udev/rules.d/60-net.rules` and `/usr/lib/udev/rename_device`, which tries to rename NICs based on `HWADDR=` in ifcfg files.

On Azure VMs with SR-IOV, both `hv_netvsc` and the VF NIC share the same MAC. This makes rename_device try to rename both to the same name (eth0). This fails with “File exists”. The failure aborts udev processing for the VF, so properties like `BBNM_UNMANAGED=1` don’t persist across reboots


supakeen:
> Code looks fine and explanation makes sense, I have not tested this on azure though

Luckily the test suite should (I think). Explanation is by @achilleas-k (it comes from the JIRA).

PR #2063, opened 2025-12-03
scripts/build-image: use libosinfo to detect installer (HMS-9801)

lzap:
When using virt-install --location disk.iso the libosinfo library is used to detect operating system and path to kernel and initramdisk on the ISO image. This is done by comparing disk label against its database of known OSes.

While most of the images I tested do work, to fully ensure this is the case and to prevent regressions, let's run this tool after image is built in case it is an ISO.

For unknown OS versions, the database contains "unknown" entries which are still recognized as unreleased versions, so this should work even if our runners will have older version of the library.

---

Closes: https://issues.redhat.com/browse/HMS-9801


lzap:
Fixed linter, let's see if all ISOs have correct OS labels. Oh wait, since this is a change in `build-image`, I need to bump seed I guess? Maybe I should move it somewhere else?

achilleas-k (approved):

lzap:
Ok, added the following to `install_dependencies`:

```shell
# latest database for libosinfo-detect checks
OSDB_DIR=$(mktemp -d -t osinfodb.XXXXXX)/osinfo-db
cleanup() {
        test "$OSDB_DIR" && rm -rf "$OSDB_DIR"
}
trap cleanup EXIT INT TERM

git clone https://gitlab.com/libosinfo/osinfo-db "$OSDB_DIR"
pushd "$OSDB_DIR"
# install into $HOME/.config/osinfo
make install -j2
popd
```

supakeen:
> > The problem is that we do, indeed, not know what is in the ISO (in the case of `image-installer`). It could be called `Everything` or `Server`; but both would be wrong.
> 
> Yeah but does it matter for the ISO label and detected OS for virtualization? Typically, this is to pickup correct drivers and find kernel location, which is always the same for all variants. Let me know to set `Everything` instead of `Unknown` to see if this fix the tests.

What do you mean by this? I think we've already established that when using `virt-install` one should use the `--cdrom` option (and not `--location`), in the former case I don't think the kernel location matters?

> > Looking at that list I assume it also doesn't really know the Fedora KDE, or any Fedora spins ISOs or how does that work? Is it just about the ordering of the fields or do we need to match those exactly?
> 
> The heuristic behavior is a bit cryptic, I could not figure out how exactly it works from the code.

I'll try out some things :)

lzap:
I like the idea of relying on `fedora-unknown.xml` but it really seems it does not have any entry for `aarch64` for some reason. Maybe I can send them a patch?

```xml
<?xml version="1.0"?>
<libosinfo version="0.0.1">
  <!-- Licensed under the GNU General Public License version 2 or later.
     See http://www.gnu.org/licenses/ for a copy of the license text -->
  <os id="http://fedoraproject.org/fedora/unknown">
    <short-id>fedora-unknown</short-id>
    <name>Fedora</name>
    <family>linux</family>
    <distro>fedora</distro>
    <upgrades id="http://fedoraproject.org/fedora/rawhide"/>
    <derives-from id="http://fedoraproject.org/fedora/rawhide"/>
    <release-status>prerelease</release-status>
    <media arch="x86_64">
      <iso>
        <volume-id>Fedora-.*-dvd-x86_64-(4[2-9]|[5-9][0-9])</volume-id>
      </iso>
      <kernel>images/pxeboot/vmlinuz</kernel>
      <initrd>images/pxeboot/initrd.img</initrd>
    </media>
    <media arch="x86_64" live="true">
      <iso>
        <volume-id>Fedora-.*-Live-(4[2-9]|[5-9][0-9]).*</volume-id>
      </iso>
      <kernel>images/pxeboot/vmlinuz</kernel>
      <initrd>images/pxeboot/initrd.img</initrd>
    </media>
    <tree arch="x86_64">
      <treeinfo>
        <family>Fedora</family>
        <version>(4[2-9]|[5-9][0-9])</version>
        <arch>x86_64</arch>
      </treeinfo>
    </tree>
    <resources arch="all">
      <minimum>
        <n-cpus>1</n-cpus>
        <cpu>2000000000</cpu>
        <ram>2147483648</ram>
        <storage>16106127360</storage>
      </minimum>
      <recommended>
        <ram>4294967296</ram>
        <storage>21474836480</storage>
      </recommended>
    </resources>
  </os>
</libosinfo>
```

supakeen:
> I like the idea of relying on `fedora-unknown.xml` but it really seems it does not have any entry for `aarch64` for some reason. Maybe I can send them a patch?
> 
> ```
> <?xml version="1.0"?>
> <libosinfo version="0.0.1">
>   
>   <os id="http://fedoraproject.org/fedora/unknown">
>     <short-id>fedora-unknown</short-id>
>     <name>Fedora</name>
>     <family>linux</family>
>     <distro>fedora</distro>
>     <upgrades id="http://fedoraproject.org/fedora/rawhide"/>
>     <derives-from id="http://fedoraproject.org/fedora/rawhide"/>
>     <release-status>prerelease</release-status>
>     <media arch="x86_64">
>       <iso>
>         <volume-id>Fedora-.*-dvd-x86_64-(4[2-9]|[5-9][0-9])</volume-id>
>       </iso>
>       <kernel>images/pxeboot/vmlinuz</kernel>
>       <initrd>images/pxeboot/initrd.img</initrd>
>     </media>
>     <media arch="x86_64" live="true">
>       <iso>
>         <volume-id>Fedora-.*-Live-(4[2-9]|[5-9][0-9]).*</volume-id>
>       </iso>
>       <kernel>images/pxeboot/vmlinuz</kernel>
>       <initrd>images/pxeboot/initrd.img</initrd>
>     </media>
>     <tree arch="x86_64">
>       <treeinfo>
>         <family>Fedora</family>
>         <version>(4[2-9]|[5-9][0-9])</version>
>         <arch>x86_64</arch>
>       </treeinfo>
>     </tree>
>     <resources arch="all">
>       <minimum>
>         <n-cpus>1</n-cpus>
>         <cpu>2000000000</cpu>
>         <ram>2147483648</ram>
>         <storage>16106127360</storage>
>       </minimum>
>       <recommended>
>         <ram>4294967296</ram>
>         <storage>21474836480</storage>
>       </recommended>
>     </resources>
>   </os>
> </libosinfo>
> ```
> 
> Edit: Looks really like an oversight on their side because RHEL has all arches. Let me do a patch into the DB.

We'd also need to get new version in Fedora, CentOS and RHEL as they all have versions of about a year or two old which don't know any other things otherwise the end result for users is the same (but not for our tests?).

lzap:
> We'd also need to get new version in Fedora, CentOS and RHEL

Let's backport this into CentOS Stream 9 and 10 perhaps? I can do that. Maybe this database is subject of regular rebasing even.

lzap:
How about this, reworded the commit message, simplified it quite a bit. Let's just do this for intel and add a patch that will land later to fix this on all arches.

PR #2064, opened 2025-12-03
test: add (basic) qemu based pxe netboot testing (HMS-9573)

lzap:
This is continuation of @mvo5 work from

The whole idea behind this simplification is that I strongly feel that we should not be testing if PXE/shim/grub/kernel work, that is something that is covered by the RHEL group and I know for fact there are automated tests for this (BIOS/EFI/HTTP-EFI).

Therefore I propose to simply use direct kernel boot and validate that the actual image configuration successfully mounts the image root via both HTTP or concatenated ISO. The initial version of my draft PR only contains test for the former, but the test case will be done also for the latter just with different kernel command line options.

The question is how to test and run host check script. My understanding is that this live OS runs from memory and support some customizations, notably SSH can be put there. So I propose to use SSH the same way as for other images.

Closes: https://issues.redhat.com/browse/HMS-9573

Edit: Applied WIP label but tests are running?!?

### DIFF
--- a/test/scripts/boot-image
+++ b/test/scripts/boot-image
@@ -307,60 +307,62 @@ def _boot_qemu_iso(arch, installer_iso_path, config_file, privkey_path):
                 qemu_cmd_scp_and_run(vm, cmd, privkey_path)
 
 
-def boot_qemu_pxe(arch, pxe_tar_path, config):
+def boot_qemu_pxe(arch, pxe_tar_path):
     with contextlib.ExitStack() as cm:
+        # unpack the tar and create a combined image
         tmpdir = cm.enter_context(TemporaryDirectory(dir="/var/tmp"))
-        # setup tftp boot dir
-        tftp_dir = pathlib.Path(tmpdir) / "tftp"
         subprocess.check_call(
-            ["cp", "-a", "/usr/share/syslinux/", os.fspath(tftp_dir)])
+            ["tar", "-C", tmpdir, "-x", "-f", pxe_tar_path])
         subprocess.check_call(
-            ["tar", "-C", tftp_dir, "-x","-f", pxe_tar_path])
+            "echo rootfs.img | cpio -c --quiet -L -o > rootfs.cpio", shell=True, cwd=tmpdir)
+        subprocess.check_call(
+            "cat initrd.img rootfs.cpio > combined.img", shell=True, cwd=tmpdir)
+
+        # start an HTTP server to serve the rootfs.img
         http_port = get_free_port()
-        # we need to serve the dir
-        cmd = subprocess.Popen([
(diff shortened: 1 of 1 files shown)


test/scripts/boot-image
```diff
@@ -304,6 +305,61 @@ def _boot_qemu_iso(arch, installer_iso_path, config_file, privkey_path):
                 qemu_cmd_scp_and_run(vm, cmd, privkey_path)
 
 
+def boot_qemu_pxe(arch, pxe_tar_path, config_file):
+    with contextlib.ExitStack() as cm:
+        # unpack the tar and create a combined image
+        tmpdir = cm.enter_context(TemporaryDirectory(dir="/var/tmp"))
+        root_dir = pathlib.Path(tmpdir)
+        subprocess.check_call(
+            ["tar", "-C", root_dir, "-x", "-f", pxe_tar_path])
```
supakeen: We use `tmpdir` in the other functions and suddenly `root_dir` here even though they refer to the same path. I understand why you convert it to a `pathlib.Path` (it's easier later on) but why not use `root_dir` everywhere, or `tmpdir` everywhere right now it flip-flops between the two (`root_dir` here, `tmpdir` on the next two lines `root_dir` on line 325, etc).

As a nitpick: `root_dir` vs `tmpdir` how about `rootdir`, `tmpdir` or `root_dir`, `tmp_dir`? ;)

supakeen (commented):
Generally OK though I'm a bit on the fence of calling this a 'PXE' test since we're not doing anything PXE except the image type? I understand the reasoning but perhaps kernel boot or such is more descriptive?

This also doesn't verify very well that we have all necessary files to set up the various network boot options in our `pxe-tar-xz` image type but that can be separate asserts I guess in a follow-up.

bcl:
> Generally OK though I'm a bit on the fence of calling this a 'PXE' test since we're not doing anything PXE except the image type? I understand the reasoning but perhaps kernel boot or such is more descriptive?

I think it's fine to call it PXE. It's testing the rootfs created by pxe-tar-xz which is the most important part.

> This also doesn't verify very well that we have all necessary files to set up the various network boot options in our `pxe-tar-xz` image type but that can be separate asserts I guess in a follow-up.

The tar purposfully doesn't contain everything needed for all the possible ways to boot it. Since there are so many different possible ways to boot things it is assumed the user knows how to setup their PXE service. We provide some help, like the grub.cfg example that needs to be edited, and the EFI tree which can be used to setup UEFI PXE booting.

So I think it's fine to just boot the kernel+initrd+rootfs.img and make sure it functions. We could add a UEFI boot test without much more work but it really wouldn't add any coverage to what we're producing, you'd really just be testing qemu and grub2 UEFI binaries in addition to kernel, etc.

supakeen (approved):
Ok then :)

PR #2067, opened 2025-12-04
boot-image: unregister in rhel network installs again

mvo5:
This commit ensures that rhel systems installed via the
unattended kickstart get unregistered after the install
again. This avoids unnecessary test systems from cluttering
the inventory.

C.f. https://github.com/osbuild/images/pull/2044/files#r2576962482


lzap (dismissed):

lzap (approved):

supakeen (approved):
So we'll only leak if the machine fails to install (e.g. before `%post`)?

PR #2068, opened 2025-12-04
pkg/depsolvednf: decouple Solver methods from API wire-format implementation to prepare for v2 Solver API (HMS-9062)

thozza:
## Overview

This PR refactors the `pkg/depsolvednf` package to introduce an `apiHandler` interface that abstracts the osbuild-depsolve-dnf API implementation. The goal is to enable easier iteration on Solver API versions by cleanly separating API wire-format handling (request building and response parsing) from the core Solver logic. The current v1 API has been extracted into a dedicated `v1Handler` implementation. This is preparatory work for adding a v2 Solver API in a follow-up PR.

## Architectural Changes

The key design decision is the introduction of the `apiHandler` interface that defines three pairs of methods for Request building and Response parsing.

A `solverConfig` struct is used to pass solver configuration to handlers without coupling them to the `Solver` type directly. The `run()` function now accepts raw `[]byte` request data instead of a typed `*Request` struct, making it agnostic to the API version being used.

This architecture allows new API versions to be added by simply implementing the `apiHandler` interface, without modifying the Solver methods themselves.

## Key Changes

- **New `apiHandler` interface** (`api.go`): Defines the contract for API version handlers with methods for building requests and parsing responses
- **New `v1Handler` implementation** (`v1.go`): Extracts all v1 API structs, request building, and response parsing logic from the main file
- **Migrated Solver methods**: `Depsolve()`, `FetchMetadata()`, and `SearchMetadata()` now use `activeHandler` instead of inline API logic
- **Refactored `run()` function**: Now accepts raw `[]byte` instead of a typed request struct
- **Replaced `Request.Hash()` with `hashRequest()`**: Caching now uses SHA256 of raw request bytes instead of custom field-based hashing
- **Extracted `applyRHSMSecrets()` helper**: The RHSM secrets override logic is now a separate, testable function

## Breaking Changes / Backward Compatibility

This PR is fully backward compatible. The external API surface of the `depsolvednf` package remains unchanged. Internal types have been moved/removed, but these were not exported. The cache hash computation has changed, which will cause a one-time cache invalidation for existing caches.

## Testing

- Added comprehensive unit tests for v1 handler request building methods (`TestV1HandlerMakeDepsolveRequest`, `TestV1HandlerMakeDumpRequest`, `TestV1HandlerMakeSearchRequest`) verifying JSON output matches expected format
- Added unit tests for `Solver.FetchMetadata()` and `Solver.SearchMetadata()` to ensure the refactoring doesn't break existing behavior
- Added unit tests for the newly extracted `applyRHSMSecrets()` function covering various RHSM/mTLS/plain repo combinations
- Existing integration tests continue to pass, now indirectly testing the v1 handler's response parsing through the Solver methods

### DIFF
--- a/pkg/depsolvednf/depsolvednf.go
+++ b/pkg/depsolvednf/depsolvednf.go
@@ -297,18 +297,8 @@ func (s *Solver) Depsolve(pkgSets []rpmmd.PackageSet, sbomType sbom.StandardType
 		return nil, err
 	}
 
-	// Apply RHSM secrets to packages.
-	// The handler sets "org.osbuild.mtls" for repos with SSLClientKey,
-	// but RHSM repos need "org.osbuild.rhsm" instead.
-	rhsmReposMap := make(map[string]bool)
-	for _, repo := range allRepos {
-		rhsmReposMap[repo.Hash()] = repo.RHSM
-	}
-	for i := range resultRaw.Packages {
-		if rhsmReposMap[resultRaw.Packages[i].RepoID] {
-			resultRaw.Packages[i].Secrets = "org.osbuild.rhsm"
-		}
-	}
+	// Apply RHSM secrets to packages from RHSM repos.
+	applyRHSMSecrets(resultRaw.Packages, allRepos)
 
 	var sbomDoc *sbom.Document
 	if sbomType != sbom.StandardTypeNone {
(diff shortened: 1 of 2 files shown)


achilleas-k (approved):
Wow, that's a lot!  Very clean PR, couldn't find anything to complain about.

LGTM

PR #2069, opened 2025-12-04
test: cross arch build/boot smoke test for ppc64le,s390x

mvo5:
test: run cross arch smoke test on ppc64le, s390x

This commit adds a minimal smoke test for centos-10/qcow2 that
will create a qcow2 and boot it. It demos how we could do more
cross arch testing and keeps the qemu code tested.

We need to decide if we want to do more tests like this, its
a trade-off. OTOH its nice to have (some) assurance that our
images boot (we did break s390/ppc64 partition tables in the
past by accident). OTOH its a bit of a pain when something fails
to figure out of it is failing because of some qemu incompatibilites
or because there is a real issue in the image.

Having this at least for local testing/validation is probably
useful though.

Note that the s390x test needs qemu-user >= 7.2-rc2 so this will
not run currently (I did run it on my local machine and it works
but the kdump service takes > 5min to startup).

---

vmtest: add cross arch boot support for s390x/ppc64le

This adds basic support to boot ppc64le/s390x machine. This
allows to do:
```
$ ./test/scripts/build-image --arch ppc64le centos-10 qcow2 cfgpath
$ ./test/scripts/boot-image ./build/centos_10-ppc64le-qcow2-empty
```
to easily boot test machines that are not easily available otherwise.

---

build-image: add support for (optional) --arch to build

This is useful to do basic testing for architectures like s390x
or ppc64le that are not available easily (or even for a quick
local aarch64 test).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -278,3 +278,41 @@ jobs:
         # yq will catch issues that yamllint will not, like duplicate anchros
         run: |
           find . "(" -iname "*.yaml" -or -iname "*.yml" ")" -exec yq . {} \+ > /dev/null
+
+  cross-arch-integration-test:
+    strategy:
+      matrix:
+        arch:
+          - ppc64le
+          - s390x
+      fail-fast: false  # if one fails, keep the other(s) running
+    name: "Cross-arch qemu based image build/boot smoke test"
+    runs-on: ubuntu-24.04
+    env:
+      # workaround for expired cert at source of indirect dependency
+      # (go.opencensus.io/trace)
+      GOPROXY: "https://proxy.golang.org|direct"
+      GOFLAGS: "-buildvcs=false"
+    container:
+      image: registry.fedoraproject.org/fedora:latest
+      options: "--privileged"
+    steps:
+      - name: Install build and test dependencies
+        run: dnf -y install python3-pytest podman go btrfs-progs-devel device-mapper-devel gpgme-devel python3-pip qemu-system-s390x-core qemu-system-ppc64 qemu-user-static osbuild osbuild-depsolve-dnf cloud-utils
+      - name: Manually start qemu-user-static
+        run: |
+          # mostly for debugging
+          mount | grep binfmt_misc || true
+          ls /proc/sys/fs/binfmt_misc/
+          sudo mount -t binfmt_misc none /proc/sys/fs/binfmt_misc
+          ls /proc/sys/fs/binfmt_misc/
+          sudo SYSTEMD_LOG_LEVEL=debug /usr/lib/systemd/systemd-binfmt
+          ls /proc/sys/fs/binfmt_misc/
+          sudo SYSTEMD_LOG_LEVEL=debug /usr/lib/systemd/systemd-binfmt /usr/lib/binfmt.d/qemu-*.conf
+          ls /proc/sys/fs/binfmt_misc/
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v6
+      - name: Cross arch integration test
+        run: |
+          pip install .
+          sudo -E pytest -rs -k ${{ matrix.arch }} -s -v ./test/test_cross_arch_integration.py
--- a/test/test_cross_arch_integration.py
+++ b/test/test_cross_arch_integration.py
@@ -1,0 +1,32 @@
+import os
+import platform
+import re
+import shutil
+import subprocess
+
+import pytest
+import scripts.imgtestlib as testlib
+
+if os.getuid() != 0:
+    pytest.skip(reason="need root to build the images", allow_module_level=True)
+
+
+@pytest.mark.images_integration
+@pytest.mark.parametrize("arch", ["ppc64le", "s390x"])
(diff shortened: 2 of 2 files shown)


bcl (approved):
Other than that one question this looks good.

test/scripts/build-image
```diff
@@ -26,7 +26,7 @@ def main():
         print(json.dumps(config, indent=2))
         config_name = config["name"]
 
-    testlib.runcmd(["go", "build", "-o", "./bin/build", "./cmd/build"])
+    testlib.runcmd(["go", "build", "-buildvcs=false", "-o", "./bin/build", "./cmd/build"])
```
thozza: Do you remember what the actual error was? This seems weird, since the binary is being built inside a Git repository checkout, so things should work. And the `go help build` says that the command should error out only when using `-buildvcs=true`. The default is `-buildvcs=auto`.

thozza (changes requested):
Thanks for this; I think that it makes sense.

That said, the default behavior of some of the scripts changes AFAICT, and I am not convinced that it is necessarily intentional. Requesting changes until this is sorted out.

test/scripts/build-image
```diff
@@ -30,14 +31,17 @@ def main():
 
     cmd = ["sudo", "-E", "./bin/build", "--output", "./build",
            "--distro", distro, "--type", image_type, "--config", config_path]
+    arch = os.uname().machine
+    if args.arch:
```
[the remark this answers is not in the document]
mvo5: Yes, excellent catch, this is handled now in build.go to only use UseBootstrapContainer if arch != currentArch

vmtest/vm.py
```diff
@@ -172,23 +171,43 @@ def __del__(self):
...
+                "qemu-system-ppc64",
+                "-machine", "pseries",
+                "-smp", "2",
+            ] + virtio_scsi_hd
+        elif self._arch == "s390x":
+            qemu_cmdline = [
+                "qemu-system-s390x",
+                "-machine", "s390-ccw-virtio",
+                "-smp", "2",
+                # sepcial disk setup
```
lzap: Special? I guess.

supakeen (approved):

cmd/build/main.go
```diff
@@ -64,8 +65,10 @@ func run() error {
 		return fmt.Errorf("invalid or unsupported distribution: %q", distroName)
 	}
 
-	archName := arch.Current().String()
-	arch, err := distribution.GetArch(archName)
+	if archName == "" {
+		archName = arch.Current().String()
+	}
+	archi, err := distribution.GetArch(archName)
```
thozza: Nitpick: I don't have any problem with the rename of `arch` -> `archi`, just wanted to point it out in case this was not intentional.

thozza (approved):
Amazing, thank you a lot!

cmd/build/main.go
mvo5: Its needed because we also import `arch` so if we have `arch` there the `arch` module is shadowed (and a few lines below iirc this code access the arch module now).

vmtest/vm.py
```diff
@@ -197,9 +216,11 @@ def _gen_qemu_cmdline(self, snapshot, use_ovmf):
             "-m", self._memory,
             "-serial", "stdio",
             "-monitor", "none",
+            "-device", f"{virtio_net_device},netdev=net.0,id=net.0",
             "-netdev", f"user,id=net.0,hostfwd=tcp::{self._ssh_port}-:22",
-            "-device", "e1000,netdev=net.0",
             "-qmp", f"unix:{self._qmp_socket},server,nowait",
+            # boot
+            "-drive", f"file={self._img},if=none,id=disk0,format=qcow2",
```
lzap: This hunk sneaked into `main` somehow and this does not appear to be correct. It renders as:

```
['qemu-system-x86_64', '-M', 'accel=kvm', '-cpu', 'host', '-device', 'virtio-scsi-pci,id=scsi', '-device', 'scsi-hd,drive=disk0', '-bios', '/usr/share/OVMF/OVMF_CODE.fd', '-m', '2048', '-serial', 'stdio', '-monitor', 'none', '-device', 'virtio-net-pci,netdev=net.0,id=net.0', '-netdev', 'user,id=net.0,hostfwd=tcp::49925-:22', '-qmp', 'unix:/tmp/vmtest-v614uhg7-disk.img/qmp.socket,server,nowait', '-drive', 'file=/var/tmp/tmpp40bxujp/disk.img,if=none,id=disk0,format=qcow2', '-nographic', '-cdrom', PosixPath('/var/tmp/tmpdk5gg6b6/netinst.iso'), PosixPath('/var/tmp/tmpp40bxujp/disk.img')]
```

Which ends with QEMU complaining about raw image not being qcow2. That can be fixed, but then the issue is the image is passed twice on the command line - once via `-device` and once via argument. When I remove the argument, then QEMU does not boot there is some problem with SCSI setup.

Another problem is few lines below when `-cdrom` is appended with Path object, but that is an easy fix with `str`.

I will look into this later next week.

PR #2071, opened 2025-12-05
pkg/manifest: remove `--server` option from rhc connect (HMS-9809)

kingsleyzissou:
The `rhc` command has a non-zero exit code for undefined flags. The `--server` flag was removed from the `rhc` cli causing the command to fail, resulting in the image not being auto-registered.

This changed with: https://github.com/RedHatInsights/rhc/pull/230

### DIFF
--- a/pkg/manifest/subscription.go
+++ b/pkg/manifest/subscription.go
@@ -128,19 +128,18 @@ func subscriptionService(
 	}
 
 	var commands []string
+	if subscriptionOptions.ServerUrl != "" {
+		commands = append(commands, fmt.Sprintf("/usr/sbin/subscription-manager config --server.hostname %s", shutil.Quote(subscriptionOptions.ServerUrl)))
+	}
+	if subscriptionOptions.Proxy != "" {
+		proxy := strings.Split(subscriptionOptions.Proxy, ":")
+		commands = append(commands, fmt.Sprintf("/usr/sbin/subscription-manager config --server.proxy_hostname %s", shutil.Quote(proxy[0])))
+		if len(proxy) == 2 {
+			commands = append(commands, fmt.Sprintf("/usr/sbin/subscription-manager config --server.proxy_port %s", shutil.Quote(proxy[1])))
+		}
+	}
+
 	if subscriptionOptions.Rhc {
-		if subscriptionOptions.ServerUrl != "" {
-			commands = append(commands, fmt.Sprintf("/usr/sbin/subscription-manager config --server.hostname %s", shutil.Quote(subscriptionOptions.ServerUrl)))
-
-		}
-		if subscriptionOptions.Proxy != "" {
-			proxy := strings.Split(subscriptionOptions.Proxy, ":")
-			commands = append(commands, fmt.Sprintf("/usr/sbin/subscription-manager config --server.proxy_hostname %s", shutil.Quote(proxy[0])))
-			if len(proxy) == 2 {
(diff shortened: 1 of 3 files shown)


supakeen (approved):
Does what it says (and we should do this).

lzap (approved):

thozza (commented):
While the change makes sense, it is very concerning that `rhc` is making such a breaking change.

osbuild-composer Cloud API defines `server_url` option for Subscription (https://github.com/osbuild/osbuild-composer/blob/c084c3d0f64e7cbb54ba443714b70224765a1343/internal/cloudapi/v2/openapi.v2.yml#L2297-L2300). Will we silently ignore it or turn it into an invalid request error?

Do you happen to know if all `rhc` versions in releases that we support have this change? I.e. RHEL-9. Should we instead check the `rhc` version in the package set and not set the option only with the versions that don't have it?

EDIT: I've just read the upstream PR. I guess not setting it at all is OK. The osbuild-composer API question will still need to be sorted out, although we control the use of it 100%, so it is probably not a big deal.

ondrejbudai (dismissed):
Marking this as "Request changes" for now as we need to make sure that we can actually register against different servers. In the frontend, we actually set this to the stage rhsm server:

https://github.com/osbuild/image-builder-frontend/blob/6d051ee8d34db6100817bc4209fad645461807bc/src/Components/CreateImageWizard/steps/Registration/components/ActivationKeysList.tsx#L84

I asked internally what's recommended approach after this change.

supakeen:
> Marking this as "Request changes" for now as we need to make sure that we can actually register against different servers. In the frontend, we actually set this to the stage rhsm server:
> 
> https://github.com/osbuild/image-builder-frontend/blob/6d051ee8d34db6100817bc4209fad645461807bc/src/Components/CreateImageWizard/steps/Registration/components/ActivationKeysList.tsx#L84
> 
> I asked internally what's recommended approach after this change.

When we discussed this previously we deemed it more important that production RHEL 10.1 systems auto-register than any (potentially) failings on stage. I think this reverses that and we'll keep 10.1 broken until we figure out what to do for stage?

croissanne:
I took the liberty of pushing a little update to support the server url again using `subcription-manager config`. @ondrejbudai wdyt? Not sure if we even need the "" check tbh.

pkg/manifest/subscription.go
```diff
@@ -128,10 +128,21 @@ func subscriptionService(
 	}
 
 	var commands []string
+	if subscriptionOptions.ServerUrl != "" {
+		commands = append(commands, fmt.Sprintf("/usr/sbin/subscription-manager config --server.hostname %s", shutil.Quote(subscriptionOptions.ServerUrl)))
+	}
```
lzap: Previously, both server hostname and base URL were set. These are two separate configuration options in rhsm:

```
  --server.hostname SERVER.HOSTNAME
                        Section: server, Name: hostname
...
  --server.proxy_hostname SERVER.PROXY_HOSTNAME
                        Section: server, Name: proxy_hostname
```

Now I only see hostname being set, this does not appear to be correct. We want to keep the ability to override both values, this is important for Satellite users who need to set both values.

What is good and step up from the previous version is that these default values are not hardcoded in the library now, this is good. Because typically this configuration is written by a specific Satellite-generated RPM package called `katello-ca.rpm`.

thozza (approved):
LGTM, but there is no unit test for the `--server.proxy_port` option being added to the config in `subscription_test.go`, which would be good to add... It can be a follow-up if you prefer it.

PR #2072, opened 2025-12-05
Add multi-resolve functions to resolvers

achilleas-k:
This PR adds convenience functions to the package, container, and ostree resolvers.  This is an old branch I had sitting around that was following up from the original introduction of the `manifestgen` package.  While working on [Add proxy support to image-builder-cli (HMS-9645)](https://issues.redhat.com/browse/HMS-9645), I remembered that I wanted to move some of the conveniences of `manifestgen` into the resolver packages themselves and this is part of that.

### Some background

The general flow of resolving content to generate a manifest is:
1. Instantiate manifest based on image type with customizations.
2. Get unresolved content descriptions (packages, containers, ostree commits) from the manifest, indexed by pipeline name.
3. Resolve each content type using an appropriate resolver, preserving the pipeline name indexing.
4. Serialize the manifest using the resolved content.

In (simplified) code, this looks like:
```go
// Instantiate manifest
manifest := imgType.Manifest(blueprint, imageOptions, repos, rngSeed)

// Get unresolved content descriptions
packageSets := manifest.GetPackageSetChains()
containers := manifest.GetContainerSourceSpecs()
commits := manifest.GetOSTreeSourceSpecs()

// Resolve each content type
solver := depsolvednf.NewSolver(...)
depsolvedSets := make(map[string]depsolvednf.DepsolveResult)
for pipeline, packageSet := range packageSets {
	depsolvedSets[pipeline] = solver.Depsolve(packageSet)
}

// ... repeat for containers and ostree commits

serializedManifest := manifest.Serialize(depsolvedSets, resolvedContainers, resolvedCommits)
```
NOTE: This isn't entirely accurate code, but is simplified to emphasise the important parts (e.g. error return values are dropped).

The [`manifestgen.Generator.Generate`](https://github.com/osbuild/images/blob/d6a6c9674060eae9ac1e4af91386d386898722b6/pkg/manifestgen/manifestgen.go#L144) function implements this flow, using the default configurations for each resolver and allows overriding them when needed.

### New stuff

This PR only provides the `ResolveAll()` convenience which performs the iteration over the unresolved content indexed by the pipeline names.  This moves some of the convenience provided by `manifestgen.Generator.Generate()` into the resolvers themselves, making them useful in cases where `manifestgen` isn't practical (e.g. in osbuild-composer where content is resolved on different workers).

### Container resolvers

For a while now, we've had two container resolver implementations, an async one (the original) and a blocking resolver (introduced in https://github.com/osbuild/images/pull/1176).  There was a short discussion on that PR about the differences between the two.  I summarised the consequences of switching to a blocking resolver [in this comment](https://github.com/osbuild/images/pull/1176#issuecomment-2621404432).  I'd like to prioritise consistency here over potential performance, but if we think we would benefit from parallel container resolution, we could go back to only having the async resolver but make its interface the same as the other resolvers (with `Resolve()` and `ResolveAll()`).

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -35,7 +35,6 @@ import (
 	"github.com/osbuild/images/pkg/distrofactory"
 	"github.com/osbuild/images/pkg/experimentalflags"
 	"github.com/osbuild/images/pkg/manifest"
-	"github.com/osbuild/images/pkg/manifestgen"
 	"github.com/osbuild/images/pkg/manifestgen/manifestmock"
 	"github.com/osbuild/images/pkg/ostree"
 	"github.com/osbuild/images/pkg/rhsm/facts"
@@ -318,7 +317,7 @@ func makeManifestJob(
 
 		var containerSpecs map[string][]container.Spec
 		if content["containers"] {
-			containerSpecs, err = manifestgen.DefaultContainerResolver(manifest.GetContainerSourceSpecs(), archName)
+			containerSpecs, err = container.NewResolver(archName).ResolveAll(manifest.GetContainerSourceSpecs())
 			if err != nil {
 				return fmt.Errorf("[%s] container resolution failed: %s", filename, err.Error())
 			}
--- a/pkg/manifestgen/manifestgen.go
+++ b/pkg/manifestgen/manifestgen.go
@@ -123,7 +123,9 @@ func New(reporegistry *reporegistry.RepoRegistry, opts *Options) (*Generator, er
 		mg.depsolve = DefaultDepsolve
 	}
 	if mg.containerResolver == nil {
-		mg.containerResolver = DefaultContainerResolver
+		mg.containerResolver = func(containerSources map[string][]container.SourceSpec, archName string) (map[string][]container.Spec, error) {
+			return container.NewBlockingResolver(archName).ResolveAll(containerSources)
+		}
 	}
 	if mg.commitResolver == nil {
 		mg.commitResolver = ostree.ResolveAll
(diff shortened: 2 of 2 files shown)


achilleas-k:
DRAFT: Will conflict with #2068 and this is lower priority.

pkg/depsolvednf/depsolvednf.go
```diff
@@ -252,6 +254,11 @@ func collectRepos(pkgSets []rpmmd.PackageSet) []rpmmd.RepoConfig {
 	return repos
 }
 
+// Set the SBOM type to generate with the depsolve.
+func (s *Solver) SBOMType(sbomType sbom.StandardType) {
+	s.sbomType = sbomType
+}
```
lzap: This should have been `SetSBOMType` because ultimately this makes impossible to create a getter: https://go.dev/doc/effective_go#Getters

pkg/container/resolver.go
```diff
@@ -100,6 +102,37 @@ func (r *asyncResolver) Finish() ([]Spec, error) {
...
+	if len(resSlice) != 1 {
+		return Spec{}, fmt.Errorf("unexpected results in container resolver: expected 1 but received %d", len(resSlice))
+	}
+
+	return resSlice[0], nil
+}
+
+func (r *asyncResolver) ResolveAll(sources map[string][]SourceSpec) (map[string][]Spec, error) {
+	containers := make(map[string][]Spec, len(sources))
+	for name, srcList := range sources {
```
lzap: Just a generic warning - map order when using `range` is not guaranteed in Go. I do not know if that is a problem tho, but I would probably drop a comment that sources can be resolved in random order. Hopefully, this will not change manifest JSON output but theoretically it could (e.g. we upgrade Go version, new map implementation is in place and some output in manifest goes into ordered slice).

pkg/container/resolver.go
```diff
@@ -160,3 +181,37 @@ func (r *blockingResolver) Finish() ([]Spec, error) {
...
+		return Spec{}, err
+	}
+
+	client.SetTLSVerify(source.TLSVerify)
+	client.SetArchitectureChoice(r.Arch)
+	if r.AuthFilePath != "" {
+		client.SetAuthFilePath(r.AuthFilePath)
+	}
+
+	return client.Resolve(context.TODO(), source.Name, source.Local)
```
lzap: Please add a proper context to this method and ideally to `ResolveAll` as well. That way, a timeout can be set globally for the whole operation. I makes sense here (the client) as well as for the async resolver where another `context.Background()` is used.

bcl (approved):
Looks fine to me, agree with @lzap's comments.

pkg/depsolvednf/depsolvednf_test.go
```diff
@@ -335,7 +334,9 @@ func TestSolverDepsolveAll(t *testing.T) {
 			assert := assert.New(t)
 			tc := testCases[tcName]
 
-			res, err := solver.DepsolveAll(tc.packageSets, tc.sbomType)
+			solver := NewSolver("platform:el9", "9", "x86_64", "rhel9.0", tmpdir)
+			solver.SBOMType(tc.sbomType)
```
thozza: I would love us to move away from this pattern of constructing an instance just to modify it in the next step. Ideally moving towards the functional options pattern.

achilleas-k: Yeah the functional options you showed me the other day looks really nice.

lzap (approved):
I would really like us utilizing context more, while it is fine adding more fire-grained timeouts you could have an overall deadline as well. Also, contexts are great for logging contexts, granted, this is more relevant for the service (per request, per org) than CLI but it has its own use there as well.

Anyways, all is fine.

bcl (approved):
Looks good!

PR #2073, opened 2025-12-08
wip: allow passing defs path

supakeen:
This is mostly a request for comments as everything here is work-in-progress. I'd really like it if we can pass the definitions path directly from `image-builder` without having to go through an environment variable but I don't *quite* like what I've done here.

Does someone know a better way to do this?

An open problem is `ParseID` which directly loads definitions without going through the loader's state.

### DIFF
--- a/pkg/distro/bootc/bootc.go
+++ b/pkg/distro/bootc/bootc.go
@@ -767,7 +767,7 @@ func newBootcDistroAfterIntrospect(archStr string, info *osinfo.Info, imgref, de
 		arch: archi,
 	}
 
-	distroYAML, err := defs.LoadDistroWithoutImageTypes("bootc-generic-1")
+	distroYAML, err := defs.LoadDistroWithoutImageTypes("", "bootc-generic-1")
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/distro/defs/id.go
+++ b/pkg/distro/defs/id.go
@@ -62,7 +62,8 @@ func matchAndNormalize(reStr, nameVer string) (string, error) {
 //
 // If no match is found it will "nil" and no error (
 func ParseID(nameVer string) (*distro.ID, error) {
-	distros, err := loadDistros()
+	// XXX defsDir?
+	distros, err := loadDistros("")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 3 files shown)


thozza:
> The thing I don't like about this is that a lot of functions now have an (only or first) argument that in many cases will be set to an empty string to get the default behaviour. I think that if the defs dir being set externally is the exception, then this should be reflected in the way the functions are called and the types are initialised. It should be simpler to get the common default behaviour than the exceptional one. So, instead of:
> 
> ```go
> d := distrofactory.NewDefault("")  // default case
> ```
> 
> ```go
> e := distrofactory.NewDefault(somevalue) // exceptional case
> ```
> 
> I find it's nicer to have
> 
> ```go
> d := distrofactory.NewDefault()  // default case
> ```
> 
> ```go
> e := distrofactory.NewDefault()
> if defsPath != "" {
>     e.DefsPath(defsPath)  // exceptional case
> }
> ```

I agree, but would like to offer two different approaches that may be nicer and more idiomatic in the long run.
Especially when setting the `DefsPath` needs to basically throw away all definitions loaded by the `NewDefault()`

## Functional options pattern

```go
package distrofactory

type option func(*Factory)

func WithDefsPath(path string) option {
    return func(c *Factory) {
        // Do the loading from the custom path here.
    }
}

func NewDefault(opts ...option) *Factory {
    f := getDefault()

    // In our case, the logic would be different, because we would probably construst a new Factory object from scratch from the different path.
    for _, opt := range opts {
        opt(f)
    }
    return f
}
```

```go
f := distrofactory.NewDefault()  // default case
```

```go
f := distrofactory.NewDefault(distrofactory.WithDefsPath("/dev/null"))  // exceptional case
```

IMHO, this approach is the most efficient, given how distro defs are loaded and how they work. This is also a standard when it comes to big libraries like cloud provider SDKs.

## Dysfunctional options pattern
(inspired by https://rednafi.com/go/dysfunctional_options_pattern/#dysfunctional-options-pattern)

```go
package distrofactory

func NewDefault() *Factory {
    f := getDefault()
    return f
}

func (f *Factory) WithDefsPath(path string) *Factory {
    f.doSomething(path)
    return f
}

```

```go
f := distrofactory.NewDefault()  // default case
```

```go
f := distrofactory.NewDefault().WithDefsPath("/dev/null")  // exceptional case
```

Not ideal, because defs would be loaded twice AFAICT.

supakeen:
Not stale.

PR #2078, opened 2025-12-10
images,distro: extract shared AnacondaInstallerBase and use it

mvo5:
We currently have a lot of duplication in our `Anaconda*Installer` code. We can reduce it by adding a shared `AnacondaInstallerBase` struct that is shared across our various installers. This way we can extract common helpers that init the common data struct.

It is pretty conservative and mostly shuffles code around, however there is a commit in the middle `image: consolidate more into initIsoTreePipeline()` that is slightly more aggressive but it does not change any of our manifests and looks safe to me - but I'm calling it out because the rest is really just shuffling things around (or removing dead code like the "ref" commit).

There is probably more potential for consolidation using this pattern but this seems a good starting point (and I'm especially pleased about the `manifestFor{legy,}ISO` cleanup (which was driving this refactor).

### DIFF
--- a/pkg/distro/bootc/bootc.go
+++ b/pkg/distro/bootc/bootc.go
@@ -439,68 +439,22 @@ func (t *BootcImageType) manifestForDisk(bp *blueprint.Blueprint, options distro
 	return &mf, nil, nil
 }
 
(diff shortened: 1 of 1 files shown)


bcl (approved):
Looks nice!

PR #2079, opened 2025-12-10
test: add `--keep-booted` option to `boot-image`

mvo5:
[*if* this is getting merged I will do a small followup to add something to test/README.md]

Add a convenience way to keep the booted image around when using `test/scripts/boot-image`. When running:
```console
$ ./test/scripts/boot-image --keep-booted <path>
```
the image will stay around after the check script was run so that the image can be inspected. It also provides a convenient output how to login into the system:
```
$ ./test/scripts/boot-image --keep-booted ./build/centos_10-x86_64-qcow2-empty/
...
***********************************
keeping the image ./build/centos_10-x86_64-qcow2-empty/qcow2/disk.qcow2 booted as requested, press enter or ctrl-c to stop
to connect run:
ssh -i /tmp/tmpxbfzwzlo/testkey -p 50635 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no osbuild@localhost
```

Especially useful with https://github.com/osbuild/images/pull/2069 when building/booting tests images for other arches.

Note that currently only qcow2 is supported but if this is considered useful we can expand it to the other image types.

For some (like ISO image-installer) this is tricky as we cannot inject credentials via cloud-init. But that is a more general problem that we would ideally tackle at some point.

### DIFF
--- a/test/scripts/boot-image
+++ b/test/scripts/boot-image
@@ -6,6 +6,7 @@ import os
 import pathlib
 import random
 import shutil
+import signal
 import string
 import subprocess
 import textwrap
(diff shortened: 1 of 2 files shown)


bcl (dismissed):
Looks good!

thozza (dismissed):
Nice, thank you!

thozza (approved):

PR #2080, opened 2025-12-10
distro: extend interface with getting image types (HMS-9642)

tkoscieln:
This is a draft proposal of how we can make accessing image types more straightforward. The usage could be following:
```
architectures := distro.ListArches()
for _, architecture := range architectures {
  imageTypes, err := distro.GetImageTypes(architecture)
  if err != nil {
	  return err
  }
  for _, imageType := range imageTypes {
	  fmt.Printf("Parent Arch: %v, Image Type: %v, Supported Blueprint Options: %v\n", imageType.Arch().Name(), imageType.Name(), imageType.SupportedBlueprintOptions())
  }
}
```

### DIFF
--- a/pkg/distro/distro.go
+++ b/pkg/distro/distro.go
@@ -183,3 +183,22 @@ func SeedFrom(p *int64) int64 {
 	}
 	return *p
 }
+
+// Helper: Returns all image types supported by the given distro for the specified architecture.
+func GetImageTypes(d Distro, archName string) ([]ImageType, error) {
+	arch, err := d.GetArch(archName)
+	if err != nil {
+		return nil, err
+	}
+
+	imageTypeNames := arch.ListImageTypes()
+	imageTypes := make([]ImageType, len(imageTypeNames))
+	for i, imageTypeName := range imageTypeNames {
+		imageType, err := arch.GetImageType(imageTypeName)
+		if err != nil {
+			return nil, err
+		}
+		imageTypes[i] = imageType
+	}
+	return imageTypes, nil
+}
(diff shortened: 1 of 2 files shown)


mvo5 (dismissed):
A quick drive-by comment - it looks like we have some duplicated code here, its probably a good idea to look into how we can avoid this, I think the fact that the three implementations are literally identical is a strong hint that something is not quite right here :) [of course maybe I'm missing something here, I did not look too deep, please correct me if I do]

ondrejbudai:
Alternatively, we can just have a "generic" helper - basically make the function live outside of the interface but instead take the `Distro` interface as an arg. Something like this:


```go
func GetAllImageTypesForArch(d Distro, archName string) ([]distro.ImageType, error) {
	arch, err := d.GetArch(archName)
	if err != nil {
		return nil, err
	}

	imageTypes := make([]distro.ImageType, len(arch.ListImageTypes()))
	imageTypeNames := arch.ListImageTypes()
	for _, imageTypeName := range imageTypeNames {
		imageType, err := arch.GetImageType(imageTypeName)
		if err != nil {
			return nil, err
		}
		imageTypes = append(imageTypes, imageType)
	}
	return imageTypes, nil
}
```

This way you just need one implementation, instead of one per each `Distro` interface implementation. Also, the interface is smaller which is always nice.

Kudos to @mvo5 for this suggestion.

lzap:
Right, my intention was not to doubt how this was done, clearly there is a lot of value from reusing `images` library. I was merely trying to understand how this works and suggesting that we probably want to take this functionality out of this interface.

lzap (dismissed):
@ondrejbudai @achilleas-k ?

supakeen (dismissed):
Could use a test.

tkoscieln:
I've added simple unit test

thozza (approved):
Thanks, LGTM.

I've added one nitpick, but that can be addressed also later if you prefer.

lzap (approved):

PR #2083, opened 2025-12-11
manifestgen: add missing options documentation

mvo5:
Quick drive-by to add missing docstrings and update requirements for using librepo.

### DIFF
--- a/pkg/osbuild/source.go
+++ b/pkg/osbuild/source.go
@@ -13,12 +13,13 @@ import (
 )
 
 // RpmDownloader specifies what backend to use for rpm downloads
-// Note that the librepo backend requires a newer osbuild.
+// Note that the librepo backend requires osbuild v138 from
+// 2025-01-15.
 type RpmDownloader uint64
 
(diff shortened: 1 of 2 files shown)


pkg/osbuild/source.go
```diff
@@ -13,12 +13,13 @@ import (
...
 // RpmDownloader specifies what backend to use for rpm downloads
-// Note that the librepo backend requires a newer osbuild.
+// Note that the librepo backend requires osbuild v138 from
+// 2025-01-15.
 type RpmDownloader uint64
 
 const (
-	RpmDownloaderCurl    = iota
-	RpmDownloaderLibrepo = iota
+	RpmDownloaderCurl = iota
```
mvo5: I do wonder if we should flip the order around and make librepo default now? The downside would be that we would potentially start generate manifests that are incompatible if we have systems that have a newer "images" library but an osbuild release that is older than a year.

supakeen (approved):

PR #1669, opened 2025-07-16
Add iso_boot_type

bcl:
Ready for review. Works, no manifest changes.

PR #2088, opened 2025-12-15
distrodefs: drop `use_syslinux` as it has no effect

mvo5:
This commit drops the `use_syslinux:` YAML key as it is not actually used and has no effect at the image manifest.

The `use_syslinux` was actually covered by https://github.com/osbuild/images/pull/1669 already so this yaml key is superfluous and confusing.

Thanks to Simon for asking about it that lead to this investigation.


supakeen:
If I get it correctly this is now set through `ISOBootType` in `installer_config`?

PR #2091, opened 2025-12-15
many: allow configuration of weak dependency installation in installer environments (HMS-9964)

supakeen:
This PR brings our `everything-network-installer` size difference from ~200 MiB to just 36 MiB. 16 of which is probably due to packages that have grown post-release, and 20 of it is that we leave `efiboot.img` lying around in the ISO. The last one I also plan to address.

```
many: install weak dependencies in installers

By default we always installed weak dependencies in installers that use
the Anaconda tree for their environment. This makes this configurable
in the installer configuration in YAML.

This is because I noticed that we can drop the installation of weak
dependencies in Fedora's network installer (at a minimum).
```

```
fedora: no weak deps on network installer

Drop the weak dependency installation on the network installer for
Fedora. This brings the filesize within 36 MiB to the currently
Lorax-produced media. Previously the difference was ~200 MiB.

The last bits will need to be addressed separately but this is already
much more manageable.
```


supakeen:
Had to force push since there were conflicts due to the `syslinux` PR #2089.

lzap (approved):
Ah conflicts again.

supakeen:
With the latest rebase it seems there are many real failures in the CI that are unrelated to artifacts touched by this PR.

I see CentOS qcow's failing on finding a disk in initramfs, and I see pxe-tar-xz failures in the 'disk.img is not in qcow2' format? @achilleas-k / @lzap just for your awareness.

lzap (approved):
I wish this was my weight over holidays...

PR #2093, opened 2025-12-15
manifest: add `set -e` to `bootc switch...` kickstart %post

mvo5:
The `bootc switch` command we use in our bootc installer ISOs stopped working [0] with the error:
```
error: Switching: Initializing storage: System not booted via bootc
```

However the install continue because while we do set `--erroronfail` the script is not setting `set -e`.

This commit fixes this and makes the bootc switch error in script a hard failure. This means that installs will error currently but its better to error than to install a system that is then not connected to any updates :(

With that the install looks like this:
```
Configuring installed system
........
Writing network configuration
.
Creating users
.....
Configuring addons
Generating initramfs
....
Storing configuration files and kickstarts
.
Running post-installation scripts
====================================================================
====================================================================
Error

  There was an error running the kickstart script at line 12.  This 
  error and installation will be aborted.  The details of this error
  
  error: Switching: Initializing storage: System not booted via boot
  

Press ENTER to exit: 
care:fatal==
```
which is not great but preferable (IMHO) over the current silent error.

[0] https://github.com/osbuild/image-builder-cli/issues/410


lzap (dismissed):

supakeen (dismissed):

thozza (dismissed):
LGTM, but the PR needs a rebase @mvo5

lzap (approved):

PR #2099, opened 2025-12-17
many: include legal and license files in ISO

supakeen:
This PR allows the copying of files from the Anaconda tree into the root of an ISO through our image definitions.

When looking through Fedora network install ISOs I noticed we're omitting the `Fedora-Legal-README.txt` and `LICENSE` files at the root of the ISO; here's what our ISO looks like:

```
€ ls image-builder-iso 
boot  EFI  images
```

While a network install ISO downloaded from Fedora shows:

```
€ sudo mount ~/tmp/Fedora-Everything-netinst-x86_64-43-1.6.iso iso
mount: /home/user/src/github.com/osbuild/image-builder-cli/iso: WARNING: source write-protected, mounted read-only.
€ ls iso 
boot  EFI  Fedora-Legal-README.txt  images  LICENSE
```

With this PR applied our ISO also includes these files at the appropriate places. Note that this is only applied to Fedora as RHEL ISOs do *not* seem to include these files:

```
€ sudo mount ~/tmp/rhel-10.0-x86_64-boot.iso iso 
mount: /home/user/src/github.com/osbuild/image-builder-cli/iso: WARNING: source write-protected, mounted read-only.
€ ls iso 
boot  EFI  images
```

Fedora Live ISOs are also out of scope as these also do not include these files.

---

```
many: copy files into ISO

Allow defining files that need to be copied to the root of an ISO. For
example current Fedora ISOs container a `Fedora-Legal-README.txt` and a
`LICENSE` file that we are not including.

This commit allows defining files to be copied, subsequent commits
implement copying and add the copying of files to Fedora's definitions.
```

```
manifest/aiit: copy iso files

Adjust the Anaconda ISO tree pipeline generator to also copy files that
are defined in the definitions. For example legal notices and/or license
files.
```

```
fedora: copy legal and license into iso

The Fedora network installer(s) as provided by Fedora include a legal
notice and license at the root of the ISO. Let's copy those files over
with the previously introduced capabilities.
```


lzap:
> I'm not the biggest fan but we use the same structure for `BootFiles` in platform :)

I was horrified when I saw it, but then it quickly grew on me when I saw the YAML. Sort of make sense and everyone should understand that.

PR #2101, opened 2025-12-18
data: import RH v4 key or rhel-10.1+ only

achilleas-k:
There is a new produce signing key (v4) that recently added [0][1] and will be used for rhel-10.1 and rhel-9.7. We do not import it for rhel-9.7 (yet) but we do add it for 10.1+.

Note that this also adds a extra newline (`\n`) after the `-----BEGIN PGP PUBLIC KEY BLOCK-----` which is omited in the upstream key but that is apparently a mistake.

[0] https://access.redhat.com/security/team/key
[1] https://security.access.redhat.com/data/6afedf8f.txt

---

Backport of #2097

---

The second commit disables reverse dependency tests because it doesn't make sense to test a release branch against the `main` branch of osbuild-composer and image-builder-cli.


achilleas-k:
https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/12487309975

download.eng.bos has been retired, so I'm pulling in a commit from main that updates the repo snapshots.

achilleas-k:
Had to update ssh keys and the CI variable name in the test configs.
See the two new commits.

PR #2103, opened 2025-12-19
scripts/boot-image: fix cross-arch smoke tests

lzap:
Cross arch testing does not work properly, the QEMU command line renders image twice. Once as device option and once as an argument. On top of that, arguments are incorrectly formatted since Path object is getting printed:

PosixPath('/var/tmp/tmpdk5gg6b6/netinst.iso'), PosixPath('/var/tmp/tmpp40bxujp/disk.img')

QEMU complains about raw image that cannot be auto-detected because it is empty. The problem is, for installer ISO we create an empty raw image and since there is zero content it cannot be autodetected. But the format= argument is passed as qcow2 which is not correct in this case.

As part of this PR, I have also modernized the CPU little bit which was required in order to use virtio. On my system with recent Intel CPU, I was getting some panics and I had to enable Q35 model and also I must not pass `-cpu host` which passed some very modern features which did not work well. Instead, I selected a relatively modern model that has x86_64-v3 set required by RHEL10 which makes it more stable across various laptops/servers we locally test on.

Also some other smaller things like `cache=unsafe` or `-cpu max` for ARM, see the individual commits.

I am pulling the terraform update which should bring new instances, hopefully this can make things a bit faster at times.

### DIFF
--- a/test/scripts/boot-image
+++ b/test/scripts/boot-image
@@ -134,6 +134,12 @@ class CannotRunQemuTest(Exception):
         self.skip_reason = skip_reason
 
 
+class MissingBootImplementation(Exception):
+    def __init__(self, skip_reason):
+        super().__init__(skip_reason)
+        self.skip_reason = skip_reason
+
+
 def ensure_can_run_qemu_test(image_path, config_file):
     """
     Check if the given image_path, config_file is capable of running a
@@ -535,6 +541,10 @@ def main():
     arch = build_info["arch"]
     image_type = build_info["image-type"]
 
+    if not testlib.can_boot_test(image_type, arch):
+        print(f"{image_type} boot tests are not supported yet")
+        return
+
     print(f"Testing image at {image_path}")
     bib_image_id = ""
     # Keep test/scripts/imagetestlib.py:CAN_BOOT_TEST in sync as it
@@ -580,8 +590,7 @@ def main():
             boot_wsl(distro, arch, image_path, build_config_path)
         case _:
             # skip
-            print(f"{image_type} boot tests are not supported yet")
-            return
+            raise MissingBootImplementation(f"{arch} {image_type} is missing a boot implementation.")
 
     print("✅ Marking boot successful")
     # amend build info with boot success
--- a/test/scripts/imgtestlib.py
+++ b/test/scripts/imgtestlib.py
@@ -292,7 +292,16 @@ def read_manifests(path):
     return manifests
 
 
+def can_boot_test(image_type, arch):
+    return image_type in CAN_BOOT_TEST.get("*", []) + CAN_BOOT_TEST.get(arch, [])
+
+
 def check_for_build(manifest_fname, build_info_dir, errors):
+    """
+    Checks if a manifest was built (and optionally booted) successfully.
+
+    This function returns True if the image needs to be built.
+    """
     build_info_path = os.path.join(build_info_dir, "info.json")
     # rebuild if matching build info is not found
     if not os.path.exists(build_info_path):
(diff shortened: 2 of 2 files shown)


supakeen:
> I only now see Fedora 42/43 on aarch64 failing to boot, something is wrong with the boot partition, could this be related to BTRFS changes we recently merged @supakeen ? What I see are pxe-tar-xz and qcow2 (all-customizations) failing being unable to find /boot/efi. But we enabled it only on cloud image, I can confirm that these images do indeed have ext4:
> 
> ```
> (venv) lzap@dev:~/images$ virt-filesystems -a build/fedora_44-aarch64-qcow2-import_rpm_gpg_keys_from_tree_fedora/qcow2/disk.qcow2 --all --long --uuid
> Name      Type       VFS  Label MBR Size       Parent   UUID
> /dev/sda1 filesystem vfat ESP   -   209489920  -        09F3-B305
> /dev/sda2 filesystem ext4 boot  -   2040373248 -        2f0db0f7-13e0-4b5f-8425-455f7bb9066d
> /dev/sda3 filesystem ext4 root  -   3094028288 -        c475c74d-3f0d-4b54-89e1-e2b6935aa755
> /dev/sda1 partition  -    -     -   209715200  /dev/sda -
> /dev/sda2 partition  -    -     -   2147483648 /dev/sda -
> /dev/sda3 partition  -    -     -   3222257152 /dev/sda -
> /dev/sda  device     -    -     -   5580521472 -        -
> 
> (venv) lzap@dev:~/images$ virt-filesystems -a build/fedora_44-x86_64-qcow2-import_rpm_gpg_keys_from_tree_fedora/qcow2/disk.qcow2 --all --long --uuid
> Name      Type       VFS     Label MBR Size       Parent   UUID
> /dev/sda1 filesystem unknown -     -   1048576    -        -
> /dev/sda2 filesystem vfat    ESP   -   209489920  -        7118-63EE
> /dev/sda3 filesystem ext4    boot  -   2040373248 -        491e50ab-b976-4b3e-92ee-787c9871f463
> /dev/sda4 filesystem ext4    root  -   3094028288 -        a416ef13-b72b-48e4-b0c4-43cb87618b57
> /dev/sda1 partition  -       -     -   1048576    /dev/sda -
> /dev/sda2 partition  -       -     -   209715200  /dev/sda -
> /dev/sda3 partition  -       -     -   2147483648 /dev/sda -
> /dev/sda4 partition  -       -     -   3222257152 /dev/sda -
> /dev/sda  device     -       -     -   5581570048 -        -
> ```
> 
> Then I also see Fedora on aarch64 running out of space on the VM image during boot, not sure how that is related. Several image configurations fail, I also see lack of RNG entropy so SSH keys might delay the boot, I am going to add a RNG passthrough to see if that helps. But since these are Fedoras this might again be related to BTRFS change?

We only changed to btrfs for the Cloud images (so: `cloud-azure`, `cloud-ec2`, `cloud-gce`, `cloud-qcow2`) on all architectures except s390x; not the `pxe-tar-xz` image type.

Looking at the latest test run I do not see any disk space errors (I think) on aarch64; I mostly see `ConnectionRefusedError: cannot connect to port 47555 after 600s` there and none of it on any `cloud-*` images so I don't think that's related here.

Can you link me to a run where an image runs out of disk space during boot?

lzap:
For example: https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/12511371522

```
systemd[1]: Failed to populate /etc with preset unit settings, ignoring: Read-only file system
```

I need to take a closer look that might be not actually full disk.

lzap:
Found out why `pxe-tar-xz` is being executed on AWS, fixed it. Not sure about the other failures tho.

lzap:
Nice, only Fedoras on aarch64 fails to boot, it has a common problem with storage. Check those three failures out @supakeen thanks.

lzap:
Good idea, squashed 1800 that is 3 times more, if this does not work, well, we might need to disable this. Yeah, this is full emulation I actually tried to solve this by adding some optimalization (multi-threading) but it did not help too much.

supakeen (dismissed):

supakeen:
Seems to still fail on the `aarch64` emulated tests; they're getting retried but that likely means that this PR will time out.

supakeen (dismissed):
I'll approve but I *think* we still have paths where we skip images (for example `CannotRunQemuTest` when `jq` is not found?).

PR #2111, opened 2026-01-06
github: fix workflows that update refs

achilleas-k:
The osbuild commit ID update workflow has been failing on this step for a while now.  It seems the checkout action isn't setting the extraheader config anymore, causing the related line to fail.

More generally, it seems that several things changed in actions/checkout@v6 which is causing the branch push and PR creation to fail.

This commit drops the usage of actions/checkout in favour of using `gh repo clone` with Schutzbot's access token.

The update-bootc-image-builder workflow hasn't been used for a while so it's being deleted.  If we need it in the future we can add it again.

Another change here is that the branch is crated in the osbuild (upstream) repository.  This lets us easily delete the branch in the PR after it's merged and avoids filling up Schutzbot's fork with [useless leftover branches](https://github.com/schutzbot/images/branches/active).


bcl (approved):
Looks good!

ondrejbudai (dismissed):
Thanks!

achilleas-k:
Alright, I think it's all fixed now.  Dropped actions/checkout completely.
https://github.com/osbuild/images/actions/runs/20779773193/job/59674206131

bcl (approved):
Ack

PR #2118, opened 2026-01-08
check-host-config: invert condition

supakeen:
Currently we're testing if the config file exists and if it doesn't we exit unsuccesfully. However, the condition checks if the file exists and if it *does* exits.

Invert the logic to check if the file does *not* exist.


croissanne (approved):

achilleas-k:
Well, that's extremely embarrassing.  Sorry for all the breakage and wasted time.

PR #2119, opened 2026-01-08
repositories: convert from JSON to YAML

lzap:
Converts repos from JSON to YAML, I used Python to do the job and then I created a formatting script that modifies the generated YAML because it was totally unusable (keys order, quoting, GPG keys not multi-line).

There are no changes made except I took the liberty to remove empty lists (e.g. `gpgkeys: []`) I quickly scanned the codebase and it should not be a problem. It looks nicer this way.

Edit: I dropped the conversion script which I initially included. Shutzbot update PR linked to this PR.

### DIFF
--- a/data/repositories/repos.go
+++ b/data/repositories/repos.go
@@ -4,5 +4,5 @@ import (
 	"embed"
 )
 
-//go:embed *.json
+//go:embed *.yaml
 var FS embed.FS
--- a/test/data/repositories/testrepos.go
+++ b/test/data/repositories/testrepos.go
@@ -7,7 +7,7 @@ import (
 	"github.com/osbuild/images/pkg/reporegistry"
 )
 
-//go:embed *.json
+//go:embed *.yaml
 var FS embed.FS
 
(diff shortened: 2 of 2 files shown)


achilleas-k:
I would personally prefer not to do this, but can be convinced otherwise if everyone else is in favour.

The downsides of keeping them in json, as far as I can tell are:
1. Long strings with linebreaks are annoying (GPG keys).
2. No comments.

I agree with 1, but 2 was never a problem for these files.

Now, the downsides of having them in yaml are:
1. No standard library support in any language we use.  That's fine for Go, we use yaml internally already, but for our other tooling, which is usually python scripts, that means we need an external dependency for every small script that needs to read or modify these files.
2. Needs a whole other script to format it to make it readable.

I don't think it's worth it.  We're rarely, if ever, modifying these by hand and constantly reading and modifying them in scripts.  I agree that the GPG keys can get annoying to copy and paste in json but it's not frequent enough nor really that difficult to do to warrant this kind of change.

PR #2120, opened 2026-01-08
sfdisk: check for extended partitions with MBR

aesteve-rh:
When validating a `dos` partition (MBR) with more than 4 partitions, check also if there is an extended partition type (0f) before actually failing.

### DIFF
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -93,6 +93,9 @@ const (
 	// Partition type ID for PRep on dos
 	PRepPartitionDOSID = "41"
 
+	// Partition type ID for extended partition on dos
+	ExtendedPartitionDOSID = "0f"
+
 	// static UUIDs for partitions and filesystems
 	// NOTE(akoutsou): These are unnecessary and have stuck around since the
 	// beginning where (I believe) the goal was to have predictable,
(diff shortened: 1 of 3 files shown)


supakeen (commented):
Let me preface with: it's been a while since I thought about extended partitions. Are the additional partitions actually contained inside the extended partition or do the follow the extended partition and the extended partition only contains the information about the (potentially multiple) following partition?

If the former then I think it'd be nicer to `len(o.Partitions) - len(PartitionsThatAreExtended)`. If the latter then, while I can probably make up more validation, we can keep it as is :)

Separate question for @alexlarsson what partition layout are you aiming for that you need more than 4 on an RPi4? Fedora images usually have 3 (or 2).

alexlarsson:
> Separate question for @alexlarsson what partition layout are you aiming for that you need more than 4 on an RPi4? Fedora images usually have 3 (or 2).

When we use UKIBoot we need at least these partitions: EFI, ukiboot_a, ukiboot_b, ukibootctl, rootfs. Then additionally we may need a separate /var.

aesteve-rh:
> I think what we want to check is that either the len is <= 4, or there is an extended partition in the first four slots.

So iiuc it should only check/iterate the first four partitions in the validate loop?

supakeen (approved):
Looks OK to me, @achilleas-k is most knowledgeable about this part so let's see what he thinks too.

lzap (approved):

alexlarsson:
Can we get this merged?

PR #2124, opened 2026-01-09
imgtestlib: support overriding detected host distro

achilleas-k:
The host distro is used to determine the 'runner' value for the build cache when checking for manifests.  Being able to override it is useful for running test scripts that depend on reading the value locally.


lzap (approved):
Code is clean.

PR #2131, opened 2026-01-13
distro/fedora: rename iot-bootable-container to generic

achilleas-k:
The bootc-based container image we have in Fedora is not "IoT", nor any other known variant or spin.  It's a generic bootc-based container image.  Rename it to match the other, similar unofficial image types we have in the distro.


achilleas-k:
This might need some work to get the container image working with newer versions of bootc and bootc-image-builder.

PR #2133, opened 2026-01-13
Add a generic bootc-generic-iso image type (HMS-10013)

ondrejbudai:
This PR introduces a new bootc-generic-iso image type that builds live ISOs directly from bootc container images without assuming any specific installer payload (e.g. anaconda).

This pairs with https://github.com/ondrejbudai/bootc-isos, which is a repository with examples of such Containerfiles.

This is clearly an experimental image type at this point. What's missing:

- kernel args are hardcoded
- we cannot embed the container image to be installed
- grub configs are hardcoded
- we cannot choose between a live ISO and a "boot.iso"-style ISO (no live environment, just booting straight to anaconda).

The most interesting point is the last one: if we manage to implement this, we might be able to drop `bootc-installer`, and just use this fully-generic type for everything. I would very much like that.


supakeen (dismissed):
Looks fine; tests can be skipped for now this is most definitely not its final form. I'd like to start contributing to these code paths so I can pick up any potential tests in my follow-ups if that's OK?

ondrejbudai:
I added both a smoke test in Go that checks for some basic stages in the manifest, and a checksum. See the last commit. Please review the tests carefully, this was the first time I touched the faked bootc checksums.

supakeen (approved):

supakeen (approved):

PR #2134, opened 2026-01-14
chore: introduce fedora-45

supakeen:
These repository URLs are currently broken but we need support to build fedora-45 in koji *before* rawhide starts reporting as fedora-45 which means it needs to be in `image-builder` *before* half of february.

I plan to later on fix this up in a more proper way but for now this is still necessary.

Not adding this to test repositories yet until fedora-45 actually exists and we snapshot it.


thozza (approved):
ACK. 

BTW, we should not need to add the repos here to make it possible for `image-builder` to build F45 in Koji. That's a common use case for RHEL and something that I have been emphasizing every time we wanted to merge repo definitions into distro definitions.

PR #2136, opened 2026-01-15
Align the bootc distro package with the generic one

achilleas-k:
This is purely a refactoring PR.  It's the first step in aligning the way we define the bootc distro, its image types, and the manifest generators with the structure and functions of the generic distro.

The end goal is to make the bootc distro package (the pipeline generators for bootc-based images) have the same interface and behave as much as possible as the other distros.  The bootc distro package was initially defined in bootc-image-builder, where we had very different requirements and assumptions about the running environment.  Since we merged it here, there's now a lot of duplication and, more importantly, the way we define the bootc-based image types and generate the manifests is very different from the traditional way the manifest generators work.

A big part of this goal is making it simpler to instantiate the manifest without needing to resolve content (resolve the container ref, inspect/run the container), or at least without that happening implicitly.  This will be very useful both for mocking (generating manifests without real content) and for having finer control over remote content resolution, like we already do for packages, containers, and ostree refs, which in turn is needed for defining jobs in the osbuild-composer and the service.

### DIFF
--- a/pkg/distro/bootc/imagetype.go
+++ b/pkg/distro/bootc/imagetype.go
@@ -136,11 +136,11 @@ func (t *imageType) checkOptions(bp *blueprint.Blueprint) []string {
 func (t *imageType) Manifest(bp *blueprint.Blueprint, options distro.ImageOptions, repos []rpmmd.RepoConfig, seedp *int64) (*manifest.Manifest, []string, error) {
 	validationWarnings := t.checkOptions(bp)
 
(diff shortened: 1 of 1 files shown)


achilleas-k:
NOTE: There will be merge conflicts here once some (higher priority) PRs get merged.

lzap (approved):

thozza (approved):
Nice and clean, :rocket: it

PR #2142, opened 2026-01-19
check: implement file/dir/kernel host checks

lzap:
SSIA

This also adds a new smoke test that runs in a container because I realized today that when there is a change in the host check code, new image rebuilds will not be triggered. This could lead to regressions so the smoke test actually runs on a real system, the idea is to use container locally to run it (and it will be executed together with other tests on CICD). I wanted to safe-guard this because for some tests system configuration will have to be changed so an environment variable must be present so no one accidentally run the smoke test locally.

There is a makefile target if you want to do that which is pretty useful and fast as well.

Follow-up:

* [x] Merge `osfuncs.go` and `utils.go`
* [x] Make user of `MustFindCheckByName` in all tests and use shortname instead full name.
* [x] Make use of `any` arguments in Pass/Fail functions
* [x] Refactor all unit tests to tabular-style
* [ ] Implement system tests for all the remaining checks

### DIFF
--- a/cmd/check-host-config/main_test.go
+++ b/cmd/check-host-config/main_test.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/osbuild/blueprint/pkg/blueprint"
+	"github.com/osbuild/images/cmd/check-host-config/check"
+	"github.com/osbuild/images/internal/buildconfig"
+)
+
+// This is a happy-path smoke test that is supposed to be executed in a
+// temporary Fedora container. It is ran on our CI/CD. To run it locally (in
+// podman), execute `make host-check-test`.
+//
+//nolint:gosec // G303: Temporary files need to be consitently named
+func TestSmokeAll(t *testing.T) {
+	if os.Getenv("OSBUILD_TEST_CONTAINER") != "true" {
+		t.Skip("Not running in container, skipping host check test")
+	}
+
+	// Prepare the container environment (cleanup not needed)
+	if err := os.Mkdir("/tmp/dir", 0700); err != nil {
+		t.Fatalf("Failed to create test directory: %v", err)
+	}
+	if err := os.WriteFile("/tmp/dir/file", []byte("data"), 0600); err != nil {
+		t.Fatalf("Failed to create test file: %v", err)
+	}
+
+	tests := []struct {
+		chk check.RegisteredCheck
+		c   blueprint.Customizations
+	}{
+		{
+			chk: check.MustFindCheckByName("Kernel Check"),
+			c: blueprint.Customizations{
+				Kernel: &blueprint.KernelCustomization{
+					Name:   "kernel",
+					Append: "root=",
+				},
+			},
+		},
+		{
+			chk: check.MustFindCheckByName("Kernel Check"),
+			c: blueprint.Customizations{
+				Kernel: &blueprint.KernelCustomization{
+					Name: "kernel-debug",
+				},
+			},
+		},
+		{
+			chk: check.MustFindCheckByName("Directories Check"),
+			c: blueprint.Customizations{
+				Directories: []blueprint.DirectoryCustomization{
+					{Path: "/tmp/dir"},
+					{Path: "/tmp/dir", Mode: "0700"},
+					{Path: "/tmp/dir", Mode: "0700", User: "root", Group: "root"},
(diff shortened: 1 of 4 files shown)


cmd/check-host-config/check/files_test.go
```diff
@@ -91,7 +166,178 @@ func TestFilesCheckWithReadFileError(t *testing.T) {
...
+			{Path: "/etc/testfile", Mode: "0755", User: uint32(0), Group: uint32(0)},
+		},
+	})
+
+	err := chk.Func(chk.Meta, config)
+	require.Error(t, err)
+	assert.True(t, check.IsFail(err))
+}
+
+func TestFilesCheckWithMatchingUid(t *testing.T) {
```
thozza: This check is basically a copy of `TestFilesCheck` with not added checks, AFAICT.

cmd/check-host-config/check/files_test.go
```diff
@@ -91,7 +166,178 @@ func TestFilesCheckWithReadFileError(t *testing.T) {
...
+			{Path: "/etc/testfile", Mode: "0644", User: uint32(1000), Group: uint32(0)},
+		},
+	})
+
+	err := chk.Func(chk.Meta, config)
+	require.Error(t, err)
+	assert.True(t, check.IsFail(err))
+}
+
+func TestFilesCheckWithNonMatchingGid(t *testing.T) {
```
thozza: This check is basically a copy of `TestFilesCheck` with not added checks, AFAICT.

cmd/check-host-config/check/files_test.go
```diff
@@ -1,26 +1,347 @@
...
 func TestFilesCheck(t *testing.T) {
 	test.MockGlobal(t, &check.Exists, func(name string) bool {
 		return true
 	})
 
+	test.MockGlobal(t, &check.Stat, func(name string) (os.FileInfo, error) {
+		if name == "/etc/testfile" {
+			return &mockFileInfo{
+				name:    "testfile",
+				mode:    0644,
```
thozza: In a real file system, `os.Stat()` would return a mode like 0100644 (regular file + 0644 permissions), but the mock is incorrectly returning just 0644.

thozza (commented):
Thanks.

I've added some comments inline. The Files check has some serious issues.

The tests are thorough, which is great, but I think we can make them more maintainable with a table-driven approach.

Right now there's quite a bit of duplication. For example, `TestFilesCheck`, `TestFilesCheckWithMatchingMode`, and `TestFilesCheckWithMatchingUid` are essentially testing the same thing ("when file attributes match, the check passes") — just with slightly different values. These could be consolidated into a single table-driven test where each case is just a row:

```go
tests := []struct {
    name        string
    fileInfo    *mockFileInfo
    customization blueprint.FileCustomization
    expectError bool
}{
    {
        name:     "matching mode user and group",
        fileInfo: &mockFileInfo{mode: 0644, uid: 0, gid: 0},
        customization: blueprint.FileCustomization{Path: "/etc/testfile", Mode: "0644", User: int64(0), Group: int64(0)},
    },
    {
        name:     "non-matching mode",
        fileInfo: &mockFileInfo{mode: 0644, uid: 0, gid: 0},
        customization: blueprint.FileCustomization{Path: "/etc/testfile", Mode: "0755", User: int64(0), Group: int64(0)},
        expectError: true,
    },
    // ... more cases
}
```

Additionally, the mock setup and other boilerplate code is unnecessarily duplicated in each unit test function.

cmd/check-host-config/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+			Customizations: &blueprint.Customizations{
+				Kernel: &blueprint.KernelCustomization{
+					// Name cannot be tested in container (rpm db does not contain any kernel)
+					Append: "vmlinuz", // Should be always present in the kernel command line
+				},
+				Files: []blueprint.FileCustomization{
+					{Path: "/etc/group"},
+					{Path: "/etc/shadow", Mode: "0000"},
+					{Path: "/etc/passwd", Mode: "0644", User: 0, Group: 0},
+					{Path: "/etc/passwd", Mode: "0644", User: "root", Group: "root"},
```
thozza: I know this works, but would it make sense to specify the string value user and group on a different file? Using the same path looks like a bug at the first sight.

thozza (commented):
Thanks for the changes. I like it much more!

The table-driven tests are much easier to read.
The smoke test inside a container is also IMO a reasonable approach.

I've added some inline comments. On top of that, there are some linter errors.

PR #2146, opened 2026-01-19
osbuild: add force option to org.osbuild.groups and always enable [HMS-9825]

achilleas-k:
The `--force` option of the groupadd command makes it possible to run with a group name or GID that already exists without failing.  This is very useful for letting users define groups in their configs without worrying about the group existing in the base system or being created from a package, which isn't always predictable.


achilleas-k:
Cross-arch qemu based image build/boot smoke test will be fixed by #2145

data/dependencies/deps_test.go
```diff
@@ -8,5 +8,5 @@ import (
 )
 
 func TestMinimumOSBuildVersion(t *testing.T) {
-	assert.Equal(t, "167", dependencies.MinimumOSBuildVersion())
+	assert.Equal(t, "169", dependencies.MinimumOSBuildVersion())
```
thozza: I don't feel too strongly about this, but I would prefer the minimum version bump to happen before or with the commit that adds the dependency on the newer version. In this case, it would be the one that adds the new stage option and makes use of it.

thozza (approved):
In general, LGTM.

I've added a few comments. I think that the behavior of the `--force` option with two groups with the same GIDs is worth addressing, but could be in a follow up or as a documentation.

data/dependencies/deps_test.go
achilleas-k: yeah, that makes sense.

PR #2150, opened 2026-01-21
vmtest: Add timeouts to start, wait_qmp_event, and wait_ssh_ready

bcl:
Defaults to 120s if timeout_sec isn't set.

Calling wait_ssh_ready defaults to 1800s, which seems excessive, and is overridden by timeout_sec when called from QEMU.run()


supakeen (approved):
Thank you, I quite like this too. We might have to fudge timeout values for cross-arch emulated boots but we'll see if that happens.

PR #2157, opened 2026-01-28
pkg/cloud/awscloud: allow specifying AWS credentials profile

FrostyX:
The `$HOME/.aws/credentials` file looks like this

    [default]
    aws_access_key_id = secretString1
    aws_secret_access_key = secretString2

    [some-profile]
    aws_access_key_id = secretString3
    aws_secret_access_key = secretString4

This allows us to specify the non-default profile.

### DIFF
--- a/pkg/cloud/awscloud/uploader.go
+++ b/pkg/cloud/awscloud/uploader.go
@@ -21,6 +21,7 @@ type awsUploader struct {
 	region     string
 	bucketName string
 	imageName  string
+	profile    string
 	tags       []AWSTag
 	targetArch arch.Arch
 	bootMode   *platform.BootMode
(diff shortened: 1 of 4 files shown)


FrostyX:

This sounds like a problem. Maybe we could somehow refactor it to not break the API?

pkg/cloud/awscloud/uploader.go
```diff
@@ -48,24 +50,30 @@ type awsClient interface {
...
+	client, err := newAwsClient(region, opts.Profile)
 	if err != nil {
 		return nil, err
 	}
 
+	profile := opts.Profile
+	if profile == "" {
+		profile = "default"
+	}
+
```
mdujava: shouln'd this block be placed before calling newAwsClient, so aws client is called with a "default" instead of empty string?

supakeen:
Needs linter and test case appeasement (function signature that changed for the latter). Otherwise looks fine to me 🙂

PR #2164, opened 2026-01-30
check-host-config: convert tests to tabular format

lzap:
This PR contains several improvements discussed in the final reviews in https://github.com/osbuild/images/pull/2142 which is nicely isolated in individual commits.

For the rewrite of unit tests to tabular style, I have decided to do it in a single commit which might be pretty unreadable due to indentation changes, so I suggest to checkout the branch and see the code directly. All of them follows the same structure which is documented in the commit:

```
check-host-config: convert tests to tabular format

Pretty much all tests except OpenSCAP checks look much cleaner in the
tabular testing format. In order to achieve maximum readability, I
introduced a helper types and functions in mock_helpers_test.go that
allows to reuse the following testing pattern:

	tests := []struct {
		name         string
		config       *blueprint.KernelCustomization
		mockExec     map[string]ExecResult
		mockReadFile map[string]ReadFileResult
		wantErr      error
	}

Each test case has a name, configuration and returns either nil or
error. It also has one or more mocks which are represented as maps with
input (typically a string but a struct for functions with multiple
arguments) and output (struct). These mock maps can be nicely formatted:

    {
        name: "fail when append does not match",
        config: &blueprint.KernelCustomization{
            Append: "debug",
        },
        mockReadFile: map[string]ReadFileResult{
            "/proc/cmdline": {Data: []byte("root=UUID=1234-5678 ro")},
        },
        wantErr: check.ErrCheckFailed,
    },

The code that performs installation of these mocks is contained in the
helper functions:

    installMockExec(t, tt.mockExec)
    installMockReadFile(t, tt.mockReadFile)
```

I plan to implement system tests for all checks in `test_main.go` but I will either do it in separate PR or stack those commits on top of this one after your review. Thanks @thozza @ondrejbudai

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -1,6 +1,6 @@
 {
   "common": {
-    "rngseed": 12,
+    "rngseed": 13,
     "bootc-image-builder": {
       "ref": "quay.io/centos-bootc/bootc-image-builder@sha256:9893e7209e5f449b86ababfd2ee02a58cca2e5990f77b06c3539227531fc8120"
     },


thozza (approved):
LGTM as is, let's not make this any bigger.

PR #2167, opened 2026-02-02
pkg/bib/container: Support use in rootless containers

alexlarsson:
When running in a rootless container (made possible via https://github.com/osbuild/bootc-image-builder/pull/1189) then generally the per-user container store is passed to bc-i-b, both because that makes sense, but also because the container doesn't have permissions to read the host store. This is typically done by passing `-v ~/.local/share/containers/storage:/var/lib/containers/storage` when running bc-i-b in podman.

Unfortunately when this happens, podman complains that the store is in the wrong place:

```
Error: database static dir "/home/alex/.local/share/containers/storage/libpod" does not match our static dir "/var/lib/containers/storage/libpod": database configuration mismatch
```

To fix this, we pass in podman arguments `--root` and `--imagestore` to set up an environment where the graphroot is some empty directory that has the "correct" location and we only load the image layers from /var/lib/containers.

To ensure this doesn't weirdly affect the regular case, this is all protected by podmanutil.IsRootless().

Note: This is similar to what is done in
https://github.com/osbuild/image-builder-cli/pull/447 and it has a local copy of podmanutil.IsRootless() from image-builder-cli.

### DIFF
--- a/pkg/bib/container/container.go
+++ b/pkg/bib/container/container.go
@@ -1,8 +1,12 @@
 package container
 
 import (
+	"bufio"
+	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"strconv"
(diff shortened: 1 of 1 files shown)


alexlarsson:
Hmm, i'm playing with automotive-image-builder using this, and the checks for rootless=1 means this isn't working when I run a container from inside a rootless container, because the inner container thinks it is rootful due to uid being 0 in the outer container....

pkg/bib/container/container.go
```diff
@@ -12,14 +16,43 @@ import (
...
 )
 
+// envPath is written by podman
+const envPath = "/run/.containerenv"
+
+// rootlessKey is set when we are rootless
+const rootlessKey = "rootless=1"
+
+// isPodmanRootless detects if we are running rootless in podman;
+// other situations (e.g. docker) will successfuly return false.
```
achilleas-k: Looking into this, I discovered it will also return false if running rootless but without `--privileged`.  Probably not important though since I don't see us getting rid of that any time soon.

supakeen (approved):

PR #2168, opened 2026-02-03
arch: add ARCH_ARM

mtudan:
This is first attempt to add the legacy `arm` i.e. `armv7` support. I'm not sure what else might be missing or requires adaption. Let's discuss it here.

### DIFF
--- a/pkg/arch/arch.go
+++ b/pkg/arch/arch.go
@@ -12,6 +12,7 @@ type Arch uint64
 
 const ( // architecture enum
 	ARCH_UNSET Arch = iota
+	ARCH_ARM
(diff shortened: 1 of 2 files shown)


supakeen (approved):
I wonder if we should call this `ARCH_ARMV7` but at this point I don't think we'll ever clash with an even older variant, it also doesn't affect much as we'll likely not use this for any package-based things.

It mostly affects bootable containers (which is what @mtudan wants to use it for) and we have some checks that want an accepted architecture which means that we won't be building any buildroots for this architecture ourselves (we use the container as buildroot).

PR #2171, opened 2026-02-04
manifest/subscription: fix selinux systemd race condition (HMS-10142)

kingsleyzissou:
On first boot, selinux-autorelabel.service and osbuild-subscription-register.service
run concurrently. The SELinux relabel triggers a system reboot upon completion,
which interrupts osbuild-subscription-register while rhc is mid-execution.

This results in:
  - partial registration (subscribed to rhsm but not registered with insights)
  - service failure on second boot ("system already connected" exit code 64)

Adding After=selinux-autorelabel.service ensures the subscription service only
runs after the relabel-triggered reboot, allowing it to complete successfully.

JIRA: [HMS-10142](https://issues.redhat.com/browse/HMS-10142)

### DIFF
--- a/pkg/manifest/subscription.go
+++ b/pkg/manifest/subscription.go
@@ -214,7 +214,12 @@ func subscriptionService(
 				Description:         "First-boot service for registering with Red Hat subscription manager and/or insights",
 				ConditionPathExists: []string{subkeyFilepath},
 				Wants:               []string{"network-online.target"},
-				After:               []string{"network-online.target"},
+				After: []string{
+					// The selinux-autorelabel reboot was killing rhc mid-execution,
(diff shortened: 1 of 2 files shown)


achilleas-k (approved):

lzap:
Nice find, isn't this actually the same problem for all services that do something? I am thinking all oneshot units like osbuild-firstoobt or osbuild-custom-firstboot?

kingsleyzissou:
> Nice find, isn't this actually the same problem for all services that do something? I am thinking all oneshot units like osbuild-firstoobt or osbuild-custom-firstboot?

Ya good point. I can look into it, maybe to be safe we can add this condition there for other units too.
Although I'm not sure if any of those are defined in images. Some are probably still in crc and the frontend

PR #2172, opened 2026-02-04
Move bootc distro and image type code into generic distro

achilleas-k:
Third part of the bootc distro package alignment with the generic distro, started in #2136 and #2151.

This PR is a much bigger change.  It significantly changes the way the bootc distro is initialised and the way the bootc container info is resolved.  Most code in `pkg/distro/bootc/` is moved into `pkg/distro/generic/`, with distro and image type variants for bootc.  A future goal is to unify at least the BootcDistro type with the generic distro implementation and perhaps the image type as well, if possible.  For now they remain separate because of slightly different behaviour.

This PR implements API-breaking changes.  The new method for initialising a bootc distro and generating a manifest can be seen clearly in the changes to `gen-manifests`.  In short, the previous flow was (ignoring errors for brevity):
```go
distribution, _ := bootcdistro.NewBootcDistro(bootcRef, nil)  // runs podman to initialise the distribution with the container info
distribution.SetBuildContainer(buildBootcRef)  // optional, if it differs from the base bootc ref - also runs podman to read container info

// continue as normal distribution
arch := distribution.GetArch(...)
imageType := arch.GetImageType(...)
```

The new flow becomes:
```go
c, _ := bootc.NewContainer(bootcRef)
cinfo := c.ResolveInfo()
distribution, _ := generic.NewBootc("bootc", bootcInfo)

// optional build container resolve
b, _ := bootc.NewContainer(buildbootcRef)
binfo := b.ResolveInfo()
distribution.SetBuildContainer(binfo)

// continue as normal distribution
```

This lets the API caller control how and when containers are run to retrieve info.  It makes it much easier to initialise and generate bootc-based manifests with mock data, or by using alternative methods to resolve information (we are considering, for example, ways to resolve container info without running the container, but by simply mounting it and reading files).

Tests have been moved (and where necessary, adapted) from `pkg/distro/bootc` to `pkg/distro/generic` to verify the new implementation behaves in a consistent way.  Test manifests have been verified to be unchanged (both mocked manifests and real ones).

### DIFF
--- a/pkg/distro/generic/bootc.go
+++ b/pkg/distro/generic/bootc.go
@@ -13,6 +13,8 @@ import (
 	"github.com/osbuild/images/pkg/disk"
 	"github.com/osbuild/images/pkg/distro"
 	"github.com/osbuild/images/pkg/distro/defs"
+	"github.com/osbuild/images/pkg/manifest"
+	"github.com/osbuild/images/pkg/runner"
 )
 
 type BootcDistro struct {
@@ -167,6 +169,15 @@ func (d *BootcDistro) Name() string {
 	return d.id.String()
 }
 
+func (d *BootcDistro) ID() distro.ID {
+	return d.id
+}
+
+func (d *BootcDistro) IDLike() manifest.Distro {
+	// not applicable or needed for bootc
+	return 0
+}
+
 func (d *BootcDistro) Codename() string {
 	return ""
 }
@@ -201,6 +212,16 @@ func (d *BootcDistro) GetArch(arch string) (distro.Arch, error) {
 
 func (d *BootcDistro) GetTweaks() *distro.Tweaks {
 	// The bootc distro does not require or support tweaks (yet)
+	return nil
+}
+
+func (d *BootcDistro) Runner() runner.RunnerConf {
+	// To get the bootc distro runner, use [bootc.GetDistroAndRunner]
+	return runner.RunnerConf{}
+}
+
+func (d *BootcDistro) ImageConfig() *distro.ImageConfig {
+	// not applicable for bootc distro
 	return nil
 }
 
@@ -254,4 +275,23 @@ func (d *BootcDistro) SetBuildContainer(cinfo *bootc.Info) error {
 	d.buildSourceInfo = cinfo.OSInfo
 
 	return nil
+}
+
+func (d *BootcDistro) BootstrapContainer(a string) (string, error) {
+	// NOTE: Return the build container ref instead and we will unify these two
+	// concepts later.
+
+	// verify that the architecture matches the bootc distro's architecture
+	// (there should be only one)
+	distroArches := d.ListArches()
+	if len(distroArches) != 1 {
+		// there should only ever be one architecture for a bootc distro
+		return "", fmt.Errorf("found %d architectures for bootc distro while getting build container: bootc distro should have exactly 1 architecture", len(distroArches))
+	}
+	if _, err := d.GetArch(a); err != nil {
+		baseArch := distroArches[0]
+		return "", fmt.Errorf("failed to get build container for bootc distro: requested container architecture %q does not match base container %q", a, baseArch)
+	}
+
+	return d.buildImgref, nil
 }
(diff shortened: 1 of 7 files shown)


pkg/distro/generic/distro.go
```diff
@@ -43,33 +43,7 @@ type distribution struct {
...
-		if err != nil {
-			// XXX: cleanup isoLabelFunc to allow error
-			panic(err)
-		}
-		return buf.String()
-	}
-}
-
-func newDistro(nameVer string) (distro.Distro, error) {
+func New(nameVer string) (distro.Distro, error) {
```
croissanne: Hm, this doesn't seem to be actually used anywhere in the following patches? (outside of the generic package).

achilleas-k: Right.  I was thinking about touching the distrofactory more down the line, but I didn't get to it in this PR and it already became a bit too large.  I can revert it so it's less noise.

supakeen (dismissed):

supakeen (approved):

pkg/distro/generic/bootc.go
```diff
@@ -0,0 +1,182 @@
...
+
+// NewBootc returns a distro initialised with the provided information. All
+// required information must be defined. There are no restrictions or
+// requirements for the name of the distro and it is only used to identify this
+// particular instance of the distribution. The name bootc is commonly used,
+// unless multiple instances are created.
+// To generate the [github.com/osbuild/images/pkg/bootc.Info] from a container
+// ref, use the [github.com/osbuild/images/pkg/bootc.Container] type and its
+// methods.
+func NewBootc(name string, cinfo *bootc.Info) (distro.Distro, error) {
```
lzap: _Accept interfaces, return structs._ Go proverb.

This should return `BootcDistro` instead, there is no need to give up on the type that early.

pkg/distro/generic/bootc.go
```diff
@@ -0,0 +1,182 @@
...
+	errPrefix := "NewBootc"
+	if cinfo == nil {
+		return nil, fmt.Errorf("%s: missing required info: container info is empty", errPrefix)
```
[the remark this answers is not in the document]
lzap: All of my comments are nitpicks, feel free to resolve anything you like.

thozza (approved):
Nice work on merging the two distro implementations! 👏 
I like where this is going...

pkg/distro/generic/distro.go
```diff
@@ -220,6 +220,21 @@ func (a *architecture) addImageType(platform platform.Platform, it imageType) er
 	return nil
 }
 
+func (a *architecture) addBootcImageType(it bootcImageType) error {
```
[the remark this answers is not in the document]
achilleas-k: I'm going to leave this for a follow-up.  I want to spend some time with all three interfaces (distro, arch, image type) and generalise them well so they make sense for all implementations (the regular old distro and bootc).

Perhaps we can break them down into smaller interfaces that make sense in limited scopes.

pkg/distro/generic/bootc_test.go
```diff
@@ -129,6 +129,7 @@ func loadImageTypes(t *testing.T, d *BootcDistro) {
 	for archName, arch := range d.arches {
 		darch := arch.(*architecture)
 		darch.imageTypes = map[string]distro.ImageType{}
+		darch.distro = d // link distro to architecture as well
```
lzap: I can't help but comment (gh really needs "non-blocking" comments feature and I hate to do this feel free to resolve as soon as you read this):

Looking at the Arch interface, I wonder how helpful it is. To me, it looks like architecture is just a tag that is associated with image type. It could have been a string field, having this as a separate nested level creates unnecessary burden as things in distro could be as simple as `ListImageTypeForArch("i386")` instead `Arch().ListImageTypes()`.

That being said, circular references are fine in Go and getting rid of Arch is not on the menu for today, it might not even make sense cos there might be some advantage or use I fail to see.

lzap: Oh by any mean did I think to refactor this right here right now. I am just exploring ideas, yeah. It is good enough, there is a lot of good here.

PR #2175, opened 2026-02-05
distro/generic: propagate kernel options to kickstart for installers

thozza:
The blueprint's customizations.kernel.append setting should configure kernel options for the installed system's bootloader, not just the installer ISO's boot menu. This was working for bootc installers but was missing from the generic distro's installer image functions.

Without this fix, users setting kernel.append in their blueprint would only see the options applied to the installer's boot menu, but after installation the system's bootloader would not have these options configured.

Add the missing KernelOptionsAppend propagation to the kickstart configuration in imageInstallerImage (minimal-installer),
iotInstallerImage (edge-installer), and networkInstallerImage (boot-iso) functions.

### DIFF
--- a/cmd/check-host-config/check/kernel.go
+++ b/cmd/check-host-config/check/kernel.go
@@ -41,7 +41,7 @@ func kernelCheck(meta *Metadata, config *buildconfig.BuildConfig) error {
 		}
 
 		if !strings.Contains(string(cmdline), expected.Append) {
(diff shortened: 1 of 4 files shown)


supakeen:
AFAIK this should've changed checksums because we're now adding additional options to the kickstart stage (and I *think* we have a test config that adds kernel options, hence the initial test failures?).

But it didn't, so what did I miss? 🙂

PR #2177, opened 2026-02-05
New image type for RHEL 9 and 10: ec2-cvm [HMS-10096]

achilleas-k:
This PR introduces a new image definition for RHEL 9.6+ and 10.0+ called ec2-cvm.  It's a variant of the ec2 base image with the CVM configurations applied like they were for the azure-cvm image.

I had to remove grub2 from the base package set of the ec2 image.  The grub2 package should be part of the platform configuration, so I moved it to the appropriate platform objects in both repos.  This resulted in gce (RHEL 9 and 10) and minimal-raw (RHEL 9 only) gaining a couple of grub2 packages in their OS pipeline: grub2-pc and grub2-pc-modules.

Otherwise, all other image definitions are unchanged.  Manifest checksums have been modified because of the different package selection (`grub2` is part of the explicit package selection, but it pulls in packages that were already part of the selection), but depsolved manifests are unchanged.


thozza (approved):
LGTM

achilleas-k:
QE discovered that `kdump-utils` was missing from the image, which I had to spend some time looking into.  Added it back to all ec2 and ami image types for RHEL 10.  See new commit for details.

ochosi:
Beautiful! We should expose this in the service, too, somehow

achilleas-k:
Rebased and cleared conflicts.

achilleas-k:
> Beautiful! We should expose this in the service, too, somehow

That's the immediate next step, yes :)

PR #2181, opened 2026-02-07
data/repositories: update f45 keys

supakeen:
With Fedora 45 now existing it has its own key. Let's import it and use it.


supakeen:
A rare double approval.

PR #2184, opened 2026-02-09
distro: move ostree default reference to image type (HMS-10151)

supakeen:
The default ostree reference that we use to fill in when no reference is passed through `ImageOptions` was set per distribution. This is problematic as different variants within the same distribution can have different templates. Amongst those are the upcoming Fedora Atomic variants which use `fedora/$version/$arch/$variant` which is notably different from the Fedora IoT variants which use `fedora/$codename/$arch/iot`.

Note that this means that our reference template for Fedora IoT has always been wrong as it contains the version instead of (`rawhide`, `devel`, `stable`); I plan to address that in a followup commit when I figure out a good way 🙁

Another thing that is different per variant is the ostree repository URL. This will also be moved onto the image types in a future PR.

Oh and let's add more interesting things; our remote names are always prefixed `rhel-` perhaps that should be templated as well in a followup.

Ok I'll stop making up more work.

---

No manifest changes.

### DIFF
--- a/pkg/distro/generic/imagetype.go
+++ b/pkg/distro/generic/imagetype.go
@@ -373,6 +373,12 @@ func (t *imageType) expandOSTreeRefTemplate(ar *architecture, id distro.ID) erro
 
 		t.ostreeRef = buf.String()
 
+		// if we're empty after templating that's an error as we can't
+		// have an empty commit
+		if t.ostreeRef == "" {
+			return fmt.Errorf("empty ostree ref after expansion")
+		}
+
 		return nil
 	}
 


pkg/distro/generic/imagetype.go
```diff
@@ -98,12 +100,22 @@ func (t *imageType) MIMEType() string {
...
+
+		if err := tmpl.Execute(&buf, t.arch.distro.ID); err != nil {
+			return "", err
+		}
+
+		return fmt.Sprintf(buf.String(), t.arch.Name()), nil
 	}
-	return ""
+	return "", nil
 }
```
lzap: Just an idea, maybe this could have been calculated when image type is being constructed just once, then the accessor would not return any error (and be more efficient as well). I do not know if this is a pattern in this code, if it is that of course stick with it.

Edit: I am commenting on the accessor that calculates the value via Go templates, it is not apparent from the hunk I pointed on.

It does appear that this is the first use of templates in this type, so consider this.

supakeen: Actually I think we can't because we don't know the architecture at that point yet? Maybe I'm not seeing it :)

lzap (dismissed):
Is okay but this would work perhaps?

https://gist.github.com/lzap/e6ef90aefe95b2a223318d997cd08aff

thozza (approved):
LGTM, Thanks!

PR #2189, opened 2026-02-11
test/scripts: don't use post-release version bump commits for osbuild

achilleas-k:
Post-release version bump commits contain the message [skip ci] which skips running CI.  These commits do not have RPMs associated with them, so they shouldn't be used to downstream testing.  Use the parent commit ID if we detect [skip ci] in the commit message on main.


achilleas-k:
Status checks seem stuck.  Clicking rebase.

PR #2193, opened 2026-02-12
ci: introduce f45

supakeen:
Enable Fedora 45 (the current rawhide) in our test repositories and generate its manifests.

Note that this is based on #2181 so we have the correct GPG keys; hence this will need that to land first.


bcl (approved):
Ack!

lzap (approved):

PR #2196, opened 2026-02-12
distro: add erofs options into iso

avitova:
This commit adds the erofs extended options into the iso configuration.


pkg/distro/defs/loader_test.go
```diff
@@ -650,7 +651,7 @@ image_types:
 	assert.Equal(t, dist.ImageConfig(), &distro.ImageConfig{
 		Locale:   common.ToPtr("C.UTF-8"),
 		Timezone: common.ToPtr("OverrideTZ"),
-		Users:    []users.User{users.User{Name: "testuser"}},
+		Users:    []users.User{{Name: "testuser"}},
```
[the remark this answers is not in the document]
supakeen: I don't know which one I like more; I think it's what you changed it to. I guess theoretically because this is an unrelated change it's best to have it in a separate commit `chore: linter appeasement` or such.

This tends to be a good idea because if someone agrees with all the other code; but doesn't agree with the above change (unlikely in this case, but let me use it as an example) it's easier to drop the offending commit/pull it out in a separate PR instead of rewriting what you already have.

supakeen (commented):
Nice, the first bits now look a lot better (and have a test!). Let's start adjusting the stuff that uses this; that's mostly `pkg/manifest/anaconda_installer_iso_tree.go` in the `NewErofsStage` function.

We should be using the `ISOCustomizations` that are now set based on the YAML there for the `erofsOptions`; since you embedded the stage options directly in the YAML this should quick to do.

Note that once you do this your manifests will likely change as you now *must* have the same values in the YAML as those that were previously set in the Go code, but you can tinker with that.

It should be possible to plumb this in *without* any manifests changing *if* the same options are present in the YAML for the image types affected 🙂.

data/distrodefs/fedora/imagetypes.yaml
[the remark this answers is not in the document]
supakeen: We want the change. You can generate the checksums and amend them onto the commit that changes the YAML.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -294,6 +294,9 @@ func (p *AnacondaInstallerISOTree) NewSquashfsStage() (*osbuild.Stage, error) {
 // the erofs root filesystem for the ISO.
 func (p *AnacondaInstallerISOTree) NewErofsStage() (*osbuild.Stage, error) {
 	var erofsOptions osbuild.ErofsStageOptions
+	if p.anacondaPipeline == nil {
+		return nil, fmt.Errorf("Anaconda pipeline not set for %s pipeline", p.name)
+	}
```
supakeen: Let's drop this commit; we don't check non-`nil`-ness of `p.anacondaPipeline` anywhere else and if we *were* to add it it should probably live in the constructor.

pkg/manifest/anaconda_installer_iso_tree.go
```diff
@@ -293,41 +293,31 @@ func (p *AnacondaInstallerISOTree) NewSquashfsStage() (*osbuild.Stage, error) {
...
 // the erofs root filesystem for the ISO.
 func (p *AnacondaInstallerISOTree) NewErofsStage() (*osbuild.Stage, error) {
-	var erofsOptions osbuild.ErofsStageOptions
+	var erofsOptions *osbuild.ErofsStageOptions
 	if p.anacondaPipeline == nil {
 		return nil, fmt.Errorf("Anaconda pipeline not set for %s pipeline", p.name)
 	}
 
+	if p.RootfsType != ErofsRootfs || p.anacondaPipeline.ISOCustomizations.ErofsOptions == nil {
+		return nil, fmt.Errorf("Rootfs not set to Erofs or options set to nil for %s pipeline, can not create erofs stage", p.name)
```
bcl: Erofs should be lower case since that's the value that is used in the yaml.

data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -2520,7 +2520,17 @@ image_types:
         - ["/usr/share/licenses/fedora-release-common/Fedora-Legal-README.txt", "/Fedora-Legal-README.txt"]
         - ["/usr/share/licenses/fedora-release-common/LICENSE", "/LICENSE"]
       install_weak_deps: false
-    iso_config: *default_iso_config
+    iso_config:
```
bcl: Typically in Fedora we don't make big changes like this to older releases so we may need to limit this to F44 and later. @achilleas-k what do you think?

supakeen: I think we tend to only do that for deliverables that are actually part of composes in the case of Fedora, the network installer isn't (yet).

supakeen (approved):

Issue #2176, opened 2026-02-05
c9s: The `minimal-raw` `/boot` is too small


thozza:
Building the `minimal-raw` with the latest c9s repos fails on the `/boot` being too small.

Reproduce by:
```
sudo go run cmd/build/* -arch x86_64 -distro centos-9 -repositories data/repositories/centos-9.json -type minimal-raw -config test/configs/empty.json
```
Result:
```
...
org.osbuild.copy: 8e8fef26f38a83c3aad3498b20464a7039708b1296454ee9867f6ce180542963 {
  "paths": [
    {
      "from": "input://root-tree/",
      "to": "mount://-/"
    }
  ]
}
device/- (org.osbuild.loopback): loop0 acquired (locked: False)
device/boot (org.osbuild.loopback): loop1 acquired (locked: False)
device/boot-efi (org.osbuild.loopback): loop2 acquired (locked: False)
mount/- (org.osbuild.xfs): mounting /dev/loop0 -> /home/thozza/devel/images/.osbuild/tmp/buildroot-tmp-sw23h3b8/mounts/
mount/boot (org.osbuild.xfs): mounting /dev/loop1 -> /home/thozza/devel/images/.osbuild/tmp/buildroot-tmp-sw23h3b8/mounts/boot
mount/boot-efi (org.osbuild.fat): mounting /dev/loop2 -> /home/thozza/devel/images/.osbuild/tmp/buildroot-tmp-sw23h3b8/mounts/boot/efi
Failed to open file "/sys/fs/selinux/checkreqprot": Read-only file system
copying '/run/osbuild/inputs/root-tree/.' -> '/run/osbuild/mounts/.'
cp: error writing '/run/osbuild/mounts/././boot/initramfs-0-rescue-ffffffffffffffffffffffffffffffff.img': No space left on device
Traceback (most recent call last):
  File "/run/osbuild/bin/org.osbuild.copy", line 29, in <module>
    r = main(_args, _args["options"])
  File "/run/osbuild/bin/org.osbuild.copy", line 22, in main
    subprocess.run(cmd + [src, dst], check=True)
  File "/usr/lib64/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['cp', '-a', '--reflink=auto', '/run/osbuild/inputs/root-tree/.', '/run/osbuild/mounts/.']' returned non-zero exit status 1.
[...]

PR #2200, opened 2026-02-16
rhel/centos-9: grow `/boot` for minimal-raw

supakeen:
Same thing that happened before happened again. Let's grow the size of the `/boot` partition some more for newer RHELs making a bigger step than before.

Inverts the condition so we can use greater than twice.

---

Closes: https://github.com/osbuild/images/issues/2176


supakeen:
Seems like there is non-determinism in the conditions with the way I wrote them. Will need more work I guess :)

thozza (approved):

lzap (approved):
Probably my longest review of a dozen of YAML lines. This is getting wild but I cannot think of a better way of doing this.

PR #2206, opened 2026-02-17
many: add GPG key resolving and source

lzap:
This uses the recently refactored `remotefile` package and adds a bit of code that can be used to resolve and add any URLs to the sources. Then, it uses this for custom repo GPG keys which are URLs to resolve them.

This by itself does not do anything but adds a new curl source and stops actually trying to import URLs in the RPM stage which fixes the bugreport. I added NVidia GPG key to validate the functionality, here is the diff:

```diff
diff -uNr /tmp/tmpdck0irw4/ref/fedora_42-x86_64-generic_ami-f42_x86_cuda_drivers.json /tmp/tmpdck0irw4/new/fedora_42-x86_64-generic_ami-f42_x86_cuda_drivers.json
--- /tmp/tmpdck0irw4/ref/fedora_42-x86_64-generic_ami-f42_x86_cuda_drivers.json 2026-02-17 16:17:10.863873257 +0100
+++ /tmp/tmpdck0irw4/new/fedora_42-x86_64-generic_ami-f42_x86_cuda_drivers.json 2026-02-17 16:17:16.722920871 +0100
@@ -571,6 +571,7 @@
         "sha256:16ea7e9de7369174eaa9fef63e90b9e8d57f495eb5006277505d5d0c0bf89c47": {
           "url": "https://rpmrepo.osbuild.org/v2/mirror/public/f42/f42-x86_64-updates-released-20260125/packages/grub2-pc-0:4-0.pkgset~os^trans~0.x86_64.rpm"
         },
+        "sha256:27e46a2d43e125859fb8a62c3b75bf798aeb95fa6f7d9bf790c1167ed9a0b39c": "https://developer.download.nvidia.com/compute/cuda/repos/fedora42/x86_64/D42D0685.pub",
         "sha256:3e5a5ce99777696ff5b30f79a2ce9edb40b021468675e8cece7a5a8184db77b5": {
           "url": "https://rpmrepo.osbuild.org/v2/mirror/public/f42/f42-x86_64-fedora-20250512/packages/exclude:geolite2-country-0:6-2.pkgset~os^trans~0.x86_64.rpm"
         },

```

I will continue on integrating this with https://github.com/osbuild/osbuild/pull/2326 which adds ability to pull GPG key from sources. This could be merged as well as a standalone patch if we want but I think I will add the functionality on top.

Fixes: https://issues.redhat.com/browse/HMS-10216


thozza:
Adding an unused source to a manifest is IMHO a bug, so maybe let's hold off with this PR until the source is actually used in the manifest by a stage?

supakeen (commented):
Needs rebase.

PR #2210, opened 2026-02-18
fedora: iot installers use erofs after 44

supakeen:
Let's switch over the Fedora IoT installers to erofs to align them with the other Fedora artifacts produced by both `kiwi` and `image-builder`.

Note that since Fedora IoT is an artifact that is produced as part of the Fedora composes we don't retroactively change previous versions.


bcl:
I like this. And I also want to point out the amount of yaml being changed vs. code :)

supakeen:
> I like this. And I also want to point out the amount of yaml being changed vs. code :)

Yea it's quite neat (if you only look at the last commit here) to be able to switch these things around *with* distro conditions in 10 lines.

lzap (approved):

PR #2211, opened 2026-02-18
fedora: live installers use erofs

supakeen:
Note that since the Fedora Live installers are not currently part of any composes this change applies to all versions.

---

Based on #2209 so we have the default shared erofs options. Keeping as draft until that lands.


supakeen:
Still relevant, just didn't have time to adapt our usage of /tmp.

PR #2212, opened 2026-02-18
many: copy ErofsOptions in manifest

lzap:
The Go race detector identified a data race occurring when `serialize` is called. Now, we do not use `t.Parallel()` (I think) but we run tests with `go test -race` and it fails. This patch resolves the issue by converting **ErofsOptions** into a value type, allowing it to be copied safely. Additionally, the argument in **NewErofsStage** has been updated to a value type for consistency.

---

While our codebase frequently uses pointers to leverage the omitempty tag in the standard encoding/json library, we should avoid pointers in non-JSON types unless strictly necessary.

Benefits of using value types here:

* Thread Safety: Eliminates shared state risks during concurrent operations.
* Performance: Go is highly optimized for copying small structs; value types can be more efficient than pointer indirection and reduce GC pressure.
* Simplicity: Removes the need for nil-checks and dereferencing.

Go 1.25/26 experimental JSON v2 introduces the omitzero flag, which will eventually allow us to omit empty structs without needing pointers. Until that is standard, moving toward value types for internal logic is the preferred approach for safety.

Fixes: https://issues.redhat.com/browse/HMS-10222

Edit: Reworded the description a bit.

### DIFF
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -492,7 +492,7 @@ func isoCustomizations(t *imageType, c *blueprint.Customizations) (manifest.ISOC
 		}
 
 		if erofsOptions := isoConfig.ErofsOptions; erofsOptions != nil {
-			isc.ErofsOptions = erofsOptions
+			isc.ErofsOptions = *erofsOptions
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


supakeen (approved):
Do we ever call serialize concurrently? Otherwise fine change and I agree 🙂

lzap:
> Do we ever call serialize concurrently?

Yeah I badly worded this, we probably do not have `t.Parallel()` it is just the race detector checking for that.

PR #2213, opened 2026-02-18
check-host-config: add RunOn filtering and re-enable modularity check

lzap:
Add a RunOn field to the Metadata struct and use it to filter checks
based on the OS ID and version. The RunOn field is a list of OS IDs to
run the check on (prefix with `!` to exclude). The OS ID and version are
extracted from the /etc/os-release file.

Uses the functionality to only run modularity check on CentOS 9.

---

Also enables the modularity test and use filter to run it only on CentOS 9.

### DIFF
--- a/cmd/check-host-config/check/modularity_test.go
+++ b/cmd/check-host-config/check/modularity_test.go
@@ -10,65 +10,62 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// dnf module list --enabled output fixtures for different RHEL versions
-func dnfModuleListOutputRHEL7() string {
-	return "Last metadata expiration check: 1:23:45 ago on Mon 01 Jan 2024 12:00:00 PM UTC.\n" +
-		"Dependencies resolved.\n" +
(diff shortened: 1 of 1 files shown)


achilleas-k (dismissed):
Very nice.  Thank you!

LGTM

cmd/check-host-config/main.go
```diff
@@ -32,8 +34,31 @@ func waitForSystem(timeout time.Duration) error {
...
+	if len(runOn) == 0 || osRelease == nil {
+		return true
+	}
+
+	currentID := strings.ToLower(strings.TrimSpace(osRelease.ID + "-" + osRelease.VersionID))
+	var inclusions, exclusions []string
+	for _, entry := range runOn {
+		entry = strings.TrimSpace(entry)
+		if after, ok := strings.CutPrefix(entry, "!"); ok {
+			exclusions = append(exclusions, strings.ToLower(after))
```
avitova:
```suggestion
if strings.ToLower(target) == currentID {
    return false
}
```

PR #2215, opened 2026-02-19
bootc: directly embed OSCustomizations

supakeen:
Instead of duplicating a bunch of options we already have let's follow the idiom we do for other pipeline generators and put the customizations directly into the relevant struct.

---

Am I missing something here?

cc @jbtrystram (since we spoke about this on Matrix and you might be working on things related to this).


bcl (approved):
Looks good

PR #2222, opened 2026-02-23
bootc: allow ignition first boot stage in blueprint

jbtrystram:
This allow setting a `Firstboot.Ignition.Enable` boolean to inject the igntition firstboot stage[1] after the container is deployed.

Requires https://github.com/osbuild/blueprint/pull/46 [1] https://github.com/osbuild/osbuild/blob/main/stages/org.osbuild.ignition

### DIFF
--- a/pkg/manifest/raw_bootc.go
+++ b/pkg/manifest/raw_bootc.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/osbuild/images/internal/common"
 	"github.com/osbuild/images/pkg/artifact"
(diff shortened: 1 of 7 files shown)


supakeen:
I've done a release of the blueprint library so you can update the dependency here and the option will be available in the code.

pkg/manifest/raw_bootc.go
```diff
@@ -265,15 +265,22 @@ func (p *RawBootcImage) serialize() (osbuild.Pipeline, error) {
...
+			ignitionStage = osbuild.NewIgnitionStage(&opts)
 		} else if p.OSCustomizations.Ignition.Empty {
-			postStages = append(postStages, osbuild.NewEmptyIgnitionStage())
+			opts := osbuild.IgnitionStageOptions{
+				Target: "mount://boot/",
+			}
+			ignitionStage = osbuild.NewIgnitionStage(&opts)
 		}
+		ignitionStage.Devices = devices
+		ignitionStage.Mounts = mounts
```
supakeen: Re-using the mounts here is the cause of your file not showing up where you expect it to. The mounts variable contains an ostree deployment mount additionally to the device mounts which overmounts whatever is in the ostree commit. This is a bit of a detail, and we might be mishandling the ostree deployment (it's necessary everywhere else I *think*).

A 'quick fix' is to re-create the mounts here locally in this scope, without appending the ostree deployment mounts:

```diff
diff --git a/pkg/manifest/raw_bootc.go b/pkg/manifest/raw_bootc.go
index 6b810b9e3..709d7a8fd 100644
--- a/pkg/manifest/raw_bootc.go
+++ b/pkg/manifest/raw_bootc.go
@@ -278,6 +278,12 @@ func (p *RawBootcImage) serialize() (osbuild.Pipeline, error) {
                        }
                        ignitionStage = osbuild.NewIgnitionStage(&opts)
                }
+
+               devices, mounts, err := osbuild.GenBootupdDevicesMounts(p.filename, p.PartitionTable, p.platform)
+               if err != nil {
+                       return osbuild.Pipeline{}, err
+               }
+
                ignitionStage.Devices = devices
                ignitionStage.Mounts = mounts
                postStages = append(postStages, ignitionStage)
```

jbtrystram: Ok I added your suggestion there with a couple of comments . Thanks

pkg/manifest/raw_bootc.go
```diff
@@ -264,16 +264,32 @@ func (p *RawBootcImage) serialize() (osbuild.Pipeline, error) {
...
 		} else if p.OSCustomizations.Ignition.Empty {
-			postStages = append(postStages, osbuild.NewEmptyIgnitionStage())
+			opts := osbuild.IgnitionStageOptions{
+				Target: "mount://boot/",
+			}
+			ignitionStage = osbuild.NewIgnitionStage(&opts)
+		}
+		var err error
+		// We cannot reuse the existing mounts because the generated ostree mounts are shadowing /boot and the file ends up
+		// in the wrong place. We reuse the bootupd mount generator as it's enough for this. We just need /boot.
```
supakeen: ❤️ for adding this in a comment as well.

pkg/manifest/raw_bootc.go
jbtrystram: > But longer term I think we should also handle it in bootupd, if other tools want to support generating bootc derivatives + ignition enabled out of the box.

Thinking about this some more I think I disagree. Bootupd touching the file means that bootc would need to learn to pass the flag to bootupd, so that would create coupling on two projects, just for an empty file.
Furthermore, when you think about it, Ignition and bootc are quite orthogonal. If we, as a distribution integrator want to offer ignition on a bootc image, fine but that's our soup to mix, and doing so with the tools we use to create the image, so here in image-builder, is the right place to do so IMHO.
But let's take this conversation further somewhere else maybe

supakeen: For what it's worth I feel that enabling ignition is by design unmanaged state; it's a toggle you set only once (either as a karg, or dropping a file) and not managed by the container (or by extension bootupd).

jbtrystram:
> This isn't tested anywhere but I trust you tried it already with your own builds.  We should add some FCOS manifests to test these things at some point.

I'll file something to track this as part of our overall effort of moving to image-builder

PR #2230, opened 2026-02-24
bootc: default kernel arguments for bootc types

supakeen:
The default kernel arguments we have defined, while helpful, get in the way of use cases such as [1].
    
Since this is a breaking change (people might rely on these arguments being present) we can't remove them entirely in one go.
    
Let's introduce an `ImageOptions` toggle to turn off the default kernel arguments. This allows us to implement a command-line toggle for this behaviour to `(bootc-)image-builder`.
    
I've opted for an `ImageOption` here because I believe that in the future we will no longer ship any default kernel arguments for bootc image types; instead having users define them [2].
    
Having the argument allows us to emit a warning in `(bootc-)image-builder` that the behavior will change in the future and to opt in to the new behavior the user can pass the toggle.
    
This fits better than a blueprint option as those try to keep backwards compatibility much stricter.
    
[1]: https://github.com/osbuild/bootc-image-builder/issues/824
[2]: https://docs.fedoraproject.org/en-US/bootc/kernel-args/

### DIFF
--- a/pkg/distro/distro.go
+++ b/pkg/distro/distro.go
@@ -162,6 +162,14 @@ type ImageType interface {
 
 type BootcImageOptions struct {
 	InstallerPayloadRef string `json:"installer_payload_ref,omitempty"`
+
+	// We introduce default kernel arguments in bootc generic image types which
+	// can clash with certain user use cases. Normally we only allow appending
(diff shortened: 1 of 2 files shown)


bcl:
This can already be set (or is it extended?) via blueprint. Why add another cmdline argument that does the same thing?

PR #2231, opened 2026-02-25
bootc: resolve much less for build containers

supakeen:
For the build container we don't need *any* information from inside the container at all. Just the information we have up front is enough.

Introduce a new resolve function for build containers specifically. Do this as a separate function because it feels nice to use this in the executables this way and it lives in the right place.

This resolves (haha) the following issue [1] after it lands in executables.

There's a larger story here that we probably shouldn't treat inspection errors as errors unless an image type *actually* needs the information.

[1]: https://github.com/osbuild/image-builder-cli/issues/464


ondrejbudai (commented):
Thanks for the fix! ❤️

PR #2232, opened 2026-02-26
boot-image: bump ISO boot timeout to 15 minutes

lzap:
I don't know about this :-) 

I feel bad.


achilleas-k:
> Well, I am merely trying to unblock CICD

Since we're not certain about the solution, either pull the snapshot updates into this PR or make the change on #2217 where boot tests are failing so we don't need to keep merging guesswork and rerunning.

PR #2235, opened 2026-02-27
vmtest: Fix timeout problems

bcl:
It ends up that when you use pull_event(wait=1.0) every call after the
first raises a QMPConnectError without waiting. This may be a qmp bug,
but I could not find reports of it.
    
So this changes the loop to implement the wait ourselves, calling
pull_event(wait=False) which will not wait and will return None if there
are no events. We sleep for 1s between calls and check for a timeout as
before.
    
This also bumps the schutzbot rngseed in order to have it regenerate the
manifests and re-trigger the tests in order to exercise this change in
the PR.

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -1,6 +1,6 @@
 {
   "common": {
-    "rngseed": 13,
+    "rngseed": 2026030200,
     "bootc-image-builder": {
       "ref": "quay.io/centos-bootc/bootc-image-builder@sha256:9893e7209e5f449b86ababfd2ee02a58cca2e5990f77b06c3539227531fc8120"
     },


supakeen:
> So apparently simply bumping rngseed doesn't cause it to generate new manifests for the boot tests? I still ended up with NullBuild, so I cherry picked the commits from #2217 in the hopes that SOME test runs and shows whether or not vmtest is fixed. 

We had quite a few prs bumping rng seed and some of them went unmerged. What I think is going on is that the number you picked already had manifests in the cache?

supakeen (approved):

lzap (dismissed):
Those CICD VM hosts must be swapping of something, this is slightly faster on my own VM at home.

PR #2236, opened 2026-02-27
imgtestlib: Make NullBuild more descriptive

bcl:
The NullBuild task doesn't actually tell you why it was generated, the logging of that is elsewhere. This changes it to actually tell you why there is no build.


achilleas-k (approved):
Good idea.  Thanks!

Issue #2149, opened 2026-01-21
Unsafe `append()` patterns in pipeline implementations


thozza:
## Problem

Several pipelines in `pkg/manifest` use `append()` on struct fields in ways that could cause subtle bugs.

In Go, `append(slice, items...)` will write to the existing backing array if it has sufficient capacity, rather than allocating a new one. This means `append(p.repos, p.ExtraRepos...)` could modify memory that other slices reference if `p.repos` has excess capacity.

Two related issues:

1. **In `getPackageSetChain()`** - `append(p.repos, p.ExtraRepos...)` returns a new slice but may have written `ExtraRepos` elements into `p.repos`'s backing array
2. **In `serializeStart()`** - `p.repos = append(p.repos, ...)` modifies the field, but `serializeEnd()` never resets it, so repos accumulate if serialization were called multiple times

## Affected Files

| File | getPackageSetChain | serializeStart |
|------|-------------------|----------------|
| `os.go` | L323, L535-536 | L506 |
| `anaconda_installer.go` | L224-226 | L249 |
| `coreos_installer.go` | L133-135 | L157 |
| `commit_server_tree.go` | L64-65 | L88 |
| `build.go` | - | L162 |

## Suggested Fix

Use `slices.Concat()` (Go 1.22+) which always allocates a new slice:

```go
// Before
Repositories: append(p.repos, p.ExtraRepos...),

// After  
Repositories: slices.Concat(p.repos, p.ExtraRepos),
```

For `serializeStart()`, either reset `p.repos` in `serializeEnd()` or store depsolve repos in a separate field.

PR #2237, opened 2026-02-28
pkg/manifest: use slices.Concat for repo slice merges

KunalKumar-1:
Replace append() with slices.Concat() when merging depsolveRepos and extra repos to avoid backing array mutation.

Fixes: #2149

### DIFF
--- a/pkg/manifest/anaconda_installer.go
+++ b/pkg/manifest/anaconda_installer.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
 
 	"github.com/osbuild/images/internal/common"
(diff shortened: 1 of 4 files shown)


supakeen (approved):
Thanks @KunalKumar-1 looks OK for me. You'll probably have to wait for the week to start for more reviews.

I think the 'fixes' is a bit optimistic, the issue mentions two problems and you address one of the two but I think it's probably better to fix the separate issue in `serializeStart` in a separate PR.

---

Were these the only locations where this pattern was used or are they the only ones mentioned in the issue?

achilleas-k (approved):
Nice.  Thank you!

PR #2244, opened 2026-03-03
fedora/atomic: include legal files in ISO root (HMS-10282)

supakeen:
Self-explanatory but we should include the legal files in the ISO root for the Atomic installers as well; we currently only do this for the network installer so let's align.

---

Note this applies to all Fedora versions as we didn't previously create these artifacts and will only start doing so in Fedora 45.


supakeen:
Force-pushed due to checksum conflicts.

avitova (approved):

PR #2248, opened 2026-03-04
Refactor how iso bootloaders are setup to make it easier to add more

bcl:
Add an interface and helper functions to make setup of bootloaders easier. No checksums were harmed in this PR except one, and it is a bug fix.


pkg/manifest/efi_boot_tree.go
```diff
@@ -118,3 +120,36 @@ func (p *EFIBootTree) serialize() (osbuild.Pipeline, error) {
 	pipeline.AddStage(grub2Stage)
 	return pipeline, nil
 }
+
+// GetISOBootStages returns stages needed to setup booting from an ISO
+// This creates the efiboot.img, copies the EFI files into it, and copies
+// them to the /EFI directory of the pipeline
```
lzap: The contract is very x86_64/arm64 centric but I guess that is the world we are living in, fine :-)

PR #2258, opened 2026-03-11
many: installer payload kickstart location (HMS-10171)

supakeen:
This PR implements the optional ability to write the *payload* bits of a kickstart into the interactive defaults kickstart. This is a special kickstart for which values contained in it do not cause the 'incomplete kickstart'.

The functionality can be toggled with a value in the YAML definitions.

When turned on any *non* payload bits of the kickstart still go into the `osbuild.ks` kickstart that is placed at the root of the ISO, and added to the kernel arguments. If there are no customizations the installer will *not* have a kickstart at the root and will not have additional kernel arguments.

Currently implemented only for the ostree payload; however, the design has now been made so that this should be relatively easy to port over to the rest of the payloads if necessary in the future.

I've also done some of the clean-up/minor fixes that were part of the previous PRs related to this. I've split the ostreeCommit and ostreeKickstart stages; something we'll be able to do for other payloads as well; I also plan to (in a follow-up) move most of the logic of 'do this or do that based on these enums' into a separate function.

For that however I need to see the full picture. The next PR related to the installers will bring flatpak embeddings. Once that is in I feel more confident that I can oversee the entire design and improve where applicable.

The last commit enables this for all Fedora Atomic installers; since this is what they currently use when built with Lorax this brings us parity. We don't deliver those as part of any compose at the moment so it's applied to everything.

In the future we might want to do the same for Fedora IoT but that will be a separate PR.

---

This will need rebasing if #1604 makes it in first.

---

Before:

Screenshot_20260311_101605

After:

Screenshot_20260311_095244


bcl (approved):
Looks good

PR #2267, opened 2026-03-16
blueprint: accept dots in datasizes

avitova:
Allows the use of dots (e.g. 1.5 GiB) in YAML and blueprint configurations. Previously, only integers were accepted.


supakeen:
> I think that `datasize.Size` should be kept `unit64` internally. We can still support parsing float strings.

I agree, let's add the ability to parse float representations but keep the type an uint.

lzap (approved):
I wonder if that works in non-US locales where decimal is not actually a _point_... :-D

Whatever, this is enough.

PR #2271, opened 2026-03-18
Add support for generating bootc manifests with remote (skopeo) container sources. (HMS-10296)

thozza:
Service backends like osbuild-composer need to pull bootc containers via osbuild's `org.osbuild.skopeo` source rather than expecting them in local podman storage (`org.osbuild.containers-storage`).

This PR adds a `UseRemoteContainerSource` option to `BootcImageOptions` and wires it through all bootc manifest generators, along with corresponding `gen-manifests` CLI flags (`--bootc-remote`, `--bootc-installer-ref`) so external CI can produce manifests for either source path. This should prepare the ground for boot-testing derived containers later in the bootc-foundry repository.

## Architectural Changes

The new `UseRemoteContainerSource` boolean on `BootcImageOptions` controls the `Local` flag on every `container.SourceSpec` created by the `manifestFor*` methods. A `useLocalStorage()` helper centralizes the nil-check and negation so each generator doesn't repeat it. The zero value (`false`) preserves the existing BIB-compatible behavior, making the change fully opt-in.

## Key Changes

- Fix mock container resolver to propagate `SourceSpec.Local` → `Spec.LocalStorage`, correcting previously incorrect skopeo-sourced bootc manifests
- Add `UseRemoteContainerSource` field to `BootcImageOptions` and wire it through all five `manifestFor*` methods (disk, ISO, generic ISO, legacy ISO, PXE)
- Pass `ImageOptions` to `manifestForLegacyISO` for consistency with all other generators
- Add `--bootc-remote` and `--bootc-installer-ref` CLI flags to `gen-manifests` for external CI use and convenience
- Add `bootc-remote-empty` and `bootc-remote-with-payload` build configs with corresponding manifest checksums

## Breaking Changes

This PR is fully backward compatible. The default value of `UseRemoteContainerSource` (`false`) preserves the existing `containers-storage` behavior. Existing manifest checksums change because the mock resolver bug fix now correctly produces `containers-storage` sources for the default (local) path.

## Testing

Unit tests cover the new option for all bootc image types by generating manifests with both `local=true` and `local=false` and asserting the expected osbuild source type (`org.osbuild.containers-storage` vs `org.osbuild.skopeo`) appears in the serialized output. New build configs (`bootc-remote-empty`, `bootc-remote-with-payload`) exercise the remote path through `gen-manifests --fake-bootc`, with checksums validated in CI. Legacy ISO types are skipped in the unit test since they require real distro definitions not available in the generic test distro, but are covered by the build config path with real distros.

### DIFF
--- a/cmd/gen-manifests/main.go
+++ b/cmd/gen-manifests/main.go
@@ -235,6 +235,7 @@ func makeManifestJob(
 	metadata bool,
 	tmpdirRoot string,
 	bootcRemote bool,
+	bootcInstallerRef string,
 ) manifestJob {
 	name := bc.Name
 	distroName := distribution.Name()
@@ -249,14 +250,19 @@ func makeManifestJob(
 	}
 
 	options := bc.Options
-	if bootcRemote {
+	if bootcRemote || bootcInstallerRef != "" {
 		if options.Bootc == nil {
 			options.Bootc = &distro.BootcImageOptions{}
 		} else {
 			bootcOpts := *options.Bootc
 			options.Bootc = &bootcOpts
 		}
-		options.Bootc.UseRemoteContainerSource = true
+		if bootcRemote {
+			options.Bootc.UseRemoteContainerSource = true
+		}
+		if bootcInstallerRef != "" {
+			options.Bootc.InstallerPayloadRef = bootcInstallerRef
+		}
 	}
 
 	var bp blueprint.Blueprint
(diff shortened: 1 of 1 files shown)


test/config-list.json
```diff
@@ -739,6 +739,38 @@
...
+    "filters": {
+      "distros": [
+        "bootc-*"
+      ],
+      "image-types": [
+        "bootc-installer",
+        "bootc-generic-iso"
+      ]
+    }
+  },
```
lzap: We really need to remove those once we add proper image tests for bootc.

I think we will have to update image configs because each distro/version has a different image ref. This is good enough for now.

PR #2274, opened 2026-03-20
Retain blueprint as `/root/blueprint.toml` in built images

ochosi:
## Why

This change follows the established patter of Anaconda which retains a kickstart in `/root` and a similar pattern in bootc disk images that retains Containerfiles in `/root`.
Today, users who receive a built image have no easy way to know what blueprint produced it or to rebuild it. The RHSM facts we already embed provide some provenance metadata (API type, blueprint ID), but not the full blueprint itself.

This PR writes the complete blueprint to `/root/blueprint.toml` in every built image. This serves two purposes:

1. **Provenance** -- anyone inspecting a running system can see exactly what customizations were requested. This is useful for auditing, debugging, and support.
2. **Reproducibility** -- a user can extract the blueprint and feed it back into the build tooling to produce a similar image. This is not SLSA3-grade provenance (the blueprint is not signed, and the exact package versions are not pinned in it), but it covers the common case of "I have this image, how do I make another one like it?"

## What

- A new `BlueprintTOML []byte` field on `OSCustomizations` carries the TOML-encoded blueprint through the manifest pipeline.
- `osCustomizations()` encodes the blueprint using `toml.Marshal()` (the `Blueprint` struct already has `toml:"..."` tags, and `BurntSushi/toml` is already a dependency).
- `OS.serialize()` writes it to `/root/blueprint.toml` via the existing inline-data copy stage mechanism (same pattern as modularity failsafe files).
- User passwords (`Customizations.User[].Password`) are redacted before marshaling -- the blueprint is deep-copied and passwords are nilled out so they never appear in the image, whether plaintext or hashed.

## Design decision

Since this is a proposal, I opted for a minimal implementation. If we decide we want to go ahead and do this, but the architecture is wrong, I'm happy to rework this PR.

## Testing

- [x] Built a test image manually with `image-builder-cli` and verified `/root/blueprint.toml` exists with expected content
- [x] Unit test verifying the serialized pipeline contains a copy stage targeting `/root/blueprint.toml`
- [x] Unit tests for TOML encoding and password redaction in `osCustomizations()`
- [x] Regenerate manifest checksums (all manifests will change due to the new stage)


supakeen (commented):
Looking at this purely from a functional standpoint, not the implementation:

This is nice to have though there are some concerns.

One is that we'd need to have this functionality gated behind an `ImageOption` (`BuildOption`) so we can turn it off for any images produced in build systems that use customizations (e.g. RHUI, LVMified images for clouds, etc). Initially off by default for the cli as we will want to plumb through/make the option available in `koji-image-builder` and `pungi` for those systems while the service can toggle it on.

The other is secrets, any file, first-boot, or many of the other customizations can potentially contain secrets. While those secrets would usually be able to be found by `root` anyhow it'd need quite some thought.

lzap:
While it is a valid security concern, Anaconda AFAIK does store KS file in `/root` including passwords. They are usually hashed tho (not sure if it accepts plaintext anymore).

And do we really want to hide it for the generic images like RHUI etc? This would be a nice "advertisment" specifically if we add a nice comment block about how to use image builder and the blueprint to build a new image :-)

PR #2285, opened 2026-04-01
installer flatpaks from blueprint (HMS-10441)

supakeen:
Uses the recently introduced `installer.payload.flatpaks` customization to allow overriding of Flatpaks that are used in an ostree installer.

This is required because Flatpaks in Fedora do not follow the normal release cadence so we can't rely on `{{.MajorVersion}}` by default and instead they need to be passed along in the `pungi` config.


supakeen:
People can take a look at the first bits here but I'll be adding a few quality of life fixes here as well that were mentioned during earlier PRs (specifically: the URL formatting).

PR #2291, opened 2026-04-06
network-installer: Do not set timezone or locale in Kickstart

bcl:
The network installer does not use these values to create a kickstart. The blueprint Locale is used for the locale of the booted iso, not the installed system. The timezone customization is not used at all. But setting these triggers creation of an empty osbuild.ks which causes Anaconda to crash.

Remove both of them, and remove locale from the list of supported blueprint options in the image definitions yaml.

Related: HMS-10469


bcl:
Moved to #2295

PR #2293, opened 2026-04-07
distro: basic variable substitution (HMS-10479)

supakeen:
For Flatpak references defined in blueprints (specifically in `pungi`) the architecture should be templatable.

This implements a very narrow approach of only replacing `$arch` and *only* on Flatpak references coming from the customizations.

In the future we can have a discussion when there's no time pressure on if this should grow into basic substitution everywhere or if we should do this differently.

```toml
[[customizations.installer.payload.flatpaks.force]]                                                                                                                                                                                                                                                                                                                                                                                       
registry = { remote_name = "fedora", url = "oci+https://registry.fedoraproject.org" }                                                                                                                                                                                                                                                                                                                                                     
references = [                                                                                                                                                                                                                                                                                                                                                                                                                            
    "runtime/org.fedoraproject.KDE6Platform/$arch/f44"                                                                                                                                                                                                                                                                                                                                                                                    
]
```

### DIFF
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/osbuild/blueprint/pkg/blueprint"
+	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/container"
 	"github.com/osbuild/images/pkg/customizations/anaconda"
 	"github.com/osbuild/images/pkg/customizations/bootc"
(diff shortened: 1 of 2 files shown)


avitova (commented):
Nice. Can you also add some tests featuring replacing the string, but also some checking that we do not replace the string outside of Flatpak? The latter one is more for the future so that we do not do this by accident.

supakeen:
> Nice. Can you also add some tests featuring replacing the string, but also some checking that we do not replace the string outside of Flatpak? The latter one is more for the future so that we do not do this by accident.

It seems a bit prohibitive to add tests to all properties that might come from the blueprint to ensure they don't replace a variable, also hard to keep track of in the future as more things get added.

I've added a basic test case for the replacement function. Testing if the correct source specs are being generated is a bunch of work that I'd prefer to do in a follow-up if I may 🙂.

lzap (approved):
I mean why not to do it everywhere, the character can be URL-escaped anyways if needed.

supakeen:
> I mean why not to do it everywhere, the character can be URL-escaped anyways if needed.

Because then I think we should put a bit more thought into it and a bit of design 🙂

avitova (approved):
Okay, great. We agreed that tests might be a followup, I am good here.

PR #2294, opened 2026-04-07
fedora: prepare for 42 EOL

supakeen:
DRAFT


brlane-rht:
This should also bump the versions used in the github workflows.

PR #2295, opened 2026-04-07
network-installer: Do not set timezone or locale in Kickstart

brlane-rht:
The network installer does not use these values to create a kickstart.
The blueprint Locale is used for the locale of the booted iso, not the
installed system. The timezone customization is not used at all. But
setting these triggers creation of an empty osbuild.ks which causes
Anaconda to crash.
    
This removes the code that was setting these in the Kickstart struct.
Locale is still set in the installer using org.osbuild.locale
    
This includes new manifest tests that only set customizations.locale in
the blueprint which will create an org.osbuild.locale stage, but NOT an
org.osbuild.kickstart stage.
    
Resolves: HMS-10469


croissanne (approved):
thank you :)

croissanne:
> Remove both of them, and remove locale from the list of supported blueprint options in the image definitions yaml.

@brlane-rht wait, is this missing from the PR? I can't see any change to the image definition yaml

brlane-rht:
> @brlane-rht wait, is this missing from the PR? I can't see any change to the image definition yaml

ugh. github doesn't update the PR description after the first push, locale is still allowed in the yaml because it can be used to set the iso's language. I'll fix the description to match the current state :)

supakeen:
> The network installer does not use these values to create a kickstart. The blueprint Locale is used for the locale of the booted iso, not the installed system. The timezone customization is not used at all. But setting these triggers creation of an empty osbuild.ks which causes Anaconda to crash.
> 
> This removes the code that was setting these in the Kickstart struct. Locale is still set in the installer using org.osbuild.locale
> 
> This includes new manifest tests that only set customizations.locale in the blueprint which will create an org.osbuild.locale stage, but NOT an org.osbuild.kickstart stage.
> 
> Resolves: HMS-10469

What I'm confused about is; don't we *want* to create a kickstart when the timezone or locale is provided in customizations? AFAIK it's OK to have a kickstart for customizations, just not if none are provided.

Shouldn't in that case the fix be the other way around, to actually propagate these values in the kickstart?

brlane-rht:
Not for the network installer, if users want to set those on the installed system they will pass it their own kickstart.

supakeen:
> Not for the network installer, if users want to set those on the installed system they will pass it their own kickstart.

Mmm; I thought we *did* want to generate a kickstart for customizations if those customizations are explicitly passed in the blueprint without dumping all of that one the user. Let's figure this out because otherwise it looks good to me so it's really just the 'what do we want'.

cc @achilleas-k can you take a look at what we want here?

brlane-rht:
I'd argue for not making any kickstart for netinst, and removing all the customizations. Once a kickstart is created that makes it harder for the user to use the netinst with their own kickstart. The goal here is to make a boot.iso that is the equivalent of the one created by Lorax.

achilleas-k:
TL;DR: Let's fix the current problem without changing behaviour and figure out the proper installer behaviour separately.  That means not adding timezone and locale to the kickstart for the network installer for now.

---

There's a question of purpose here, for the image type and the IB experience of building it more generally.

Our first ISOs (edge and image installer) always had a kickstart because they were built with predetermined payloads, the kickstart was already there, so adding customizations to it was low friction and was compatible with the idea of building a single-purpose installer.  It also let users build ISOs that would run automated, non-interactive installations.

> The goal here is to make a boot.iso that is the equivalent of the one created by Lorax.

This now is addressing a very different problem or request than the previous ones.  If the net installer is targeted only at this requirement (building the official network installer), then it's ok to not support any customizations (like some internal-only cloud images we have).  But as soon as we make it user-facing, we would want to make some things configurable.  Supporting the locale makes sense in more than one case: As a distro/spin maintainer, you may want to build multiple versions of the general-purpose installer with different default locales.  But if you're an end-user configuring a single- / specific-purpose installer, you might want the locale to also be part of the kickstart, setting a default locale for the system-to-be-installed as well.

> Once a kickstart is created that makes it harder for the user to use the netinst with their own kickstart.

True, and with the other installers, we've tried a few things to make this less of a problem, by separating the kickstart into necessary bits (a payload kickstart) and the customizations (the user kickstart), or by adding the `kickstart.contents` blueprint option that lets users write their kickstart directly during a build.  That second one is both a convenience that saves users from needing to modify an ISO (or the boot kernel args) after it's built and, in the case of payload ISOs, also saves them from needing to learn how the user kickstart might interact with the payload one.

The question for me is how do we reconcile the two requirements of customizing the installation experience itself (i.e. the Anaconda environment language, keyboard, timezone) and the target environment (the defaults for the installed system).  In the single-user, single-purpose case, it makes sense to assume that the keyboard layout they want during installation should probably be the default for the target system.  In the distro-maintainer case, it doesn't.  I think the approach we took for payload installers, which were all our installers so far, makes sense as it's easy to argue that those are single-purpose installers.  The network installer muddies this a lot and we need to think about it more generally.

For now I think let's fix the current problem (empty kickstart) with as little change to the behaviour as possible.  That would mean not adding the locale to the net installer kickstart.  It might be confusing for users of the other installers we build, but I'm more leaning towards being conservative with what we support initially and opening it up later to avoid needing to do the opposite, which can be more disruptive later.

supakeen (approved):
Thanks; in that case let's go ahead with this.

achilleas-k (approved):

PR #2296, opened 2026-04-09
anaconda: use `bare-user` for ostree (HMS-10504)

supakeen:
While doing scratch builds of the Fedora Atomic variants in Koji the ostree repository embedded in the root filesystem gets corrupted when embedding a rawhide commit.

The corrupted file is labelled `unlabeled_t` in our repository while in a Lorax-produced ISO it is labeled `accountsd_share_t`.

This is likely because the host system (or we) don't know about this SELinux label and fall back to a default; I'm not entirely sure on the exact mechanism.

Previously this wasn't a problem since we only stored `ostree` commits outside the root filesystem and the ISO filesystem itself doesn't do any labeling.

Switching to `bare-user` allows us to store SELinux contexts in a way that the host systems enforcement or knowledge does not matter. As a secondary benefit it also allows us to pull the `ostree` commits rootless; doesn't matter at the moment, it might in the future.

This only affects Fedora Atomic variants as they are the only ones embedding the commit in the root filesystem at the moment.

---

I've tested this by rebuilding an ISO that was previously affected and it now passes. I've also tested that the flatpaks still function.


thozza (approved):
LGTM

croissanne (approved):

PR #2297, opened 2026-04-09
pkg: add rpmlist writer for koji format

avitova:
There are still tests missing, and I want to think if we can unify this somehow with the sbom writer and reuse some lines.

Adds parsing of packages in order to provide koji rpm list in the format it expects. this works similarly to how the sbom json file is created.

### DIFF
--- a/pkg/manifestgen/manifestgen.go
+++ b/pkg/manifestgen/manifestgen.go
@@ -22,6 +22,7 @@ import (
 	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/ostree"
 	"github.com/osbuild/images/pkg/reporegistry"
+	"github.com/osbuild/images/pkg/rpmlist"
 	"github.com/osbuild/images/pkg/rpmmd"
 	"github.com/osbuild/images/pkg/sbom"
 )
@@ -82,6 +83,8 @@ type Options struct {
 	// Use the a bootstrap container to buildroot (useful for e.g.
 	// cross-arch or cross-distro builds)
 	UseBootstrapContainer bool
+
+	RPMListWriter RPMListWriterFunc
 }
 
 // Generator can generate an osbuild manifest from a given repository
@@ -105,6 +108,7 @@ type Generator struct {
 	overrideRepos []rpmmd.RepoConfig
 
 	useBootstrapContainer bool
+	rpmlistWriter         RPMListWriterFunc
 }
 
 // New will create a new manifest generator
@@ -126,6 +130,7 @@ func New(reporegistry *reporegistry.RepoRegistry, opts *Options) (*Generator, er
 		customSeed:             opts.CustomSeed,
 		overrideRepos:          opts.OverrideRepos,
 		useBootstrapContainer:  opts.UseBootstrapContainer,
+		rpmlistWriter:          opts.RPMListWriter,
 	}
 	if mg.depsolve == nil {
 		mg.depsolve = DefaultDepsolve
@@ -245,7 +250,8 @@ func (mg *Generator) Generate(bp *blueprint.Blueprint, imgType distro.ImageType,
 		return nil, err
 	}
 
-	if mg.sbomWriter != nil {
+	if mg.sbomWriter != nil || mg.rpmlistWriter != nil {
+		uniquePackages := make(map[string]rpmmd.Package)
 		// XXX: this is very similar to
 		// osbuild-composer:jobimpl-osbuild.go, see if code
 		// can be shared
@@ -259,20 +265,56 @@ func (mg *Generator) Generate(bp *blueprint.Blueprint, imgType distro.ImageType,
 			}
 			// XXX: sync with image-builder-cli:build.go name generation - can we have a shared helper?
 			imageName := fmt.Sprintf("%s-%s-%s", dist.Name(), imgType.Name(), a.Name())
-			sbomDocOutputFilename := fmt.Sprintf("%s.%s-%s.%s", imageName, pipelinePurpose, plName, defaultSBOMExt)
-
-			var buf bytes.Buffer
-			enc := json.NewEncoder(&buf)
-			if err := enc.Encode(depsolvedPipeline.SBOM.Document); err != nil {
-				return nil, err
+			if mg.sbomWriter != nil {
+				sbomDocOutputFilename := fmt.Sprintf("%s.%s-%s.%s", imageName, pipelinePurpose, plName, defaultSBOMExt)
+				var buf bytes.Buffer
+				enc := json.NewEncoder(&buf)
+				if err := enc.Encode(depsolvedPipeline.SBOM.Document); err != nil {
+					return nil, err
+				}
+				if err := mg.sbomWriter(sbomDocOutputFilename, &buf, depsolvedPipeline.SBOM.DocType); err != nil {
+					return nil, err
+				}
 			}
-			if err := mg.sbomWriter(sbomDocOutputFilename, &buf, depsolvedPipeline.SBOM.DocType); err != nil {
+
+			if mg.rpmlistWriter != nil && pipelinePurpose == "image" {
+				addUniquePackagesFromPipeline(uniquePackages, depsolvedPipeline)
+			}
+		}
+
+		if mg.rpmlistWriter != nil {
+			if err := writeRPMList(mg.rpmlistWriter, uniquePackages); err != nil {
 				return nil, err
 			}
 		}
 	}
 
 	return mf, nil
+}
+
+func addUniquePackagesFromPipeline(unique map[string]rpmmd.Package, pipeline depsolvednf.DepsolveResult) {
+	for _, pkg := range pipeline.Transactions.AllPackages() {
+		var key string
+		key = pkg.Checksum.Value
+		if pkg.Checksum.Value == "" {
+			key = fmt.Sprintf("%s-%s-%s.%s", pkg.Name, pkg.Version, pkg.Release, pkg.Arch)
+		}
+		if _, exists := unique[key]; !exists {
+			unique[key] = pkg
+		}
+	}
+}
+
+func writeRPMList(writer RPMListWriterFunc, unique map[string]rpmmd.Package) error {
+	var packages rpmmd.PackageList
+	for _, pkg := range unique {
+		packages = append(packages, pkg)
+	}
+	rpmListJSON, err := rpmlist.EncodePackages(packages)
+	if err != nil {
+		return err
+	}
+	return writer("rpmlist.json", rpmListJSON)
 }
 
 func xdgCacheHome() (string, error) {
@@ -319,4 +361,6 @@ type (
 	FlatpakResolverFunc func(flatpakSources map[string][]flatpak.SourceSpec) (map[string][]flatpak.Spec, error)
 
 	SBOMWriterFunc func(filename string, content io.Reader, docType sbom.StandardType) error
+
+	RPMListWriterFunc func(filename string, content io.Reader) error
 )
--- a/pkg/rpmlist/rpmlist.go
+++ b/pkg/rpmlist/rpmlist.go
@@ -1,0 +1,61 @@
+package rpmlist
+
+import (
+	"bytes"
+	"encoding/json"
+
+	"github.com/osbuild/images/pkg/rpmmd"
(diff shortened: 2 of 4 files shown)


pkg/rpmmd/koji.go
```diff
@@ -0,0 +1,57 @@
...
+	Release     string `json:"release"`
+	Epoch       *uint  `json:"epoch"`
+	Arch        string `json:"arch"`
+	BuildTime   int64  `json:"buildtime"`
+	Size        uint64 `json:"size"`
+	PayloadHash string `json:"payloadhash"`
+}
+
+func PackageToKojiRpmListEntry(p Package) KojiRpmListEntry {
+	// hash in kiwi is calculated for a header
```
supakeen: You should probably compare to what we do in `osbuild-composer`. This comment isn't entirely clear what you mean 🙂

Also should we check if hash is still empty after line 25?

pkg/rpmlist/rpmlist.go
supakeen, line 1: I would personally put these in a new `pkg/rpmlist` instead of inside `rpmmd` and calling them `koji`.

pkg/manifestgen/manifestgen.go
```diff
@@ -259,15 +251,28 @@ func (mg *Generator) Generate(bp *blueprint.Blueprint, imgType distro.ImageType,
...
+				}
+				if err := mg.sbomWriter(sbomDocOutputFilename, &buf, depsolvedPipeline.SBOM.DocType); err != nil {
+					return nil, err
+				}
 			}
-			if err := mg.sbomWriter(sbomDocOutputFilename, &buf, depsolvedPipeline.SBOM.DocType); err != nil {
-				return nil, err
+
+			if mg.rpmlistWriter != nil {
+				rpmListDocOutputFilename := fmt.Sprintf("%s.%s-%s.%s", imageName, pipelinePurpose, plName, rpmListExt)
```
supakeen: We need to think about this because having files per-purpose/per-pipeline means the logic of which files we deserialize has to live inside the `koji-image-builder` plugin which might not have the knowledge to do so.

It'd be fancier if we could only have a single `rpmlist` output here that includes all packages that were used in any image-related pipelines (perhaps a naive skipping of the build pipeline and merging all others would work, cc @thozza?).

pkg/rpmmd/koji.go
avitova: But I was a little bit confused about the kind of hash kiwi expects as well: https://github.com/koji-project/koji/blob/67cc2cdbac97d3af96ac2b2a4a34b5fe84a3435e/plugins/builder/kiwi.py#L276

Looking at the comment, I think it is deprecated. Previously, I used the header checksum, but later I came to the conclusion that `sigmd5` refers to the RPM package checksum, and not its header. Good point, I will remove that comment.

avitova: About what we do in `osbuild-composer`, I think that might be a little bit different. For example, we have a "signature" in the struct, while it is not even shown by Koji, and kiwi implementation does not have it either https://github.com/koji-project/koji/blob/master/plugins/builder/kiwi.py#L433.
Also, we do not use `Buildtime` there. What we can use as well is deduplication, although I am not really sure if that is needed.
So I believe if we pass this data down all the way to our koji image builder plugin in this format, we are good. I do not see why deduplication would be needed yet, so I did not put it there.

pkg/manifestgen/manifestgen.go
avitova: Ooooh, I understand. Yeah, I do not quite understand why we need to know the pipeline purpose for sbom. Why do we skip build pipelines, though?

Let's see what Tomas says, but I think it would make sense to just collect all the packages without any context of the pipeline, in this case also deduplicate them, and then create the JSON file with RPM packages. :)

avitova:
So I did [this](https://github.com/osbuild/images/pull/2297/changes/6e2233d8cbe977c2cf40776ebb84ee54c55df06b), and we can discuss which way we like more.
I think I prefer your idea of putting rpm packages all into one file.

pkg/manifestgen/manifestgen.go
supakeen: We can discuss this but as I understand it the RPM list in Koji would contain the RPMs that are contained inside the artifact; not the ones use to produce it.

Let's double check with @Conan-Kudo / @AdamWill?

Conan-Kudo: Yes. It's for what's _in_ the image, not what _made_ the image. Koji itself already knows what _made_ the image.

avitova: Thank you, that makes sense:)

supakeen (commented):
Looks OK to me; feel free to undraft so others can take a look as well and then I'll re-review too.

Your last commit has the wrong description; the title is 'only build rpms' but you're including ony 'image' RPMs.

brlane-rht (commented):
Is there documentation for what format koji expects? If so I'd add a link to it in the module docstring for koji.go -- or at least a pointer to the koji code that will be reading the output if there isn't proper documentation.

avitova:
As far as I was able to find out, we are aiming for something like [this](https://riscv-koji.fedoraproject.org/koji/taskinfo?taskID=136875#).
I did not find any information on this in the documentation. Kiwi then does something like [this](http://github.com/koji-project/koji/blob/master/plugins/builder/kiwi.py#L263).

pkg/manifestgen/manifestgen.go
```diff
@@ -275,6 +291,31 @@ func (mg *Generator) Generate(bp *blueprint.Blueprint, imgType distro.ImageType,
...
+func writeRPMList(writer RPMListWriterFunc, unique map[string]rpmmd.Package) error {
+	var packages rpmmd.PackageList
+	for _, pkg := range unique {
+		packages = append(packages, pkg)
+	}
+	rpmListJSON, err := rpmmd.EncodePackages(packages)
+	if err != nil {
+		return err
+	}
+	return writer("rpmlist.json", rpmListJSON)
```
lzap, line 317: It is a pity that using an interface it means the buffer will be held in memory until read. We already fight with increased memory consumption around package lists and SBOMs. But not something for this particular PR.

lzap (dismissed):
Code looks clean, I wish we could do this with less allocations tho but that is not on the agenda for this PR.

avitova:
Sorry, I ended up adding those references to the comment before EncodePackages function. I hope that's okay:)

pkg/rpmlist/rpmlist.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/osbuild/images/pkg/rpmmd"
+)
+
+// rpmlist item entry expected by koji
+type KojiRpmListEntry struct {
+	Name        string `json:"name"`
+	Version     string `json:"version"`
+	Release     string `json:"release"`
+	Epoch       *uint  `json:"epoch,omitempty"`
```
supakeen: We probably don't want the omitempty; having the `null` there is useful when we read it in on the Pyton side.

avitova: fixed!

lzap:
Checksums...

supakeen (dismissed):
There aren't actually checklist changes when I check out this PR locally; and the commit that the CI is pointing at only adds a comment.

I've rebased and am approving expecting the check to pass.

---

Mmm check didn't pass; but still passes locally:

```
€ ./tools/gen-manifest-diffs
no revision given as argument, running against merge base
found upstream merge base '3be6df487c52ef1f2d56e601f8905535528f46f2'

Note that this diff does *not* include depsolved rpms,
so upstream dependency changes will not be caught

calculating diff against '3be6df487c52ef1f2d56e601f8905535528f46f2'
Cloning into '/tmp/tmpu5nb_rnx/images-3be6df487c52ef1f2d56e601f8905535528f46f2'...
done.
Note: switching to '3be6df487c52ef1f2d56e601f8905535528f46f2'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 3be6df487 network-installer: Do not set timezone or locale in Kickstart
no diff found to '3be6df487c52ef1f2d56e601f8905535528f46f2'
images € 
```

I think I've seen this before but forgot to take a note, cc @achilleas-k?

brlane-rht (dismissed):
Looks good. I also don't know what's up with that checksum test. Works fine here.

pkg/rpmmd/koji.go
```diff
@@ -0,0 +1,57 @@
...
+	"bytes"
+	"encoding/json"
+)
+
+// rpmlist item entry expected by koji
+type KojiRpmListEntry struct {
+	Name        string `json:"name"`
+	Version     string `json:"version"`
+	Release     string `json:"release"`
+	Epoch       *uint  `json:"epoch"`
```
thozza: Why is this a pointer? Epoch is always `0` if not set. Using pointer here is IMHO weird.

pkg/rpmmd/koji.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	buildTime := int64(0)
+	if !p.BuildTime.IsZero() {
+		buildTime = p.BuildTime.Unix()
+	}
+	return KojiRpmListEntry{
+		Name:        p.Name,
+		Version:     p.Version,
+		Release:     p.Release,
+		Epoch:       &p.Epoch,
```
thozza: So `Epoch will always be non-`nil`.

pkg/rpmmd/koji.go
```diff
@@ -0,0 +1,57 @@
...
+		Arch:        p.Arch,
+		BuildTime:   buildTime,
+		PayloadHash: hash,
+		Size:        p.DownloadSize,
+	}
+}
+
+// EncodePackages returns JSON in a format that Koji expects for the given package list
+func EncodePackages(packages PackageList) (*bytes.Buffer, error) {
```
thozza: This is a very unfortunate function name, especially given that it is in the `rpmmd` module. So we end up with `rpmmd.EncodePackages()` that serializes into Koji-specific format. That is quite unexpected from my PoV.

I would suggest to either rename the function to something that clearly expresses what is its purpose or, my preferred solution, move this to a `koji` or `rpmlist` standalone module.

pkg/rpmmd/koji.go
```diff
@@ -0,0 +1,57 @@
+package rpmmd
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+// rpmlist item entry expected by koji
+type KojiRpmListEntry struct {
```
thozza: There is no need to export this type, it is not used anywhere outside of the module.

thozza (changes requested):
Thanks for the work and sorry for the late review, especially given that I'm going to request changes.

Some of my comments are inline and one is fixed by later commits (the move of the new code to a separate module, which is great).

On a general note, I would propose to split the first commit into "adding the new functionality" and in the second commit "wire up things in the `manifestgen` package". Also some changes should be really squashed into a single commit and the whole PR would ideally be just two commits, so that we have a cleaner history. I'm not opposed to have fixup commits appended to a PR, but IMO it makes sense mostly for complex PRs. This one is fairly readable.

pkg/rpmmd/koji.go
thozza: This is modified in later commits, so not that relevant.

pkg/rpmmd/koji.go
avitova: Hm, I thought I'd use nil, but looking at it, you are right, and there is no good reason to use the pointer. Thanks for catching this:)

avitova:
Thank you for the review, Tomáš. I believe I addressed everything.

avitova:
Aaaand, I have updated commits based on our discussion. :) I do not see any manifest changes. :shrug:

achilleas-k:
> Mmm check didn't pass; but still passes locally:

Seems like a flake.  Not sure how it might have happened though.

thozza (approved):
Amazing, ship it! 🚀

achilleas-k (approved):

PR #2300, opened 2026-04-11
fedora: raw disks for atomic variants (HMS-10515)

supakeen:
For the Atomic variants that have an OOBE we should also provide raw disk images aside from the QCOW2's.


data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -1269,21 +1269,28 @@
         - "customizations.timezone"
         - "customizations.locale"
 
-  atomic_disk_common: &atomic_qcow2_common
+  atomic_disk_common: &atomic_disk_common
     <<: *rpm_ostree_imgtype_common
     filename: "image.qcow2"
     mime_type: "application/x-qemu-disk"
```
lzap: Were you planning to drop those two lines? This is not `qcow2` anymore and they are copied on line 1291-92 down below.

lzap (changes requested):

data/distrodefs/fedora/imagetypes.yaml
supakeen: Yes, I was! Thanks

supakeen: Done.

brlane-rht (approved):
Looks good!

lzap (approved):

PR #2304, opened 2026-04-13
pkg/distro: remove separate rpmostree boolean (HMS-10298)

avitova:
Remove rpm_ostree separate boolean and deduce the value based on the ostree struct values.

https://redhat.atlassian.net/browse/HMS-10298

### DIFF
--- a/pkg/distro/defs/loader.go
+++ b/pkg/distro/defs/loader.go
@@ -451,8 +451,6 @@ type ImageTypeYAML struct {
 	// XXX: or iso_variant?
 	Variant string `yaml:"variant"`
 
-	RPMOSTree bool `yaml:"rpm_ostree"`
-
 	OSTree struct {
 		Name       string `yaml:"name"`
 		RemoteName string `yaml:"remote_name"`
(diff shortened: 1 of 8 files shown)


brlane-rht (dismissed):
Looks good to me, and it didn't change the manifest checksums :)

lzap (dismissed):
But the conflict.

pkg/distro/defs/loader.go
```diff
@@ -488,6 +486,10 @@ type ImageTypeYAML struct {
 	name string
 }
 
+func (it *ImageTypeYAML) HasOSTreeDefinition() bool {
+	return it.OSTree.Name != "" || it.OSTree.RemoteName != "" || it.OSTree.Ref != "" || it.OSTree.URL != ""
+}
```
achilleas-k: The `RPMOSTree` boolean property was there to identify ostree-based image types.  I don't know if `HasOSTreeDefinition` is the right way to describe that.  At the very least, I'd expect the function to be `IsSomething()` rather than `HasSomething()` (is this an ostree-based image type?).

I think `HasSomething()` would make sense if we had some kind of polymorphic image type that contained both ostree and non-ostree definitions, but that's not the case.

achilleas-k (changes requested):
I seem to recall we used to do something similar in the past, deciding if an image type is ostree-based from the `Ref` being non-empty, but the git log is telling me that might be a false memory.

Anyway, I like this.  Not sure about the function name though (see below).

achilleas-k (approved):
Thank you :)

brlane-rht (approved):
Ack ack

PR #2317, opened 2026-04-30
fedora: enable bootloader-update.service for IoT images

kgiusti:
The pre-built IoT images available from the Fedora project have the bootloader-update.service enabled, but images generated using image-builder do not. This patch enables the update service by default.

See https://github.com/fedora-iot/iot-distro/issues/135


kgiusti:
While there is inconsistencies between the IoT images, my main motivation for this PR is to ensure that the keys used for secure boot are automatically kept up to date.

I noticed the IoT discrepancy while prototyping a solution, but my primary concern is keeping the shim up to date for IoT and rhel-edge.

supakeen (approved):
Thank you, as discussed elsewhere this is OK.

brlane-rht (approved):
Thanks!

PR #2318, opened 2026-05-03
import ELN (HMS-10620, HMS-10621, HMS-10623)

supakeen:
Introduction of a new distribution separate from Fedora and RHEL-alikes called 'ELN' (Enterprise Linux Next). Initially created from a copy of the RHEL-10 definitions.

ELN is to be the next RHEL. We have a major version of 11 set on it even if it is meant to behave like a stream. When CentOS Stream 11 forks off of ELN we will copy over the ELN-11 definitions to RHEL-11; drop ELN-11 and introduce ELN-12 (using the same repositories); the process will then repeat for RHEL-12, etc.

We'll be handling the stream-bits by allowing the omission of the major version in `image-builder`'s `--distro` argument (e.g. `image-builder build --distro eln ...`).

Any changes to image types we wish to make in RHEL-11 can be done in ELN-11 and it also lets us find unavailable packages and other changes earlier. This initial import does very little of the sort.

I have used the new 'split imagetypes' introduced by @avitova to organize things a bit better but more can be done.

The intent is to (eventually) build these images in Fedora's infrastructure with `image-builder` so ELN images that have RHEL counterparts are built the same way as RHEL and CentOS do.

---

Related GitHub issues:

- https://github.com/fedora-eln/eln/issues/502
- https://github.com/fedora-eln/eln/issues/411


supakeen:
RPM repo snapshots already existed. I've updated the commit to use them.

supakeen:
Undrafted; this is a pretty rough gem at the moment but getting it in gives us a good opportunity to iterate on this. I have a local branch that does a bunch of deduplication in the installers and disk images (but also changes them) which I'm leaving out of this PR but will come after.

supakeen:
CI was failing on the installer images. The `20260426` snapshots that I used were affected by https://github.com/fedora-eln/eln/issues/494 so I've updated to `20260503` which have a newer Anaconda and pykickstart in them.

supakeen:
With the newer `20260503` snapshots we have a failure to boot the ISO. This is similar to the issue `rawhide` had with a broken `glibc`. I'll see if I can do some more debugging.

---

Seems that todays compose has the broken things untagged so I'm going to do a new rpmrepo snapshot.

supakeen:
With the latest snapshot it seems Azure is the only thing to not boot; I'll take a look tomorrow.

supakeen:
Going to temporarily pause this PR until ELN kernels get into less choppy waters.

supakeen:
Validate manifest checksums here is a false positive (?); it's pointing at the CI-change commit but no manifests changed there.

lzap:
> Validate manifest checksums here is a false positive (?); it's pointing at the CI-change commit but no manifests changed there.

Would maybe having `git diff` or `git diff --cached` help in that script?

supakeen:
I've rebased and undrafted; this is a good starting point to start iterating in followups and CI is about to go green.

pkg/distro/defs/loader.go
```diff
@@ -399,6 +399,7 @@ func mergeImageTypeConfigs(d *DistroYAML, configs []imageTypesYAML) error {
 type imageTypesYAML struct {
 	ImageTypes map[string]ImageTypeYAML `yaml:"image_types"`
 	Common     map[string]any           `yaml:".common,omitempty"`
+	Shared     map[string]any           `yaml:".shared,omitempty"`
 }
```
lzap, line 403: Should Common be deprecated here or something?

lzap (dismissed):
Nicely documented!

pkg/distro/defs/loader.go
supakeen: No, we allow both for now for some freedom for users; honestly this split-imagetypes like this is all very experimental and since it only lives in our repository at the moment this is (IMHO) fine.

data/distrodefs/eln-11/disk.yaml
```diff
@@ -82,14 +70,6 @@ image_types:
...
             - "rng-tools"
             - "udisks2"
-          conditions:
-            "add insights pkgs on rhel":
-              when:
-                distro_name: "rhel"
-              append:
-                include:
-                  - "insights-client"
-                  - "subscription-manager-cockpit"
```
thozza: One can argue that condition like this one will be later missed when we fork el11/c11s from eln... It may be a good idea to keep those here or add them to the base package set, to prevent us from missing this addition later.

thozza (commented):
In general, LGTM, but I have one comment....

data/distrodefs/eln-11/disk.yaml
supakeen: I'm not entirely sure about this either; we can do either. I felt it to be OK to re-add them when we import but let's see what others think in this case since I'm pretty neutral as well.

thozza: We can add them later for sure. The point I'm trying to make is how can we not forget to add them in the future? Keeping the condition in the definition is harmless and it would remove the need to add anything later.

supakeen: Yea so I'm thinking ahead where these definitions might not live inside images but in a separate repository potentially not even under our control; as we stabilize the definition format.

In that case they stick out a bit that the maintainers of ELN need to know what to include for the other EL's. Maybe it makes sense, maybe it doesn't.

I'm happy to drop the commit and then this can be discussed at a later time?

supakeen: Ok; I've dropped the commit.

thozza: Ehm... The point I was trying to make was really only about the additional packages installed on `rhel` distro. The removal of conditions for specific RHEL (minor) versions IMHO made a lot of sense. Anyway, I'll approve the PR as is and you can remove those in a follow-up, up to you.

thozza (approved):

lzap (approved):

PR #2319, opened 2026-05-03
experimental: use yamlplus

supakeen:
Controversial to submit upstream, but I thought I'd give it a try since it's fully gated behind an experimental flag. I've duplicated the entire `loadImageTypeConfigs`-function so it's clear that there's no touching of the normal path.

---

This is a bit far out but over the past few weeks I've written a small YAML library that handles `!xref` tags in YAML. This commit enables the use of that library conditionally on an experimental boolean.

The following command would allow combining this additional behavior which will not be enabled by default.

`IMAGE_BUILDER_EXPERIMENTAL="yamlplus,yamldir=..." image-builder ...`

I'd like to use this *without* forking `images` and `image-builder` in my little remix to get a feel for how this works and how nice of an organization I can whip up for definitions.

If we decide to go down this path, and that's a big if, the `yamlplus` package will likely be interned into the `images` library and dropped from my personal GitHub.

---

Repository: https://github.com/supakeen/yamlplus
Documentation: https://pkg.go.dev/github.com/supakeen/yamlplus@v1.1.0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,9 +38,10 @@ require (
 	github.com/spf13/cobra v1.9.1
 	github.com/spf13/pflag v1.0.7
 	github.com/stretchr/testify v1.11.1
+	github.com/supakeen/yamlplus v1.1.0
 	github.com/ubccr/kerby v0.0.0-20230802201021-412be7bfaee5
 	github.com/vmware/govmomi v0.52.0
-	go.yaml.in/yaml/v3 v3.0.3
+	go.yaml.in/yaml/v3 v3.0.4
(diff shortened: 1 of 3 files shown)


lzap (approved):
I am fine with this as long as you give contributor/admin to some team members just in case.

My only concern is that this could possibly make loading even slower, this stack is already extremely slow like listing all image types in CLI takes like 4 seconds to load and I see some recursive registration which could make things worse in the future as the amount of files and directories expand. Would be interested in small benchmark that measures loading

thozza (approved):

PR #2325, opened 2026-05-08
defs/bootc: set XBOOTLDR GUID

supakeen:
We were (accidentally?) setting the wrong GUID on the XBOOTLDR default partition for `bootc`. This causes sealed systems to not boot correctly as they have duplicate root partitions.


brlane-rht (approved):
Looks good

thozza (approved):

PR #2326, opened 2026-05-09
many: bootc sealed images (composefs, and bootloader) (HMS-10628)

supakeen:
This PR introduces support for bootc 'sealed' images. Sealed images use the composefs backend, together with a UKI, signed systemd-boot, and chain it all together to ensure only a trusted composefs is booted.

This has implications for how we deal with the disk images produced.

For starters; currently for bootc we can assume an image is sealed based on `bootc container inspect` returning if the kernel is unified. This then implies that the composefs backend must be used. It also implies that systemd must be used as the bootloader.

For the latter we expect the container to configure this and don't assume it based on the presence of a unified kernel.

For now I've called the boolean for this 'unified'. If in the future there comes a possibility where we want composefs to be standalone from sealed then we can split the boolean out.

This is only implemented for disk images, it is currently out of scope for ISO's and PXE images; for the latter we *might* want to error out when a unified kernel is detected in the container?

We skip all customizations in the pipeline if they touch the filesystem. We cannot adjust the filesystem that was created during the container build as its verity hashes are recorded in the UKI, the system will not boot if we modify the filesystem in any way.

Similarly we cannot accept any kernel argument customizations, including those in our own definitions, as they are encoded in the UKI at container build time.

For now these are silently ignored; we have no precedent for being able to deny customizations only after we've resolved content and I could only come up with bad ways to error out later.

---

Related: https://github.com/osbuild/image-builder/issues/2427


supakeen:
I've opted to move the rest of the work here to follow-ups as they can be worked around in the container:

1. https://github.com/osbuild/images/issues/2337
2. https://github.com/osbuild/images/issues/2336

This was discussed in: https://github.com/osbuild/image-builder/issues/2427#issuecomment-4441383966

lzap (approved):

pkg/bootc/resolver.go
```diff
@@ -332,6 +338,33 @@ func (c *Container) InitrdModules(kver string) ([]string, error) {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to run bootc container inspect: %w, output:\n%s", err, output)
+	}
+
+	var bootcInspect struct {
+		Kargs  []string `json:"kargs"`
+		Kernel struct {
+			Version string `json:"version"`
+			Unified bool   `json:"unified"`
+		} `json:"kernel"`
```
thozza, line 358: Nitpick: the `kargs` and `version` don't seem to be used anywhere or even unit-tested. I think that it is useful to unmarshal all this information while we are at it, but technically this is a dead code, because on the `kernel.unified` value is ever returned from this method.

lzap:
The patch series looks fine, well, this is the build pipeline for unikernel bootc:

- prepare stage
- install to filesystem
- finish stage

Not much that can be leveraged.

thozza (approved):
Nice work!

pkg/bootc/resolver.go
supakeen: Yea this is true; I had some plans to start using them. I'll try to do that in a follow-up. For example we have our own kernel-scanning code at the moment for the version and now that `bootc` can give it we might be able to move some bits around.

PR #2329, opened 2026-05-11
test: close filesystem/disk customization coverage gaps with osbuild-composer

ochosi:
## Summary

The `osbuild-composer` repository has integration tests for filesystem and disk customizations (`filesystem.sh`, `regression-no-explicit-rootfs-definition.sh`) that verify partitioning layouts, forbidden mount paths, and edge cases. This PR brings equivalent coverage to the `images` repository so tests are not lost as functionality migrates.

### Changes

1. **check-host-config: add filesystem mountpoint verification**
   Add a new `filesystem` check to `check-host-config` that verifies all expected mountpoints from both `customizations.filesystem` and `customizations.disk` (plain, LVM, btrfs) are present in the booted VM by reading `/proc/mounts` and `/proc/swaps`. Previously, builds were verified to boot but never asserted the requested partition layout actually appeared.

2. **test: add filesystem-customizations integration config**
   Add a `filesystem-customizations.json` test config exercising the full set of 16 mountpoints from `osbuild-composer`'s `filesystem.sh` (`/`, `/var`, `/var/log`, `/var/log/audit`, `/usr`, `/tmp`, `/var/tmp`, `/home`, `/home/shadowman`, `/opt`, `/srv`, `/app`, `/data`, `/boot`, `/boot/firmware`, `/foobar`).

3. **pkg/policies: add mountpoint policy tests for forbidden and allowed paths**
   Add explicit unit tests verifying `CheckMountpointsPolicy` rejects all 14 forbidden paths (`/etc`, `/sys`, `/proc`, `/dev`, `/run`, `/bin`, `/sbin`, `/lib`, `/lib64`, `/lost+found`, `/sysroot`, `/usr/bin`, `/var/run`, `/var/lock`) and accepts the full valid set. This is the `images`-equivalent of the negative test in `osbuild-composer`'s `filesystem.sh`.

4. **test: add filesystem-no-explicit-rootfs regression config**
   Add a `filesystem-no-explicit-rootfs.json` config that specifies `/data` without an explicit `/` entry, covering the RHBZ#2049500 regression where such composes would fail. The build must succeed with the default root filesystem implicitly created.

## Test plan

- [x] `go test ./cmd/check-host-config/check/ -run TestFilesystem` — all 7 subtests pass
- [x] `go test ./pkg/policies/ -run TestCheckMountpointsPolicy` — all 30 subtests pass
- [x] `python3 test/scripts/validate-config-list` — both new configs validated against known build configurations


cmd/check-host-config/check/filesystem.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read /proc/mounts: %w", err)
+	}
+
+	for _, line := range strings.Split(string(data), "\n") {
+		fields := strings.Fields(line)
+		if len(fields) >= 2 {
+			mounts[fields[1]] = true
+		}
+	}
```
achilleas-k: Parsing `/proc/mounts` isn't very nice.  It's not likely the format will change any time soon, but `lsblk -J` is a machine-readable alternative, so let's use that instead.

test/config-list.json
```diff
@@ -597,6 +597,36 @@
...
+      "image-types": [
+        "generic-qcow2",
+        "server-qcow2"
+      ],
+      "distros": [
+        "centos-10",
+        "centos-9",
+        "fedora*",
+        "rhel-10*",
+        "rhel-9*"
```
achilleas-k: `generic-qcow2` and `server-qcow2` are the Fedora image type names.  CentOS and RHEL should be move to the next item alongside RHEL 8.

pkg/policies/policies_test.go
```diff
@@ -79,6 +85,73 @@ func TestMountpointPolicies(t *testing.T) {
 	}
 }
 
+// TestCheckMountpointsPolicyForbiddenPaths verifies that a blueprint containing
+// forbidden mountpoints (the same set tested in osbuild-composer's filesystem.sh
+// fail case) is properly rejected by CheckMountpointsPolicy.
+func TestCheckMountpointsPolicyForbiddenPaths(t *testing.T) {
```
achilleas-k: This seems to be duplicating the tests above in the same file only going through the blueprint method rather than directly calling the policies method.  I'm not sure how useful it is.  Perhaps having a generic test in the blueprint library would be good, one that sets an arbitrary policy and test if it works as expected.

test/config-list.json
```diff
@@ -627,6 +627,36 @@
...
+      ]
```
achilleas-k: Same as above.

achilleas-k (changes requested):
Thanks.  Love to see more stuff being moved from osbuild-composer down into images again.
Needs some changes though (the image type names in the config at least).  Comments inline.

cmd/check-host-config/check/filesystem.go
ochosi: Very true, amended!

achilleas-k:
I noticed that the changes addressing my comments were squashed into the last commit instead of fixing up the relevant ones so I cleaned up the history a bit and force pushed.
In the process I also realised we don't need to separate out the RHEL 8 cases in the config-list, so I merged those too.

I also added manifest checksums for the new configurations.  Unfortunately, our manifest checksum check doesn't fail when there are _missing_ checksums, only when there are changes that weren't committed.

achilleas-k (approved):

achilleas-k:
200 new configs is a bit much though.  Maybe we don't need to test all distros here.  (but maybe let's let this one build and merge and we can trim things down later)

supakeen (approved):
Building 200 extra images every time we refresh snapshots or osbuild commit is going to be ... fun. Like @achilleas-k says, it's probably pretty high priority to cut that down but let's see if it actually makes CI slow first.

lzap (approved):
Hmm why don't we merge some image configs together?

supakeen:
So we missed this, but this PR did not bump the seed and thus these tests didn't actually get ran on all configs. They are now failing in other PRs as the columns requested in `lsblk` do not exist in all RHEL version. Notably all RHEL 8 configs that try to run this check:

```
🔴 filesystem: fail: failed to read active mountpoints: failed to run lsblk: exit status 1 (exit 1, stderr: lsblk: unknown column: MOUNTPOINTS,FSTYPE
```

PR #2330, opened 2026-05-12
disk: partition table policies

supakeen:
Introduces the concept of partition table policies. The policy, much like the options, defines rules on how partition tables should be created.

It lives separately from the options because it can be used both for filesystem customizations and disk customizations. It's also meant to be configurable from image definitions, including those potentially stored inside a container (in `disk.yaml`).

Initially we add a boolean to be able to disable the creation of the /boot mountpoint (generally XBOOTLDR). This is a necessary bit for the conversion of sealed bootable containers. Creating a non-VFAT /boot there leads to broken boot (as systemd auto-mounting verifies that /boot is FAT nowadays?) and in general /boot isn't wanted on these.

This carries over to package mode where it's very possible we want to create images that do *not* introduce a `/boot` by default as everything would be stored in the ESP for some cases.

---

Note that since we currently do define a `/boot` in our default partition tables for bootc image types you can only test the behavior with an embedded `disk.yaml` that omits `/boot` (or temporarily adjust the definitions) until those files are included in the upstream containers.

---

I'm a bit torn about the options vs policy split here; it kinda fits, but it also kinda feels like this /should/ live in options instead though options and policies are different and sourced from different places. The main argument for me is that image types (through their base partition table) know the bootloader(s) involved and what is necessary to boot.

I'm *also* unsure about the abstraction level of this. It makes sense to make these tunables really low level; it *also* makes sense to have a higher level "Unified" or whatever. I've opted for the lower level here because a unified kernel doesn't imply there to never be an XBOOTLDR.

In fact we might want to expand the policy in a follow-up to have a tunable to actively *disallow* adding XBOOTLDR that is non-VFAT in diskcust or fscust. Adding one makes systemd's auto mounting break if it's of the wrong filesystem type. Again something only the image type knows.

---

Related: https://github.com/osbuild/image-builder/issues/2427

### DIFF
--- a/pkg/disk/disk_test.go
+++ b/pkg/disk/disk_test.go
@@ -1081,3 +1081,40 @@ func TestCreatePartitionTableDefaultFs(t *testing.T) {
 	assert.Equal(t, "/var/stuff", mpt.Partitions[nParts-2].Payload.(*disk.Filesystem).Mountpoint)
 	assert.Equal(t, "ext4", mpt.Partitions[nParts-2].Payload.(*disk.Filesystem).Type)
 }
+
+func TestNewPartitionTablePolicyNoXBOOTLDR(t *testing.T) {
+	/* #nosec G404 */
+	rng := rand.New(rand.NewSource(13))
+	noBootPolicy := &disk.PartitionTablePolicy{EnsureXBOOTLDR: false}
+
+	t.Run("lvm-noboot-errors", func(t *testing.T) {
+		pt := testdisk.TestPartitionTables()["plain-noboot"]
+		pt.Policy = noBootPolicy
+		_, err := disk.NewPartitionTable(&pt, []blueprint.FilesystemCustomization{
+			{Mountpoint: "/home", MinSize: 1 * GiB},
+		}, 13*MiB, partition.AutoLVMPartitioningMode, arch.ARCH_X86_64, nil, "", rng)
+		assert.ErrorContains(t, err, "policy says to not create one")
+	})
+
+	t.Run("btrfs-noboot-succeeds", func(t *testing.T) {
+		pt := testdisk.TestPartitionTables()["plain-noboot"]
+		pt.Policy = noBootPolicy
+		mpt, err := disk.NewPartitionTable(&pt, []blueprint.FilesystemCustomization{
+			{Mountpoint: "/home", MinSize: 1 * GiB},
+		}, 13*MiB, partition.BtrfsPartitioningMode, arch.ARCH_X86_64, nil, "", rng)
+		require.NoError(t, err)
+		assert.Nil(t, disk.EntityPath(mpt, "/boot"), "no /boot should be created")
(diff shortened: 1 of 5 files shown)


achilleas-k (commented):
Some quick thoughts.

I really like this idea but I'd like to see it go a bit further.  I think some of the existing things we do (ensuring sizes, require root, require esp) can fit into this too and can use those requirements to think about how the policy can work more broadly.  It would be a great way to encode things that are hidden in the partition table generation functions (`NewPartitionTable()` and `NewCustomPartitionTable()`) and aren't very visible from the outside ("outside" here meaning as an observer of the `disk` package API without looking into the functions themselves).

For example, the required filesystem sizes are hard coded right now.  For bootc, we have the 2x container size rule.  So if we have a default policy for the size of `/` and `/usr` in the non-bootc defaults, the bootc manifest generation function can update/override the policy with the result of the 2x rule before generating the manifest.

For `/boot`, esp, and MBR type things, we make those decisions in the PT generators based on the boot mode.  But I thin it would be clearer if the boot mode was an input to a policy initialiser (or policy modifier function).  That would make our rules more obvious: The boot mode is one of the inputs that creates a policy about which partitions are needed and all these rules are resolved before a PT generator is called.

We might need to think a it about what's a policy and what's a customization, how to differentiate the two, but I think we can figure it out.

This is a great idea in general!

supakeen:
> Some quick thoughts.
> 
> I really like this idea but I'd like to see it go a bit further. I think some of the existing things we do (ensuring sizes, require root, require esp) can fit into this too and can use those requirements to think about how the policy can work more broadly. It would be a great way to encode things that are hidden in the partition table generation functions (`NewPartitionTable()` and `NewCustomPartitionTable()`) and aren't very visible from the outside ("outside" here meaning as an observer of the `disk` package API without looking into the functions themselves).
> 
> For example, the required filesystem sizes are hard coded right now. For bootc, we have the 2x container size rule. So if we have a default policy for the size of `/` and `/usr` in the non-bootc defaults, the bootc manifest generation function can update/override the policy with the result of the 2x rule before generating the manifest.
> 
> For `/boot`, esp, and MBR type things, we make those decisions in the PT generators based on the boot mode. But I thin it would be clearer if the boot mode was an input to a policy initialiser (or policy modifier function). That would make our rules more obvious: The boot mode is one of the inputs that creates a policy about which partitions are needed and all these rules are resolved before a PT generator is called.
> 
> We might need to think a it about what's a policy and what's a customization, how to differentiate the two, but I think we can figure it out.
> 
> This is a great idea in general!

The question is (I think) if we want to address that immediately as a big bang, or slowly over time. We kinda need it on the short term for sealed bootc images but I'd also like to expand on this.

achilleas-k (approved):
Let's get this merged to start building images from sealed bootc containers.

lzap (approved):
I agree with @achilleas-k that this could be really useful. But this is done cleanly and it can be easily extended at any point, no need to hold this back.

PR #2331, opened 2026-05-12
fedora: remove vc4 module blacklist from ostree kernel options

pcdubs:
There is no reason to blacklist the vc4 driver. This removes modprobe.blacklist=vc4 from ostree_deployment_kernel_options, which affects IoT and Atomic disk images.


supakeen:
Do you recall why we had it blacklisted initially?

pcdubs:
> Do you recall why we had it blacklisted initially?

I think it was unreliable and crashed periodically. It's been working well for sometime and the removal was overlooked.

supakeen (approved):

achilleas-k (approved):
Nice.  Thank you!

PR #2333, opened 2026-05-12
Update RHEL 9 and 10 OCI image definitions [HMS-10328, HMS-10472]

achilleas-k:
Update RHEL 9 and RHEL 10 OCI image definitions:
- Add a dracut config file called `oci.conf` that adds the `mlx5_core` driver [HMS-10472].
- Change the default partitioning to LVM [HMS-10328].

I made some decisions about the details of the partitioning layout.  See the the commit message(s) for details.

[HMS-10472]: https://redhat.atlassian.net/browse/HMS-10472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HMS-10328]: https://redhat.atlassian.net/browse/HMS-10328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ochosi (dismissed):
After having read through the tickets and the Oracle Linux docs, this generally looks good. I think there are some implicit assumptions/choices that we could document in the commit message, like

- 64 GB default size is what we use for other LVM cloud images (but this is defined in pungi)
- 30 GB for / is a consequence of the default image size and the recommendation from the Oracle docs (half of the overall disk size)
- omitting the oled directory makes sense (but we may have to add it later)

ochosi (approved):

supakeen (approved):
Does OCI have something to resize the pvs/lvs?

achilleas-k:
Standard tooling should work to grow the LV with `lvextend` and then `xfs_growfs` for the filesystem.  There's also `oci-growfs` as part of [oci-utils](https://github.com/oracle/oci-utils) which, afaik, isn't available in RHEL, but it might be made available as part of the whole RHEL on OCI effort.

PR #2335, opened 2026-05-14
Add bootc-foundry boot test infrastructure (HMS-10336)

thozza:
Enable boot testing of disk images built from bootc-derived containers in the bootc-foundry repository. Teach `cmd/build` to accept a bootc container ref directly (deriving distro identity from the container's os-release), consolidate shared dependency tracking in the Schutzfile under `common.dependencies`, and harden path resolution so test scripts work when the images repo is cloned as a subdirectory of another project. Add bootc-specific support to the boot-image test script, including ISO kickstart merging and bootc manifest detection.

## Architectural Changes

Introduce a `RequiresBootc` metadata flag in the check-host-config framework, following the established `RequiresBlueprint`/`RequiresCustomizations` pattern. This allows bootc-specific host checks (like `bootc status`) to be cleanly skipped for non-bootc images without special-casing in the check functions themselves.

Consolidate the osbuild dependency commit from per-distro entries into `common.dependencies.osbuild` in the Schutzfile. The fallback logic (distro-specific first, then common) preserves the ability to override per-distro when needed, while reducing duplication and enabling external repos like bootc-foundry to discover the correct osbuild commit for any distro.

## Key Changes

- Add `-bootc-ref`, `-bootc-build-ref`, and `-bootc-remote` flags to `cmd/build`, making `-distro` unnecessary (and mutually exclusive) when building from a bootc container
- Restructure Schutzfile to track shared dependencies (bib, osbuild) under `common.dependencies` and update all consuming scripts
- Resolve the Schutzfile and pip install target via absolute paths derived from the script location, so they work when the repo is a subdirectory
- Fix image file discovery to find the actually exported pipeline directory instead of assuming the last pipeline in the manifest
- Add bootc manifest detection (`_is_bootc_manifest()`) to skip ssh-server/jq checks that don't apply to container-based images
- Support boot-testing bootc ISOs with embedded kickstarts by merging the original kickstart content with the generated unattended directives

## Breaking Changes

This PR is fully backward compatible. The Schutzfile restructuring updates all consuming scripts in the same commits, and the `get_osbuild_commit()` fallback ensures existing distro-specific overrides still take precedence.

## Testing

- Add table-driven tests for `RequiresBootc` skip logic in `check-host-config`, covering nil config, nil `Options.Bootc`, and valid bootc config
- Add unit tests for the `bootc-status` check using mock exec to verify pass/fail behavior without requiring a real bootc environment
- Add early environment variable validation (`ensure_env_vars`) to the boot-image script so missing cloud credentials produce a clear error instead of a confusing traceback mid-test
- Cross tested in https://github.com/osbuild/bootc-foundry/pull/85 via https://gitlab.com/redhat/services/products/image-builder/ci/bootc-foundry/-/pipelines/2525058069


lzap (approved):

test/scripts/imgtestlib.py
```diff
@@ -321,18 +336,19 @@ def can_boot_test(manifest_fname, manifest_data, image_type, arch, distro, bluep
...
+            if not any("ssh-server" in url for url in urls):
+                # This can happen e.g. when an image is build with the "minimal: true" customization.
+                # We could use guestfs to inject keys, see PR#1995
+                print(f"  not bootable: ssh-server not found in manifest {manifest_fname} ({arch} {image_type})")
+                return False
+            # We need jq in the image many images do not have it
+            # (e.g. centos-9/rhel-9 with releasever config) so skip those too
+            if not any("jq" in url for url in urls):
+                print(f"  not bootable: jq not found in {manifest_fname} ({arch} {image_type})")
+                return False
```
achilleas-k, line 351: FOR FOLLOW-UP

This predates the check-host-config Go binary and was required when we were running a shell script on the host and using jq to parse the build config.  It shouldn't be necessary now.  Let's drop it in a later PR.

achilleas-k (approved):
This is great.  Nice improvements all around.

LGTM

test/scripts/imgtestlib.py
thozza: Yeah, that's what I thought as well, but didn't want to touch it as part of this PR...

PR #2334, opened 2026-05-13
flatpak: use container resolver to resolve flatpak refs

lzap:
We need to optimize memory consumption and we ideally want to do this just once, for container refs. Flatpak did have its own resolver, let's reuse what we have.

---

Initially I wanted to solve the memory problem described here: https://github.com/osbuild/image-builder/issues/2426

Only to figure out that fedora registry does not support server-side filtering, at least not the one we need here. I think we really need to optimize the parsing. I believe @achilleas-k has some ideas already.

---

Update: I did have one idea which was confirmed - aren't we querying the same image over and over again just with different inputs? And this was confirmed. A simple patch (2nd commit) that implements a trivial "cache" shaves the processing time significantly. It just remembers the last tag and parsed data so it can be immediately reused, no need for fancy LRU caching or something like that.

Before:

```
time go run ./cmd/image-builder -v manifest --distro fedora-44 silverblue-installer
real	2m44.135s
user	0m21.812s
sys	0m9.427s
```

After:

```
time go run ./cmd/image-builder -v manifest --distro fedora-44 silverblue-installer
real	0m49.459s
user	0m10.343s
sys	0m4.078s
```

Now, I really hope this is not just `quay.io` being slow for me, please test this too :-)

---

Reminder for myself: rebase after https://github.com/osbuild/osbuild-composer/pull/5136 and https://github.com/osbuild/osbuild-composer/pull/5138

PR #2343, opened 2026-05-19
CI: run tests on Fedora 44

achilleas-k:
SSIA


Schutzfile
```diff
@@ -9,48 +9,48 @@
         "commit": "6a00632cbaaa211abf6ec43e46ddd14774745054"
```
thozza: you need to also update osbuild ref, because the one in Schutzfile is AFAICT too old and it won't have F44 RPM builds.

See https://github.com/osbuild/osbuild/blob/6a00632cbaaa211abf6ec43e46ddd14774745054/.gitlab-ci.yml#L38-L41

achilleas-k: Thank you!!  Fixed.

thozza (dismissed):

achilleas-k:
Manifest checksum check is failing but I can't reproduce the failure locally.  I pulled in the `git status` commit from #2334 to see what's happening.

achilleas-k:
Alright, I rebased on `main`, bumped the terraform ID and osbuild commit to their current `main`, and regenerated the gitlab-ci config.  Let's go!

achilleas-k:
Fedora 44 osbuild RPMs on `main` aren't ready.  Will retry when the job is done.

achilleas-k:
Building RHEL 7 on Fedora 44 is going to be a problem it seems.  Every package fails to verify with `package _ does not verify: no digest`.

I think I'll switch those to build on RHEL 9 for now.

achilleas-k:
Of course, the install dependencies script doesn't work on RHEL because it also installs non-RHEL packages.

achilleas-k:
Ok so installing more dependencies trips up something in the container unit tests.

achilleas-k:
I want to get #2343 merged first and then rebase this

brlane-rht (approved):
Ack. Ack. Ack! Thanks for sorting all this out.

supakeen (approved):

PR #2344, opened 2026-05-19
boot-azure: switch machine type

supakeen:
The D1 SKU we were using doesn't correctly start cloud-init 26.1 in ELN. Switching to an AMD SKU makes mcelog fail because the driver seems to be missing (will investigate).

This instance type, determined by brute force, seems to work.

### DIFF
--- a/Schutzfile
+++ b/Schutzfile
@@ -1,6 +1,6 @@
 {
   "common": {
-    "rngseed": 2026051400,
+    "rngseed": 2026051900,
     "dependencies": {
       "bootc-image-builder": {
         "ref": "quay.io/centos-bootc/bootc-image-builder@sha256:9893e7209e5f449b86ababfd2ee02a58cca2e5990f77b06c3539227531fc8120"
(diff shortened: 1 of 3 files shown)


achilleas-k (dismissed):
TY

croissanne (approved):

croissanne (changes requested):
needs a seed bump, this touches every azure test (except wsl & aarch)

croissanne (approved):
great

achilleas-k (dismissed):

croissanne (dismissed):
again?!

supakeen:
RHEL 8 is failing because https://github.com/osbuild/images/pull/2329 introduced a check which uses `lsblk` with the `MOUNTPOINTS,FSTYPE` columns. These (or one of these) do not exist in RHEL 8.

We forgot to bump the seed in that PR so the boot tests weren't re-ran.

supakeen:
I've added a commit that makes the test not run on RHEL-8. I think we can keep the configs for now as the check can later be adjusted to work on RHEL-8.

ochosi (approved):
Thank you for cleaning up my whoopsie 🫣 🙏

achilleas-k (approved):

achilleas-k:
> Thank you for cleaning up my whoopsie 🫣 🙏

I think this one's on me.  I should have realised that we're adding a test and not running it and asked you to bump the rngseed for the tests.

PR #2345, opened 2026-05-19
gen-manifests: respect SIGINT/SIGTERM

lzap:
It was not possible to interrupt the generation of manifests by sending
SIGINT/SIGTERM.

---

Plus a little change in the YAML loader.

### DIFF
--- a/pkg/depsolvednf/depsolvednf.go
+++ b/pkg/depsolvednf/depsolvednf.go
@@ -16,6 +16,7 @@ package depsolvednf
 import (
 	"bytes"
 	"cmp"
+	"context"
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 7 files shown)


pkg/depsolvednf/depsolvednf.go
```diff
@@ -635,8 +642,11 @@ func run(dnfJsonCmd []string, reqData []byte, stderr io.Writer) ([]byte, error)
...
 	err = cmd.Wait()
 	output := stdout.Bytes()
+	if ctx.Err() != nil {
+		return output, ctx.Err()
+	}
 	if runError, ok := err.(*exec.ExitError); ok && runError.ExitCode() != 0 {
 		return output, fmt.Errorf("depsolve failed with exit code %d", runError.ExitCode())
 	}
-	return output, nil
+	return output, err
```
thozza: Nitpick: there is no added value in not returning `nil` explicitly for an error here.

thozza (commented):
The first commit could have been a separate PR and you could have addressed also `loadImageTypeConfigs()` and `loadImageTypeConfigsPlus()` in the same file, which also call `defer f.Close()`.

I've added one style nitpick, but I think that it makes sense to fix it and be consistent with the project style - return `nil` for an error if there is none at the end of a function.

The rest looks good to me.

lzap:
Rebased, dropped the commit that was irrelevant.

thozza (approved):

pkg/depsolvednf/depsolvednf.go
```diff
@@ -635,8 +642,11 @@ func run(dnfJsonCmd []string, reqData []byte, stderr io.Writer) ([]byte, error)
 
 	err = cmd.Wait()
 	output := stdout.Bytes()
-	if err != nil {
-		return output, fmt.Errorf("running the depsolver failed: %w", err)
+	if ctx.Err() != nil {
+		return output, ctx.Err()
+	}
+	if runError, ok := err.(*exec.ExitError); ok && runError.ExitCode() != 0 {
+		return output, fmt.Errorf("depsolve failed with exit code %d", runError.ExitCode())
```
thozza, line 649: This reverts Ondrej's changes. This change should be dropped.

thozza (changes requested):

lzap:
I am gonna close this for now, I have a different PR this will conflict.

PR #2346, opened 2026-05-19
container: resolve containers using skopeo [RHEL-56367]

achilleas-k:
This is the first part towards removing our dependency on the `containers/image` Go library by shelling out to tools like skopeo and podman.  Ideally, we would only rely on skopeo, but one specific use case uses podman and I think we can drop it.  More details on this in the code and relevant commit.

The referenced ticket requests removing our dependence on GnuPG.  There are workarounds to using a different backend, but the current approach aligns with something I've wanted for a while so this is an opportunity to resolve multiple issues at once.  When we finally remove the containers dependencies, we will be able to drop our dependency on the following development packages:
- gpgme-devel
- btrfs-progs-devel (or, in the case of RHEL and CentOS, the `exclude_graphdriver_btrfs` build flag).
- device-mapper-devel

This will simplify our code and dependency chain by essentially "outsourcing" the work to skopeo.  Note that skopeo is already an indirect dependency through osbuild as it's used for the `org.osbuild.skopeo` source for containers.

I've been working on this on and off for a few weeks now and I was hoping to have it done a while ago.  The big part that remains is the container upload command and internal functionality.  I decided to open the PR with the current state to get some reviews and perhaps merge this while I work on the rest.


achilleas-k:
@kingsleyzissou Added you as a reviewer since you've worked on the local manifest resolution.

achilleas-k:
I missed the test failures when running locally because they only happen with `-race`.

Also I didn't consider that adding our test containers to `all-customizations` would fail because we only have x86_64 and aarch64 versions.

achilleas-k:
Oh, this is interesting:
```
🔴 filesystem: fail: failed to read active mountpoints: failed to run lsblk: exit status 1 (exit 1, stderr: lsblk: unknown column: MOUNTPOINTS,FSTYPE)
```
Another case where we changed a test but didn't run it because of caching.

This is a RHEL 8 failure.

achilleas-k:
> Oh, this is interesting:
> 
> ```
> 🔴 filesystem: fail: failed to read active mountpoints: failed to run lsblk: exit status 1 (exit 1, stderr: lsblk: unknown column: MOUNTPOINTS,FSTYPE)
> ```
> 
> Another case where we changed a test but didn't run it because of caching.
> 
> This is a RHEL 8 failure.

Getting fixed in #2344.

pkg/container/client_test.go
```diff
@@ -105,3 +113,531 @@ func TestClientAuthFilePath(t *testing.T) {
...
+				authFile, err := os.Create(authFilepath)
+				require.NoError(err)
+				if tc.authFileReadable {
+					expected = authFilepath
+				} else {
+					require.NoError(authFile.Chmod(0o000))
+				}
+			}
+
+			require.Equal(expected, container.GetDefaultAuthFile())
```
thozza, line 226: Nitpick: Personally, I find it a bit odd that the expected value is not part of the test case definition. The test case definition is more of a setup definition and the actual expected values are determined in the test runner. I mean, all of this looks reasonable, but I have a hard time double-checking what the test actually verifies. FWIW, you do not need to test the full path set in the ENV variable, you could just define the default and then a different filename for each scenario (or env variable). The temporary path is IMHO not that important if the filename is always different.

achilleas-k:
New failure seems to only happen when I run the tests in a container

achilleas-k:
> New failure seems to only happen when I run the tests in a container

The good news is they're not resolver-related.  The new tests that were introduced in the first commit fail to run inside a container.  I'll need to fix them up.

achilleas-k:
Still need to fix the container situation for s390x and ppc64le but the rest should be good.

thozza (approved):
LGTM

kingsleyzissou (approved):
Ohhh this is very nice

pkg/container/client_test.go
achilleas-k: That's a very good point.  I think another way around the issue, now that I think about it, is the test can have a global tmpdir root and the test cases can be set up after creating the tmpdir, so an element in the test matrix could be
```go
{
    name: "authfile",
    authEnvFile: filepath.Join(tmpdir, "really-cool-authfile.json"),
},
```

I think that would take care of the readability while also matching full paths.

PR #2349, opened 2026-05-20
ci: include flatpaks

supakeen:
Some image types in the ostree branch of things have flatpaks. Those should be resolved same as commits; if they're not their checksums do not match which in turn causes Kinoite and Silverblue installers to be continually rebuilt.


achilleas-k (approved):
ah yes!  Of course!!

Thanks!

thozza (approved):

PR #2351, opened 2026-05-20
distro/rhel-9: fix OCI partitioning

achilleas-k:
Commit b480f7991d89620dbcc2f635aab68361e4b39814 claims it changed the default OCI partitioning to be LVM-based.  The claim is wrong.  Instead it changed the partitioning for the edge-container.  Which does not have, nor does it need, a partition table.

This commit fixes the previous error.


supakeen (approved):
Actually clicked the up-arrows and it looks right now 🙂

brlane-rht (approved):
Looks good. I notice the  oci checksum files changed :)

thozza (approved):

PR #2352, opened 2026-05-20
cicd: run gobump per commit

lzap:
Latest version of `gobump` tool was updated to create multiple commits. This was done to overcome issued when calling `prepare-source.sh` or `go mod vendor` created changes that were breaking other version bumps. This is now fixed, every dependency bump always starts with a clean git checkout. This also means that there will be no commit at the end of the operation (before push) therefore this was removed.

While working on it, I discovered that `skopeo` was not installed, so tests were failing, therefore gobump would never create a PR because not a single dependency bump could ever pass. I separated a task that runs tests first so the job does not bother to run lengthy gobump if tests are not passing for any reason on `main`.

---

Pushed to the upstream so I can test this on CICD.

PR made from this branch:

* https://github.com/osbuild/images/pull/2368

### DIFF
--- a/.github/workflows/gobump.yml
+++ b/.github/workflows/gobump.yml
@@ -10,33 +10,38 @@ on:  # yamllint disable-line rule:truthy
 jobs:
   update-and-push:
     runs-on: ubuntu-latest
-    container: registry.fedoraproject.org/fedora:42
+    container: registry.fedoraproject.org/fedora:43
     steps:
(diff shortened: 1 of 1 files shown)


brlane-rht (dismissed):
Looks good

lzap:
Rebased, can I get more ACKs please?

supakeen (approved):

brlane-rht (approved):
Ack!

PR #2354, opened 2026-05-20
Make EFIPartitionTable function public, move it into disk

brlane-rht:
This will allow it to be used from other places more easily.


supakeen (dismissed):

supakeen (approved):

supakeen:
The function could probably be named something that scopes it to ISO/PXE things as it's quite a small ESP but that's not that important to me at the moment 🙂

lzap (approved):

PR #2359, opened 2026-05-21
Drop gitlab dynamic pipelines [HMS-9712]

### LINKED ISSUE #1703: Reduce runner resource usage when doing full rebuilds in CI
PRs like #1698 always need some CI babysitting.  Updating the osbuild commit ID causes CI to rebuild all test images.  This always creates issues due to concurrent runner limits or, more commonly, AWS task limits when boot testing.

```
Register(): ResourceCountLimitExceeded: Task failed to initialize - task limit (40) exceeded.
	status code: 400, request id: 2be9e766-86bc-4a9d-bd11-7c70ba2b7849
exit status 1
```

We need to somehow reduce the number of concurrent tasks, especially when booting images in AWS.

Some initial ideas:
- Catch the task limit error, wait and retry.
- Query AWS for the number of current tasks, if possible, and wait until the number is a lot lower than the limit (probably racy).
- Split off the boot tests into a separate job in the pipeline.  Do multiple boot tests sequentially in the same job.

achilleas-k:
This PR simplifies image building and testing in Gitlab CI by removing the dynamic pipeline generation and instead builds and tests all images for a given distribution and architecture on the same runner.

The imgtestlib has been refactored into a module with multiple files for easier navigation, as it was getting too big for a single file.

Some further improvements I'd like to do after this is merged:
- Async "touch" for S3 objects.
- Async boot tests.
- Return errors from build and boot functions.  Currently the test functions rely on the `sp.run()` shell commands failing to fail a build.  I'd like to capture those errors instead and handle them gracefully.  That way we can generate clean failure messages.  Also it would make it possible to continue with other image builds when a build or boot test fails.
- Merge vmtest into imgtestlib.

Closes #1703


achilleas-k:
The PR moves the core parts of the test scripts into the imgtestlib module.  The `boot-image` script uses Python's `match` statement, which isn't available on EL9.  This wasn't an issue before because `boot-image` was only ever run on the CI runners, which are Fedora 42.  Now that the core functionality is part of the importable module though, we need to rewrite it to run on older Python versions.

We should be testing builds on EL9 as well, so I should do this regardless.

supakeen:
So; since there are a lot of failures I went through them:

1. 7 jobs succeeded.
2. 20+ jobs got their instance killed.
3. Jobs fail when testing installers, as they need access to KVM and it isn't available.
4. A few failures due to: `time="2026-05-21T16:30:35Z" level=fatal msg="Error parsing image name \"docker://None\": invalid reference format: repository name must be lowercase"`

achilleas-k:
> So; since there are a lot of failures I went through them:

Thanks for going through them!!

>     1. 7 jobs succeeded.

Not great.


>     2. 20+ jobs got their instance killed.

I suspect this will be the biggest issue with this change.


>     3. Jobs fail when testing installers, as they need access to KVM and it isn't 

Ugh, right, yeah.  I guess we're going to need to run everything on KVM-enabled runners since every distro has an installer.

>     4. A few failures due to: `time="2026-05-21T16:30:35Z" level=fatal msg="Error parsing image name \"docker://None\": invalid reference format: repository name must be lowercase"`

I think I fixed that?  Anyway, definitely fixable.

lzap:
Observation: average job time was 1 hour and the slowest one was 4 hours.

> 2\. 20+ jobs got their instance killed.

We must start tracking these, I wonder if we pay actually more than if we were not using spot. Because when a spot instance is killed by AWS for capacity reasons, we still pay the time on the clock. AWS sends a signal 2 minutes before the term/kill so we can mark those jobs for later inspection and statistics.

achilleas-k:
Rebased on `main` but deleted `.gitlab-ci.yml`.  I want to try a few things before rerunning the pipelines.  Setting to draft.

achilleas-k:
I'll rebase this on #2383 and start experimenting with doing some things async

achilleas-k:
Current state captures all output from the build and turns saves it as a job artifact.  The boot tests also generate a lot of output though, so we're still reaching the limits of the job log viewer.

Doing boot tests async and saving the output as an artifact as well will help.  I'm not sure if doing multiple _local_ (KVM) boot tests in parallel is a good idea though.  Our runners would probably get overloaded quickly if we start boot testing an ISO and a couple of qcows at the same time.

PR #2360, opened 2026-05-21
distro: Make isoCustomizations usable by bootcImageType

brlane-rht:
The function only depends on ISOLabel() and getDefaultISOConfig(). Make it take a new interface, ISOImageType, so that it can be shared between regular images and bootc images.

Related: HMS-10627

### DIFF
--- a/pkg/distro/generic/bootc_imagetype.go
+++ b/pkg/distro/generic/bootc_imagetype.go
@@ -130,6 +130,14 @@ func (t *bootcImageType) RequiredBlueprintOptions() []string {
 	return nil
 }
 
+func (t *bootcImageType) getDefaultISOConfig() (*distro.ISOConfig, error) {
+	if !t.ImageTypeYAML.BootISO {
+		return nil, fmt.Errorf("image type %q is not an ISO", t.Name())
+	}
+	d := t.Arch().Distro()
+	return t.ISOConfig(d.ID(), t.arch.arch.String()), nil
+}
+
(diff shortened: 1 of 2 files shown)


supakeen (approved):

pkg/distro/generic/bootc_imagetype.go
```diff
@@ -130,6 +130,14 @@ func (t *bootcImageType) RequiredBlueprintOptions() []string {
 	return nil
 }
 
+func (t *bootcImageType) getDefaultISOConfig() (*distro.ISOConfig, error) {
+	if !t.ImageTypeYAML.BootISO {
+		return nil, fmt.Errorf("image type %q is not an ISO", t.Name())
+	}
+	d := t.Arch().Distro()
+	return t.ISOConfig(d.ID(), t.arch.arch.String()), nil
```
thozza, line 138: Please use named fields when initializing the struct, to minimize confusion and make the code a bit more explicit. We never use positional struct initialization for any other cases than slices of primitive types.

thozza (commented):
Functionally LGTM, but there is one style / common practice inconsistency.

pkg/distro/generic/bootc_imagetype.go
thozza: My bad, there is a Type with the same name as an `ImageTypeYAML` method. This is fine.

thozza (approved):
I got confused by a struct and method of the same name. All good.

PR #2361, opened 2026-05-21
Add BootcRootFS pipeline and use it for bootc pxe-tar-xz

brlane-rht:
This moves the rootfs.img setup into its own pipeline so that it can be shared between PXE and ISO bootc creation.

One slightly ugly bit is that I'm using ISOCustomizations to control the rootfs compression type and optional erofs config settings. It's not ideal, but I think it's better than also changing everything around. But it is a good candidate for future cleanup.

~This set builds on top of #2360 and needs osbuild v182 for the squashfs/erofs mount support so it's currently a daft.~


supakeen:
> This moves the rootfs.img setup into its own pipeline so that it can be shared between PXE and ISO bootc creation.
> 
> One slightly ugly bit is that I'm using ISOCustomizations to control the rootfs compression type and optional erofs config settings. It's not ideal, but I think it's better than also changing everything around. But it is a good candidate for future cleanup.
> 
> ~This set builds on top of #2360 and needs osbuild v182 for the squashfs/erofs mount support so it's currently a daft.~

Yea I think I initially had something like `crofs` (a name people didn't like) that would control the "compressed read only filesystem" pipeline or stages if they exist. We need to rethink a bit how to pass customizations/definitions on a per-pipeline basis without encoding pipeline names into it.

supakeen (dismissed):

supakeen (dismissed):

pkg/manifest/bootc_rootfs.go
```diff
@@ -0,0 +1,161 @@
...
+		Paths: []osbuild.CopyStagePath{
+			{
+				From: "mount://-/boot/efi/EFI",
+				To:   "tree:///EFI",
+			},
+		},
+	}
+	copyStage := osbuild.NewCopyStage(copyStageOptions, nil, devices, mounts)
+	pipeline.AddStage(copyStage)
+	*/
```
achilleas-k: Keeping this?

achilleas-k (dismissed):
Thanks.  LGTM.
Minor comment about that disabled code block.

pkg/manifest/bootc_rootfs.go
brlane-rht: Good catch, forgot to delete that. The EFI copy happens in the tree stage and isn't needed here.

supakeen (approved):

achilleas-k (approved):

PR #2366, opened 2026-05-25
config-list: narrow down the matrix

lzap:
Reduces the test configuration matrix by eliminating redundant coverage and keeping only representative test cases for each customization type.

### Changes

  **Drop dedicated firewall.json entry**
  - Remove 24 minimal-raw firewall test jobs across RHEL 8/9 and CentOS 9
  - Firewall functionality remains covered via all-customizations boot test paths

  **Drop oscap-rhel9 in favor of JSON tailoring**
  - Remove plain `oscap-rhel9` test configuration (subset of the kept variant)
  - Keep `oscap-rhel9-with-json-tailoring` on `rhel-9.4 ami` as representative

  **Combine centos-9 modularity into one config**
  - Merge `modularity_in_packages` and `modularity_enabled_modules` into single `modularity.json`
  - Single test now validates both the packages shorthand and enabled_modules code paths
  - Two birds, one strike

Edit: Dropped some of the proposals. I think the (1) is no brainer, the other two are for discussions.


achilleas-k:
What's up with the commit authors?  Also, was all of this done with cursor?  I'd rather a human go through and check the coverage of each test config, check why it was added when it was and if it's actually redundant rather than rely on an LLM for it.

achilleas-k (changes requested):
Commit authors are messed up.

lzap:
> What's up with the commit authors?

I managed to set my config to Schutzbot last week when I was working on some `gobump` changes. Fixed.

> Also, was all of this done with cursor?

Not at all, I used Cursor to analyze each and every config and out of about 20 recommendations I selected the candidates hoping we can discuss in the PR what can be merged. Well, git history was not always useful but you are right that for "releasever" it nicely showed that it is actually significant and I dropped this commit.

I also dropped two "narrowing" commits, I was too optimistic with these I think I know where we would end up already. Rebased.

lzap:
Forgot to push, dropped commits there were maybe too much. Simplified the PR description.

supakeen (commented):
Please fix the commit authors, these commits weren't created by Schutzbot. Otherwise this is fine in my book.

lzap:
> Please fix the commit authors, these commits weren't created by Schutzbot. Otherwise this is fine in my book.

OMG I forgot to fix the damn thing. Rebased. What a week, Captain!

achilleas-k (approved):
Nice.  Let's get this merged.

supakeen (approved):

PR #2369, opened 2026-05-25
Fix non-deterministic manifest generation (HMS-10718)

lzap:
This PR fixes non-deterministic manifest generation that was causing CI checksum validation failures. The issue manifested as random differences in generated manifests across runs, particularly in flatpak container references swapping between architectures (x86_64 ↔ aarch64).

### Problem

Manifest generation was non-deterministic. I made checksum generation faster and then wrote a tool that allows for quick re-testing for consistency. This allowed Claude to iterate and confirm various ideas.

### Solution

This PR implements a layered fix addressing each root cause:

- **gen-manifests: use canonical JSON for checksums** - Added `-checksums-only` flag to generate only checksums that marshals, unmarshal and marshal again to take advantage of Go stdlib JSON library that will automatically sort keys. As a bonus, this is faster than doing this externally in a shell script (2.0x speedup: 10.3s → 5.3s).
- **tools: add test-manifest-checksums.sh** - Testing tool that generates manifests N times and diffs them to detect non-determinism. This helped to validate the correctness of the solution.
- **cicd: show diff on validate checksum error** - Enhanced error reporting to show actual differences when checksums mismatch. This will help us in the future to directly see the problem, it is not ideal because theoretically manifest generation can fail and then manifest diff can miss the flakyness and show nothing but it is better to try.
- **gen-manifests: fix race condition in workerQueue** - Added `sync.Mutex` to protect concurrent map access in the worker queue. This did not fix the issue yet but it is a race condition that should be fixed.
- **distro: fix InstallerConfig deep copy for Flatpaks template expansion** - The mutex alone wasn't sufficient; created shallow copy of InstallerConfig struct + deep copy of Flatpaks slice before template expansion to prevent concurrent modification of shared YAML-loaded data

The deep copy is necessary because `ExpandTemplates()` modifies `Flatpaks[i].References[j]` in place to expand templates like `{{.Arch}}`. Without the copy, multiple architectures would modify the same slice, causing random architecture values.

### Verification

Tested with `tools/test-manifest-checksums.sh` generating manifests 10 times consecutively:
- **Before**: Failed on run 2-3 with random flatpak architecture mismatches
- **After**: ✓ All 10 runs produced identical manifests

Edit: I reordered commits and added temporary disablement of manifest checksum check.


pkg/distro/installer_config.go
```diff
@@ -59,6 +59,69 @@ func (c *InstallerConfig) InheritFrom(parentConfig *InstallerConfig) *InstallerC
...
+					regCopy := *fp.Registry
+					fpCopy.Registry = &regCopy
+				}
+				if fp.References != nil {
+					fpCopy.References = append([]string(nil), fp.References...)
+				}
+				copy.Flatpaks[i] = &fpCopy
+			}
+		}
+	}
```
supakeen: The comment isn't very nice and perhaps we should move the types from inline to public so they don't get duplicated here.

supakeen:
Nice. I had indeed found that the flatpak/installerconfig templating was the culprit a few days ago but hadn't had time to address it.

Could you update this PR? The title doesn't really clearly reflect the contents anymore. Also left one comment.

This also needs to attach updated manifest checksums in case they were wrongly generated initially. Probably amended to the deepcopy commit.

lzap:
I have reordered commits, added a special tool to test for mfgen flakyness, also sped up manifest generation by 50%. I was able to consistently reproduce using the new script. Then I retested every single commit and I have found that we only need one change actually - that is the template expanding. I also slightly changed the deep copy to `WithExpandTemplates` approach let me know if you like it better.

pkg/distro/installer_config.go
lzap: I changed for a slightly different solution, if you do not like the new one I can go back to this.

supakeen (dismissed):
I guess the checksum failures here are expected since the 'wrong' files are in `main`?

lzap:
> I guess the checksum failures here are expected since the 'wrong' files are in `main`?

Yes, I need to do one more rebase I missed go linter tho. Locally I am unable to see any flakyness:

```
$ tools/test-manifest-checksums.sh 20
Checking if gen-manifests compiles
github.com/osbuild/images/cmd/gen-manifests
Generating and comparing manifests 20 times...
  Run 1/20
  Run 2/20 - comparing...
  Run 3/20 - comparing...
  Run 4/20 - comparing...
  Run 5/20 - comparing...
  Run 6/20 - comparing...
  Run 7/20 - comparing...
  Run 8/20 - comparing...
  Run 9/20 - comparing...
  Run 10/20 - comparing...
  Run 11/20 - comparing...
  Run 12/20 - comparing...
  Run 13/20 - comparing...
  Run 14/20 - comparing...
  Run 15/20 - comparing...
  Run 16/20 - comparing...
  Run 17/20 - comparing...
  Run 18/20 - comparing...
  Run 19/20 - comparing...
  Run 20/20 - comparing...
✓ All 20 runs produced identical manifests
```

Rebased, also added the new deterministic tool into `.github/workflows/validate-checksums.yml` to be executed as the last step. But yeah, this might need an admin merge.

supakeen:
> > I guess the checksum failures here are expected since the 'wrong' files are in `main`?
> 
> Yes, I need to do one more rebase I missed go linter tho. Locally I am unable to see any flakyness:
> 
> ```
> $ tools/test-manifest-checksums.sh 20
> Checking if gen-manifests compiles
> github.com/osbuild/images/cmd/gen-manifests
> Generating and comparing manifests 20 times...
>   Run 1/20
>   Run 2/20 - comparing...
>   Run 3/20 - comparing...
>   Run 4/20 - comparing...
>   Run 5/20 - comparing...
>   Run 6/20 - comparing...
>   Run 7/20 - comparing...
>   Run 8/20 - comparing...
>   Run 9/20 - comparing...
>   Run 10/20 - comparing...
>   Run 11/20 - comparing...
>   Run 12/20 - comparing...
>   Run 13/20 - comparing...
>   Run 14/20 - comparing...
>   Run 15/20 - comparing...
>   Run 16/20 - comparing...
>   Run 17/20 - comparing...
>   Run 18/20 - comparing...
>   Run 19/20 - comparing...
>   Run 20/20 - comparing...
> ✓ All 20 runs produced identical manifests
> ```
> 
> Rebased, also added the new deterministic tool into `.github/workflows/validate-checksums.yml` to be executed as the last step. But yeah, this might need an admin merge.

The manifest checksums are not required for merging.

supakeen (dismissed):

cmd/gen-manifests/main.go
```diff
@@ -768,6 +797,20 @@ func main() {
 		}
 		exit = 1
 	}
+
+	// If checksums-only mode and no errors, write checksums to files
+	if checksumsOnlyMode && exit == 0 {
+		// Write checksum files to the output directory
+		for filename, checksum := range checksums {
+			checksumPath := filepath.Join(outputDir, filename)
+			if err := os.WriteFile(checksumPath, []byte(checksum+"\n"), 0644); err != nil { //nolint:gosec // G306 - checksum files are not sensitive
```
achilleas-k, line 824: The newline at the end of the file was an artifact of the previous generation method.  It would be cleaner to not have it I think.

cmd/gen-manifests/main.go
```diff
@@ -59,8 +59,8 @@ func (e *panicError) Error() string {
 // Global state for checksums-only mode
 var (
 	checksumsOnlyMode bool
-	checksums        = make(map[string]string)
-	checksumsMu      sync.Mutex
```
achilleas-k: Why did the indentation change here?

achilleas-k (dismissed):
Thanks for the fix.  A couple of small notes but nothing serious.

I think a large part of the cause here is the architecture of the config loading and sharing from the YAML files.  The loader layer is very leaky.  In the long run, we should rearchitect things so that image type instantiation either loads the configs every time or gets a copy of some internal cached state.  The lowest level configs loaded from the YAML files should be immutable.

cmd/gen-manifests/main.go
achilleas-k: Never mind.  Outdated diff.

achilleas-k:
> But yeah, this might need an admin merge.

Not sure I follow the reason for this?  I see the same issue locally.  Modifying the offending commit should fix it, no?

lzap:
> Not sure I follow the reason for this?

Yeah sorry I had impression manifest check is required. My bad.

lzap:
Let's merge so things are clean in `main` and I will do a followup.

cmd/gen-manifests/main.go
lzap: I wanted to keep the parity for this fix, I can do a followup. But I am in the "each text file should end with a newline" camp :)

lzap:
Wait @supakeen said manifest validation is not required but I do see it required?

image

supakeen:
> Wait @supakeen said manifest validation is not required but I do see it required?
> image

So do I, thanks must've changed.

lzap:
> Modifying the offending commit should fix it, no?

Not sure what you mean, manifest generation is currently flaky. What I could do is reorder commits so the main fix comes first, that should do it.

lzap:
I reordered commits so the main fix comes first, that should fix the CI.

achilleas-k:
Snyk was complaining about the use of `sha1`.  I dismissed the issue.

achilleas-k:
> Update - Actions is experiencing degraded performance. We are continuing to investigate.
> May 26, 2026 - 12:17 UTC

Swell.

achilleas-k (dismissed):

lzap:
All Systems Operational! Made an empty rebase.

lzap:
So initially I had two more commits which Claude identified as possible offenders but I retested everything commit by commit with the test script 20 times and it seemed that the other two commits fixed nothing - at least I could not find any reproducer locally so I dropped them both.

Now it turns out that on CICD there are still non-deterministic results while locally I am unable to reproduce it. In both environments we use Go 1.24.12 on x86_64, I am not sure what is different, maybe kernel settings? Anyways, I am bringing the commit back to see if that fixes CICD. And if not I will bring the other commit as well. The first one is random iteration order of Go maps for our manifest mocks - pretty straightforward nobrainer.

lzap:
And I have to add the other commit as well, that one solves also ordering of maps in Go, but for sources. Similar problem, different part of the codebase. Initially Claude came up with this but I could not reproduce locally but CI clearly have a problem with that, I do not know what is different maybe CPU or number of cores?!? Not sure. Adding it back.

Now, since there are three distinct commits that fix three different issues with undeterministic behavior, I am going to add one small patch that will temporarily turn off failure in `.github/workflows/validate-checksums.yml`, then the three main fixes will follow, then another commit will turn that on and the rest of the patch series (5 commits) will be tested each. The last commit will finally also experience additional `test-gen-manifests.sh`.

achilleas-k:
@lzap This is getting a little messy and comments like
> Claude came up with this but I could not reproduce locally but CI clearly have a problem with that, I do not know what is different maybe CPU or number of cores?!? Not sure

aren't really helping make any of it clearer.  I'm also seeing commits authored by Schutzbot again.

I'm not a fan of the new marshallers either, tbh.  We shouldn't need that level of key ordering, really.  osbuild manifests should have the same ID regardless of key order for objects.  Our checksumming is a test and validation measure and hashing the raw manifests was a quick and dirty way to achieve that.  So far it's been fine but, if it's not any more, I'd rather we didn't spread the messiness by having these custom marshallers and instead make the test and validation things work properly.  The downside is that getting the _real_ manifest ID can be slow (`osbuild --inspect`).  Fortunately, we can probably get it to be fast enough by doing them in parallel.

lzap:
> This is getting a little messy

Duh, well, since I do not see the behavior locally I am on trial and error here. But you got a point, I did not like these commits either I was merely trying to test if they actually help.

I am gonna draft this I will need few more rebases, I have an idea how to calculate a better checksum without too much trouble and then I can drop those ordering commits.

achilleas-k:
Can we cherry pick the flatpak fix at least?  That change on its own should work without issues and fix our problem of non-deterministic checksums.

lzap:
> Can we cherry pick the flatpak fix at least?

How this is going to get in, tests fail constantly.

achilleas-k:
> > Can we cherry pick the flatpak fix at least?
> 
> How this is going to get in, tests fail constantly.

https://github.com/osbuild/images/pull/2373

lzap:
This PR diverged quite a bit, @achilleas-k cherry picked the most important commit and I got stuck. But some of the ideas I want to carry over to a new PR.

PR #2370, opened 2026-05-26
bootc: assert non-nil kernel info

supakeen:
We use the kernel information we read from the image for ISOs. Sometimes it is not available as we can't find the kernel in the container image.

If this happens we panicked on a `nil`, let's print a slightly more helpful error.

---

This 'resolves' https://github.com/osbuild/image-builder/issues/2424 in that the panic is no longer shown, but we still *require* `/usr/lib/modules/*/(vmlinuz|initramfs.img)` to exist as per: https://osbuild.org/docs/developer-guide/projects/image-builder/advanced/bootc/isos/#generic


thozza (approved):

brlane-rht (approved):
Looks good

lzap (approved):
Restarted failed tests.

PR #2371, opened 2026-05-26
stage/grub2: allow VFAT for boot fs

supakeen:
The grub2 stage implementation here panicked on non-UUID values for the boot filesystem. When `/boot` is VFAT the UUID won't be an actual UUID.

In `osbuild` we already accept both values. Let's allow both values here as well but still validate the UUID as we did before when it's *not* VFAT.

Note that according to the bootloader specification [1] /boot 'SHOULD' be VFAT.

Also note that this only helps with XBOOTLDR being VFAT. We probably need to adjust the code more in the future when we mount the ESP to /boot. Probably by expanding it to check if it's the ESP in which case we *don't* inject the boot fs uuid at all?

### DIFF
--- a/pkg/osbuild/grub2_stage.go
+++ b/pkg/osbuild/grub2_stage.go
@@ -18,7 +18,7 @@ import (
 type GRUB2StageOptions struct {
 	CompatVersion      int          `json:"compat_version,omitempty"`
 	RootFilesystemUUID uuid.UUID    `json:"root_fs_uuid"`
-	BootFilesystemUUID *uuid.UUID   `json:"boot_fs_uuid,omitempty"`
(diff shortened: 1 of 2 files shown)


thozza (dismissed):
A unit test would be great, otherwise LGTM.

supakeen:
> A unit test would be great, otherwise LGTM.

Ah yea, there weren't any yet but I'll push an additional commit for it.

thozza (dismissed):
Thanks!

lzap (dismissed):
Saw on Slack, yeah.

supakeen:
Mmm, I could've sworn I ran `go fmt`; needs re-acks because I added a `<tab>` 😞

thozza (approved):

lzap (approved):

PR #2372, opened 2026-05-26
container: guess at mime type

supakeen:
When switching to the skopeo based resolver it was missed that the media type property is optional. If this is the case then the `containers/image` library guesses at the type since we don't always have access to the transfer information.

Let's call that function directly here; but only if it wasn't set initially.

When we drop `containers/image` dependencies entirely this function will need to be interned.

---

Resolves: https://github.com/osbuild/image-builder-cli/issues/527

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -497,7 +497,6 @@ func (cl *Client) resolveRawManifest(ctx context.Context, rm RawManifest, local 
 			return resolvedIds{}, nil, nil
 		}
 		imageID = m.ConfigInfo().Digest
-
 	default:
 		return resolvedIds{}, nil, fmt.Errorf("unsupported manifest format '%s'", rm.MimeType)
 	}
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
Thanks for looking into this (and fixing it, of course).

achilleas-k:
I would suggest a test but we can add some test cases when we add our own version of the function.

supakeen:
> I would suggest a test but we can add some test cases when we add our own version of the function.

It's hard to test; it seems that *sometimes* the registry responds without the mediaType and sometimes with? Maybe depending on the exact hash?

achilleas-k:
I meant more of a unit test, which isn't really necessary now since the "unit" to test isn't our own code.  But we could prepare some for when we replace it.  Can do it in a follow-up though.

supakeen:
Let's do follow-up.

thozza (approved):

lzap (approved):
Manifest fix is in the queue now, crossing fingers.

PR #2373, opened 2026-05-26
distro/loader: expand installerConfig template by copying (HMS-10718)

achilleas-k:
Deep copy the InstallerConfig when expanding templates to avoid leaking state on the base (loaded) object between consecutive applications of the variables in the template.


lzap (approved):
And of course you get green light right from the get-go.

achilleas-k:
> And of course you get green light right from the get-go.

Well, the only problem on `main` right now is the installer config leakage.  This should stabilise the issue.  Everything else can be addressed slowly and cleanly.

lzap:
Sure I had this commit together with pretty irrelevant other commits and it failed multiple times. I dunno, maybe the multi-commit validation is not right or something. I am gonna pick it up after this is merged.

brlane-rht (approved):
Looks good, but at some point I think the struct needs a nice big NOTICE added so that future changes make sure they add to the deep copy if needed.

PR #2374, opened 2026-05-26
gitlab: exit with error code when manifests fail to validate

achilleas-k:
The generate-manifests jobs in gitlab CI are meant to catch invalid manifest generation for the two architectures we don't build test: ppc64le and s390x.  The way the test is written however, errors are printed but do not cause the execution to fail.

Fix the test script to exit with an error code.  The exit code is the number of manifests that failed to validate.


achilleas-k:
The CI tests should now fail for the network installer: https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/14548808415

lzap (dismissed):
LGTM but this will fail, huh? You just want to confirm?

achilleas-k:
> LGTM but this will fail, huh? You just want to confirm?

Yeah, I wanna make sure the fix works by getting a test failure before fixing the problem.

brlane-rht (dismissed):
Looks ok, but that sure it a lot of almost identical scripting.

achilleas-k:
> Looks ok, but that sure it a lot of almost identical scripting.

The `.gitlab-ci.yml` file is generated from the python script, so the real change is only in one place.

lzap (approved):
Makes sense.

brlane-rht (approved):
Looks good

PR #2375, opened 2026-05-26
New Manifest Checksum Caching Implementation

lzap:
## Overview

We have a "fake" manifest checksum implementation that can be fragile sometimes. We always had a plan to use `osbuild --inspect` to do proper ID resolution. This patch series bring this.

But there is a snag - calling `osbuild` is costly, about 300ms per call. That is too much, a full manifest generation takes about a minute on my system. Therefore, part of this patch, there is a simple cache implementation. We keep calculating sha sums from the manifest blobs (upgraded from sha1 to sha256) but this is only used as a cache key. This speeds up checksum validation by ~100x (from ~200ms to ~2.5ms per cached manifest).

Manifest checksum format changes, instead of some undefined sha, it is now manifest ID for each pipeline in this format:

```
d7e0c83abe99e164736d5c4b68dda1442cfcb82779b4dc5174f708fb3cb2594a: build
d0348416e3643ee607fde1d54c51f134ea8a4d45313276ceec701f360a110662: os
50bcbf77d3030db338b2ef32eaedfe0fbee950388884f525842d158ad9baeee0: image
4e5aefe649fd4c051fcbf453d1d9d8b3fc1c68399e90d1c3d21c57e412ac1451: xz
```

If I understand how `osbuild --inspect` works, this should give us less conflicts since it will ignore some (source) portions of data and ordering will be always correct. Pipelines can differ depending on the image type of course.

Further optimalizations were done by moving some code from shell script into Go which means that overall, this implementation is about 50% faster when cache is fully pre-populated. Regeneration of roughly 10-30 manifest then gives the same performance as before, which is a good result I think - we get better checksums for no cost other than a small cache.

An extra feature of tracking history of each checksum file (as long as it is in the cache) was added. When you pass an existing file, it will show all available entries in the cache:

```
lzap@dev:~/images$ go run ./cmd/gen-manifests -history test/data/manifest-checksums/rhel_10.3-aarch64-ec2-empty
2026-05-27 06:24:32 UTC 28e7ea69057c58a9c47d7749056cc46b280ca14745d4475a42abdbf60774707f
```

When you pass a sha256, it will output the actual output from `osbuild --inspect`. This way you can always take a quick look on the latest (or any historical version that is *in your own cache*):

```
lzap@dev:~/images$ go run ./cmd/gen-manifests -history 28e7ea69057c58a9c47d7749056cc46b280ca14745d4475a42abdbf60774707f
{ ... }
```

Therefore it should be able to compare history with something like (pseudo command): `diff (< gen-manifest -history XXX | jq) (< gen-manifest -history YYY | jq)`. I know the ideal would be some sort of central database (git repo) of all manifests or have them directly in `images` but that would be probably a lot of data. The current cache is around 10 MiB and we often rotate it. A central solution is out of scope for this initial refactoring, but I am open to suggestion.

Osbuild version is added to the key calculation and the version itself is cached too because calling `osbuild --version` itself takes 200ms. This ensures that an upgrade will invalidate cache entries. There is no automatic garbage collection because of the history feature that can be useful so I would rather keep it on my system forever. But if needed, simple deletion will do the trick.

### Cache System
- **Location**: `~/.cache/osbuild-gen-manifests` (XDG-compliant)
- **Cache key**: `SHA256(canonical JSON + osbuild version)`
- **Structure**: 256 subdirectories (first 2 hex chars) for efficient filesystem performance
- **Automatic invalidation**: Cache keys include osbuild version, so upgrades naturally use new cache entries
- Version cached in `~/.cache/osbuild-gen-manifests/version`
- Timestamp-based invalidation: only calls `osbuild --version` if binary is newer than cache file
- Cache inspection: `-history` flag shows cache entries per checksum file or contents of actual (historical) osbuild output

### Nitpicks
- Default workers: CPU count (was hardcoded to 16 I think)
- Ability to create SMC label in commit message to skip commit per commit manifest check

### GitHub Actions Integration
- Added official cache action to `.github/workflows/validate-checksums.yml`
- Cache key includes osbuild version and workflow file hash
- Shared cache across PR commits for faster CI runs

### Statistics
Displays after checksum generation:
- Cache hits and misses
- Average manifest generation time
- Average osbuild inspect call duration (when cache misses occur)


cmd/gen-manifests/main.go
```diff
@@ -411,11 +420,45 @@ func save(ms manifest.OSBuildManifest, depsolved map[string]depsolvednf.Depsolve
...
+		h.Write(canonicalBytes)
+		checksum := hex.EncodeToString(h.Sum(nil))
+
+		// Store checksum with filename (without .json extension)
+		checksumKey := strings.TrimSuffix(filename, ".json")
+		checksumsMu.Lock()
+		checksums[checksumKey] = checksum
+		checksumsMu.Unlock()
+		return nil
+	}
```
achilleas-k, line 453: It's a little confusing that this block is in the `save()` function, which normally writes the manifests to file, but instead of writing the files it collects the checksums to write them later.  What's the reason for this difference in the flow of the two cases?  If we write the files immediately here, we wouldn't need the map (or it's mutex).

cmd/gen-manifests/cache.go
```diff
@@ -0,0 +1,222 @@
...
+	if versionInfo, err := os.Stat(versionPath); err == nil {
+		// Version file exists, check if it's newer than the binary
+		if versionInfo.ModTime().After(binaryModTime) {
+			// Cache is newer, use it
+			cachedVersion, err := os.ReadFile(versionPath)
+			if err == nil && len(cachedVersion) > 0 {
+				return strings.TrimSpace(string(cachedVersion)), nil
+			}
+		}
+	}
```
achilleas-k, line 68: What's the value in comparing the mod times?  I would assume a single call to `osbuild --version` for each run of `gen-manifests` wouldn't be too much of an issue and would be perfectly reliable.
Consider, for example, that on ostree / bootc based systems, the mod time of the binary is always epoch.

cmd/gen-manifests/main.go
```diff
@@ -444,7 +452,40 @@ func parseInspectedJSON(inspectedJSON []byte) string {
...
+		// Cache miss - run osbuild --inspect
+		var err error
+		inspectedJSON, err = osbuild.OSBuildInspect(manifestJSON)
+		if err != nil {
+			// Propagate osbuild errors instead of silently returning empty string
+			return "", fmt.Errorf("osbuild --inspect failed: %w", err)
+		}
+
+		// Cache the raw JSON output for next time
+		_ = writeCache(cacheKey, string(inspectedJSON))
```
achilleas-k, line 484: I guess we ignore the error because we don't want a cache write failure to be fatal?  Could we print it as a warning to avoid confusion when things don't work as expected?

achilleas-k: Also, it's a bit strange that in all of this effort to optimise, we write back the manifest to the cache even when in cases where we just had a cache hit a few lines before.

cmd/gen-manifests/cache.go
```diff
@@ -0,0 +1,222 @@
...
+	var sums []string
+	for _, line := range strings.Split(string(data), "\n") {
+		line = strings.TrimSpace(line)
+		if line != "" {
+			sums = append(sums, line)
+		}
+	}
+
+	return sums, nil
+}
```
achilleas-k, line 189: What's the value of the history function?  I can sort of imagine it being useful to check how a manifest changed during the development of a branch, but just having a history of checksums without them being associated with something like a commit ID wouldn't tell us much.

.github/workflows/validate-checksums.yml
```diff
@@ -57,12 +57,22 @@ jobs:
...
           # commit ID of target branch
           base_sha: ${{ github.event.pull_request.base.sha }}
         run: |
+          # Check if last commit has [SMC] marker (Squashed Manifest Commit)
+          last_commit_msg=$(git log -1 --pretty=%s)
+          smc_mode=""
+          if echo "$last_commit_msg" | grep -q "SMC"; then
+            smc_mode="true"
+            echo "Last commit contains 'SMC' marker - validation will not fail on checksum differences"
+          fi
```
achilleas-k, line 74: Not sure I like this.  We either want manifest checksums changing with the commit that modified them or we don't.  I don't see a good reason to have a special case to skip the check.

The commit message says
> This is useful for large patch series where manifest regeneration is
done in a final squashed commit to avoid rebasing complexity and
reduce CI runtime.

But a large patch series is exactly the scenario where I would want to see how manifests change at every step.

achilleas-k (commented):
Some nice work and ideas overall, but there's a lot of side-features of unclear value.

Also, the description in one of the commits (`gen-manifests: add caching for osbuild inspect results`) seems to be a squashed version of multiple commits.  Can you clear that up, please?

thozza (changes requested):
Thanks for looking into this, but I have a few basic questions and a concern that is a blocker from my PoV.

> ## Overview
> We have a "fake" manifest checksum implementation that can be fragile sometimes. We always had a plan to use `osbuild --inspect` to do proper ID resolution. This patch series bring this.

Can you be specific on what was **fragile**? The justification is a bit too hand-wavy and jumps to fixing a problem that is not even clearly defined here. Moreover, the solution is quite complex and convoluted for what should be a straightforward checksum mechanism -- it's becoming a project of its own (caching system with 256-bucket subdirectories, cache invalidation via binary mod-time comparison, history tracking, statistics, SMC markers, GitHub Actions cache integration, etc.). Each layer exists to support the previous one: `osbuild --inspect` is slow, so we need caching, which needs invalidation, which needs version caching, and so on.

Additionally, can you point to the source of "We always had a plan to use `osbuild --inspect`"? Because the comment in the script you're modifying clearly states otherwise:

```
# NOTE: 'osbuild --inspect' is generally a better way to calculate a manifest
# fingerprint, because it ignores things like pipeline names, source URLs, and
# generally things that don't affect the build output.
# For mocked manifests though we want those things to be visible changes, so we
# calculate the checksum of the file directly. Also it's faster.
```

This comment is still present in the updated script, but the script now contradicts it by using `--checksums-only` (which runs `osbuild --inspect`) instead of `sha1sum` of the file.

## The blocker

The switch from `sha1sum` of the full manifest JSON to `osbuild --inspect` stage IDs breaks detection of repo snapshot updates. Stage IDs only depend on stage definitions (type, options, input references), not on the sources section where download URLs live.

At the same time, the repo pseudo-packages in `manifestmock.Depsolve()` explicitly strip the snapshot date from the URL before computing their checksum, so a date-only change (e.g., `-20240819` -> `-20250201`) doesn't produce a different pseudo-package checksum either. The result is that both detection paths are closed -- source URLs are ignored by `osbuild --inspect`, and the date is stripped before it can reach the stage inputs.

This means updating repo snapshots in `test/data/repositories/` produces identical checksum files. CI sees no difference, no rebuild is triggered, and regressions from new RPM package versions in the updated snapshot go undetected. We use pinned repo snapshots specifically to retest images against new package versions, so this silently defeats that workflow.

Note that the date stripping in `manifestmock` was fine *before* this change, because the old `sha1sum` approach still caught date changes through the regular package `RemoteLocations` in the curl sources (those include the full date-stamped URL). The date stripping only becomes a problem in combination with the switch to `osbuild --inspect`, which ignores sources entirely.

A straightforward fix would be to stop stripping the date suffix from repo pseudo-package URLs in `manifestmock.Depsolve()`. That way the snapshot date becomes part of the pseudo-package checksum, flows into RPM stage input references, and naturally changes the `osbuild --inspect` stage IDs when snapshots are updated.

Even with the blocker addressed, I'm not convinced we need all of this added infrastructure and features (caching, history, statistics, SMC markers) to solve a problem that hasn't been clearly defined yet. I'd prefer to start with a concrete description of what's broken today and the minimal change that fixes it.

achilleas-k:
I agree with the whole comment above but especially with this part:

> Even with the blocker addressed, I'm not convinced we need all of this added infrastructure and features (caching, history, statistics, SMC markers) to solve a problem that hasn't been clearly defined yet. I'd prefer to start with a concrete description of what's broken today and the minimal change that fixes it.

I will say though that I like the idea of pipeline IDs as manifest checksums.  It gives a bit more granularity to the change detection without going all the way towards having whole manifests checked in[^1].  But I do agree that we need to consider what exactly we want the change detection to detect.  I think for the purposes of this repo, things that do not affect the manifest _logically_ probably do matter.  For example, if we change the name of the OS pipeline, we would want to see a list of manifests affected by the change, which manifest IDs wouldn't show.  So if we go this way, perhaps we should consider having both.

[^1]: Having whole manifests checked in to the repo (mocked or otherwise) would be the best solution in terms of change detection and visibility, it would possibly remove a lot of sources of conflict, but it is the worst solution in terms of noise, large diffs, and raw repo data.  See [this old commit](https://github.com/osbuild/images/commit/40719798d01c4f222073ea949f5d531aefaed130) as an example.

lzap:
Good points, I somewhat took for granted that `osbuild --inspect` is the best possible outcome. I am gonna close this but one small patch from here is useful - moving away from forking `shasum` process to simply calculating this in the process - this speeds up generation by 50 %. Also I could perhaps upgrade from sha1 to sha256, or take it the opposite direction and just do CRC64? All we need to know is something has changed.

achilleas-k:
> moving away from forking `shasum` process to simply calculating this in the process - this speeds up generation by 50 %. Also I could perhaps upgrade from sha1 to sha256, or take it the opposite direction and just do CRC64? All we need to know is something has changed.

Yeah, that seems like a good change.
Any reason to prefer CRC64 over sha256?  I have no preference, it's just that we use sha256 generally (in osbuild, containers, etc).

thozza:
My preference is to use sha256, so that we do not risk the unlikely event of changing the manifest and ending up with unchanged checksum...

lzap:
Closing, I extracted the commits that were just making calculation faster: https://github.com/osbuild/images/pull/2382

Re: sha256 vs sha1 - I am not doing this in the linked PR because I want to keep things simple and clean there, but I have an idea which I will share in the opened issue.

PR #2377, opened 2026-05-27
pxe-tar-xz: Use the qcow2 packageset

brlane-rht:
In order to make the booted system consistent with other images use the generic_base_pkgset on Fedora. On RHEL 9 and 10 move the qcow2 packageset to its own section, 'qcow2_base_pkgset' and use that for qcow2 and pxe-tar-xz.

This results in a rootfs that is around 175-200MB larger but include that standard set of tools.


lzap (approved):
IDK I just think qcow2 is not the best starting point for PXE type. We can add easily stuff that simply does not belong in there. I actually think PXE should be separate, just add `@core` that should be enough. Also all those excludes are probably not relevant? Only git blame knows, have you reviewed these?

Having said all that, you are not making it worse too much, we are missing some basic tools like `sudo` on the PXE image, so yeah. 🤷🏻

ochosi:
I don't know that "not making it worse" should be the baseline.

lzap:
> I don't know that "not making it worse" should be the baseline.

Probably not, I am not yet too confident with blocking any work in images honestly. But my observation is that this is a pattern in distrodefs - we would often do something like this. And I think this PR is not the right place to do it - we should probably do a review as a separate refactoring ticket.

achilleas-k:
> IDK I just think qcow2 is not the best starting point for PXE type. We can add easily stuff that simply does not belong in there. I actually think PXE should be separate, just add `@core` that should be enough. Also all those excludes are probably not relevant? Only git blame knows, have you reviewed these?

I agree with this.  If nothing else, it'd be nice to have something like a `base_pkgset`, which could be an alias for `qcow2` to begin with, but makes it a bit clearer that pxe is not really based on qcow2.  So basically, the same thing we have for Fedora.

PR #2378, opened 2026-05-28
loader: allow custom path

supakeen:
Currently our definition loader always uses the embedded data filesystem exposed by `dataFS()`. It is possible to switch what `dataFS()` returns with an experimental environment flag.

We want to make loading of definitions a little bit less experimental by allowing a definition path to be set by calling code; instead of the library reading the environment.

A hurdle was that `distro.ParseID` loaded the definitions directly.

In this commit I introduce a new `Loader` type that embeds an `fs.FS`, `ParseID` becomes a method on this `Loader`. Where necessary old constructors that do not take a loader will automatically call the new WithLoader variants by passing the `BuiltinLoader` (which gives the same behavior).

This will allow us to pass custom paths in `image-builder` and `osbuild-composer` through (for example) a `--force-defs-path` and the potential to start shipping our definitions in a subpackage that installs into `/usr` which would allow for much easier downstream patching *and* other distributions shipping their own custom definition packages.

### DIFF
--- a/pkg/distro/defs/export_test.go
+++ b/pkg/distro/defs/export_test.go
@@ -13,3 +13,11 @@ func MockDataFS(path string) (restore func()) {
 		defaultDataFS = saved
 	}
 }
+
+func LoaderForTest(path string) *Loader {
+	return NewLoader(os.DirFS(path))
+}
+
+func ClearLoader(d *DistroYAML) {
+	d.loader = nil
+}
--- a/pkg/distro/defs/loader.go
+++ b/pkg/distro/defs/loader.go
@@ -43,14 +43,26 @@ var (
 var defaultDataFS fs.FS = distrodefs.Data
 
 func dataFS() fs.FS {
-	// XXX: this is a short term measure, pass a set of
-	// searchPaths down the stack instead
 	dataFS := defaultDataFS
 	if overrideDir := experimentalflags.String("yamldir"); overrideDir != "" {
 		olog.Printf("WARNING: using experimental override dir %q", overrideDir)
 		dataFS = os.DirFS(overrideDir)
 	}
 	return dataFS
+}
+
+type Loader struct {
+	fs fs.FS
+}
+
+func NewLoader(fsys fs.FS) *Loader {
+	return &Loader{fs: fsys}
+}
+
+// BuiltinLoader returns a Loader for the embedded distrodefs FS,
+// with the experimental yamldir override applied if it exists
+func BuiltinLoader() *Loader {
+	return NewLoader(dataFS())
 }
 
 // distrosYAML defines all supported YAML based distributions, since this can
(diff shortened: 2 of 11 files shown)


pkg/distro/defs/export_test.go
```diff
@@ -13,3 +13,11 @@ func MockDataFS(path string) (restore func()) {
...
 	}
 }
+
+func LoaderForTest(path string) *Loader {
+	return NewLoader(os.DirFS(path))
+}
+
+func ClearLoader(d *DistroYAML) {
+	d.loader = nil
+}
```
thozza, line 23: This could have been simply
```go
func (d *DistroYAML) ClearLoader() {
	d.loader = nil
}
```

pkg/distro/defs/loader.go
```diff
@@ -43,8 +43,6 @@ var (
 var defaultDataFS fs.FS = distrodefs.Data
 
 func dataFS() fs.FS {
-	// XXX: this is a short term measure, pass a set of
-	// searchPaths down the stack instead
```
thozza, line 47: I actually think that this would be a much nicer idea, than a fully forced override.

thozza (commented):
Thanks for this kick-off PR.

While the change does what it claims to do, I wonder what is the long term plan here? I think that a fully forced override of distro definitions is a feature that has a use case of its own (i.e. build systems). That said, I still think that the deleted code comment claiming that we want to have search paths is something that we should not just abandon. I think that a hierarchical override, like we allow for repo configs, is a really useful feature. It would allow selectively override just specific distro or image type, while allowing us to keep the fallback to the built-in definitions.

supakeen:
> Thanks for this kick-off PR.
> 
> While the change does what it claims to do, I wonder what is the long term plan here? I think that a fully forced override of distro definitions is a feature that has a use case of its own (i.e. build systems). That said, I still think that the deleted code comment claiming that we want to have search paths is something that we should not just abandon. I think that a hierarchical override, like we allow for repo configs, is a really useful feature. It would allow selectively override just specific distro or image type, while allowing us to keep the fallback to the built-in definitions.

The long term plan is as of yet to be defined. This change doesn't preclude the potential of having multiple search paths (`Loader` would then probably be changed to contain a `[]fs.FS` instead) but doing that in this PR would require solving the merging of definitions.

I assume in the future we will have `[]fs.FS` and in `image-builder` have both `--force-defs-dir` (to override all built-in definitions) and `--extra-defs-dir`, mirroring what we have for repositories though we'll start off with `--force-defs-dir` only and set the 'default' definition directory to a path in `/usr/share`.

thozza (approved):

pkg/distro/defs/export_test.go
supakeen: Mmm yes, not a big issue; I think I only have a use case for it in the testcases to make it easier to compare the structs and then my brain got tangled up and it ended up like this 🙂

brlane-rht (approved):
Looks good!

brlane-rht:
FWIW I like the idea of search paths, but don't like trying to merge things. I think it'll make things overly complex, and make it difficult for users to reason about the results. It's more clear if you just override the defaults with the content of the new path.

PR #2381, opened 2026-05-28
many: fix non-determistic manifest generation

ondrejbudai:
We want manifests to be generated in a deterministic way. A common source of non-determinism is Go is iterating over maps - Go actually doesn't guarantee the iteration order. Thus, when creating an array from a map, we always need to sort the map deterministically before creating the array.

My research found 3 places like this in the codebase, and this commit fixes them.

### DIFF
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -2,7 +2,9 @@ package generic
 
 import (
 	"fmt"
+	"maps"
 	"math/rand"
+	"slices"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


thozza (approved):

achilleas-k (approved):

supakeen (approved):
This kinda feels like the wrong places to do the sorting? It *feels* like this should live in the stages themselves or the manifest generator when it's serializing but it's also fine if it resolves an ongoing issue.

lzap (commented):
Nice find, I swear to God I was running into random checksums on CI that I could not reproduce locally this might be it.

lzap:
This got me curious how iteration in Go works through maps and it looks like Go always iterates in a slightly random order. This means that if there was any randomness, well, it is gone. I created a small test script and executed checksum regeneration 50 times and all good: https://github.com/osbuild/images/pull/2382

PR #2383, opened 2026-05-28
Break down imgtestlib module and  use log sections

achilleas-k:
This PR splits out the commits from #2359 that do not change the test pipelines.

## Changes

- Break down `imgtestlib` into multiple files.  The single file was getting a bit too big and harder to navigate.
- Move the core of some test scripts into the library.  Scripts like `build-image` and `boot-image` are now thin wrappers around functions in `imgtestlib` that handle command line arguments.
- Use foldable log sections when running in Gitlab CI.  This should make test logs easier to read.


achilleas-k:
The cross-arch test keeps build x86_64 images.  Hang on...

brlane-rht (dismissed):
Looks really good!

lzap (dismissed):
Very nice, the context is quite important because unclosed sections mess up the output quite a bit. I would also recommend to make a unique ID because if someone logs `Test section` twice it is very messed up. I think you could use some internal Python thing maybe stackframe or something to calculate some hash or something.

achilleas-k:
> I would also recommend to make a unique ID

It creates a UUID for each section / context.

lzap:
Ah now I see it, my implementation was bash based, this has a context! Clever. Nice.

achilleas-k:
RHEL 10.3 beta repos are experiencing the same issue we say when we first added the `network-installer` tests (https://github.com/osbuild/images/pull/2044#issue-3667039388).
Disabling 10.3 as well now.  Will check back on them later: https://redhat.atlassian.net/browse/HMS-10750.

I'll also open a follow-up PR to enable testing on 10.1 and 10.2 since those should be working normally now.

supakeen (approved):
Looks better.

lzap (approved):

PR #2384, opened 2026-05-29
distro/eln: clean up part 1 (HMS-10764)

supakeen:
Take a broom to the new ELN distro to try and form a good base for a future RHEL-11. Does a lot of deduplication, renames some image types (that hopefully get dropped in the future) and takes away all RHEL version conditions.

It might even fix a bug or two in our RHEL definitions.

---

Related: https://github.com/fedora-eln/eln/issues/530


supakeen:
We can land this if we agree. There will be one follow-up PR to this where I look at actual package sets and changes.

lzap (dismissed):
Nice. For the rename, that could go with aliases.

supakeen:
> Nice. For the rename, that could go with aliases.

Since it's a new distro I'd prefer to start fresh without aliases.

achilleas-k (changes requested):
LGTM in general, but I'd like to revert the last commit and do this separately.  I like the idea of renaming image types and moving towards more consistent naming, but I'm not sure if we want to go towards naming them by format (raw, qcow2, vhd) or target (ami/ec2, guest-image, azure).

supakeen:
Dropped the last commit, created an issue to track the rename: https://redhat.atlassian.net/browse/HMS-10769

achilleas-k (approved):
Thank you!

lzap (approved):

PR #2392, opened 2026-06-01
many: initial `systemd-boot` support

supakeen:
This PR implements the new/changes stages for `bootctl.install.root` and `fix-bls`. It then introduces a new bootloader `systemd`.

When/if `systemd-boot` is selected as the bootloader in a platform we automatically determine the boot-root in the way the BLS dictates us to do. We don't install `systemd-boot` inside the `os` pipeline as it does checks on the paths it is given to determine if they are actual mountpoints and/or filesystems. We pass the ESP/XBOOTLDR explicitly but they could also be auto-detected if we want that.

This works in my testing for YAML definitions that have only an ESP, or both an ESP and XBOOTLDR.

There's probably plenty we can improve here such as moving the find*Mountpoint functions into `pkg/disk` and generalizing them; however I'd like to do that as part of more sweeping change to `pkg/disk` where I'll also cover auto-mountpoints, auto-types, based on the recently added policies.

Let's not have a boot test yet until I've used this for a bit and am sure it's stable and doesn't need to change. There are likely functionalities missing that I might like still and haven't thought about yet.

### DIFF
--- a/pkg/manifest/raw.go
+++ b/pkg/manifest/raw.go
@@ -6,6 +6,7 @@ import (
 	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/artifact"
 	"github.com/osbuild/images/pkg/osbuild"
+	"github.com/osbuild/images/pkg/platform"
 )
 
 // A RawImage represents a raw image file which can be booted in a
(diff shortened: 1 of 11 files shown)


pkg/osbuild/bootctl_install.go
```diff
@@ -0,0 +1,22 @@
...
+	RandomSeed         string `json:"random-seed,omitempty"`
+	MakeEntryDirectory string `json:"make-entry-directory,omitempty"`
+	EntryToken         string `json:"entry-token,omitempty"`
+}
+
+func (BootctlInstallStageOptions) isStageOptions() {}
+
+func NewBootctlInstallStage(opts *BootctlInstallStageOptions, devices map[string]Device, mounts []Mount) *Stage {
+	return &Stage{
+		Type:    "org.osbuild.bootctl.install",
```
brlane-rht: You named the stage `org.osbuild.bootctl.install.root` :)

pkg/manifest/raw.go
```diff
@@ -113,6 +114,33 @@ func (p *RawImage) serialize() (osbuild.Pipeline, error) {
...
+
+		// ESP is required
+		espMountpoint, err := findESPMountpoint(p.treePipeline.PartitionTable)
+		if err != nil {
+			return osbuild.Pipeline{}, err
+		}
+		opts.ESPPath = espMountpoint
+
+		// XBOOTLDR is optional
+		xbootldrMountpoint, err := findXBootLDRMountpoint(p.treePipeline.PartitionTable)
```
brlane-rht: This could all be simplified into assigning `opts.BootPath` instead of using the intermediate `xbootldrMountpoint` checking and setting empty strings since it returns that on error anyway.

brlane-rht (changes requested):
Looks pretty good to me, one minor comment and one error I assume is because you changed your mind about naming the stage :)

pkg/osbuild/bootctl_install.go
supakeen: Wow, nice catch. I had the old name in my `/usr/lib/osbuild/stages` still! I've renamed the file, functions, structs, and this string and force-pushed.

pkg/manifest/raw.go
supakeen: Addressed this as well, added a comment for the explanation too.

brlane-rht (dismissed):
Looks good!

supakeen:

brlane-rht (approved):
Ack!

lzap (approved):

PR #2394, opened 2026-06-02
Add Custom menus to X86 and PPC64 ISO bootloaders

brlane-rht:
Adds support for setting custom menus, with tests. Will be used by later commits.

Related: HMS-10627

### DIFF
--- a/pkg/image/iso_bootloaders.go
+++ b/pkg/image/iso_bootloaders.go
@@ -9,6 +9,10 @@ import (
 type ISOBootloaders struct {
 	InstallerCustomizations *manifest.InstallerCustomizations
 	ISOCustomizations       *manifest.ISOCustomizations
+
+	// Grub2 menu customization for x86, ppc64, uefi iso menus, not supported by syslinux
+	// Setting these turns off all of the default menus
+	Custom []manifest.ISOGrub2MenuEntry
 }
 
 func (img *ISOBootloaders) Bootloaders(buildPipeline manifest.Build, platform platform.Platform, kernelOpts []string) []manifest.ISOBootloader {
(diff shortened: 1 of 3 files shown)


pkg/image/iso_bootloaders.go
```diff
@@ -53,13 +75,22 @@ func (img *ISOBootloaders) Bootloaders(buildPipeline manifest.Build, platform pl
...
-		bootloaders = append(bootloaders, bootTreePipeline)
+		efiPipeline := manifest.NewEFIBootTree(buildPipeline, img.InstallerCustomizations.Product, img.InstallerCustomizations.OSVersion)
+		efiPipeline.Platform = platform
+		efiPipeline.UEFIVendor = platform.GetUEFIVendor()
+		efiPipeline.ISOLabel = img.ISOCustomizations.Label
+		efiPipeline.DefaultMenu = img.InstallerCustomizations.DefaultMenu
+		efiPipeline.KernelOpts = kernelOpts
+
+		for _, entry := range img.Custom {
+			efiPipeline.MenuEntries = append(efiPipeline.MenuEntries, manifest.ISOGrub2MenuEntry{
```
supakeen, line 86: Since we already have the same concept in the EFI pipeline generator; is there a reason to call them `MenuEntries` there and `Custom` here. Could we maybe use `MenuEntries` in both of them (and `manifest/iso_bootloaders.go`)?

pkg/image/iso_bootloaders_test.go
```diff
@@ -36,20 +36,31 @@ func TestBootType(t *testing.T) {
 
 	type testCase struct {
 		bootType manifest.ISOBootType
+		custom   []manifest.ISOGrub2MenuEntry
 		expected []results
 	}
 
+	var noCustomMenus []manifest.ISOGrub2MenuEntry
```
supakeen, line 43: I think we don't need the `var` here.

supakeen (commented):
Thanks, naming question but functionally cool.

pkg/image/iso_bootloaders.go
brlane-rht: I called the new ones Custom because that matches the osbuild stage names. I think the old efi naming of MenuEntries is the odd duck (it even copies them into grubOptions.Custom).

pkg/image/iso_bootloaders_test.go
brlane-rht: In practice it doesn't matter, but I thought it made it more clear that it was an uninitialized list. I can change it to `noCustomMenus := []manifest.ISOGrub2MenuEntry{}` if you think that's better.

supakeen: Personally I like it more; but your explanation also makes sense so feel free to leave as is 🙂

pkg/image/iso_bootloaders.go
supakeen: Ack, let's leave as-is then and we can address it at some point to either make it all `Custom` or all `MenuEntries`; perhaps when we generalize to `systemd-boot` as well 🙂

achilleas-k (approved):

supakeen (approved):

PR #2393, opened 2026-06-01
imgtestlib: support bootc-foundry image types

lzap:
It looks like recent refactoring broke the boot tests for bootc-foundry images. This commit fixes the issue by updating the image type detection logic to handle the new format.

---

Solves:

```
$ test/boot.sh
Booting image from: /home/fedora/~/builds/redhat/services/products/image-builder/ci/bootc-foundry/build/bootc_hummingbird_20251124-aarch64-qcow2-bootc_foundry
Build config: /home/fedora/~/builds/redhat/services/products/image-builder/ci/bootc-foundry/build/bootc_hummingbird_20251124-aarch64-qcow2-bootc_foundry/config.json
Booting image
00:00
Traceback (most recent call last):
  File "/home/fedora/~/builds/redhat/services/products/image-builder/ci/bootc-foundry/_images/./test/scripts/boot-image", line 25, in <module>
    main()
    ~~~~^^
  File "/home/fedora/~/builds/redhat/services/products/image-builder/ci/bootc-foundry/_images/./test/scripts/boot-image", line 21, in main
    testlib.boot_image(search_path, build_config_path, keep)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.14/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/home/fedora/~/builds/redhat/services/products/image-builder/ci/bootc-foundry/_images/test/scripts/imgtestlib/boot.py", line 527, in boot_image
    image_path = find_image_file(search_path)
  File "/home/fedora/~/builds/redhat/services/products/image-builder/ci/bootc-foundry/_images/test/scripts/imgtestlib/core.py", line 373, in find_image_file
    files = os.listdir(os.path.join(build_path, last_pipeline))
FileNotFoundError: [Errno 2] No such file or directory: '/home/fedora/~/builds/redhat/services/products/image-builder/ci/bootc-foundry/build/bootc_hummingbird_20251124-aarch64-qcow2-bootc_foundry/gce'
```

---

I do not understand why we do this for bootc manifests, so unsure if this is a proper solution to the problem.

PR #2395, opened 2026-06-04
many: all the compression pipelines, all the time (HMS-10789)

supakeen:
Bit out of left field but it's been discussed a few times and I have a use case for it now with my efforts to simplify the image type list we have. This is by no means a final form as there's a bunch of follow-up effort and implementation effort to go.

However, I'm opening this PR to see what *breaks*. I've tested this on `image-builder` and it works fine there. I don't know if anything else isn't using the export interface.

This decouples the *first* part of the sextuple of information that is encoded inside a single image type string.


supakeen:
Tests are failing, I have an amended commit locally but I'd like `Schutzbot` to build the images first.

---

Ok; this requires a fix for the 'assume last pipeline is exported' similar to #2393. I'm going to rebase on top of that for now and drop the failing testcase from it.

brlane-rht:
I like where this is going. What's the plan for other settings like compression level?

supakeen:
> I like where this is going. What's the plan for other settings like compression level?

Mmm, I haven't thought about that yet. For now it's only going to lead to selectable compression in `image-builder` so we can rid ourselves a few image types and have slightly prettier names for others. They'll keep their default compression but users can override.

Perhaps another buildoption could be added to select specific compression options later on 🙂.

---

This PR will be slightly annoying since it changes all disk manifests, lots of conflicts all the time on checksums so I'm not going to keep rebasing it for now.

PR #2398, opened 2026-06-05
distro/eln: installers use erofs (HMS-10634)

supakeen:
Switch the ELN installers (`image-installer`, and `network-installer`) to use an erofs root filesystem.


achilleas-k (approved):

brlane-rht:
Looks like whatever mirror this is picking for eln content is terrible. Hits 1G of download in 3 minutes then slows way down until it times out.

brlane-rht (approved):
Ack!

supakeen:
> Looks like whatever mirror this is picking for eln content is terrible. Hits 1G of download in 3 minutes then slows way down until it times out.

Yea it seems a bit sucky/slow and to keep picking the same mirror.

PR #2404, opened 2026-06-08
oracle: Add support for Oracle Linux 9 and 10

lrochfort:
This PR adds new distro entries for Oracle Linux 9 and 10.

Support for the following targets has been successfully tested internally using image-builder:

- qcow2
- tarball
- bootc
- Oracle Cloud (OCI)
- Amazon AWS
- MS Azure
- Google GCP


supakeen:
@lrochfort very nice! I'll take a deeper look during the week.

pkg/manifest/os.go
```diff
@@ -783,6 +784,14 @@ func (p *OS) serialize() (osbuild.Pipeline, error) {
...
 	}
 
+	if p.OSCustomizations.Hostonly {
+		pipeline.AddStage(osbuild.NewDracutStage(&osbuild.DracutStageOptions{
+			Kernel:         []string{p.kernelVer},
+			EarlyMicrocode: true,
+			Reproducible:   true,
+		}))
+	}
+
```
supakeen: This should move into the partition table case below. It should then also be merged with the FIPS case there. Perhaps something like (untested):

```
dracutOptions := &osbuild.DracutStageOptions{
  Kernel: []string{p.kernelVer}
}

if p.OSCustomizations.Hostonly {
  dracutOptions.EarlyMicrocode = true
  dracutOptions.Reproducible = true
}

if p.OSCustomizations.FIPS {
  kernelOptions = append(kernelOptions, osbuild.GenFIPSKernelOptions(p.PartitionTable)...)
  dracutOptions.AddModules = []string{"fips"}
}

// Only regenerate the initramfs if it should be hostonly or we're FIPS enabled
// options have been set previously if that's the case
if p.OSCustomizations.Hostonly || p.OSCustomizations.FIPS {
  pipeline.AddStage(osbuild.NewDracutStage(dracutOptions)
}
```

Otherwise we'll regenerate the initramfs twice and the last one will win (FIPS in this case).

Alternatively if FIPS and hostonly are incompatible we can error out if they're both set (I don't think they're incompatible?).

pkg/distro/image_config.go
```diff
@@ -98,6 +98,7 @@ type ImageConfig struct {
 
 	Files       []*fsnode.File
 	Directories []*fsnode.Directory
+	Hostonly    *bool `yaml:"hostonly,omitempty"`
```
supakeen, line 101: So this is all implemented correctly from the image config down to the pipeline (though I have some remarks over there about merging it with the FIPS bits) however: none of the image types are actually using this new host only option thus it's dead code at the moment. Did you forget to enable it in the image definitions in this commit?

pkg/distro/host.go
```diff
@@ -62,6 +62,14 @@ func GetHostDistroName() (string, error) {
 		}
 	}
 
+	if osrelease["ID"] == "ol" {
+		versionParts := strings.Split(osrelease["VERSION_ID"], ".")
+		if len(versionParts) < 2 {
```
supakeen, line 70: Since this only indexes into `versionParts[0]` do we want the check here to be `< 1` instead? I assume we still want `< 2` to validate that there the `os-release` file is valid; but a comment explaining that would probably be a good idea.

pkg/runner/ol.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+func (o *Ol) GetBuildPackages() []string {
+	packages := []string{
+		"glibc",           // ldconfig
+		"platform-python", // osbuild
+		"python3",
+	}
+	return packages
+}
```
supakeen, line 20: This will require adding the runner to `osbuild` as well; probably symlinking `org.osbuild.ol9` -> `org.osbuild.centos9` is enough there. 

You will also want to update `pkg/distro/bootc/shared.go` with a `case "ol"`.

data/distrodefs/rhel.yaml
```diff
@@ -135,6 +140,43 @@ distros:
...
+    product: "Oracle Linux"
+    os_version: "10"
+    module_platform_id: "platform:el10"
+    # Vendor is used in bootloader UEFI setup
+    # OL retains the redhat component in the EFI path
+    # Would be inherited from &rhel10, but state explicitly for clarity
+    vendor: "redhat"
+    iso_label_tmpl: "OracleLinux-{{.Distro.MajorVersion}}-{{.Distro.MinorVersion}}-{{.Arch}}-dvd"
+    runner:
+      name: org.osbuild.ol
```
supakeen: Use `org.osbuild.ol10`.

data/distrodefs/rhel.yaml
```diff
@@ -272,6 +319,43 @@ distros:
...
+    product: "Oracle Linux"
+    os_version: 9
+    module_platform_id: "platform:el9"
+    # Vendor is used in bootloader UEFI setup
+    # OL retains the redhat component in the EFI path
+    # Would be inherited from &rhel9, but state explicitly for clarity
+    vendor: "redhat"
+    iso_label_tmpl: "OracleLinux-{{.Distro.MajorVersion}}-{{.Distro.MinorVersion}}-{{.Arch}}-dvd"
+    runner:
+      name: org.osbuild.ol
```
supakeen: Use `org.osbuild.ol9`.

pkg/customizations/oscap/oscap.go
```diff
@@ -42,6 +42,8 @@ const (
 	defaultCentos8Datastream  string = "/usr/share/xml/scap/ssg/content/ssg-centos8-ds.xml"
```
supakeen: Wait for @kingsleyzissou to confirm this but I *think* you can drop this commit; the `Default.*Datastream` functions here are dead code that was apparently left over as a distraction to any reader. The values now come from the image definitions; where you have already set them for the distribution.

supakeen (changes requested):
I've done a first pass over the Go-code bits that were adjusted; let's address those first (I think most things are relatively minor, and some questions) before I dive into the YAML.

pkg/customizations/oscap/oscap.go
kingsleyzissou: Ya it looks like it's dead code, the default datastreams now come from the `imageConfig` image definitions as @supakeen suggested

pkg/manifest/os.go
lrochfort: Thank you.  AFAICT FIPS and hostonly are compatible.  Testing an image on OCI with your suggestion  booted successfully.

I updated my fork.

pkg/distro/image_config.go
lrochfort: Yes, I did forget to add it.  I added it in my fork.

pkg/distro/host.go
lrochfort: Yes, I wanted to check the os-release is valid. I don't think we're likely to change it, but having the check seems a good idea.

I added a comment in my branch.

pkg/runner/ol.go
lrochfort: Good catch.  We added a symlink in our osbuild RPM, but I forgot to do that for upstream osbuild.

I'll submit a PR.

I added a new case and associated test in my fork.

data/distrodefs/rhel.yaml
lrochfort: Done

data/distrodefs/rhel.yaml
lrochfort: Done

pkg/customizations/oscap/oscap.go
lrochfort: Thank you. I dropped the commit in my fork.

lrochfort:
@supakeen, thank you for reviewing. I've updated my fork and pushed.  Could you please take a look when you have time.

supakeen:
> @supakeen, thank you for reviewing. I've updated my fork and pushed. Could you please take a look when you have time.

Thanks I'll take another look and focus on the YAML this time (and the other bits that were addressed).

pkg/runner/ol.go
supakeen: Thanks, I'll leave this comment open until we have the `osbuild` PR.

data/distrodefs/rhel-10/imagetypes.yaml
```diff
@@ -956,21 +956,33 @@ image_types:
...
+                exclude:
+                  - "redhat-release-eula"
+            "include redhat-release-eula exclude dracut-config-rescue when not Oracle Linux":
+              when:
+                not_distro_name: "ol"
+              append:
+                include:
+                  - "redhat-release-eula"
+                exclude:
+                  - "dracut-config-rescue"
```
lzap, line 985: This pattern repeats quite a bit, I wonder if we could implement a new loader condition `DistroNameIn` that takes a list of strings. The `not_distro_name: "ol"` could be replaced with `distro_name_in: ["rhel", "fedora", "centos"]`. I am assuming the list is finite and we know if @supakeen ? The idea is this would scale better.

supakeen: I concur but I'd like to do a full review of the YAML so @lrochfort can address everything in one batch.

pkg/runner/ol.go
lrochfort: osbuild PR submitted:
https://github.com/osbuild/osbuild/pull/2476

data/distrodefs/rhel-9/imagetypes.yaml
```diff
@@ -1567,6 +1632,21 @@ image_types:
...
                   - "subscription-manager-cockpit"
+            "exclude rhn packages on Oracle Linux":
+              when:
+                distro_name: "ol"
+              append:
+                exclude:
+                 - "rhn-check"
+                 - "rhn-setup"
+                 - "rhnsd"
+                 - "dnf-plugin-spacewalk"
```
supakeen: The linter is complaining that this needs to be indented one more space:

```
data/distrodefs/rhel-9/imagetypes.yaml
  1640:18   error    wrong indentation: expected 18 but found 17  (indentation
```

data/distrodefs/rhel-10/imagetypes.yaml
```diff
@@ -1147,6 +1180,7 @@ image_types:
           config:
             add_drivers:
               - "mlx5_core"
+      hostonly: true
```
supakeen: While it's probably correct that we want this everywhere this needs to be gated behind an 'if distro `ol`' initially. Just to make this PR *only* touch Oracle Linux image types so it's easier for us to verify that nothing else is affected.

Basically we want a clean manifest diff that doesn't touch *any* image types that aren't oracle linux. I'll provide a diff in my review that you can apply to fix this.

data/distrodefs/rhel-9/imagetypes.yaml
```diff
@@ -1444,20 +1484,35 @@ image_types:
             - "python3-jsonschema"
             - "qemu-guest-agent"
             - "redhat-release"
-            - "redhat-release-eula"
             - "rsync"
             - "tar"
             - "tcpdump"
             - "tuned"
           exclude:
             - "dracut-config-rescue"
```
supakeen: Let's drop the exclude here since it's in the condition branches below. Having it in exclude by default then including it in oracle linux leads to dependency solving issues as it's explicitly included *and* excluded.

data/distrodefs/rhel-9/imagetypes.yaml
```diff
@@ -1642,6 +1722,7 @@ image_types:
           config:
             add_drivers:
               - "mlx5_core"
+      hostonly: true
```
supakeen: Same as RHEL-10 let's conditionalize on Oracle Linux for now.

supakeen (changes requested):
Ok, I've gone through the YAML. Most important to me is that this change doesn't affect any other image types. Currently it does affect things and OL10 doesn't build due to a package being both included and excluded at the same time. I've attached a patch that addresses the comments I've made.                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                          
For me *personally* I don't care about a bit of the duplication here in the conditions, we are planning/plotting to address this in the future but that burden shouldn't particularly fall on @lrochfort.                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                          
See attached patch, please apply with `git am < ...` and then squash it into your own commits that changes the relevant image definitions before you force-push.

[0001-review-fix-up-Oracle-Linux-definitions.patch](https://github.com/user-attachments/files/28870422/0001-review-fix-up-Oracle-Linux-definitions.patch)

---

After applying that patch everything dependency solves correctly *and* at least Oracle Linux 9's QCOW2 image builds successfully on my system. 

Oracle Linux 10 does *not* because it seems like we're getting `sha1` hashes for the packages which we do not accept; I'll look into that today to see where those come from (from OL repositories, or if it's internal to us).

lrochfort:
> Oracle Linux 10 does _not_ because it seems like we're getting `sha1` hashes for the packages which we do not accept; I'll look into that today to see where those come from (from OL repositories, or if it's internal to us).

Re the hashes, we work around this on our side by running image-builder with --use-librepo=False.  If you have any suggestion as how to resolve it better, then we'd certainly appreciate it.

lrochfort:
> See attached patch, please apply with `git am < ...` and then squash it into your own commits that changes the relevant image definitions before you force-push.
> 
> [0001-review-fix-up-Oracle-Linux-definitions.patch](https://github.com/user-attachments/files/28870422/0001-review-fix-up-Oracle-Linux-definitions.patch)

Ack, thank you @supakeen 

Applied and pushed to my fork.

supakeen:
> > Oracle Linux 10 does _not_ because it seems like we're getting `sha1` hashes for the packages which we do not accept; I'll look into that today to see where those come from (from OL repositories, or if it's internal to us).
> 
> Re the hashes, we work around this on our side by running image-builder with --use-librepo=False. If you have any suggestion as how to resolve it better, then we'd certainly appreciate it.

I looked into this and it seems that OL10 [repository metadata](yum.oracle.com/repo/OracleLinux/OL10/baseos/latest/x86_64/repodata/b289db45730216911e1c580c69f291c2ef8b5e4b-primary.xml.gz) contains SHA1 checksums while OL9 repodata contains SHA256 checksums. Fixing the EL10 repository metadata to use SHA256 checksums would make everything work.

Using `--librepo=false` indeed works but isn't very nice. It'd be better to address this in the repository meta data for OL10. I guess other tooling might also want it.

Otherwise we *could* accept SHA1 checksums in the `librepo` source if we want to work around it but I don't know if we want to do that.

pkg/runner/ol.go
supakeen: Merged, so I'll resolve this conversation.

data/distrodefs/rhel-9/imagetypes.yaml
```diff
@@ -1488,8 +1488,6 @@ image_types:
             - "tar"
             - "tcpdump"
             - "tuned"
-          exclude:
-            - "dracut-config-rescue"
```
supakeen, line 1453: This hunk needs to go into the previous commit (`3b9570264a8a8b9ac286fd15039f99fd43b4e820`).

supakeen:
One last fix to the definitions which should make CI go green afterwards (if I got everything right).

That leaves only the repository checksums to resolve.

data/distrodefs/rhel-9/imagetypes.yaml
lrochfort: Updated and pushed

lrochfort:
@supakeen, the manifest checksum test is failing for my second commit:

```
panic: yaml: unmarshal errors:
	  line 1199: field hostonly not found in type distro.ImageConfig with distro rhel-10.0
```
	  
I'm not sure the best way to handle that, given we wanted to restrict hostonly to Oracle Linux for now.

supakeen:
> @supakeen, the manifest checksum test is failing for my second commit:
> 
> ```
> panic: yaml: unmarshal errors:
> 	  line 1199: field hostonly not found in type distro.ImageConfig with distro rhel-10.0
> ```
> 
> I'm not sure the best way to handle that, given we wanted to restrict hostonly to Oracle Linux for now.

If you reorder the commits and put [32aa54a](https://github.com/osbuild/images/pull/2404/commits/32aa54ac8ba25220633327a808b044491b33b9b1) *after* [65deabb](https://github.com/osbuild/images/pull/2404/commits/65deabb8bc16db194fbad867b09b2d712efca058) things should work.

Be sure to regenerate the checksums at each commit and squash them into the commit if anything changed 🙂

supakeen (approved):
Looks like that resolved everything and no checksums changed. Now the only question is if we want to merge this *before* or *after* the `osbuild` release that contains the SHA1 allowance for librepo.

And if we wait until after; if we want to bump the dependency.

Otherwise this looks good to me!!

lrochfort:
> Looks like that resolved everything and no checksums changed. Now the only question is if we want to merge this _before_ or _after_ the `osbuild` release that contains the SHA1 allowance for librepo.
> 
> And if we wait until after; if we want to bump the dependency.
> 
> Otherwise this looks good to me!!

Entirely up to you. From our side, it would be nice to have the dependency bumped so we can sync both our packages with upstream and inherit the changes

supakeen:
> > Looks like that resolved everything and no checksums changed. Now the only question is if we want to merge this _before_ or _after_ the `osbuild` release that contains the SHA1 allowance for librepo.
> > And if we wait until after; if we want to bump the dependency.
> > Otherwise this looks good to me!!
> 
> Entirely up to you. From our side, it would be nice to have the dependency bumped so we can sync both our packages with upstream and inherit the changes

Let's do that then; there should be a release of `osbuild` later today which includes the `ol9` runner and the `sha1` librepo fixup. I'll push a commit to this PR myself to bump the dependencies.

achilleas-k (approved):
Nice work, thank you!
LGTM

achilleas-k:
Flake-failure on the merge queue.  Sorry about that.  Re-queueing.

PR #2406, opened 2026-06-09
imgtestlib: handle missing sources key in tests

avitova:
Prevent a KeyError crash when accessing org.osbuild.curl for missing sources.

This is where it fails. https://gitlab.com/redhat/services/products/image-builder/ci/bootc-foundry/-/jobs/14762798107#L1511

That's just because we are getting `KeyError: 'org.osbuild.curl'` for x86_64.

PR #2408, opened 2026-06-09
imgtestlib: handle manifests without org.osbuild.curl sources

lzap:
The can_boot_test() function assumed network-installer manifests always have org.osbuild.curl sources containing ssh-server packages. However, bootc-based images don't use curl sources and have a different manifest structure.

https://github.com/osbuild/bootc-foundry/pull/96

### DIFF
--- a/test/scripts/imgtestlib/core.py
+++ b/test/scripts/imgtestlib/core.py
@@ -428,6 +428,9 @@ def can_boot_test(manifest_fname, manifest_data, image_type, arch, distro, bluep
                   "FATAL: FIPS integrity test failed")
             return False
         # Note that this needs adjustment when we switch to librepo
+        # Handle manifests that don't have org.osbuild.curl sources (e.g., bootc images)
+        if "org.osbuild.curl" not in manifest_data.get("sources", {}):
+            return True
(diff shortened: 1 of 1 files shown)


test/scripts/imgtestlib/core.py
```diff
@@ -428,6 +428,9 @@ def can_boot_test(manifest_fname, manifest_data, image_type, arch, distro, bluep
                   "FATAL: FIPS integrity test failed")
             return False
         # Note that this needs adjustment when we switch to librepo
+        # Handle manifests that don't have org.osbuild.curl sources (e.g., bootc images)
+        if "org.osbuild.curl" not in manifest_data.get("sources", {}):
+            return True
```
avitova, line 433: Shouldn't we return False? https://github.com/osbuild/images/pull/2406/changes

thozza:
maybe let's use https://github.com/osbuild/images/pull/2335/changes/da4ea14a73609f3a1eabcf7a7496b8e5f89abfdd

test/scripts/imgtestlib/core.py
lzap: Bootc is actually bootable without any curl source. Data come in a different way (container repo ref).

lzap:
Since there is prior art, I am closing in favour to https://github.com/osbuild/images/pull/2406

PR #2414, opened 2026-06-11
bootc generic-iso using ostree rootfs

brlane-rht:
This changes the bootc generic-iso so that the rootfs is the same as the bootc PXE image type.

~Note that this currently doesn't actually work right. Something is wrong with how dracut or ostree is setting up the booted filesystem, it doesn't mount things correctly so on boot a number of services fail:~
FIXED -- ends up dmsquash-live treats `LiveOS/rootfs.img` differently, switched to using `LiveOS/squashfs.img` and now it boots the same as bootc PXE tar.

This also renames it to `generic-iso`, and it drops the support for embedding a container payload since this isn't an installer iso and is not a raw filesystem (I think the user can do the same thing when setting up their bootc container, but have not explored that).


supakeen:
Will need some careful review; at least the name change will immediately break things for users and the non-embedding of containers (while I agree this can probably be done inside the bootable container itself and it's likely better there) is the same.

>  FIXED -- ends up dmsquash-live treats LiveOS/rootfs.img differently, switched to using LiveOS/squashfs.img and now it boots the same as bootc PXE tar.

In what way? I'm interested what the behavioral difference is.

I'm also tagging in @ondrejbudai since he's done a bunch with the ISOs for `bootc` too 🙂.

brlane-rht:
> Will need some careful review; at least the name change will immediately break things for users and the non-embedding of containers

Not sure what to do about that.

> > FIXED -- ends up dmsquash-live treats LiveOS/rootfs.img differently, switched to using LiveOS/squashfs.img and now it boots the same as bootc PXE tar.
> 
> In what way? I'm interested what the behavioral difference is.

When it is named squashfs.img it uses overlayfs and when it is rootfs.img it uses device-mapper:

https://github.com/dracut-ng/dracut/blob/main/modules.d/70dmsquash-live/dmsquash-live-root.sh#L330

is the section. With device-mapper the ostree root setup doesn't properly mount things and then a whole mess of services don't start (it actually still boots, which is surprising).

brlane-rht:
I've fixed some issues I was having with fstab mounts (have to pass fstab=no to the kernel/systemd), and added a README that is read from the file storage and written to the ISO, similar to the one in the PXE tar.

I've also experimented with running `podman pull ...` in the Containerfile and my simple tests work inside the booted system so you can still embed other containers.

brlane-rht:
Not sure what's up with the manifest check. Running `./tools/gen-manifest-checksums.sh` works fine locally (and I've updated the checksums).

lzap:
> Not sure what's up with the manifest check. Running `./tools/gen-manifest-checksums.sh` works fine locally (and I've updated the checksums).

There are some compilation errors, thus manifests are not generated correctly?!

```
Error: pkg/image/container_based_iso.go:106:56: cannot use osTreePipeline (variable of type *"github.com/osbuild/image-builder/pkg/manifest".OSFromContainer) as "github.com/osbuild/image-builder/pkg/manifest".TreePipeline value in argument to manifest.NewISOTree: *"github.com/osbuild/image-builder/pkg/manifest".OSFromContainer does not implement "github.com/osbuild/image-builder/pkg/manifest".TreePipeline (missing method Platform)
Error: pkg/image/container_based_iso.go:106:72: cannot use bootTreePipeline (variable of type *"github.com/osbuild/image-builder/pkg/manifest".EFIBootTree) as []"github.com/osbuild/image-builder/pkg/manifest".ISOBootloader value in argument to manifest.NewISOTree
Error: pkg/image/container_based_iso.go:111:18: isoTreePipeline.RootfsCompression undefined (type *"github.com/osbuild/image-builder/pkg/manifest".ISOTree has no field or method RootfsCompression)
Error: pkg/image/container_based_iso.go:115:18: isoTreePipeline.KernelOpts undefined (type *"github.com/osbuild/image-builder/pkg/manifest".ISOTree has no field or method KernelOpts
```

I think I know what is happening: unit and build tests do pass because they only test the tip of the branch. However, manifest generation is performed on every single commit individually. One of your commits does not compile correctly.

brlane-rht:
~Ah, right, I'd forgotten it tried on every commit. Problem is I don't want to make every commit compile, they're easier to read as distinct changes that then converge.~
Ends up compile failures are skipped, this was failing because the checksums needed to be updated :)

data/files/iso/generic/README
supakeen, line 1: It would be lovely to instead merge this into `doc/20-advanced/20-bootc/10-isos.md`.

data/distrodefs/bootc-generic/imagetypes.yaml
```diff
@@ -180,13 +188,17 @@ image_types:
   # image. Additionally, the image type is not tied to anaconda as
   # bootc-installer is. This allows users to create a very customized ISOs.
   # TODO: Discuss names of container-based isos: anaconda-iso,
-  # bootc-generic-iso and bootc-installer are all confusing.
-  bootc-generic-iso:
+  # generic-iso and bootc-installer are all confusing.
+  generic-iso:
```
supakeen: Could we keep an alias `bootc-generic-iso`?

data/files/iso/generic/README
```diff
@@ -0,0 +1,149 @@
...
+# Override using composefs for ostree (it is incompatible with the erofs rootfs)
+RUN cat <<EOF > /usr/lib/ostree/prepare-root.conf
+[composefs]
+enabled = no
+[sysroot]
+readonly = true
+EOF
+
+# Include the dmsquash-live module in the initramfs
+RUN cat <<EOF > /usr/lib/dracut/dracut.conf.d/40-pxe.conf
```
supakeen: Let's call it `40-iso.conf`.

supakeen (commented):
Some minor bits; currently building this PR and trying it out in a few different scenarios.

supakeen:
Tried some bits mostly seems to work OK aside from `bootc` itself being really unhappy about it all.

Poking @jbtrystram / @joelcapitao to take a look as well (RPMs are available in this PR, so is the docs) as they initially requested this feature.

Would also still really like @ondrejbudai's input on this since it changes the contract for ISOs a bit and perhaps we need to go onto composefs if that's at all possible on ISOs (followup).

doc/20-advanced/20-bootc/10-isos.md
```diff
@@ -1,132 +1,149 @@
...
+RUN cat <<EOF > /usr/lib/ostree/prepare-root.conf
+[composefs]
+enabled = no
+[sysroot]
+readonly = true
+EOF
+
+# Include the dmsquash-live module in the initramfs
+RUN cat <<EOF > /usr/lib/dracut/dracut.conf.d/40-iso.conf
+compress="xz"
```
travier, line 43: We now default to zstd. Why would we use `xz` here?

doc/20-advanced/20-bootc/10-isos.md
```diff
@@ -1,132 +1,149 @@
...
-    plymouth \
-    default-fonts-core-sans \
-    default-fonts-other-sans \
-    google-noto-sans-cjk-fonts
+```
+FROM quay.io/fedora/fedora-bootc:latest
+RUN dnf -y install grub2-efi-*-cdboot xorriso isomd5sum dracut-live erofs-utils shim && dnf clean all
+RUN mkdir /boot/efi && cp -r /usr/lib/efi/shim/*/EFI /boot/efi && cp -r /usr/lib/efi/grub2/*/EFI/* /boot/efi/EFI/
+
+# Override using composefs for ostree (it is incompatible with the erofs rootfs)
```
travier, line 33: That sounds weird. Not sure why that would not work. I think we do that with the CoreOS LiveISO.

brlane-rht: It seems to be related to how dmsquash-live sets up the root, it does that first, then the ostree module takes over. Without this it fails to mount.

doc/20-advanced/20-bootc/10-isos.md
brlane-rht: Because that's what the boot.iso uses.

supakeen: Yea the ISOs still tend to use xz/lzma for their initramfs compression as they mostly care about size and (usually) only get built once, same reason most of the disk images in Koji get xz compressed and not zstd; only the size matters since they're only built once.

We *could* switch that but I think that would be a separate thing to consider, it would make some form of sense for containers when iterating on an ISO 🙂

supakeen:
Just to make sure for any watchers; this is still ongoing with a lot of chats and such around the approach and CoreOS testing with integrating this :)

pkg/manifest/iso_bootloaders.go
```diff
@@ -81,9 +83,10 @@ func (boot *Grub2X86Boot) GetISOBootStages(inputName string, _ *disk.PartitionTa
 	stages := make([]*osbuild.Stage, 0)
 
 	var grub2config *osbuild.Grub2Config
-	if boot.DefaultMenu > 0 {
+	if boot.DefaultMenu > 0 || boot.MenuTimeout > 0 {
```
avitova, line 86: I think this might end up accepting values like `DefaultMenu = 1` and `MenuTimeout = -1`, right? Should we make this a bit more strict?

pkg/manifest/iso_bootloaders.go
```diff
@@ -160,9 +165,10 @@ func (boot *Grub2PPC64Boot) GetISOBootStages(inputName string, _ *disk.Partition
...
```
avitova, line 168: Same concern here.

avitova (commented):
I have one concern, might not be that likely to happen, but I'd rather split these two checks so that we do not assign a negative value to either of these.

pkg/manifest/iso_tree.go
```diff
@@ -17,103 +18,47 @@ type ISOTree struct {
...
-	KernelOpts []string
-
-	// Kernel and initramfs paths in the tree pipeline
+	// Kernel, initramfs, and rootfs paths in the tree pipeline
 	KernelPath    string
 	InitramfsPath string
+	RootfsPath    string
+
+	// Optionally set the ostree= argument (substitute the boot path for @OSTREE@ in grub.cfg)
+	SetOSTREE bool
```
achilleas-k: Nitpick: Weird capitalisation here.  In other places we do `OSTree`, so I'd expect this to be `SetOSTree`.

achilleas-k (dismissed):
Small nitpick but wont block.  Anna's concerns are also valid, so I'd like to see that addressed here or in a separate PR.

LGTM

achilleas-k:
My main concern here, even though I approved, is that these images aren't being tested right now.  I trust the work here is all good and that it's probably gone through some general manual testing, but I'd feel a lot better about it if it came after getting build and boot tests for these images.

pkg/manifest/iso_bootloaders.go
brlane-rht: We could, but I don't think it gains much. The goal of this block is to decide if it needs to use a `Grub2Config` or leave it nil, not to check the values. We generally don't do much in the way of sanity checking since the osbuild stage metadata is really what decides whether or not a value is allowed, and duplicating that information here just makes it harder to keep things in sync.

avitova: Okay, I understand the reasoning. I guess as long as this does not end up with a panic (which it hopefully should not), we are good?

brlane-rht: If anything fails it will return an error from osbuild when it processes the manifest.

jbtrystram:
I quickly tried to build the coreOS liveiso with this but this fail with : 
`blueprint validation failed for image type "iso": customizations.ignition: not supported`, so I'll have to fix that first then try again.

removing the ignition blueprint I get : 
```
image-builder build iso           --bootc-build-ref $BUILDER           --bootc-ref $TARGET_FCOS_IMAGE           --output-dir fedora-coreos           --output-name fedora-coreos-rawhide           --with-buildlog           --with-manifest           --with-metrics           --blueprint /srv/blueprints/generated/iso-x86_64.toml
Using disk.yaml root filesystem (xfs)
Manifest generation step
Building manifest for bootc-based-iso
error: error depsolving package sets for "build": DNF error occurred: InvalidRequest: No 'repos' or 'root_dir' specified
```

supakeen:
> I quickly tried to build the coreOS liveiso with this but this fail with : `blueprint validation failed for image type "iso": customizations.ignition: not supported`, so I'll have to fix that first then try again.
> 
> removing the ignition blueprint I get :
> 
> ```
> image-builder build iso           --bootc-build-ref $BUILDER           --bootc-ref $TARGET_FCOS_IMAGE           --output-dir fedora-coreos           --output-name fedora-coreos-rawhide           --with-buildlog           --with-manifest           --with-metrics           --blueprint /srv/blueprints/generated/iso-x86_64.toml
> Using disk.yaml root filesystem (xfs)
> Manifest generation step
> Building manifest for bootc-based-iso
> error: error depsolving package sets for "build": DNF error occurred: InvalidRequest: No 'repos' or 'root_dir' specified
> ```

Isn't the image type `generic-iso`; not `iso`?

jbtrystram:
> Isn't the image type `generic-iso`; not `iso`?

Ah yes, sorry :D 
`image-builder` list did not contain any hints unfortunately.
I'll try again

brlane-rht:
~Also, you have to run this as `bootc-image-builder`, you can just symlink that to the `image-builder` binary. This changes its behavior to match bib.~
nevermind, I was thinking of something else.

supakeen:
> Also, you have to run this as `bootc-image-builder`, you can just symlink that to the `image-builder` binary. This changes its behavior to match bib.

Erm, that sounds funky? We tell people to migrate to `image-builder` (and CoreOS has) why does this have to run as `bootc-image-builder`?

achilleas-k:
> Also, you have to run this as `bootc-image-builder`, you can just symlink that to the `image-builder` binary. This changes its behavior to match bib.

This _shouldn't_ matter any more.  Running `image-builder build --bootc-ref ...` should be doing the same thing as `bootc-image-builder build ...`.  If not (and I'm aware there are some small differences right now), we should be working towards making it the same to lower the migration friction as much as possible.

brlane-rht:
Rebased and re-tested locally. Only major change is dropping my MenuTimeout patch and using pointer based MenuTimeout added by @supakeen

pkg/manifest/iso_tree_test.go
```diff
@@ -0,0 +1,116 @@
...
+	// Check for the kernel/initrd/rootfs from the ostree deployment
+	assert.Contains(t, fromPaths, "input://tree/vmlinuz")
+	assert.Contains(t, fromPaths, "input://tree/initrd.img")
+	assert.Contains(t, fromPaths, "input://tree/rootfs.img")
+	assert.Contains(t, fromPaths, "input://root-tree/EFI")
+
+	// Check for final paths for kernel, initrd, rootfs
+	assert.Contains(t, toPaths, "tree:///images/pxeboot/vmlinuz")
+	assert.Contains(t, toPaths, "tree:///images/pxeboot/initrd.img")
+	assert.Contains(t, toPaths, "tree:///LiveOS/rootfs.img")
```
achilleas-k: Pairing up the expected from+to paths to not just test that they exist but that they're also correctly matched would be a minimal change here and would catch silly mistakes (like from copy-pasting during a refactoring) that this current test wouldn't catch.

achilleas-k (commented):
This all looks good to me.  Tiny note below.

We should have some kind of boot test for this.  In #2635 I'm struggling a bit to get the pxe boot tests working and seeing that we're switching the generic ISO pipelines to match that, I'm a bit worried that something might be off.  And if it isn't now, it might be later given how much we rely on bootc and how quickly it changes.

pkg/manifest/iso_tree_test.go
brlane-rht: Fixed in a new commit.

supakeen:
I should have time next week (hopefully) to dive in. I'm particularly interested in how this is going to work when bootc wants to use its composefs backend(s).

brlane-rht:
Fixed a couple issues:
* The X86 iso was only UEFI not BIOS+UEFI
* The X86 BIOS grub.cfg wasn't replacing the ostree= path

Caught thanks to new tests added by @supakeen for aarch64 changes.

PR #2415, opened 2026-06-12
Test: stop testing EOL RHEL releases

thozza:
RHEL 10.1 and 9.7 are not EUS and are no longer supported. Let's reduce the test matrix by not testing them.


achilleas-k (approved):
TY!

croissanne (approved):

PR #2439, opened 2026-06-18
many: release machinery (HMS-10850)

supakeen:
Updates the first bits of the release machinery; specfile gets an updated version, makefiles are merged. Various bits and bobs updated to name the right things. Some things removed.

After this we can change see what happens when we push a manually tagged release on `main`.


achilleas-k:
If all the Makefile targets are merged into one, let's also remove `Makefile.image-builder`

Makefile
```diff
@@ -1,10 +1,58 @@
...
+#
+BUILDDIR ?= .
+SRCDIR ?= .
+
+RST2MAN ?= rst2man
+
+# see https://hub.docker.com/r/docker/golangci-lint/tags
+# v1.55 to get golang 1.21 (1.21.3)
+# v1.53 to get golang 1.20 (1.20.5)
+GOLANGCI_LINT_VERSION=v2.3.0
```
brlane-rht, line 25: osbuild-composer is using v2.9 may as well update that here as well (I've been using it locally and it's happy with things).

supakeen: Probably follow-up; trying to get this in ASAP so we can start releasing from the repository first 🙂.

supakeen:
Ok; I *think* with these changes in `main` we should be able to tag a release and see if a downstream release appears.

supakeen:
Needs: https://github.com/osbuild/release-action/pull/47

achilleas-k (approved):
🎉

avitova (approved):

PR #2440, opened 2026-06-18
Update package name and import paths to image-builder [HMS-10853]

achilleas-k:
Update the main package name (in go.mod) to image-builder (from images).

Because of the bidirectional import with blueprints, we need two workarounds here:
1. A new package, called repomigration, that copies functions from blueprint so they can be used temporarily until the blueprint package is updated with new import paths.  This avoids the issue where a blueprint function requires a type from "images" and we try to pass the same type from the new project "image-builder".
2. Renaming the C function `gnu_ext_crypt()` to `gnu_ext_crypt_migration()`.  The function name causes issues with the same function being built by the blueprint dependency as it imports its images dependency which defines the same function.


supakeen (dismissed):

supakeen:
RPM builds failing is expected here.

avitova (dismissed):

lzap (dismissed):
Ate all my RAM until I gave up and pulled the PR locally.

schuellerf (dismissed):

achilleas-k:
Added an extra commit that disables the `s3_touch()` call on cache hits.

Updating timestamps on large files takes too long.  This slows down the filtering for the dynamic pipeline creation.
I think this will greatly reduce runner killing issues because the `generate-build-config` step will be much quicker.

brlane-rht (approved):
Ack!

ochosi (approved):

avitova:
:partying_face:

PR #2441, opened 2026-06-18
test/imgtestlib: don't update file timestamps on cache hit

achilleas-k:
Updating timestamps on large files takes too long.  This slows down the filtering for the dynamic pipeline creation.  We have some workarounds planned but, for now, let's speed things up.

Note that when osbuild uses an image in its testing, it will still get updated, so the original problem we were solving with this operation (that images were garbage collected by the retention policy while osbuild was still using them) will likely still not occur.


achilleas-k:
Closing and pushing the extra commit to #2440

PR #2442, opened 2026-06-18
test: use the runner distro for the builder when checking the cache

achilleas-k:
Build config generators download the cache, generate manifests, and check the cache for the same combination of osbuild version, runner distro, architecture, and manifest ID.  If the runner distro for the build config generator differs from the one for the build itself, then we never get cache hits.

Currently, RHEL 7.9 config generation runs on Fedora 44, which made the cache checking function use the `runner-fedora-44` as a path component, whereas the cached image and info was in `runner-centos-9`.

This commit fixes the problem by making the cache check use the builder runner distro.

### DIFF
--- a/test/scripts/imgtestlib/core.py
+++ b/test/scripts/imgtestlib/core.py
@@ -10,7 +10,8 @@ from .build import get_manifest_id
 from .cache import dl_build_info, gen_build_info_dir_path_prefix, touch_s3
 from .gitlab import log_section
 from .run import runcmd
-from .testenv import get_bib_ref, host_container_arch, rng_seed_env
+from .testenv import (get_bib_ref, get_ci_runner_for, host_container_arch,
+                      rng_seed_env)
 
 TEST_CACHE_ROOT = ".cache/osbuild-images"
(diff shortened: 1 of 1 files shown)


achilleas-k:
Gonna need #2440 to be merged first and I don't to pile on even more commits to that.

brlane-rht (dismissed):
Ack!

supakeen (dismissed):

achilleas-k:
Hang on.  I should have tested this better.  Something is clearly wrong here.
```
Build info not found: .cache/osbuild-images/s3configs/builds/osbuild-ref-f103e77c4595dc565635f20589862eca4a5f0a0b/runner-aws/fedora-44/fedora-44/aarch64/manifest-id-0f93d55ac9ccbb2427e16bd3e1fd167827ff045f57c0eed4d879776482da774a/info.json
```

yeah, `runner-aws`, no kidding.

achilleas-k:
oh, of course.  runners are defined as `<cloud>/<distro>`.

supakeen (approved):

avitova (approved):

PR #2443, opened 2026-06-18
image-builder fixups: use non-root Makefile target

avitova:
The .packit.yaml uses the make show-version and release_artifacts which are not part of the root Makefile, and therefore tests fail.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -17,9 +17,9 @@ srpm_build_deps:
 
 actions:
   # v1 just the default prior to having tags
-  get-current-version: bash -c "make show-version"
+  get-current-version: bash -c "make -f Makefile.image-builder show-version"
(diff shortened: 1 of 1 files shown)


avitova:
Make it greeeeen

achilleas-k:
Rebase'll fix it.

#2439 will need to switch it back (cc @supakeen), but it seems that one might need more work.

achilleas-k (approved):

avitova:
I did not notice that the PR changes makefiles too. We can as well juts wait in that case:)  depends on how much time this will be

achilleas-k:
> I did not notice that the PR changes makefiles too. We can as well juts wait in that case:)

Your call.  Or Simon's, idk.

supakeen:
Let's do this in #2439.

PR #2447, opened 2026-06-19
imgtestlib: handle missing sources key in tests

avitova:
Prevent a KeyError crash when accessing org.osbuild.curl for missing sources.

This commit also reconciles dropped changes in imgtestlib.

During the time when the imgtestlib split PR was open [1], some changes were merged and the PR had to be rebased. During the rebase, I (AK) missed some of the changes that were already merged, essentially reverting them.

I went through all functions that were in the monolithic imgtestlib and made sure to bring in all changes that were dropped and adapted them to newer changes.

[1] https://github.com/osbuild/images/pull/2383

### DIFF
--- a/test/scripts/imgtestlib/core.py
+++ b/test/scripts/imgtestlib/core.py
@@ -39,6 +39,7 @@ CAN_BOOT_TEST = {
         "image-installer", "minimal-installer", "network-installer",
         "qcow2", "generic-qcow2", "cloud-qcow2",
         "wsl", "generic-wsl",
+        "bootc-generic-iso",
     ]
 }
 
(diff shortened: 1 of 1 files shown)


achilleas-k (dismissed):
Thanks for cleaning up the mess :)

lzap (dismissed):

achilleas-k:
Ugh, sorry for breaking your PR (again).  It seems I missed something in a recent change to the install-dependencies script.
https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/14975989981#L2902

RHEL 7 builds on CentOS 9 machines, and boot testing the qcow2, which we're now doing in this PR with the removal of the `jq` check, requires cloud-localds, which isn't being installed on the runner.

I'll re-add the `jq` check, even though it's not needed, so we can get this fixed and everything else unblocked.

achilleas-k (approved):

lzap (approved):
Thank you for looking into this.

PR #2454, opened 2026-06-22
github: drop the reverse dependency check with image-builder-cli

achilleas-k:
We are image-builder.


achilleas-k:
This action runs on `pull_request_target`, so it uses the configuration from `main` and the tests will still show up in this PR until it's merged.

supakeen (approved):

brlane-rht (approved):
Ack

PR #2456, opened 2026-06-23
many: plumb sdboot options

supakeen:
We introduced additional options for systemd-boot configuration but never actually plumbed them through into the configuration or definitions.

Let's do that now so they can be set.

### DIFF
--- a/pkg/distro/defs/loader_test.go
+++ b/pkg/distro/defs/loader_test.go
@@ -702,6 +702,39 @@ image_types:
 	})
 }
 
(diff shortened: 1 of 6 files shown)


brlane-rht:
I think it needs tests :)

supakeen:
> I think it needs tests :)

I've force-pushed a YAML definition test but since we don't have any actual image types that use this (at least, internally) I don't think I can add more than that, unless we want to add a boot test. The stage options were already tested JSON-wise.

brlane-rht (approved):
Looks good!

achilleas-k (approved):

PR #2457, opened 2026-06-23
osbuild: ddi mount

supakeen:
`osbuild` gained the ability to mount Discoverable Disk Images through `systemd-dissect` [1]. Let's do the necessary plumbing here to land that.

There are no users of this codepath yet, but that also makes it nice and small to review.

The required `osbuild` version is already set correctly; DDI mounts were introduced in `osbuild` 183.

[1]: https://github.com/osbuild/osbuild/pull/2459

### DIFF
--- a/pkg/osbuild/ddi_mount.go
+++ b/pkg/osbuild/ddi_mount.go
@@ -1,0 +1,21 @@
+package osbuild
+
+type DDIMountOptions struct {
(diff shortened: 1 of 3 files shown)


brlane-rht (approved):
Looks good

lzap (approved):

PR #2458, opened 2026-06-23
many: pipeline mounts

supakeen:
Provide machinery to set default mounts and devices on pipelines. Any stage added through `.AddStage(s)` will be given these options.

This mostly saves (currently) in cases where we need to pass mounts *every* time.

Perhaps we want to be stricter?

---

In the *future* I might want to use this in the `os` pipeline; or a new pipeline, to invert the order of operation so that we *first* create the image and *then* write to it; similar to what we currently do for the `bootc` pipeline(s) but that's for later, it does give some background.

### DIFF
--- a/pkg/osbuild/osbuild.go
+++ b/pkg/osbuild/osbuild.go
@@ -31,6 +31,8 @@ type Pipeline struct {
 	// Sequence of stages that produce the filesystem tree, which is the
 	// payload of the produced image.
 	Stages []*Stage `json:"stages,omitempty"`
+
+	mounts []Mount
 }
 
 // SetBuild sets the pipeline and runner for generating the build environment
@@ -39,10 +41,26 @@ func (p *Pipeline) SetBuild(build string) {
 	p.Build = build
 }
 
+// SetMounts sets pipeline-level mounts that are appended to every stage
+// added via AddStage or AddStages.
+func (p *Pipeline) SetMounts(mounts ...Mount) {
+	p.mounts = append(p.mounts, mounts...)
+}
+
 // AddStage appends a stage to the list of stages of a pipeline. The stages
 // will be executed in the order they are appended.
 // If the argument is nil, it is not added.
 func (p *Pipeline) AddStage(stage *Stage) {
+	if stage != nil {
+		stage.Mounts = append(stage.Mounts, p.mounts...)
+		p.Stages = append(p.Stages, stage)
+	}
+}
+
+// AddStageDirect appends a stage without merging pipeline-level mounts or
+// devices. Use this for stages that intentionally bypass the pipeline
+// mounts (e.g. stages that write outside the deployment).
+func (p *Pipeline) AddStageDirect(stage *Stage) {
 	if stage != nil {
 		p.Stages = append(p.Stages, stage)
 	}
--- a/pkg/osbuild/osbuild.go
+++ b/pkg/osbuild/osbuild.go
@@ -32,7 +32,8 @@ type Pipeline struct {
 	// payload of the produced image.
 	Stages []*Stage `json:"stages,omitempty"`
 
-	mounts []Mount
+	mounts  []Mount
+	devices map[string]Device
 }
 
 // SetBuild sets the pipeline and runner for generating the build environment
@@ -47,12 +48,29 @@ func (p *Pipeline) SetMounts(mounts ...Mount) {
 	p.mounts = append(p.mounts, mounts...)
 }
 
+// SetDevices sets pipeline-level devices that are merged into every stage
+// added via AddStage or AddStages.
+func (p *Pipeline) SetDevices(devices map[string]Device) {
+	p.devices = devices
+}
+
 // AddStage appends a stage to the list of stages of a pipeline. The stages
 // will be executed in the order they are appended.
 // If the argument is nil, it is not added.
 func (p *Pipeline) AddStage(stage *Stage) {
 	if stage != nil {
 		stage.Mounts = append(stage.Mounts, p.mounts...)
+		if len(p.devices) > 0 {
+			if stage.Devices == nil {
+				stage.Devices = make(map[string]Device, len(p.devices))
+			}
+			for k, v := range p.devices {
+				if _, exists := stage.Devices[k]; exists {
+					panic(fmt.Sprintf("stage %q already defines device %q which conflicts with pipeline device", stage.Type, k))
+				}
+				stage.Devices[k] = v
+			}
+		}
 		p.Stages = append(p.Stages, stage)
 	}
 }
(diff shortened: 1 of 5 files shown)


achilleas-k:
> to invert the order of operation so that we _first_ create the image and _then_ write to it

Not the topic here, but I'm not fully onboard with this just yet.  The current order has some benefits, like exporting the OS pipeline result for inspection.  And there's also the potential future benefit of reproducibility with `mkfs.ext4 -d`, which I'd love to explore at some point.

pkg/osbuild/osbuild.go
```diff
@@ -31,6 +31,8 @@ type Pipeline struct {
 	// Sequence of stages that produce the filesystem tree, which is the
 	// payload of the produced image.
 	Stages []*Stage `json:"stages,omitempty"`
+
+	defaultMounts []Mount
```
achilleas-k: Given the description of the behaviour in the commit message, I think the term `default` doesn't work here.  Perhaps `base` or just `mounts`?  I was thinking `piplineMounts`, but since they're a field of the pipeline struct, just `pipeline.mounts` will do.

My thinking is that the term `defaultMounts` could imply that defining any other set of mounts on a stage overrides them, instead of appending to them.  But that's not the case.  They're global pipeline mounts, for all stages, unconditionally.

pkg/osbuild/osbuild.go
```diff
@@ -32,7 +32,8 @@ type Pipeline struct {
 	// payload of the produced image.
 	Stages []*Stage `json:"stages,omitempty"`
 
-	defaultMounts []Mount
+	defaultMounts  []Mount
+	defaultDevices map[string]Device
```
achilleas-k: Same comment about naming.

pkg/osbuild/osbuild.go
```diff
@@ -47,12 +48,28 @@ func (p *Pipeline) SetDefaultMounts(mounts ...Mount) {
...
 	if stage != nil {
 		stage.Mounts = append(stage.Mounts, p.defaultMounts...)
+		if len(p.defaultDevices) > 0 {
+			if stage.Devices == nil {
+				stage.Devices = make(map[string]Device, len(p.defaultDevices))
+			}
+			for k, v := range p.defaultDevices {
+				if _, exists := stage.Devices[k]; !exists {
+					stage.Devices[k] = v
+				}
```
achilleas-k: I'm not sure this is the correct behaviour here.  I think if a stage defines a device with the same name as the global pipeline devices, it might make sense for that to be an error.  If we really want the ability to override the device configuration for a stage by adding a device with the same name, we should document it in the docstrings, both here and in the method where the pipeline devices are defined.

achilleas-k (commented):
Generally a good idea.  Some comments below about naming and behaviour details.
I initially missed that the way to opt out of adding stages without applying the defaults is to append to the pipeline's stages directly.  That's a bit messy.  Not a huge problem, we do similar things when we want to _prepend_ stages, but it might be nicer to make the Pipeline API more useful by having methods that do these things (appending stages, prepending stages, adding with or without defaults, etc) in a consistent way with sanity checking so that we don't have to be appending elements to slices directly.

I'd be curious to see how many stages need opting out and how much trouble/code this saves us in the end.  It might be interesting to consider alternatives, like a little closure at the top of the pipeline generator that does the same work, e.g.:
```go
func (p *SomePipeline) serialize() (osbuild.Pipeline, error) {

	...

	mounts := ...  // make pipeline mounts
	devices := ...  // make pipeline devices
	func addStageWithDefaults(stage *Stage) {
		stage.Mounts = mounts
		stage.Devices = devices
		pipeline.AddStage(stage)
	}

	...

	addStageWithDefaults(osbuild.NewWhateverStage(...))

}
```

or adding an extra method to the pipeline for adding stages with mounts and devices:
```go
func (p *SomePipeline) serialize() (osbuild.Pipeline, error) {

	...

	mounts := ...  // make pipeline mounts
	devices := ...  // make pipeline devices

	pipeline.AddStageWithMounts(osbuild.NewWhateverStage(...), mounts, devices)
}
```

pkg/osbuild/osbuild.go
supakeen: Just `mounts` / `devices` is probably the better name here.

pkg/osbuild/osbuild.go
supakeen: Let's make it error now. If we get a use case where we would need this we can cross the bridge then.

supakeen:
> Generally a good idea. Some comments below about naming and behaviour details. I initially missed that the way to opt out of adding stages without applying the defaults is to append to the pipeline's stages directly. That's a bit messy. Not a huge problem, we do similar things when we want to _prepend_ stages, but it might be nicer to make the Pipeline API more useful by having methods that do these things (appending stages, prepending stages, adding with or without defaults, etc) in a consistent way with sanity checking so that we don't have to be appending elements to slices directly.
> 
> I'd be curious to see how many stages need opting out and how much trouble/code this saves us in the end. It might be interesting to consider alternatives, like a little closure at the top of the pipeline generator that does the same work, e.g.:
> 
> ```go
> func (p *SomePipeline) serialize() (osbuild.Pipeline, error) {
> 
> 	...
> 
> 	mounts := ...  // make pipeline mounts
> 	devices := ...  // make pipeline devices
> 	func addStageWithDefaults(stage *Stage) {
> 		stage.Mounts = mounts
> 		stage.Devices = devices
> 		pipeline.AddStage(stage)
> 	}
> 
> 	...
> 
> 	addStageWithDefaults(osbuild.NewWhateverStage(...))
> 
> }
> ```
> 
> or adding an extra method to the pipeline for adding stages with mounts and devices:
> 
> ```go
> func (p *SomePipeline) serialize() (osbuild.Pipeline, error) {
> 
> 	...
> 
> 	mounts := ...  // make pipeline mounts
> 	devices := ...  // make pipeline devices
> 
> 	pipeline.AddStageWithMounts(osbuild.NewWhateverStage(...), mounts, devices)
> }
> ```

I think the little closure mostly makes sense because there currently aren't any other functions provided to add with/without mounts/devices. I was quite unsure about doing those because it's a bit opinionated and there was a good opt out possibility by appending to stages directly.

I'll do an `AddStageWithMountsDevices` instead and see how that looks 🙂, it should be the non-default behavior at least.

supakeen:
Now has `AddStageDirect` instead of flipping the behavior the default behavior is to append pipeline mounts/devices and when `*Direct` is used they're bypassed.

We also now panic on adding a stage with a device that's pre-existing. This avoids changing like 250 call sites.

Writing all of this however made me realize that it's /probably/ better to rewrite the entire stage adding mechanics to get rid of all the 'must be in the correct order' and to allow better control over everything so I'll close this and think about that.

PR #2459, opened 2026-06-23
disk: check all partitions for boot partition requirement

lzap:
Function needsBoot exited early if it found a plain root or boot partition or a btrfs root with a boot subvolume. This was not correct, we need to check all partitions to avoid:

panic: the device name "boot" has been generated for two different devices

Example customization that caused the panic:

        [[customizations.disk.partitions]]
        type = "plain"
        fs_type = "vfat"
        minsize = "1 GiB"
        mountpoint = "/boot/efi"

        [[customizations.disk.partitions]]
        name = "vg"
        type = "lvm"
        minsize = "1 GiB"

        [[customizations.disk.partitions.logical_volumes]]
        name = "lv_root"
        fs_type = "xfs"
        minsize = "20 GiB"
        mountpoint = "/"

        [[customizations.disk.partitions]]
        type = "plain"
        fs_type = "xfs"
        minsize = "2 GiB"
        mountpoint = "/boot"

Reordering the partitions would fix the problem, but this was a insights customer who were running into this.

---

https://redhat.atlassian.net/browse/HMS-10805

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -1697,46 +1697,58 @@ func needsBoot(disk *blueprint.DiskCustomization) bool {
 	}
 
 	var foundBtrfsOrLVM bool
+	var hasPlainRoot bool
+	var hasPlainBoot bool
+	var lvmOrBtrfsRootNeedsBoot bool
+	var hasBtrfsRootWithBootSubvol bool
+
+	// Do not use early returns in the loop below, we need to check all partitions.
 	for _, part := range disk.Partitions {
 		switch part.Type {
 		case "plain", "":
-			if part.Mountpoint == "/" {
-				return false
-			}
-			if part.Mountpoint == "/boot" {
-				return false
+			switch part.Mountpoint {
+			case "/":
+				hasPlainRoot = true
+			case "/boot":
+				hasPlainBoot = true
 			}
 		case "lvm":
 			foundBtrfsOrLVM = true
-			// check if any of the LVs is root
 			for _, lv := range part.LogicalVolumes {
 				if lv.Mountpoint == "/" {
-					return true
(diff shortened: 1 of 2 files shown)


supakeen (approved):

achilleas-k (approved):
Nice work.  LGTM.
That function is getting a bit complicated but the `disk` package has always been a place that requires some complex logic.

One small note: Does the function docstring cover everything still or should we expand it to more clearly describe all scenarios?

supakeen:
> Nice work. LGTM. That function is getting a bit complicated but the `disk` package has always been a place that requires some complex logic.
> 
> One small note: Does the function docstring cover everything still or should we expand it to more clearly describe all scenarios?

I think it's clear-ish; from the current docstring for the function it just says that it will find out if there's a boot thingy nothing about the ordering 🙂

supakeen:
Also, do we later re-order the partitions at all? Having `/boot` be the last partition on disk would get in the way of `cloud-init` / `systemd-repart` or hyperscaler-specific guest utilities automatically resizing the root partition.

achilleas-k:
> Also, do we later re-order the partitions at all? Having `/boot` be the last partition on disk would get in the way of `cloud-init` / `systemd-repart` or hyperscaler-specific guest utilities automatically resizing the root partition.

For disk customizations, we explicitly made it so the user defined order is respected and documented it as such: https://osbuild.org/docs/user-guide/partitioning/#general-principles and https://osbuild.org/docs/user-guide/partitioning/#order.

Yes, that would interfere with resize tooling, but when we were designing the customizations we were always thinking about it in terms of "user makes their own PT", as long as it's valid.

Breaking cloud-init and repart is a valid concern, but I also think there are valid use cases for letting users lay things out exactly the way they want them, for whatever reason they need.

PR #2461, opened 2026-06-24
many: allow `copr://` URLs in `--(force|extra-repo)`

supakeen:
A long time ago we had discussed making it easier to pass certain types of repositories into `image-builder`. A super common use case is to want to build with a package from a COPR repository.

This PR implements just that. When `--(force|extra)-repo copr://@osbuild/osbuild` is passed the COPR API is used to look up the corresponding chroot (based on the distro/arch combination) and a repoconfig is generated from it and used.

The implementation here is slow to the point of it being barely usable when a COPR URL is passed. This is because we look up COPR things for the distro*arch combination for all distros when newRegistryImpl is called. This could be addressed a few different ways and I'd like input.

1. Just accept it for now, merge the PR, address later now that we have a reason to do so (this has my preference).
2. Provide a cache to `parseRepoURLs` (I tested this; it's way faster).
3. Move the resolving of COPR URLs later into the process (somewhere in reporegistry?) after we know the actual arch/distro we're targetting.
4. Don't have extraRepos be a part of that step at *all*, in that case things are only slow if force-repos is passed.

Which one has preference? The comment that is there seems to hint at the latter but it also feels kinda funky to deal with anything else but `RepoConf`.

All of this boils down to us needing the reporegistry to determine what we can build.

### DIFF
--- a/cmd/image-builder/copr.go
+++ b/cmd/image-builder/copr.go
@@ -1,0 +1,168 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/osbuild/image-builder/pkg/rpmmd"
+)
+
+var coprBaseURL = "https://copr.fedorainfracloud.org"
+
+type coprProject struct {
+	FullName    string
+	Owner       string
+	Name        string
+	ChrootRepos map[string]string
+	GPGKeyURL   string
+}
+
+type coprAPIResponse struct {
+	FullName    string            `json:"full_name"`
+	Ownername   string            `json:"ownername"`
+	Name        string            `json:"name"`
+	ChrootRepos map[string]string `json:"chroot_repos"`
+}
+
+var coprHTTPClient = http.DefaultClient
+
+// parseCoprURL extracts the owner and project from a parsed
+// copr:// URL. The expected form is copr://owner/project.
+func parseCoprURL(u *url.URL) (owner, project string, err error) {
+	// TODO: do we want to require hostnames or something? This
+	// TODO: all seems to be a bit iffy as we're working around the
+	// TODO: url.URL kinda?
+
+	// url.Parse("copr://@osbuild/osbuild") gives:
+	//   Opaque: "//@osbuild/osbuild"
+	// url.Parse("copr://user/project") gives:
+	//   Host: "user", Path: "/project"
+
+	// Rebuild the path from the raw form to handle both uniformly.
+	raw := strings.TrimPrefix(u.String(), "copr://")
+
+	parts := strings.SplitN(raw, "/", 2)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", "", fmt.Errorf("invalid copr URL %q, expected copr://owner/project (e.g. copr://@osbuild/osbuild)", u)
+	}
+
+	return parts[0], parts[1], nil
+}
+
+func fetchCoprProject(owner, project string) (*coprProject, error) {
+	apiURL := fmt.Sprintf("%s/api_3/project?ownername=%s&projectname=%s",
+		coprBaseURL,
+		url.QueryEscape(owner),
+		url.QueryEscape(project),
+	)
+
+	resp, err := coprHTTPClient.Get(apiURL)
+	if err != nil {
+		return nil, fmt.Errorf("cannot fetch copr project %s/%s: %w", owner, project, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, fmt.Errorf("copr project %s/%s not found", owner, project)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("copr API returned status %d for project %s/%s", resp.StatusCode, owner, project)
+	}
+
+	var apiResp coprAPIResponse
+	if err := json.NewDecoder(resp.Body).Decode(&apiResp); err != nil {
+		return nil, fmt.Errorf("cannot decode copr API response for %s/%s: %w", owner, project, err)
+	}
+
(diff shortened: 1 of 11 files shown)


supakeen:
Going to undraft this, we can accept the slowness but I'd be interested in the discussion around it.

cmd/image-builder/copr.go
```diff
@@ -0,0 +1,168 @@
...
+	parts := strings.SplitN(raw, "/", 2)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", "", fmt.Errorf("invalid copr URL %q, expected copr://owner/project (e.g. copr://@osbuild/osbuild)", u)
+	}
+
+	return parts[0], parts[1], nil
+}
+
+func fetchCoprProject(owner, project string) (*coprProject, error) {
+	apiURL := fmt.Sprintf("%s/api_3/project?ownername=%s&projectname=%s",
```
brlane-rht, line 56: Where is the API documented? Should be included in a comment somewhere in this file.

brlane-rht:
I'm a bit on the fence. Mostly because it's adding complexity. But running really slow I think is a show-stopper, at the least I think it needs to only be run once it has all the needed info.

But also, what does this really gain? Looking up copr repos is pretty easy already, the use probably already knows which arch/distro they want and can pull the baseurl from copr and pass it to `--extra-repo`.

supakeen:
> I'm a bit on the fence. Mostly because it's adding complexity. But running really slow I think is a show-stopper, at the least I think it needs to only be run once it has all the needed info.
> 
> But also, what does this really gain? Looking up copr repos is pretty easy already, the use probably already knows which arch/distro they want and can pull the baseurl from copr and pass it to `--extra-repo`.

Looking up COPR repo's isn't *that* easy; to give you a workflow: I want to build c10s and all Fedora's for aarch64 and x86_64 with some additional packages from several COPR repositories to test a new systemd/uboot/whatever.

If you try to go through those steps in your head of what you need to do:

1. Look up the COPR repo's on COPR.
2. Click on their repo files.
3. Copy/paste the URLs out of that repofile.

Then for each arch/distro combination fill in the values yourself. You spend a lot of time on it.

When doing a single COPR for a single distro/arch sure that's straightforward enough; but there's also no real reason not to make it *easier* to do that. Same as I want to make it easier to say 'build an image with this Bodhi update enabled' which is super useful during test/freeze periods in Fedora.

---


If it *wasn't* so slow, I think a user-friendly feature like this would probably be not controversial, same goes for the Bodhi-variant :)

I agree that with the current implementation it's not worth it because it slows down the way we do repositories so badly. But now that this PR is open, it does show a problem in how we do our repository registry and what effect it has so we'll have to discuss that first 🙁

cmd/image-builder/copr.go
supakeen: Will do!

lzap:
This grows the CLI a bit more beyond the UNIX philosophy: one command does one thing and I do not see a problem that this tries to solve. A simple shell script enables COPR and then passes the repository into the CLI. So this gets 👎🏻 from me and I really think we should be heading the opposite direction - removing unnecessary things from CLI into separate binaries like image uploads (yeah there is metadata and I really think it should be a sidecar file stored alongside the image).

supakeen:
> This grows the CLI a bit more beyond the UNIX philosophy: one command does one thing and I do not see a problem that this tries to solve. A simple shell script enables COPR and then passes the repository into the CLI. So this gets 👎🏻 from me and I really think we should be heading the opposite direction - removing unnecessary things from CLI into separate binaries like image uploads (yeah there is metadata and I really think it should be a sidecar file stored alongside the image).

This 'UNIX philosophy' doesn't apply here; if every one of our users has to go create their own shell scripts and figure out the COPR API, Bodhi API, etc that really doesn't scale well nor does it give a good experience to BYOSH just to enable a Bodhi update in the `build` command.

The upload command is something else entirely and I agree that it /could/ probably be split out. Though I don't know how much benefit that gives.

It'd be cool to be able to do `image-builder build --extra-repo copr://@supakeen/uboot --extra-repo bodhi://FEDORA-2026-00f1ec3c28 raw` vs `image-builder build --extra-repo $(~/bin/my-copr-resolver @supakeen/uboot fedora-45 x86_64) --extra-repo $(~/bin/my-bodhi-resolver FEDORA-2026-00f1ec3c28 fedora-45 x86_64)`

This isn't a rare or unusual case, it's something people do semi-regularly especially when working on kernel patches, hardware support, or testing pending updates, that sort of stuff 🙂

lzap:
Both COPR and Bodhi projects have command line tools available and I just feel like the mentioned features are a really great candidates to be added there, not in image builder. By the way, you just showed the power of UNIX/Linux right there, I love the latter solution of leveraging shell to do things!

supakeen:
Not stale.

cmd/image-builder/copr.go
achilleas-k, line 1: Doing all this work in `cmd/image-builder/` feels very wrong to me.  The module is already overloaded with functionality that doesn't belong in the `cmd/`, let's not add more.  This module should, eventually, only be concerned with parsing and resolving command line subcommands and options and handing off any logic to the internal modules.

Architecturally speaking, this kind of functionality is essentially a resolver.  But it's a particularly interesting resolver because its output feeds into another resolver, the depsolver.  I think for this we should either
- create a new resolver that takes a copr URL (with `distro` and `arch`) and resolves a repository config that gets added to the repos for the depsolver, or
- add it as a pre-processing step to the depsolver, so it first resolves the copr URL to a repo config and then does the depsolve.

achilleas-k (changes requested):

cmd/image-builder/copr.go
supakeen: Since in general there doesn't seem to be a lot of shared interest in this feature I think it's better if I just let it go instead of trying to get it right at the cost of a lot of work on all sides. Maybe a later RFE can build on top of the idea.

PR #2462, opened 2026-06-24
imgtestlib: use image-builder cmd, drop cmd/build/ (HMS-10298)

avitova:
I saw more possible approaches, such as splitting configs into two files: blueprint, and the options part. However, creating a temporary file for blueprint in /tmp sounds like a good option and keeps things clean. Let me know what you guys think.


test/scripts/imgtestlib/build.py
```diff
@@ -19,10 +19,16 @@ def build_image(distro, arch, image_type, config_path):
...
 
-    runcmd(["go", "build", "-o", "./bin/build", "./cmd/build"])
-
-    cmd = ["sudo", "-E", "./bin/build", "--output", "./build", "--checkpoints", "build",
-           "--distro", distro, "--arch", arch, "--type", image_type, "--config", config_path]
+    runcmd(["go", "build", "-o", "./bin/image-builder", "./cmd/image-builder"])
+
+    cmd = [
+        "sudo", "-E", "./bin/image-builder", "build", image_type,
+        "--config", config_path,
```
supakeen: Where does `--config` come from?

supakeen:
You'll probably want to bump the Schutzbot seed if changing the build script I think?

test/scripts/imgtestlib/build.py
avitova: Yeah, you are right, this was never supported. But tests passed locally previously; that is odd. I believe we ignore non-existent flags then?

cmd/image-builder/main.go
```diff
@@ -776,6 +786,8 @@ operating systems like Fedora, CentOS and RHEL with easy customizations support.
 	manifestCmd.Flags().String("bootc-installer-payload-ref", "", `bootc installer payload ref`)
 	manifestCmd.Flags().String("bootc-default-fs", "", `default filesystem to use for the bootc install (e.g. ext4)`)
 	manifestCmd.Flags().Bool("bootc-no-default-kernel-args", false, `don't use the default kernel arguments`)
+	manifestCmd.Flags().Bool("bootc-remote", false, `use org.osbuild.skopeo sources instead of containers-storage`)
```
achilleas-k: I think using the osbuild stage names here isn't very helpful.  We don't expect users of the CLI to know about osbuild internals so it's a bit of a layer violation, even though they appear in the output.
It'd be better to describe it in general terms, like:
"Pull bootc container from remote location instead of using it from local container storage"

The option could also use some workshopping.  Maybe `bootc-pull-container`?

achilleas-k:
While working on #2467, I saw this in the old bootc-image-builder github actions:
```yaml
      - name: Workaround podman issues in GH actions
        run: |
          # see https://github.com/osbuild/bootc-image-builder/issues/446
          sudo rm -rf /var/lib/containers/storage
          sudo mkdir -p /etc/containers
          echo -e "[storage]\ndriver = \"overlay\"\nrunroot = \"/run/containers/storage\"\ngraphroot = \"/var/lib/containers/storage\"" | sudo tee /etc/containers/storage.conf
```

Might be what's needed to get that cross-arch build to work.
Still doesn't explain why it started failing now though.

avitova:
I am just trying to make Shutzbot tests not fail. :face_holding_back_tears:

test/scripts/imgtestlib/build.py
```diff
@@ -1,10 +1,44 @@
 import json
 import os
-from typing import Dict
+import tempfile
+from typing import Dict, List
 
 from .gitlab import log_section
 from .run import runcmd, runcmd_nc
-from .testenv import get_host_distro, get_osbuild_commit, rng_seed_env
```
achilleas-k, line 7: Is the test rng seed ignored now?  That would invalidate the caching and cause constant rebuilds.  We have a `--seed` are for the build subcommand, so it should be added to build the exact, predictable manifest that will be checked for in later runs.

achilleas-k (changes requested):
Nice work.  Found one issue though.

test/scripts/imgtestlib/build.py
supakeen: We should verify that `--seed` is properly plumbed through, I think it is but I also seem to recall it might not be.

avitova: Oh, I see! Let me fix that!

supakeen (approved):
I would probably mark the new options as hidden but that's kinda minor I think.

achilleas-k (approved):
Excellent work.  Thank you!

LGTM

PR #2463, opened 2026-06-25
WIP: include oca in rhel9, 10 image definitions

src-up:
In the case of the cloud-agent, it would be better to have the package in all the images by default


supakeen:
You'll need to regenerate checksums by running `./tools/gen-manifest-checksums.sh` and squashing it into your commit.

We use that to ensure each commit only changes expected manifests (and the checksums are also used for caching in our CI so we know if we've already built and boot tested a given image) 🙂 

Also tagging in @achilleas-k because he's been working with the virt people on the OCI images if I recall correctly.

data/distrodefs/rhel-10/imagetypes.yaml
```diff
@@ -1248,6 +1248,10 @@ image_types:
                     label: "root"
                     mountpoint: "/"
                     fstab_options: "defaults"
+    package_sets:
```
supakeen, line 1252: Doing it like this overwrites the entire package set that come from the `*qcow2` anchor that this definition inherits from; you'd want to change it to:

```
package_sets:
  <<: *qcow2_package_sets
  guests:
    - include:
      - "oracle-cloud-agent"
```

And then in the `qcow2` `package_sets` definition create that anchor (line 1033 at the moment):

```
package_sets: &qcow2_package_sets
```

supakeen (commented):
Some adjustments to make things work.

Is `oracle-cloud-agent` shipped in RHEL/CentOS/Alma/Rocky/OL repositories? They all build from the RHEL definitions. If it's not available then all the changes still apply but it'd probably be easier to have a `conditions` on the qcow2 package sets that includes this *only* for those distributions that ship the package (and then work on shipping the package elsewhere).

data/distrodefs/rhel-10/imagetypes.yaml
src-up: Thanks.  This helps. I wasnt aware that other distros also use the same yaml.

supakeen (changes requested):
Ok, so it seems the main concern here is that `oracle-cloud-agent` is not available in RHEL Base/AppStream repositories. Have you already engaged with people to get that in the repositories?

Without it we can't merge this PR as it breaks builds during depsolving for RHEL for any image type that inherits from the `qcow2`.

For now I think we can either:

1. Include it only for distros where it's available. This seems to be only Oracle Linux at the moment.
2. Put a hold on this PR until the package is potentially available in RHEL (and maybe CentOS Stream, Alma Linux, Rocky Linux (and potentially Fedora?)).

achilleas-k:
> Ok, so it seems the main concern here is that `oracle-cloud-agent` is not available in RHEL Base/AppStream repositories.

We've been talking with Simran (and Lars) about the details of the image and the agent already.  I told him the agent should be part of the base image since we want it in all OCI images, not just marketplace ones.  But I didn't realise (or more like, I was absent minded and didn't make the connection) that the repository wouldn't be available outside of brew builds.  So it seems we can't have it as part of the base image unless it gets distributed in Red Hat repos, which I don't think it will be.

supakeen:
@achilleas-k / @src-up any progress on this? I was speaking with Fedora folks about this as well and it'd be nice if it could be packaged there as well 🙂 do I need to reach out to the Oracle Linux people for this or do we want to close this and only add the package in `brew` (feels weird).

src-up:
Hey @supakeen - yeah i think we need to do this in brew for now and we can reopen this later ? If its okay with you, can we close this ?

supakeen:
> Hey @supakeen - yeah i think we need to do this in brew for now and we can reopen this later ? If its okay with you, can we close this ?

Yep! I'll do it for you.

PR #2465, opened 2026-06-25
rhel-10: reintroduce minimal-raw-xz

supakeen:
Reintroduces the minimal image type to all EL-variants. This is a useful image type for hardware enablement and other quick testing things.

Previously I tried to reintroduce it in [1] but there's a bunch of missing packages. After some discussion with Peter Robinson we've decided to just drop `initial-setup` for the initial implementation and then go from there. We can tell users that they must pass a blueprint to set up users and/or root passwords.

Enables only the empty blueprint for testing to reduce the drag; I don't think it's super useful to test all customizations?

[1]: https://github.com/osbuild/image-builder/pull/2241


brlane-rht (approved):
Looks good!

lzap (approved):

PR #2469, opened 2026-06-26
workflows: Add a link to the osbuild-composer API unit test results

brlane-rht:
This will make it easier to find the results.


brlane-rht:
NOTE: This will not change the workflow until it is merged to main according to the excellent comments at the top of the file. So I don't know how to be 100% sure I got it right...

achilleas-k (approved):
Neat.  Thank you!

supakeen (approved):

PR #2473, opened 2026-06-29
cmd/image-builder: mock simple integration tests (HMS-10857)

supakeen:
Mock the integration tests that are easy and straightforward to mock. This reduces the test time significantly as we're not performing real dependency solving anymore in most of the tests.

---

Note this doesn't *really* make things faster that much. The test suite still takes long as almost every test calls `main.Run()` which loads distribution definitions which takes 2-3 seconds for *every* test.

We also can't make the tests parallel without rewriting the entire mocking strategy since we're mocking globals 🙁

### DIFF
--- a/cmd/image-builder/export_test.go
+++ b/cmd/image-builder/export_test.go
@@ -25,25 +25,25 @@ var (
 
 type DescribeImgYAML describeImgYAML
 
-func MockOsArgs(new []string) (restore func()) {
+func MockOsArgs(args []string) (restore func()) {
 	saved := os.Args
-	os.Args = append([]string{"argv0"}, new...)
+	os.Args = append([]string{"argv0"}, args...)
 	return func() {
 		os.Args = saved
 	}
 }
 
-func MockOsStdout(new io.Writer) (restore func()) {
+func MockOsStdout(w io.Writer) (restore func()) {
 	saved := osStdout
-	osStdout = new
+	osStdout = w
 	return func() {
 		osStdout = saved
 	}
 }
 
-func MockOsStderr(new io.Writer) (restore func()) {
+func MockOsStderr(w io.Writer) (restore func()) {
 	saved := osStderr
-	osStderr = new
+	osStderr = w
 	return func() {
 		osStderr = saved
 	}
@@ -102,10 +102,18 @@ func MockDirSize(f func(string) (int64, error)) (restore func()) {
 	}
 }
 
-func MockManifestgenDepsolver(new manifestgen.DepsolveFunc) (restore func()) {
+func MockManifestgenDepsolver(fn manifestgen.DepsolveFunc) (restore func()) {
 	saved := manifestgenDepsolver
-	manifestgenDepsolver = new
+	manifestgenDepsolver = fn
 	return func() {
 		manifestgenDepsolver = saved
+	}
+}
+
+func MockManifestgenContainerResolver(fn manifestgen.ContainerResolverFunc) (restore func()) {
+	saved := manifestgenContainerResolver
+	manifestgenContainerResolver = fn
+	return func() {
+		manifestgenContainerResolver = saved
 	}
 }
--- a/cmd/image-builder/main_test.go
+++ b/cmd/image-builder/main_test.go
@@ -17,6 +17,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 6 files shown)


achilleas-k (dismissed):
Very nice.

I'm sure we can get more improvements to the loading times once we're done with the merge and can focus on the loader.

supakeen:
It seems that with some (new?) tests we now have downstream failures in RPM builds or is this caused by this PR. I'll take a look in the morning.

supakeen:
Rebased to resolve conflicts.

achilleas-k (dismissed):

supakeen:
Something broke during the rebase; let's only ack when it actually goes green this time 😞

achilleas-k (dismissed):

cmd/image-builder/main_test.go
```diff
@@ -947,7 +953,7 @@ func TestBuildCrossArchSmoke(t *testing.T) {
 		pipelines, err := manifesttest.PipelineNamesFrom(manifest)
 		assert.NoError(t, err)
 		crossArchPipeline := "bootstrap-buildroot"
-		crossArchWarning := `WARNING: using experimental cross-architecture building to build "aarch64"`
+		crossArchWarning := fmt.Sprintf(`WARNING: using experimental cross-architecture building to build %q`, crossArch.String())
```
avitova: No reason to call `String()`, I have no problem with this other than it is inconsistent with other `fmt.Sprintf` where you do nto call `.String()`

cmd/image-builder/main_test.go
```diff
@@ -1053,10 +1058,36 @@ func fakeDepsolve(solver *depsolvednf.Solver, cacheDir string, depsolveWarningsO
...
+		containerSpecs[plName] = containers
+	}
+	return containerSpecs, nil
+}
+
+func sha256For(s string) string {
+	h := sha256.New()
+	h.Write([]byte(s))
+	return fmt.Sprintf("%x", h.Sum(nil))
+}
```
avitova: This is already defined in manifestgen_test.go. If we figure out a way to share this, maybe from some helpers.go file for tests, it would look cleaner, but not a blocker.

avitova (dismissed):

supakeen:
Had to rebase for conflicts.

achilleas-k (dismissed):

cmd/image-builder/export_test.go
```diff
@@ -109,3 +109,11 @@ func MockManifestgenDepsolver(new manifestgen.DepsolveFunc) (restore func()) {
 		manifestgenDepsolver = saved
 	}
 }
+
+func MockManifestgenContainerResolver(new manifestgen.ContainerResolverFunc) (restore func()) {
```
brlane-rht: We shouldn't be using `new` as a variable name. It's a built-in function, I'm surprised the linter didn't complain.

cmd/image-builder/upload_test.go
```diff
@@ -19,8 +19,25 @@ import (
 
 	main "github.com/osbuild/image-builder/cmd/image-builder"
 	"github.com/osbuild/image-builder/internal/testutil"
+	testrepos "github.com/osbuild/image-builder/test/data/repositories"
 )
 
+func skipIfNoAMISupport(t *testing.T) {
+	t.Helper()
+	switch arch.Current() {
+	case arch.ARCH_X86_64, arch.ARCH_AARCH64:
```
brlane-rht, line 28: Maybe throw a return in here? My `C` brain briefly wants to read this as a fall-through to default :)

brlane-rht (dismissed):
Looks good other than using `new`.

cmd/image-builder/main_test.go
supakeen: Moved into `testutil`.

cmd/image-builder/main_test.go
supakeen: Removed `.String()` call.

cmd/image-builder/export_test.go
supakeen: I was being consistent with the rest of `export_test.go`. I've renamed all of them which slightly increases the diff size.

cmd/image-builder/upload_test.go
supakeen: Done.

achilleas-k (approved):
TY

avitova (approved):

PR #2476, opened 2026-06-30
many: use `image-builder upload` in tests (HMS-10856)

supakeen:
Replace the upload parts of `boot-(aws|azure)` with `image-builder upload`. Had to add some options and move all the Azure upload code since it didn't exist yet.

### DIFF
--- a/cmd/boot-aws/main.go
+++ b/cmd/boot-aws/main.go
@@ -8,16 +8,12 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
-	"github.com/osbuild/image-builder/internal/common"
 	"github.com/osbuild/image-builder/internal/test"
-	"github.com/osbuild/image-builder/pkg/arch"
 	"github.com/osbuild/image-builder/pkg/cloud/awscloud"
-	"github.com/osbuild/image-builder/pkg/platform"
 )
 
 // exitCheck can be deferred from the top of command functions to exit with an
@@ -49,8 +45,6 @@ ssh_authorized_keys:
 // resources created or allocated for an instance that can be cleaned up when
 // tearing down.
 type resources struct {
-	AMI           *string `json:"ami,omitempty"`
-	Snapshot      *string `json:"snapshot,omitempty"`
 	SecurityGroup *string `json:"security-group,omitempty"`
 	InstanceID    *string `json:"instance,omitempty"`
 }
@@ -87,20 +81,7 @@ func newClientFromArgs(flags *pflag.FlagSet) (*awscloud.AWS, error) {
 	return awscloud.New(region, keyID, secretKey, sessionToken)
 }
 
-// getOptionalStringFlag returns the value of a string flag if it's set, or nil
-// if it's not set.
-func getOptionalStringFlag(flags *pflag.FlagSet, name string) (*string, error) {
-	value, err := flags.GetString(name)
-	if err != nil {
-		return nil, err
-	}
-	if value == "" {
-		return nil, nil
-	}
-	return &value, nil
-}
-
-func doSetup(a *awscloud.AWS, filename string, flags *pflag.FlagSet, res *resources) error {
+func doSetup(a *awscloud.AWS, flags *pflag.FlagSet, res *resources) error {
 	username, err := flags.GetString("username")
 	if err != nil {
 		return err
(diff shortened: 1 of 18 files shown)


supakeen:
This all *probably* won't work as it likely needs secrets we can't currently pass on the commandline of `image-builder`.

supakeen:
Ok; this needs to wire up `boot-mode` for AWS in the `image-builder upload` command to make it work for older RHELs which were BIOS-only.

test/scripts/imgtestlib/boot.py
```diff
@@ -381,8 +381,23 @@ def boot_qemu_pxe(arch, pxe_tar_path):
             http_server.wait()
 
 
+def get_boot_mode(distro, arch, image_type):
+    """Get the boot mode for an image type from image-builder describe."""
+    stdout, _ = runcmd(["go", "run", "./cmd/image-builder", "describe",
+                        image_type, "--distro", distro, "--arch", arch])
+    for line in stdout.splitlines():
+        if line.startswith("bootmode:"):
```
supakeen, line 389: ... sorry

supakeen:
I cannot see the Snyk issue but otherwise this is ready for review.

supakeen:
Ok; this is only failing on the dnf5 depsolver tests now; let's see if a rebase fixes that.

supakeen:
Snyk errors are not relevant, the idiom already existed.

cmd/image-builder/bib_main_test.go
```diff
@@ -184,13 +184,13 @@ func (fa *fakeAwsUploader) Check(status io.Writer) error {
...
 
-func (fa *fakeAwsUploader) UploadAndRegister(r io.Reader, size uint64, status io.Writer) error {
+func (fa *fakeAwsUploader) UploadAndRegister(r io.Reader, size uint64, status io.Writer) (*cloud.UploadResult, error) {
 	fa.uploadAndRegisterCalls++
 	_, err := io.Copy(&fa.uploadAndRegisterRead, r)
 	if err != nil {
 		panic(err)
 	}
-	return fa.uploadAndRegisterErr
+	return &cloud.UploadResult{Provider: "aws"}, fa.uploadAndRegisterErr
```
brlane-rht: Shouldn't the upload result be nil when `uploadAndRegisterErr` is non-nil in order to match the actual function's behavior?

cmd/image-builder/upload.go
```diff
@@ -286,6 +288,30 @@ func cmdUpload(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	format, err := cmd.Flags().GetString("format")
+	if err != nil {
+		return err
+	}
+	switch format {
+	case "", "yaml":
```
brlane-rht, line 382: Since this is getting printed I think it should look the same as other yaml/json output and be indented like you did with the system status output.

pkg/cloud/azure/uploader.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/osbuild/image-builder/pkg/cloud"
+)
+
+const uploaderStorageContainer = "images"
+
+var _ cloud.Uploader = &azureUploader{}
+
+type azureUploader struct {
+	client        *Client
+	subscription  string
```
brlane-rht: subscription is stored in Client, does it need to be in 2 places?

cmd/boot-aws/main.go
```diff
@@ -8,16 +8,12 @@ import (
 	"path/filepath"
```
brlane-rht, line 8: Commit message should include `image-builder upload --to=aws` to make it clear how to upload.

test/scripts/imgtestlib/boot.py
```diff
@@ -381,8 +381,24 @@ def boot_qemu_pxe(arch, pxe_tar_path):
...
 
 
+def get_boot_mode(distro, arch, image_type):
+    """Get the boot mode for an image type from image-builder describe."""
+    stdout, _ = runcmd(["go", "run", "./cmd/image-builder", "describe",
+                        image_type, "--distro", distro, "--arch", arch])
+    for line in stdout.decode().splitlines():
+        if line.startswith("bootmode:"):
+            mode = line.split(":", 1)[1].strip()
+            return {
```
brlane-rht: Maybe catch KeyError here? To make it easier to debug when a new bootmode is added and this isn't updated.

brlane-rht (commented):
Looks pretty good, just a few small questions.

test/scripts/imgtestlib/boot.py
supakeen: Sure it can have friendlier output.

pkg/cloud/azure/uploader.go
supakeen: No, it can be only on Client if I add a getter to it.

cmd/image-builder/upload.go
supakeen: I'll use `MarshalIndent` here as well.

supakeen:
> Looks pretty good, just a few small questions.

I've addressed all of them 🙂

brlane-rht (approved):
Looks good!

cmd/image-builder/upload.go
```diff
@@ -235,6 +242,58 @@ func uploaderForCmdIbmCloud(cmd *cobra.Command, targetArchStr string, bootMode *
...
+	requiredArgs := []string{"azure-client-id", "azure-client-secret", "azure-tenant", "azure-subscription", "azure-resource-group", "azure-image-name"}
+	for _, argName := range requiredArgs {
+		arg, err := cmd.Flags().GetString(argName)
+		if err != nil {
+			return nil, err
+		}
+		if arg == "" {
+			missing = append(missing, fmt.Sprintf("--%s", argName))
+		}
+	}
```
avitova, line 281: Since all of the args are required, we could also remove whatever is in the function before, and just do:
```suggestion
    usedArgs := make(map[string]string)
	requiredArgs := []string{"azure-client-id", "azure-client-secret", "azure-tenant", "azure-subscription", "azure-resource-group", "azure-image-name"}
	for _, argName := range requiredArgs {
		arg, err := cmd.Flags().GetString(argName)
		if err != nil {
			return nil, err
		}
		if arg == "" {
			missing = append(missing, fmt.Sprintf("--%s", argName))
		} else {
		  usedArgs[argName] = arg
		}
	}
```

And then we return values from the map. Not necessary, might complicate things once we start using arguments that are not required, but I think worth thinking about?

achilleas-k (approved):

achilleas-k:
Very nice and clean.  Thank you!

supakeen:
Failing with data race detections in the merge queue.

supakeen:
I think we found this race here because we actually run with the race detector in this branch; I've opened a PR to address it and we'll need to rebase this PR (and likely others) after that lands: #2487

PR #2478, opened 2026-06-30
fedora: drop slirp4netns from the IoT image

nullr0ute:
The slirp4netns usage for rootless containers hasn't been the default for quite some time, we've only kept it around for upgrade paths, but now it's no longer supported in podman 6 at all so drop it. People should have long migrated by now.


supakeen:
Was this also dropped from the `ostree` commit that's produced downstream and do you want to drop it from all supported Fedora's or only rawhide? 🙂.

supakeen:
I've force-pushed your branch to amend the missing manifest checksums so this can go green.

supakeen:
Test failure will be fixed with #2487.

achilleas-k (approved):
Cheers.  LGTM!

supakeen (approved):

PR #2479, opened 2026-07-01
depsolvednf: don't require solver.json in test

achilleas-k:
Drop the hard requirement for solver.json to be on the system and for it to contain the use_dnf5 property.

----

https://github.com/osbuild/image-builder/pull/2475#discussion_r3501097332


avitova (approved):

brlane-rht (approved):
Ack

PR #2480, opened 2026-07-01
ova: set os type and virtual hardware version (HMS-3248)

croissanne:
Setting the appropriate os type and virtual hardware version can dramatically impact performance on platforms like vSphere.

---

I tested this on vcenter 8, they all boot just fine.


pkg/distro/defs/loader.go
```diff
@@ -555,6 +555,9 @@ type ImageTypeYAML struct {
 
 	DiskImageVPCForceSize *bool `yaml:"disk_image_vpc_force_size"`
 
+	DiskImageVMWareOVFOSType                 string `yaml:"disk_image_vmware_ovf_os_type"`
+	DiskImageVMWareOVFVirtualHardwareVersion string `yaml:"disk_image_vmware_ovf_virtual_hardware_version"`
+
```
supakeen: Let's put them under `DiskConfig` (line 519)? Perhaps even under a `VMWare` struct inside it.

I've been trying to find some logical grouping to all the configs we have but this one seems like it has a good place to go.

croissanne: wdyt?

supakeen:
I think you missed adjusting this for ELN?

pkg/distro/defs/loader.go
supakeen: Much better ❤️

achilleas-k (dismissed):

data/distrodefs/rhel-9/imagetypes.yaml
```diff
@@ -2068,6 +2068,10 @@ image_types:
     platforms:
       - <<: *x86_64_bios_platform
         image_format: "ova"
+    disk_config:
+      ovf_vmware:
+        os_type: "rhel9_64Guest"
+        virtual_hardware_version: "vmx-18"
```
thozza, line 2074: Out of curiosity, why not `"vmx-21"`, since all rhel9 cases elsewhere use that?

thozza (dismissed):
LGTM

data/distrodefs/rhel-9/imagetypes.yaml
lzap: I do not understand neither 18 nor 21. I would expect all hw versions to match the vSphere version that introduced the official support in the UI except RHEL 10 which is nicely explained in the comment above.

Image

croissanne: What's the source on that table @lzap ? Clicking in vsphere the rhel9_64Guest is available from ESXi 7.0 U1 (vmx-18) and up. So I just picked the minimum. We could pick vmx-21 though for the nvme stuff, I'm open to either. Long term it should probably become a customization tbh.

croissanne: @thozza wdym with "all rhel9 cases elsewhere" ?

lzap: The image is Gemini with instructions to show vmx version when RHEL release is available in the UI.

The official docs actually allows for support even older versions, but in that case the process is to select the older version. So I guess they mean "it runs": https://compatibilityguide.broadcom.com/detail?program=software&productId=55714&persona=live

In general, I agree with keeping the version as the minimum possible that works. I would not bother with NVMe stuff if we can.

Edit: I think vmx-18 is in fact a good choice here.

lzap (dismissed):
Checksums.

supakeen (dismissed):

achilleas-k (dismissed):

achilleas-k:
I thought we fixed this: https://gitlab.com/redhat/services/products/image-builder/ci/images/-/jobs/15983803543

achilleas-k:
also, they shouldn't be building to begin with :thinking:

achilleas-k:
Well I fixed it.  Kinda.

achilleas-k:
> Well I fixed it. Kinda.

Actually it got fixed by #2599, but then we got checksum conflicts again.

achilleas-k (approved):

supakeen (approved):

PR #2482, opened 2026-07-03
distro/eln: use dnf5 and update package lists

yselkowitz:
* dnf5 replaces dnf, yum, and microdnf in RHEL 11.
* glycin includes SVG support by default, so rsvg-pixbuf-loader is obsolete.
* use wsl-setup and set the WSL distro name.

https://github.com/fedora-eln/eln/issues/530

/cc @supakeen


data/distrodefs/eln-11/wsl.yaml
```diff
@@ -21,72 +21,24 @@ image_types:
...
-          shallow_merge:
-            cloud_init:
-              - filename: "99_wsl.cfg"
-                config:
-                  datasource_list:
-                    - "WSL"
-                    - "None"
-                  network:
-                    config: "disabled"
+    distribution_config:
```
supakeen: You'll want to change this to:

```
wsl: *wsl_config
```

And add the `default_name` at the top of the file; or since this anchor is only used once in this file you can inline its contents entirely here.

Currently this is failing because it's not under a `wsl` key.

supakeen (changes requested):
When changes are made to the YAML files you'll need to run `./tools/gen-manifest-checksums.sh` and squash the (potential) changes under `test/data` into the same commit.

We use this for CI to only rebuild actually changed manifests and it makes it explicit which artifacts are affected by a certain change.

---

Let me know if you have no time to deal with our idiosyncrasies and I can handle it for you 🙂

yselkowitz:
Modified, but please double check.

supakeen (dismissed):
Thanks, looks good to me; let's see what the build/boot thinks about it in CI.

data/distrodefs/eln-11/wsl.yaml
```diff
@@ -21,72 +10,30 @@ image_types:
...
-
+      wsl:
+        config:
+          boot_systemd: true
+        distribution_config:
+          shortcut:
+            enabled: true
+            icon: /usr/share/pixmaps/fedora-logo.ico
+          oobe:
+            default_name: Fedora-ELN
```
supakeen: So CI uncovered that this requires there to be a `%s` so if you could change it to `Fedora-ELN-%s` for *now* then I'll do a follow-up to drop it, but I'll need to change some code.

supakeen:
I've force-pushed to your branch to amend the checksums after your last change so we can get this merged today and land it in time for release.

supakeen (approved):

achilleas-k (approved):
Thank you!

PR #2483, opened 2026-07-03
Add aliases for bootc ami and qcow2 image types (HMS-10949)

kingsleyzissou:
## Summary

Add `aliases` entries to the bootc-generic distrodefs so that `cockpit-image-builder` can resolve user-facing image type names without translating them at the frontend layer.

## Changes

- Add `"aws"` alias for the `ami` bootc image type
- Add `"guest-image"` alias for the `qcow2` bootc image type

### DIFF
--- a/data/distrodefs/bootc-generic/imagetypes.yaml
+++ b/data/distrodefs/bootc-generic/imagetypes.yaml
@@ -125,10 +125,12 @@ image_types:
 
   "ami":
     <<: *raw_image_type
+    aliases: ["aws"]
     filename: "disk.raw"
 
   "qcow2": &qcow2_image_type
     <<: *raw_image_type
+    aliases: ["guest-image"]
     filename: "disk.qcow2"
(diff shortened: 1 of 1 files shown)


data/distrodefs/bootc-generic/imagetypes.yaml
```diff
@@ -125,10 +125,12 @@ image_types:
 
   "ami":
     <<: *raw_image_type
+    aliases: ["aws"]
```
supakeen, line 128: The field is called `name_aliases`.

kingsleyzissou: My bad, I rushed it. Weirdly it worked though...

Ahhh I just clocked, I was using an old composer. This might not actually be needed at all

kingsleyzissou:
This actually isn't needed, I was on an older version of composer without realising it

PR #2467, opened 2026-06-26
Enable basic bootc-image-builder tests [HMS-10851]

achilleas-k:
Move bootc-image-builder related files, enable them, and clean up duplicate fixtures.

This PR creates a new top-level directory called `bootc-image-builder` which holds anything related to the tool (whenever possible).  It now contains the Containerfile, test scripts, and the old bootc-image-builder README.

---

This PR only enables one test, `test_image_is_generated()`, for Fedora 44 and CentOS Stream 10 on both x86_64 and aarch64.  It's run on GitLab CI on AWS runners for simplicity, since the GitHub actions environment gets a bit too fiddly for image building.

PR #2493, opened 2026-07-07
progress: add file progress (HMS-10977)

croissanne:
This progress "bar" creates a file in the output directory ending with
    ".progress". The file only contains the last status, meaning it is
    truncated each time. Each line is a valid JSON struct. Regular messages
    without progress are ignored. Finally, no subprogress is printed without
    the context of the progress above (for instance the progress of overal
    pipelines and the stages within the currently running pipeline will be
    contained in a single statement).

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -244,15 +244,8 @@ type cmdManifestWrapperOptions struct {
 // used in tests
 var manifestgenDepsolver manifestgen.DepsolveFunc
 
-func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []string, w io.Writer, wd io.Writer, wrapperOpts *cmdManifestWrapperOptions) (*imagefilter.Result, error) {
-	if wrapperOpts == nil {
-		wrapperOpts = &cmdManifestWrapperOptions{}
-	}
+func getImage(cmd *cobra.Command, args []string) (*imagefilter.Result, error) {
 	repoDir, err := cmd.Flags().GetString("force-repo-dir")
-	if err != nil {
-		return nil, err
-	}
-	rpmmdCacheDir, err := cmd.Flags().GetString("rpmmd-cache")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


pkg/progress/progress.go
```diff
@@ -363,3 +374,84 @@ func (b *debugProgressBar) SetProgress(subLevel int, msg string, done int, total
...
+func (b *fileProgressBar) Start() (err error) {
+	b.w, err = os.Create(b.path)
+	return err
+}
+
+func (b *fileProgressBar) Write(p []byte) (n int, err error) {
+	return b.w.Write(p)
+}
+
+func (b *fileProgressBar) Stop() {
```
supakeen, line 419: In the other implementation we guard against `b.w` being `nil`. We should probably do the same here? It can be `nil` if `Start()` wasn't called.

pkg/progress/progress.go
```diff
@@ -363,3 +374,84 @@ func (b *debugProgressBar) SetProgress(subLevel int, msg string, done int, total
...
+			b.levels = b.levels[:level+1]
+		}
+		b.levels[level] = &item
+	}
+
+	for i := 0; i < len(b.levels)-1; i++ {
+		b.levels[i].SubProgress = b.levels[i+1]
+	}
+
+	err := b.w.Truncate(0)
```
supakeen: This is not atomic, whatever is reading this file will have to be able to deal with partial content.

cmd/image-builder/main.go
```diff
@@ -431,10 +347,6 @@ func cmdManifestWrapper(pbar progress.ProgressBar, cmd *cobra.Command, args []st
 			return nil, err
 		}
 		img = &imagefilter.Result{ImgType: imgType, Repos: nil}
-		// XXX: hack to skip repo loading for the bootc image.
-		// We need to add a SkipRepositories or similar to
-		// manifestgen instead to make this clean
-		forceRepos = []string{"https://example.com/not-used"}
```
supakeen, line 437: Eh, can we remove this? I don't think we did what this comment mentions.

supakeen (commented):
Took a cursory look.

pkg/progress/progress.go
croissanne: Oh right of course, yea I'll rewrite this to use `os.Rename`

achilleas-k (approved):

supakeen (approved):

avitova (approved):

achilleas-k:
Adding to the merge queue again.  The failure should be fixed by #2467 (which is already on the queue right now).

PR #2495, opened 2026-07-08
test: fix build info cache path creation to use correct runner_distro

achilleas-k:
In commit e24d7994d69c2e70afacb6779c85086e4b655c5b I attempted to fix the problem of having the build config generator run on a different distro than the build itself.  The problem was that by using the host distro to determine the runner path component in the cache meant that building on one distribution (CentOS Stream 9 for RHEL 7 images) and generating the configs on a different distribution (Fedora 44 for all) would always miss and cause rebuilds.  The attempted fix sets the runner distro to the appropriate one when filtering manifests for the rebuild. However, this fix neglected an important step: The cache download that precedes the filtering was still using the host distro path component. So when generating build configs for RHEL 7, the following occurred:
- Download build info from cache with runner distro Fedora 44 (the host distro).
- Check for build info using the a runner distro CentOS 9 (the runner for the build for RHEL 7).
- Miss every check.

This commit fixes the gen_build_info_dir_path_prefix() function to always consult the Schutzfile for the runner distro path component, essentially moving the previous fix into the function that's always used to construct cache paths, so all cache-related functionality uses the correct paths.

### DIFF
--- a/test/scripts/imgtestlib/cache.py
+++ b/test/scripts/imgtestlib/cache.py
@@ -8,7 +8,7 @@ from .build import (gen_build_name, get_manifest_id, read_build_info,
                     write_build_info)
 from .gitlab import log_section
 from .run import runcmd, runcmd_nc
-from .testenv import get_host_distro, get_osbuild_commit
+from .testenv import get_ci_runner_for, get_osbuild_commit
 
 S3_BUCKET = "s3://" + os.environ.get("AWS_BUCKET", "images-ci-cache")
 S3_PREFIX = "images/builds"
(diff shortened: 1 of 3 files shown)


supakeen (dismissed):

achilleas-k:
Rebase needed.

supakeen (dismissed):

croissanne (dismissed):

achilleas-k:
I saw this had conflicts in the merge queue so I rebased it on the queue branch.

supakeen (dismissed):

croissanne (dismissed):

achilleas-k:
this will need a rebase after the other PRs are merged.  sorry for the mess

croissanne (approved):

supakeen (approved):

PR #2500, opened 2026-07-09
ci: enable rawhide tests

supakeen:
Enable rawhide unit testing.


supakeen:
Same issue here; we don't have rawhide runners yet.

PR #2505, opened 2026-07-10
installers: grub2 timeout everywhere

### LINKED ISSUE #2223: blueprint's grub2 timeout is not hooked up to most image types
There is a grub2 menu timeout setting available as `customizations.installer.bootloader.grub2` `menu-timeout` but it appears to only be used in `pkg/distro/generic/bootc_imagetype.go` and `pkg/image/container_based_iso.go`
It should also work with the other installer types, including `minimal-installer` and `everything-network-installer`

Additionally it may be helpful to add a menu entry option as well, to make it possible to select other grub2 menus, eg. when building images for testing it would be nice to be able to skip the isomd5sum check to make things run faster. But that will require blueprint changes to be coordinated.

supakeen:
When `installer.bootloader.grub2.menu-timeout` is set in the blueprint make sure it applies to *all* installer types. Closes #2223.

### DIFF
--- a/pkg/manifest/iso_bootloaders.go
+++ b/pkg/manifest/iso_bootloaders.go
@@ -63,6 +63,8 @@ type Grub2X86Boot struct {
 
 	// Default Grub2 menu on the ISO
 	DefaultMenu int
+	// Default Grub2 menu timeout on the ISO
+	MenuTimeout *int
 	// Custom menu entries (overrides all default entries)
 	Custom []ISOGrub2MenuEntry
 }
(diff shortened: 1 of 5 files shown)


avitova (approved):
Nice. I don't see this used in `Grub2ISOLegacyStageOptions`, do we need to add it somewhere to that structure as well? It uses the `options` struct which has a `Grub2Config` too.

pkg/manifest/iso_bootloaders.go
```diff
@@ -63,6 +63,8 @@ type Grub2X86Boot struct {
 
 	// Default Grub2 menu on the ISO
 	DefaultMenu int
+	// Default Grub2 menu timeout on the ISO
+	MenuTimeout *int
```
brlane-rht, line 67: FWIW I have this as part of PR #2414 where I just used an int and compare with 0, the same way we do for DefaultMenu.

supakeen: Yea but I think for the timeout 0 is a valid value; while it isn't for DefaultMenu?

brlane-rht: Hrm, now that I re-read the docs I guess so (I just hate to see new pointers introduced). 0 == boot now, -1 == wait forever. Default is 5s without it set at all. I guess I'll have to change #2414 then.

brlane-rht: Ok, went to fix this and realized why I did it that way in the first place. The stages use omitempty so 0 isn't a valid setting. The osbuild code has its own defaults when it isn't set (60s for grub2.iso and legacy, 0 for the non-iso grub stages. So I used a non-pointer int and 0 to be consistent with this.

achilleas-k (approved):

PR #2510, opened 2026-07-11
disk: systemd-repart compatibility

supakeen:
A few commits that make our partition tables compatible with what `systemd-repart` expects. The important commit here is the 'align footer' one.

Currently gated behind flags (which could be dropped if we consider the behavior safe); can be changed and each commit can land independently.

I had closed this PR accidentally when removing my fork during the rename.


brlane-rht (approved):
Looks good

avitova (approved):
Very cool, I learned new things! TY!

PR #2514, opened 2026-07-13
Generate manifests with librepo sources for packages

achilleas-k:
Builds in CI now use image-builder, which by default generates manifests with the librepo instead of the curl source.  We need to generate manifests in a compatible way, otherwise we will keep rebuilding every image in every CI run.


croissanne (approved):
I want my runners back!

supakeen (approved):

achilleas-k:
Included in #2516 (merged).

PR #2515, opened 2026-07-13
Add hidden pkgsearch subcommand for querying available packages (HMS-11011)

kingsleyzissou:
Add `pkgsearch` subcommand for querying available packages

## Summary

Adds a hidden `pkgsearch` subcommand to `image-builder` that searches RPM package metadata for a given distro and architecture, outputting results as JSON. This enables tooling like cockpit-image-builder to query available packages without shelling out to dnf directly.

## Changes

- Add `cmdPkgSearch` handler with distro/arch/type-based repo resolution and a `pkgSearchFormatter` interface for extensible output formats
- Register `pkgsearch` as a hidden subcommand with `--distro`, `--arch`, `--type`, `--format`, and `--rpmmd-cache` flags
- Extract the solver call into a mockable `pkgSearcher` variable and add table-driven tests covering the full command flow, including zero-args invocation and `--rpmmd-cache` passthrough

JIRA: [HMS-11011](https://issues.redhat.com/browse/HMS-11011)

[HMS-11011]: https://redhat.atlassian.net/browse/HMS-11011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/image-builder/pkgsearch.go
+++ b/cmd/image-builder/pkgsearch.go
@@ -57,6 +57,13 @@ func (*jsonPkgFormatter) Output(w io.Writer, pkgs rpmmd.PackageList) error {
 	return enc.Encode(result)
 }
 
+// pkgSearcher performs the actual package search. It is a variable so
+// tests can replace it with a fake that doesn't require osbuild-depsolve-dnf.
+var pkgSearcher = func(d distro.Distro, archStr, cacheDir string, repos []rpmmd.RepoConfig, packages []string) (rpmmd.PackageList, error) {
+	solver := depsolvednf.NewSolver(d.ModulePlatformID(), d.Releasever(), archStr, d.Name(), cacheDir)
+	return solver.SearchMetadata(repos, packages)
+}
+
 func cmdPkgSearch(cmd *cobra.Command, args []string) error {
 	format, err := cmd.Flags().GetString("format")
 	if err != nil {
@@ -155,8 +162,7 @@ func cmdPkgSearch(cmd *cobra.Command, args []string) error {
 		cacheDir = defaultCacheDir()
 	}
 
-	solver := depsolvednf.NewSolver(d.ModulePlatformID(), d.Releasever(), archStr, d.Name(), cacheDir)
-	results, err := solver.SearchMetadata(searchRepos, args)
+	results, err := pkgSearcher(d, archStr, cacheDir, searchRepos, args)
 	if err != nil {
 		return err
 	}
--- a/cmd/image-builder/pkgsearch.go
+++ b/cmd/image-builder/pkgsearch.go
@@ -146,8 +146,15 @@ func cmdPkgSearch(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// XXX set sensible cache dir
-	cacheDir := ""
+	cacheDir, err := cmd.Flags().GetString("rpmmd-cache")
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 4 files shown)


kingsleyzissou:
I based this on the work that was done here:
https://github.com/osbuild/image-builder-cli/pull/268

cmd/image-builder/pkgsearch.go
```diff
@@ -0,0 +1,160 @@
...
+
+	distroString, err := cmd.Flags().GetString("distro")
+	if err != nil {
+		return err
+	}
+
+	archString, err := cmd.Flags().GetString("arch")
+	if err != nil {
+		return err
+	}
```
supakeen: Normally we default to the current architecture if not set for example:

```go
if archString == "" {
  archString = arch.Current().String()
}
```

Also we call this `archStr` in `cmd/main.go`.

cmd/image-builder/pkgsearch.go
```diff
@@ -0,0 +1,160 @@
...
+
+	forceRepos, err := cmd.Flags().GetStringArray("force-repo")
+	if err != nil {
+		return err
+	}
+
+	distroString, err := cmd.Flags().GetString("distro")
+	if err != nil {
+		return err
+	}
```
supakeen: We should default to the current `distro`, something like:

```go
distroString, err := cmd.Flags().GetString("distro")
if err != nil {
	return err
}

distroString, err = findDistro(distroString, "")
if err != nil {
	return err
}
```

Is used in for example `describe` or other places.

cmd/image-builder/pkgsearch.go
```diff
@@ -0,0 +1,160 @@
...
+
+	imageType, err := cmd.Flags().GetString("type")
+	if err != nil {
+		return err
+	}
+
+	repoOpts := repoOptions{
+		RepoDir:    repoDir,
+		ExtraRepos: extraRepos,
+		ForceRepos: forceRepos,
```
supakeen: This field isn't used here so better to not pass it.

cmd/image-builder/pkgsearch.go
kingsleyzissou: Thanks, that's good insight, been a while since I've touched this code

supakeen (approved):
Functionally seems to work. I don't particularly like the error I get when I pass no packages at all:

```
€ ./image-builder pkgsearch                                     
No distro name specified, selecting "fedora-44" based on host, use --distro to override
error invalid request
InvalidRequest: Search command must contain at least one package specification
error: DNF error occurred: InvalidRequest: Search command must contain at least one package specification
```

But since this is a hidden command for internal use (for now) I'm not too upset over it.

kingsleyzissou:
> Functionally seems to work. I don't particularly like the error I get when I pass no packages at all:
> 
> ```
> € ./image-builder pkgsearch                                     
> No distro name specified, selecting "fedora-44" based on host, use --distro to override
> error invalid request
> InvalidRequest: Search command must contain at least one package specification
> error: DNF error occurred: InvalidRequest: Search command must contain at least one package specification
> ```
> 
> But since this is a hidden command for internal use (for now) I'm not too upset over it.

Ya that is a bit of a weird error for that. I can sort that out in a follow up

croissanne (approved):
thank you!

PR #2519, opened 2026-07-14
test: mock cache dir

supakeen:
On my system when I run tests they fail all the time because I actually have a populated cache dir with a different value from what the tests expect.

Mock the cache dir for the tests that are failing.


achilleas-k (approved):
LGTM

(I had the same issue a few days ago and ignored it; thanks for being as lazy)

avitova (approved):
Oh, cool! Thanks!

PR #2520, opened 2026-07-14
Run more bootc-image-builder tests in gitlab [HMS-10854]

achilleas-k:
Enable more bootc-image-builder tests in gitlab.

This PR expands the testing of the existing gitlab job and adds two jobs (per architecture) that run more of the existing tests using pytest.

This change will add a lot more work to the gitlab testing pipelines that will be run on every PR and in the merge queue.  See [this comment on the relevant ticket](https://redhat.atlassian.net/browse/HMS-10854?focusedCommentId=17582961) for the full plan.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1073,6 +1073,12 @@ fail:
   extends: .terraform
   variables:
     RUNNER: aws/fedora-44-x86_64
+  rules:
+    # 1. Skip the job if we're on the github merge queue
+    - if: '$CI_COMMIT_BRANCH =~ /^gh-readonly-queue/'
+      when: never
+    # 2. Fallback rule for any other branch name
+    - if: '$CI_COMMIT_BRANCH'
   script:
     - sudo ./test/scripts/install-dependencies
     - sudo mkdir -pv /var/tmp/bootc-image-builder-tests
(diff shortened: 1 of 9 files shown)


achilleas-k:
I'd like to hear what others think of the last commit, skipping the bootc-image-builder tests on the merge queue.

avitova:
Well, if we had two PRs in the queue and the first one broke the latter, we'd not find out, and it would just get merged, right? But I think that might not happen as often, and we'd eventually find out, so why not skip those tests if they are expensive/take long time to finish.
If I understand correctly, there are a bunch of tests for bootc that we run in GitHub Actions anyway.

thozza (dismissed):
LGTM. Skipping the tests on merge queue is IMO acceptable as an interim state.

avitova (dismissed):
Ooops, I meant to approve too.

achilleas-k:
Glad we're all on board with the merge queue skip.

Some of the new tests are failing though, so I'll switch to draft while I look into them.

Thanks for the initial reviews.

achilleas-k:
I think I caught everything (for now).  There's one test that's failing, `bootc-image-builder/test/test_build_disk.py::test_image_build_without_se_linux_denials`, but I think that's a storage space issue.  It builds too many images and runs out of disk space.

I'll see if I can get the same test to run with a fixture that creates fewer images.  Otherwise, I'll disable it until we sort it out.  The image build fixture is a little convoluted and could require some refactoring to make it more manageable.

achilleas-k:
> I'll see if I can get the same test to run with a fixture that creates fewer images. Otherwise, I'll disable it until we sort it out. The image build fixture is a little convoluted and could require some refactoring to make it more manageable.

I did reduce the images, but the selinux denial issue is not resolved yet, so we're kicking this can down the road again.

brlane-rht (approved):
Yay!

thozza (approved):
LGTM

PR #2521, opened 2026-07-14
image-builder: add hidden `--with-upload-result` option

croissanne:
This saves the upload result in `$basename.upload-result`, if no uploader is involved, just create a default upload result containing the absolute path of the image.

This option is hidden for now, as its only usecase is cockpit-image-builder.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/osbuild/image-builder/pkg/arch"
 	"github.com/osbuild/image-builder/pkg/bootc"
+	"github.com/osbuild/image-builder/pkg/cloud"
 	"github.com/osbuild/image-builder/pkg/customizations/subscription"
 	"github.com/osbuild/image-builder/pkg/distro/generic"
 	"github.com/osbuild/image-builder/pkg/imagefilter"
@@ -578,6 +579,10 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	withUploadResult, err := cmd.Flags().GetBool("with-upload-result")
+	if err != nil {
+		return err
+	}
 	withMetrics, err := cmd.Flags().GetBool("with-metrics")
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


achilleas-k (dismissed):

brlane-rht (dismissed):
Ack!

cmd/image-builder/main.go
```diff
@@ -683,9 +687,26 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 
 	fmt.Fprintf(osStdout, "Image build successful: %s\n", imagePath)
 
+	// Default upload result to write out in case no uploader was specified
+	uploadResult := &cloud.UploadResult{
+		Provider: "None",
+		ImageID:  imagePath,
+	}
```
thozza, line 695: Do you see this local path to be used anywhere? If not, then it would make sense to keep this `nil` and not set it at all if no uploader is specified. If yes, then the upload results could still contain a local path in addition to the cloud location. In such case, something like `LocalPath` instead of `None` would be a bit more descriptive.

The commit message also claims that the path would be absolute, but that depends on what the user passes as the output directory. So it can be relative as well. A minor thing, but the commit message is misleading in this regard.

Note that `ImageID` is not sufficient for anything else than AWS, but I guess that this is simply the limitation of the uploader interface which is AWS-centric since the beginning. That said, reusing `ImageID` for the local image path may not be the best idea long-term, since we should really have target-specific upload results.

thozza (dismissed):
I'm OK with getting this in as is and addressing my comments later, so that cockpit-image-builder is not blocked. That said, it would be good to resolve the format of the upload result, before cockpit-image-builder gets too dependent on it. Ideally, we should not try to fit all different targets results into the `ImageID` just because the initial uploader interface didn't consider anything else than AWS.

cmd/image-builder/main.go
croissanne: Yea I agree this needs to be fixed in general, even with AWS it would be nice to have the region there.

I'm not sure what the best pattern would be though, something like we do for the targets? Or just one field per cloud? (so like uploadResult.AWS.(ImageID|Region)).

lucasgarfield:
I must be a bit of the loop, what's the specific use case we need this for in Cockpit Image Builder? General support for uploads, or something specific to Image Mode, or...?

croissanne:
> I must be a bit of the loop, what's the specific use case we need this for in Cockpit Image Builder? General support for uploads, or something specific to Image Mode, or...?

Having the upload results in the output directory allows us to construct a correct compose status  (for instance for aws images).

cmd/image-builder/main.go
croissanne: But let's do this in followups, currently cockpit-image-builder only supports AWS, and it's actually fine to change it in future (as long as 2 versions which work land at the same time in rhel etc.)

thozza:
> > I must be a bit of the loop, what's the specific use case we need this for in Cockpit Image Builder? General support for uploads, or something specific to Image Mode, or...?
> 
> Having the upload results in the output directory allows us to construct a correct compose status (for instance for aws images).

On top of that, you need this information to locate the image in the target cloud.

thozza (approved):
LGTM as the first step.

brlane-rht (approved):
Looks good!

PR #2522, opened 2026-07-14
test: fix vm.py import error [HMS-11012]

avitova:
Current imports are incorrect, and bootc-foundry fails with am error `AttributeError: module 'qemu.qmp' has no attribute 'legacy'`. This commit should fix it, as it imports the QEMUMonitorProtocol from qemu.qmp.legacy directly.

Attempting to fix https://gitlab.com/redhat/services/products/image-builder/ci/bootc-foundry/-/jobs/15303940592.

### DIFF
--- a/test/scripts/imgtestlib/vm.py
+++ b/test/scripts/imgtestlib/vm.py
@@ -15,10 +15,11 @@ import uuid
 from io import StringIO
 
 try:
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
THANK YOU!

Linking this to HMS-11012 since this is part of that task.

supakeen (approved):

PR #2524, opened 2026-07-15
pkg/disk.yaml: add an option to not grow root

jbtrystram:
Add a `grow_root_to_fill_disk` boolean option to allow not growing the root partition to use all available space. Defaults to `true` for backwards compatibility.

This allows creating disk images with empty available space, so users can create additionnal partitions or grow the root at first boot via e.g. `cloud-utils-growpart`.

See discussion in https://github.com/coreos/fedora-coreos-tracker/issues/2188

Assisted-By: Opencode.ai <Opus 4.6>

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -587,7 +587,7 @@ func (pt *PartitionTable) relayout(size datasizes.Size) uint64 {
 	// add the extra padding specified in the partition table
 	footer += datasizes.Size(pt.ExtraPadding)
 
-	// If the sum of all partitions is bigger then the specified size,
+	// If the sum of all partitions is bigger than the specified size,
 	// we use that instead. Grow the partition table size if needed.
 	end := pt.AlignUp(datasizes.Size(root.Start) + footer + root.Size)
 	if end > size {
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -52,6 +52,21 @@ type PartitionTablePolicy struct {
 	// readable by firmware (LVM, btrfs). When set to false an XBOOTLDR partition
 	// will not be created even for those filesystems.
 	EnsureXBOOTLDR bool `json:"ensure_xbootldr" yaml:"ensure_xbootldr"`
+
+	// Controls whether the root partition (the one containing "/") is
+	// grown to fill the remaining disk space during relayout. Defaults
+	// to true (nil means true) to preserve backward compatibility. Set
+	// to false to keep the root partition at its specified size.
+	GrowRootToFillDisk *bool `json:"grow_root_to_fill_disk,omitempty" yaml:"grow_root_to_fill_disk,omitempty"`
+}
+
+// growRootToFillDisk returns whether the root partition should be grown
+// to fill remaining disk space. Defaults to true when not explicitly set.
+func (pt *PartitionTable) growRootToFillDisk() bool {
+	if pt.Policy == nil || pt.Policy.GrowRootToFillDisk == nil {
+		return true
+	}
+	return *pt.Policy.GrowRootToFillDisk
 }
 
 var _ = MountpointCreator(&PartitionTable{})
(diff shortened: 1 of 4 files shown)


pkg/disk/partition_table.go
```diff
@@ -534,13 +552,14 @@ func (pt *PartitionTable) relayout(size datasizes.Size) uint64 {
...
+	root.Size = pt.AlignUp(root.Size)
 
 	// add the extra padding specified in the partition table
 	footer += datasizes.Size(pt.ExtraPadding)
 
-	// If the sum of all partitions is bigger then the specified size,
+	// If the sum of all partitions is bigger than the specified size,
 	// we use that instead. Grow the partition table size if needed.
-	end := pt.AlignUp(datasizes.Size(root.Start) + footer + root.Size)
+	end := pt.AlignUp(datasizes.Size(root.Start) + root.Size + footer)
```
achilleas-k: Why the order swap here?

pkg/disk/partition_table.go
```diff
@@ -534,13 +552,14 @@ func (pt *PartitionTable) relayout(size datasizes.Size) uint64 {
 	root := &pt.Partitions[rootIdx]
 	root.Start = start
 	root.fitTo(root.Size)
+	root.Size = pt.AlignUp(root.Size)
 
 	// add the extra padding specified in the partition table
 	footer += datasizes.Size(pt.ExtraPadding)
 
-	// If the sum of all partitions is bigger then the specified size,
+	// If the sum of all partitions is bigger than the specified size,
```
achilleas-k, line 590: Thanks for the typo fix but can you separate it into a different commit?  It's not really related to the rest of the changes.

pkg/disk/partition_table.go
```diff
@@ -36,6 +36,12 @@ type PartitionTable struct {
 	// Dictates if certain bits and bobs are required or not; uses the default
 	// policy if not set.
 	Policy *PartitionTablePolicy `json:"policy,omitempty" yaml:"policy,omitempty"`
+
+	// Controls whether the root partition (the one containing "/") is
+	// grown to fill the remaining disk space during relayout. Defaults
+	// to true (nil means true) to preserve backward compatibility. Set
+	// to false to keep the root partition at its specified size.
+	GrowRootToFillDisk *bool `json:"grow_root_to_fill_disk,omitempty" yaml:"grow_root_to_fill_disk,omitempty"`
```
achilleas-k: I don't really like booleans defaulting to true.  We do have it in a few places, but that's usually when interacting with external configs and their defaults.  If we're making our own, I'd rather avoid it.

Negative bools (like `DisableX = true`) are also a bit tricky and error prone.  But this whole discussion is moot anyway.

achilleas-k (changes requested):
A few minor comments below but more generally:

I'm not sure where this option fits better.  I'll have to think about it some more but I think my preference right now is to make not growing the default behaviour for image definitions.  We would have to keep the blueprint behaviour the same, so users don't get different behaviour, but the image definitions make no stability guarantees.  So the new rules would be:
- Partition tables in image definitions get created as they are described, with no growing.
- Customizing a partition table from the blueprint does automatic things, including root growing.
- Blueprint adds an option, somewhere, that prevents growing root to fill.
    - This could be a boolean, but it might also become an option that lets users control _which_ filesystem to grow.  In other words, it could be `grow_to_fill = "<mountpoint>|None"`, with default value `/`, and `None` disabling the behaviour.
- Grow to fill option is also added to image definitions for similar behaviour.

This is just a bit of brainstorming, so don't take any of these as instructions.  We need to consider all side effects and how this fits into potential future behaviour first.

pkg/disk/partition_table.go
jbtrystram: I made it default to true so it keep the same existing behavior when not specified, to not change people that haven't changed their configs.

pkg/disk/partition_table.go
jbtrystram: That order felt more natural (start + size + footer) as it reads in the same order. But it's just additive quantities anyway so I can revert

pkg/disk/partition_table.go
```diff
@@ -47,6 +47,12 @@ type PartitionTable struct {
...
```
supakeen: Please move this property into the `Policy` struct; otherwise this is fine.

supakeen:
Summary from our chat about this: let's move the property under the `Policy` struct and then this can be merged.

supakeen (dismissed):

supakeen:
@jbtrystram I already approved this; but there seems to be a conflict that needs to be resolved :)

achilleas-k:
I rebased and cleared the conflict

achilleas-k (approved):

jbtrystram:
sorry Achilleas, I did not see you had already fixed the conflict and pushed the exact same content :grin:

supakeen (approved):

PR #2526, opened 2026-07-16
Improve error message for pkgsearch with no packages

kingsleyzissou:
Improve error message for pkgsearch with no packages

## Summary

Return a clear error when `pkgsearch` is invoked without any package names instead of silently returning all packages.

## Changes

- Add an early argument check in `cmdPkgSearch` that returns an explicit error when no package names are provided
- Update the existing "zero args" test case to assert the new error message instead of expecting a successful result

### DIFF
--- a/cmd/image-builder/pkgsearch.go
+++ b/cmd/image-builder/pkgsearch.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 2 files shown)


kingsleyzissou:
Follow up for https://github.com/osbuild/image-builder/pull/2515#pullrequestreview-4702112906

lucasgarfield (approved):
I think one nice improvement on this is that the failure path is much, much shorter.

The old test was a bit dodgy as well, what with 2 expectedPkgs... now the failure is properly tested.

brlane-rht (approved):
Ack!

PR #2527, opened 2026-07-16
Restore anaconda-iso/iso type to bootc-image-builder

brlane-rht:
It was removed, but is still used by some, so this adds the missing code and adjusts it to work with the current image-builder

Related: HMS-10969

### DIFF
--- a/cmd/image-builder/bib_legacy.go
+++ b/cmd/image-builder/bib_legacy.go
@@ -1,0 +1,474 @@
+package main
+
+import (
+	"fmt"
+	"math/rand"
+	"slices"
+	"strconv"
+	"strings"
+
+	"github.com/osbuild/blueprint/pkg/blueprint"
+	"github.com/sirupsen/logrus"
+
+	"github.com/osbuild/image-builder/internal/cmdutil"
+	"github.com/osbuild/image-builder/pkg/arch"
+	"github.com/osbuild/image-builder/pkg/bib/osinfo"
+	"github.com/osbuild/image-builder/pkg/container"
+	"github.com/osbuild/image-builder/pkg/customizations/anaconda"
+	"github.com/osbuild/image-builder/pkg/customizations/kickstart"
+	"github.com/osbuild/image-builder/pkg/depsolvednf"
+	"github.com/osbuild/image-builder/pkg/disk"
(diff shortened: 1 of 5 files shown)


supakeen:
I'm confused why this was removed in the first place; it definitely wasn't supposed to be 🙂, did something go wrong during the merge/previous import? cc @achilleas-k.

achilleas-k:
> I'm confused why this was removed in the first place; it definitely wasn't supposed to be 🙂, did something go wrong during the merge/previous import? cc @achilleas-k.

I don't know for certain.  I discovered it wasn't working as expected when I enabled the BIB tests here.  It's possible we never tested it with bootc-image-builder built from the image-builder-cli repo.

cmd/image-builder/bib_main.go
```diff
@@ -135,6 +135,12 @@ func bibManifestFromCobra(cmd *cobra.Command, args []string, pbar progress.Progr
 	// 1. the bootc disk manifests contains exports for all supported image types
 	// 2. the bootc legacy types (iso, anaconda-iso) always do a single build
 	imgTypeStr := imageTypes[0]
+
+	// The anaconda-iso code is different enough for a separate function
+	if imgTypeStr == "anaconda-iso" || imgTypeStr == "iso" {
+		return manifestFromCobraForLegacyISO(imgref, buildImgref, imgTypeStr, rootFs, rpmCacheRoot, config, useLibrepo, cntArch)
+	}
```
achilleas-k, line 142: I like that this is separated completely from every other image type.  Makes it easier to work with this specific image type, which is very different from the rest, without needing to think about potential interactions with anything else.

cmd/image-builder/bib_legacy.go
achilleas-k, line 1: It looks like this file contains functions that were already copied over and distributed to other files.  But I'm fine with merging this as is and cleaning up after.  Now that we have the ISO tests passing, we can much refactor things safely.  I'd like for the refactoring to happen soon though, to minimise the chance that we start making changes, fixing bugs, and causing confusion and drift between the duplicate functions.

achilleas-k (approved):
LGTM!

Thanks for taking care of this on short notice.  A comment below for a follow-up, but good to merge.

supakeen (approved):

PR #2528, opened 2026-07-17
manifest/raw_bootc: support subscription registration on first boot (HMS-10897)

lucasgarfield:
Wire up `customizations.subscription` for bootc disk images (qcow2, ami, ...).
The shared subscriptionService() machinery now runs in the RawBootcImage
pipeline: it creates the first-boot registration unit in /etc (not /usr, which
is immutable ostree content) and reruns Insights collection on boot so the
console reflects the new commit hash after upgrades — same options the
standalone Subscription pipeline uses for ostree systems.

Only the disk variant is wired for now.

While in there, URI-backed file customizations now return a clear error at
manifest generation time — they were previously silently broken in this
pipeline (missing manifest source).

### DIFF
--- a/pkg/manifest/raw_bootc.go
+++ b/pkg/manifest/raw_bootc.go
@@ -49,6 +49,8 @@ type RawBootcImage struct {
 
 	// DiskCustomizations can influence things in the base OS tree
 	DiskCustomizations DiskCustomizations
+
+	inlineData []string
 }
 
 func (p RawBootcImage) Filename() string {
@@ -261,6 +263,38 @@ func (p *RawBootcImage) serialize() (osbuild.Pipeline, error) {
 			postStages = append(postStages, stages...)
 		}
 
+		if p.OSCustomizations.Subscription != nil {
+			subStage, subDirs, subFiles, subServices, err := subscriptionService(
+				*p.OSCustomizations.Subscription,
+				&subscriptionServiceOptions{
+					// ostree based: unit in /etc (not /usr commit content),
+					// insights on boot
+					InsightsOnBoot: true,
+					UnitPath:       osbuild.EtcUnitPath,
+					// no-op for now: manifestForDisk never sets OSCustomizations.PermissiveRHC
+					PermissiveRHC: common.ValueOrEmpty(p.OSCustomizations.PermissiveRHC),
+				})
+			if err != nil {
+				return osbuild.Pipeline{}, err
+			}
+
+			subFileStages, err := p.genFileStagesAndRecordInlineData(subFiles)
+			if err != nil {
+				return osbuild.Pipeline{}, err
+			}
+			stages := []*osbuild.Stage{subStage}
+			stages = append(stages, osbuild.GenDirectoryNodesStages(subDirs)...)
+			stages = append(stages, subFileStages...)
+			stages = append(stages, osbuild.NewSystemdStage(&osbuild.SystemdStageOptions{
+				EnabledServices: subServices,
+			}))
+			for _, stage := range stages {
+				stage.Mounts = mounts
+				stage.Devices = devices
+			}
+			postStages = append(postStages, stages...)
+		}
+
 		// First create custom directories, because some of the custom files may depend on them
 		if len(p.OSCustomizations.Directories) > 0 {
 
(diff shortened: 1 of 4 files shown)


pkg/manifest/raw_bootc.go
```diff
@@ -261,6 +265,36 @@ func (p *RawBootcImage) serialize() (osbuild.Pipeline, error) {
...
 
+		if p.OSCustomizations.Subscription != nil {
+			subStage, subDirs, subFiles, subServices, err := subscriptionService(
+				*p.OSCustomizations.Subscription,
+				&subscriptionServiceOptions{
+					// ostree based: unit in /etc (not /usr commit content),
+					// insights on boot
+					InsightsOnBoot: true,
+					UnitPath:       osbuild.EtcUnitPath,
+					PermissiveRHC:  common.ValueOrEmpty(p.OSCustomizations.PermissiveRHC),
```
supakeen: This isn't actually ever set right? Not a blocker but also just a no-op at the moment.

supakeen (dismissed):

supakeen:
Needs linter appeasement otherwise OK.

pkg/manifest/raw_bootc.go
lucasgarfield: Correct. I went ahead and added a comment (// no-op for now: manifestForDisk never sets OSCustomizations.PermissiveRHC) but we could probably also wire it up without too much effort.

pkg/manifest/raw_bootc.go
```diff
@@ -273,14 +308,31 @@ func (p *RawBootcImage) serialize() (osbuild.Pipeline, error) {
...
 			for _, stage := range stages {
 				stage.Mounts = mounts
 				stage.Devices = devices
 			}
 			postStages = append(postStages, stages...)
 		}
 
+		// single point where enabled services are collected, as in the OS
+		// pipeline; OSCustomizations.EnabledServices is not honoured yet
+		var enabledServices []string
```
brlane-rht: Why have 2 lists for enabled services? You have already populated `subscriptionEnabledServices` and all this does is copy it.

pkg/manifest/raw_bootc.go
```diff
@@ -333,16 +385,24 @@ func (p *RawBootcImage) serialize() (osbuild.Pipeline, error) {
...
-		inlineData = append(inlineData, string(file.Data()))
+// genFileStagesAndRecordInlineData returns the stages that create the given
+// files and records their content as inline data for getInline().
+func (p *RawBootcImage) genFileStagesAndRecordInlineData(files []*fsnode.File) ([]*osbuild.Stage, error) {
+	for _, file := range files {
+		// files that come via an URI are not inline data, they
+		// would need to be added to the manifest sources via a
+		// fileRefs() implementation like the one in the OS pipeline
+		if file.URI() != "" {
+			return nil, fmt.Errorf("cannot create file %q from %q: files from an URI are not supported for bootc disk images yet", file.Path(), file.URI())
```
brlane-rht: I'd drop the `yet`, no need to speculate in error messages and get people's hopes up :)

pkg/manifest/raw_bootc_test.go
```diff
@@ -409,3 +411,163 @@ func TestRawBootcPXE(t *testing.T) {
...
+		opts := s.Options.(*osbuild.SystemdUnitCreateStageOptions)
+		if opts.Filename == "osbuild-subscription-register.service" {
+			return opts.UnitPath
+		}
+	}
+	require.Fail(t, "no osbuild-subscription-register.service unit.create stage found")
+	return ""
+}
+
+func collectMkdirPaths(stages []*osbuild.Stage) []string {
```
brlane-rht, line 624: May as well use this in the TestRawBootcPXE function, probably in a new commit just to keep things clean.

brlane-rht:
Is there a reason you didn't add this to PXE and ISO types?

pkg/manifest/raw_bootc.go
lucasgarfield: Done.

pkg/manifest/raw_bootc.go
lucasgarfield: Originally I thought I would add in “futures” for easily enabling Customizations.Services for Image Mode, and this is how it is implemented in OS.serialize(). But now that I think about it… I’m not sure that we would ever want to enable Customizations.Services for Image Mode. That seems like something that the containerfile should own… it could be really counter-intuitive if someone later adjusted the same services in the containerfile because our customization would go into /etc/ and over-ride the containerfile during a 3 way /etc/ merge...

Maybe for now it would be safer to just do everything related to Subscriptions in the same code block - I've updated the PR. If someone wants to add support for Customizations.Services in the future, a little bit more friction might be a good thing.

pkg/manifest/raw_bootc_test.go
lucasgarfield: Nice catch, done.

lucasgarfield:
> Is there a reason you didn't add this to PXE and ISO types?

I will do so in a follow-up. I’m still quite new to hacking on `image-builder` so I wanted to keep things simple for this pull request. Also, I don't really grok how the ISO works yet. :wink:

brlane-rht (approved):
Looks good!

supakeen (approved):

PR #2533, opened 2026-07-20
Fedora ISO Modernization (HMS-9965)

supakeen:
This is the implementation of the Fedora 45 change proposal [1] which drops the use of Lorax to create `boot.iso`-style media in Fedora 45.

Behavior is gated on the definition of `LoraxTemplatePackage`. If it is not present we perform certain steps that Lorax used to perform ourselves albeit it being a much more minimal set of actions (see the documentation section of the change proposal for the steps).

Some of the steps we already performed unconditionally so these are untouched (user creation, selinux permissions).

[1]: https://fedoraproject.org/wiki/Changes/ModernizeBootISO

---

Some bits aren't the nicest, especially the handling of root filesystem excludes for non-installer image types (`bootc` and `pxe`). I'd like to leave that for a follow-up as it isn't strictly worse but might require some more foundational changes to root filesystem options and configs.


avitova (approved):
lgtm:)

croissanne (approved):
makes sense to move the excludes to the image defs. thanks!

PR #2535, opened 2026-07-20
eln: set runner to eln11

supakeen:
The `rhel11` runner doesn't actually exist and because of bugs in `osbuild`'s runner detection this would fall back to `rhel7` and not `rhel82`.

Regardless of this; we *should* be using the `eln11` runner which is specific to ELN. This likely fixes the issues we've been seeing in the snapshot updates [1].
    
Because the rhel7 runner that's being erroneously selected doesn't set up either minimal passwd *or* call sysusers the relevant user doesn't exist and thus skopeo fails.

[1]: https://github.com/osbuild/image-builder/pull/2511


avitova (approved):
Quick fix! 🤞

achilleas-k (approved):

PR #2538, opened 2026-07-21
eln: disk images are LVM by default (HMS-11050, HMS-10803)

supakeen:
This PR does a bunch of clean-up in the beginning, aligning the two partition tables that were in effect closer and closer; it then gets rid of the non-LVM version and switches all disk images over to use an LVM-based layout.

---

For passers-by. This isn't a final thing. Discussions are still ongoing surrounding:

- What should the default LVM layout look like.
- Should we offer LVM *and* non-LVM image types (`ec2`, `ec2-lvm`, `azure`, `azure-lvm`, for example).


supakeen:
Not stale just still under discussion how we want to provide this.

supakeen:
Not stale just still under discussion how we want to provide this.

PR #2539, opened 2026-07-21
Generate bootc-image-builder test jobs dynamically and enable all tests [HMS-11073]

achilleas-k:
Glob the test paths under bootc-image-builder/test/ to dynamically create one gitlab job for each test file.  We separate these because a single job for all tests would take too long.

With this change, we can be sure that no bootc-image-builder test is skipped.

This also enables all[*](https://github.com/osbuild/image-builder/pull/2539#issuecomment-5048005984) bootc-image-builder tests.


achilleas-k:
Needs work.  Drafting.

achilleas-k:
Some tests are still disabled.  I created https://redhat.atlassian.net/browse/HMS-11073 with subtasks for the remaining ones.

bootc-image-builder/test/_test_build_pxe.py
avitova, line 1: Do you want to leave a comment also into the file that these will be skipped, so that the naming is not confusing?

avitova (approved):
Yeah, we agreed that this is a work in progress, so I agree with skipping some tests conditionally or with renaming the file so that it does not get grabbed by the generator. Thank you, this was an interesting introduction to how we generate tests for Gitlab!
Looks good! One comment but not necessary.

bootc-image-builder/test/_test_build_pxe.py
achilleas-k: I don't mind either way.  I think since I have the task open, I'll want to get to it soon, so I don't expect it to stay like that for long (_famous last words_).

supakeen (approved):

PR #2540, opened 2026-07-21
bib_legacy: Use functions from bootc

brlane-rht:
Use the shared versions of LabelForISO, LoraxTemplates, LoraxTemplatePackage, and GetDistroAndRunner from bootc.

Related: HMS-10969

### DIFF
--- a/cmd/image-builder/bib_legacy.go
+++ b/cmd/image-builder/bib_legacy.go
@@ -3,9 +3,6 @@ package main
 import (
 	"fmt"
 	"math/rand"
-	"slices"
-	"strconv"
-	"strings"
 
 	"github.com/osbuild/blueprint/pkg/blueprint"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


cmd/image-builder/bib_legacy.go
```diff
@@ -402,7 +351,7 @@ func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, erro
 
 	mf := manifest.New()
 
-	foundDistro, foundRunner, err := getDistroAndRunner(c.SourceInfo.OSRelease)
+	foundDistro, foundRunner, err := bootc.GetDistroAndRunner(c.SourceInfo.OSRelease)
```
avitova, line 354: Just saying that bootc getDistroAndRunner also accepts osRelease.ID == "ol" which this function did not, so this changes behaviour a little bit.

avitova (commented):
Looks good to me, nice cleanup. 🧹 I am not approving just to confirm that the change in behaviour is intentional.

cmd/image-builder/bib_legacy.go
brlane-rht: Yeah, I didn't mention that because the two functions drifted while the anaconda-iso support was gone, the version in the bootc package is the correct one.

supakeen (approved):

supakeen:
Nice cleanup indeed!

lzap (approved):

cmd/image-builder/bib_legacy.go
avitova: Thanks!

avitova (approved):

PR #2542, opened 2026-07-21
image-builder: distro detection for `build` and `manifest`

supakeen:
See https://github.com/osbuild/image-builder/issues/2541 for the bug this resolves. In https://github.com/osbuild/image-builder/commit/40c680562189265f544680864b98a736e6b2855d we restructured to a `getImage` call for the `build` and `manifest` subcommands. This `getImage` call does not perform distribution detection when no distribution is passed.

We don't pass a distribution in Koji nor OpenQA (we detect it on the mock root we run in) thus this bug breaks a bunch of things.

First commit introduces failing testcase, second bug fixes it.

---

This will need a follow-up with more extensive testing and unification of these codepaths between the other subcommands.

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -349,6 +349,22 @@ func getImage(cmd *cobra.Command, args []string) (*imagefilter.Result, error) {
 		}
 		img = &imagefilter.Result{ImgType: imgType, Repos: nil}
 	} else {
+		var bpDistroName string
+		blueprintPath, err := cmd.Flags().GetString("blueprint")
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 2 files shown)


brlane-rht (approved):
Looks good

ondrejbudai (approved):
The double-call to `findDistro` is unfortunate as it causes `image-builder` to print `No distro name specified, selecting ... based on host` twice but I think this is acceptable for a hotfix.

supakeen:
> The double-call to `findDistro` is unfortunate as it causes `image-builder` to print `No distro name specified, selecting ... based on host` twice but I think this is acceptable for a hotfix.

Yep, necessary evil for now.

PR #2543, opened 2026-07-21
bib: Replace logrus usage with olog

brlane-rht:
logrus was removed in commit dbdc5c30572aab20568d92250f712a1ed7a5b888
but snuck back in with the bib merge. This removes it and replaces it
with olog.
    
The log level now accepts any string, which simply turns on all the
logging output in addition to using the progress bar. This is different
from the verbose flag which turns on all the output from the osbuild
stages.

### DIFF
--- a/bootc-image-builder/test/test_opts.py
+++ b/bootc-image-builder/test/test_opts.py
@@ -80,28 +80,6 @@ def test_opts_arch_is_same_arch_is_fine(tmp_path, build_fake_container, target_a
         assert expected_err in res.stderr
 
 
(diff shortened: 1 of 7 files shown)


croissanne (dismissed):

avitova (dismissed):

lzap (dismissed):
Sorry, wrote a grep that never worked, tested it, yeah it was a pass :-)

achilleas-k (approved):

supakeen (approved):

PR #2544, opened 2026-07-22
anaconda: enable shell shenanigans for anaconda

supakeen:
When not calling Lorax we need to set up the shell for Anaconda itself. This is temporary as I'd like to move the enablement of this into Anaconda itself. However; since this is currently breaking Fedora's OpenQA tests and the other approach needs a bit more work we should handle it ourselves for now.

See: https://github.com/rhinstaller/anaconda/pull/7193 for what will replace this.


croissanne (approved):
lgtm

achilleas-k (approved):

PR #2546, opened 2026-07-22
fedora: create 46

supakeen:
Branching is happening in about 3 days; `image-builder` should recognize `fedora-46` as a valid distribution. Add repository definitions for it because that's what we use to determine if a given distribution is valid.


schuellerf (approved):
LGTM

avitova (approved):

PR #2550, opened 2026-07-23
pkg/koji/upload: session credentials to headers

avitova:
In the release 1.31, Koji moved from accepting session id and key as part of URL into accepting it in an HTTP header. This commit moves what we previously passed in session credentials and encoded into URL to header.

Relevant: https://docs.pagure.org/koji/release_notes/release_notes_1.31/
Relevant: https://pagure.io/koji/pull-request/3426


croissanne (approved):
amazing, thanks!

schuellerf (approved):

PR #2551, opened 2026-07-23
gitlab: remove setup-show option for pytest

achilleas-k:
In f58502be3442ae2a24d8412a85256539d49cbd69 I added the --setup-show option to the pytest call, which prints the fixture setup and teardown during a run.  This turns out to be very ugly and makes it harder to spot failed tests.  Reverting.


brlane-rht (approved):
Good idea

avitova (approved):
Let's remove it. ❎

PR #2552, opened 2026-07-23
disk: take the ESP size from the image type's partition table (RHEL-214147)

lucasgarfield:
When a blueprint contains a partitioning customization we build the partition table with NewCustomPartitionTable(), which derives almost everything from the customization and only inherits the partition table *type* from the image type's own table. The EFI system partition that addPartitionsForBootMode() adds when the customization doesn't declare one was therefore hardcoded to 200 MiB, silently discarding whatever the image type defines.

A user who customizes their disk without mentioning /boot/efi is not opting out of the image type's ESP; they are expressing no opinion about it, and should get the same ESP an uncustomized build would give them. The definitions vary more than the hardcoded value suggests: rhel-8 defaults to 100 MiB, the rhel-8/9 edge types use 127 MiB, rhel-9/10 "oci" and "azure" use 500 MiB, Fedora IoT and the generic bootc types use 501 MiB, and the CVM types use 512 MiB. Adding any disk customization silently replaced all of those with 200 MiB.

For most image types the mismatch is only cosmetic, since the ESP holds just shim and grub, which are a few MiB. It is a real bug for UKI-based image types: "ec2-cvm" and "azure-cvm" (rhel-9 and rhel-10) set bootloader: uki, which redirects the kernel install root to the ESP, so the kernel and initrd live there instead of in /boot. Those types declare a 512 MiB ESP for that reason. Building one with any partitioning customization that doesn't happen to mention /boot/efi produced a 200 MiB ESP instead, leaving no room for a kernel update on the running system.

Thread the size through CustomPartitionTableOptions, as we already do for RequiredMinSizes, and have both callers pass the ESP size of the image type's base partition table. Add PartitionTable.ESPSize() to look it up, identifying the ESP by partition type rather than by mountpoint so tables that mount it elsewhere still work. Three types count as an ESP: the GPT GUID, the DOS ID "ef", and plain FAT16 "06", which the aarch64 DOS tables (IoT, minimal-raw) use for firmware that reads the FAT partition directly.

The 200 MiB default is kept as a fallback for callers that don't set the option and for image types whose base table has no ESP. Customized builds now follow the image type's declaration in both directions -- rhel-8 shrinks to its declared 100 MiB just as the CVM types grow to 512 MiB -- and an image type can change its ESP size purely in YAML and have it apply to customized builds too.

### DIFF
--- a/pkg/disk/partition_table.go
+++ b/pkg/disk/partition_table.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"path/filepath"
 	"slices"
+	"strings"
 
 	"github.com/google/uuid"
 
@@ -730,6 +731,23 @@ func (pt *PartitionTable) FindMountableOnPlain(mountpoint string) Mountable {
 	return path[0].(Mountable)
 }
 
+// ESPSize returns the size of the EFI system partition in the PartitionTable.
+// Returns 0 if the table has no ESP. The ESP is identified by its partition
+// type, not by its mountpoint, so that tables which mount it somewhere other
+// than /boot/efi are still detected. Besides the GPT GUID and the DOS ID,
+// plain FAT16 ("06") also counts: the aarch64 DOS tables (IoT, minimal-raw)
+// type their ESP that way for firmware that reads the FAT partition directly.
+func (pt *PartitionTable) ESPSize() datasizes.Size {
+	for _, part := range pt.Partitions {
+		if strings.EqualFold(part.Type, EFISystemPartitionGUID) ||
+			strings.EqualFold(part.Type, EFISystemPartitionDOSID) ||
+			part.Type == FAT16BDOSID {
+			return part.Size
+		}
+	}
+	return 0
+}
+
 func clampFSSize(mountpoint string, size datasizes.Size) datasizes.Size {
 	// set a minimum size of 1GB for all mountpoints
 	// with the exception for '/boot' (= 500 MB)
@@ -1197,20 +1215,30 @@ func hasESP(disk *blueprint.DiskCustomization) bool {
 	return false
 }
 
+// defaultESPSize is the size of the automatically created EFI system partition
+// when the caller does not specify one in CustomPartitionTableOptions.
+const defaultESPSize = 200 * datasizes.MiB
+
 // addPartitionsForBootMode creates partitions to satisfy the boot mode requirements:
 //   - BIOS/legacy: adds a 1 MiB BIOS boot partition.
-//   - UEFI: adds a 200 MiB EFI system partition.
+//   - UEFI: adds an EFI system partition of options.ESPSize (or defaultESPSize).
 //   - Hybrid: adds both.
 //
 // The function will append the new partitions to the end of the existing
 // partition table therefore it is best to call this function early to put them
 // near the front (as is conventional).
-func addPartitionsForBootMode(pt *PartitionTable, disk *blueprint.DiskCustomization, bootMode platform.BootMode, architecture arch.Arch) error {
+func addPartitionsForBootMode(pt *PartitionTable, disk *blueprint.DiskCustomization, options *CustomPartitionTableOptions) error {
+	bootMode := options.BootMode
 	if bootMode == platform.BOOT_NONE {
 		return nil
 	}
 
-	switch architecture {
+	espSize := options.ESPSize
+	if espSize == 0 {
+		espSize = defaultESPSize
+	}
+
+	switch options.Architecture {
 	case arch.ARCH_PPC64LE:
 		// UEFI is not supported for PPC CPUs so we don't need to
 		// add an ESP partition there
@@ -1226,7 +1254,7 @@ func addPartitionsForBootMode(pt *PartitionTable, disk *blueprint.DiskCustomizat
 	case arch.ARCH_AARCH64, arch.ARCH_RISCV64:
 		// (our) aarch64/riscv64 only supports UEFI right now
 		if !hasESP(disk) {
-			part, err := mkESP(200*datasizes.MiB, pt.Type)
+			part, err := mkESP(espSize, pt.Type)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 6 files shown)


pkg/disk/partition_table.go
```diff
@@ -730,6 +731,23 @@ func (pt *PartitionTable) FindMountableOnPlain(mountpoint string) Mountable {
...
+// Returns 0 if the table has no ESP. The ESP is identified by its partition
+// type, not by its mountpoint, so that tables which mount it somewhere other
+// than /boot/efi are still detected. Besides the GPT GUID and the DOS ID,
+// plain FAT16 ("06") also counts: the aarch64 DOS tables (IoT, minimal-raw)
+// type their ESP that way for firmware that reads the FAT partition directly.
+func (pt *PartitionTable) ESPSize() datasizes.Size {
+	for _, part := range pt.Partitions {
+		if strings.EqualFold(part.Type, EFISystemPartitionGUID) ||
+			strings.EqualFold(part.Type, EFISystemPartitionDOSID) ||
+			part.Type == FAT16BDOSID {
```
achilleas-k: I wonder if there are valid cases where there might be a FAT partition but also an ESP later in the table.  Perhaps this should be held as a fallback in case the partition type search ends without finding one, and not interrupt the search immediately on the first FAT partition.

achilleas-k (dismissed):
Nice work.  Thanks!  Much simpler than I thought it would be :)

Small note below but wont block.

LGTM!

pkg/disk/partition_table.go
lucasgarfield: Good point, it feels sloppy right now and I don't like it. :sweat_smile:  I added a second commit to handle it better. Feels like quite a bit of extra code to handle a little Raspberry Pi edge case but I guess that's edge cases for you!

pkg/disk/partition_table.go
```diff
@@ -730,6 +731,38 @@ func (pt *PartitionTable) FindMountableOnPlain(mountpoint string) Mountable {
...
+// partition" and nothing else, so matching them is unambiguous.
+//
+// FAT16 ("06") is different: it names a filesystem, not a role, so a table
+// could legitimately carry a FAT16 partition that is not an ESP. It is
+// accepted here only because the aarch64/riscv64 DOS tables (IoT,
+// minimal-raw) deliberately give their ESP this non-standard type for
+// Raspberry Pi compatibility. To keep that workaround from matching an
+// unrelated FAT16 partition, a "06" partition counts only when it is also
+// mounted at /boot/efi, and only if no unambiguously typed ESP exists
+// anywhere in the table.
```
supakeen: This comment is super verbose; saying 'for DOS pts we pick `ef` or `06` *if* mounted at `/boot/efi`' would be enough. The latter is still a workaround. Note super bad or anything but since it mentions a bunch of images that might change this there will be churn or the comment will become outdated fast.

Feel free to leave it if you like.

pkg/disk/partition_table.go
```diff
@@ -730,6 +731,38 @@ func (pt *PartitionTable) FindMountableOnPlain(mountpoint string) Mountable {
...
+func (pt *PartitionTable) ESPSize() datasizes.Size {
+	var fat16ESP datasizes.Size
+	for _, part := range pt.Partitions {
+		if strings.EqualFold(part.Type, EFISystemPartitionGUID) ||
+			strings.EqualFold(part.Type, EFISystemPartitionDOSID) {
+			return part.Size
+		}
+		if part.Type == FAT16BDOSID && fat16ESP == 0 {
```
supakeen: You use `strings.EqualFold` two lines above to compare DOS ID's but then go for a normal equality check here; why not be consistent?

pkg/disk/partition_table.go
```diff
@@ -730,6 +731,38 @@ func (pt *PartitionTable) FindMountableOnPlain(mountpoint string) Mountable {
...
+			if mnt, ok := part.Payload.(Mountable); ok && mnt.GetMountpoint() == "/boot/efi" {
+				fat16ESP = part.Size
+			}
+		}
+	}
+	return fat16ESP
+}
```
supakeen, line 751: We have the opportunity here to check if multiple ESPs are defined and error early; doesn't have to be in this PR but it'd be a cute follow-up; or maybe we already do that elsewhere in the codebase. I only read the changes.

pkg/disk/partition_table.go
lucasgarfield: Agreed, way too verbose and likely to become stale.

pkg/disk/partition_table.go
lucasgarfield: Good point!

pkg/disk/partition_table.go
lucasgarfield: I don't think a normal user could define multiple ESPs - the ESP is hardcoded in the image definition, right? So through customizations.disk a user can resize the predetermined ESP but not change it.

I'm up for putting it in (maybe in a follow-up)... it seems like the main utility would be protecting distro maintainers, right?

supakeen: Follow-up is absolutely fine, yes. Feel free to resolve.

supakeen (approved):

pkg/disk/partition_table.go
achilleas-k: You're right, it does feel a bit fiddly.  Thinking about it more, this is meant to be used to find the ESP from the built-in image definition.  Perhaps we should ignore any poorly defined PTs that don't set the correct partition type, right?  If we can make sure, somehow, that this function will never be reused to scan a user-defined PT (from a blueprint), then we can have higher expectations for validity here.  Even if we're assuming a case where a user is defining image types and partition tables using drop-in definition files, or a `disk.yaml` in bootc images, we can have stricter rules there because it's the _I know what I'm doing_ interface.

achilleas-k (approved):

achilleas-k:
Good to merge on my end.  I had a comment above about the FAT detection again.  I wont ask for a change.  We can revisit it if we want to later.

PR #2562, opened 2026-07-28
disk: align footer to grain by default

supakeen:
In a previous PR [1] I introduced several compatibility options that make our partition tables (more) compatible with systems that have `systemd-repart` enabled on boot to grow or create partitions.

I introduced the behaviors controlled by flags. That way users can test the behavior without us switching over the defaults.

The most important one is the size of the last partition on disk. The default behavior we have at the moment is that we grow the last partition to take up all remaining size on the disk and then shrink it by the size of the GPT header (which is also at the end of the disk).

Since that footer isn't a multiple of grain size we get a partition that's slightly *larger* than expected.

If we instead align the footer to the grain size first and then grow the partition to that footer it becomes the exact requested size [2].

This PR enables the behavior by default. Nothing *should* depend on the specific alignment of the footer but it might be worth discussing. All tests are adjusted to the new offsets.

[1]: https://github.com/osbuild/image-builder/pull/2510
[2]: https://github.com/osbuild/image-builder/pull/2510/changes/4a3867ebbe433b7dd17694d758779a68196d4689


achilleas-k (approved):
I agree, nothing should depend on the exact alignment of the start ("leftmost" byte) of the backup header, as long as it's at the end of the disk.

I ran a couple of sample disk images (one build from `main` and one from this PR) through a few verification tools and everything seems fine.  The one thing that changes, other than the end sector of the last partition, is a note that the new partition table has some free sectors:

```
› sfdisk -V main/image/image/disk.img
main/image/image/disk.img:
No errors detected.
Header version: 1.0
Using 4 out of 128 partitions.
A total of 0 free sectors is available in 0 segments (the largest is 0 B).

› sfdisk -V PR/image/image/disk.img
PR/image/image/disk.img:
No errors detected.
Header version: 1.0
Using 4 out of 128 partitions.
A total of 2015 free sectors is available in 1 segment.
```

which is expected since we're now saving 1 MiB space for the backup PT and that creates a `1 MiB - headersize` gap between it and the last partition (and the size of that gap is `1 MiB - (512 B + (128 * 128) B) = 16896 B = 2015 sectors`.

Anyway, just verifying things.

LGTM :)

achilleas-k:
I was thinking about this again this morning and I'm wondering _why_ we need this change.

The description on this PR and the previous one it links to talk about being compatible with `systemd-repart` and what it expects.  Commit 88301ec9ef17f9d2b5fc56f58f5ca06841dbf132 also talks about having the last partition be exactly the expected size.  Now, I realise there's a lot of cases where we might not want to grow the last partition to fill the space on the disk (we should talk about that at some point find a nice way to expose that option to users).  But if we _are_ going to grow it to fill the space, why leave a gap?  At the end of the day, the size of the disk is the same in both cases.  With the change in this PR, the only difference is that you have about 16 KiB less space.  It's not a waste or anything, but since we know exactly the size of the disk and all the partitions, why leave the tiny gap there?  It feels a bit odd to me.  Like if I was creating partitions on a disk with some interactive tool and, while I was creating one last partition, asked the tool to make that partition take up the rest of the (usable) space, I'd probably be confused if the partitioning tool I was using was leaving, what looked like, an arbitrary number of bytes unused (which I realise is what `systemd-repart` does by default).

If this isn't breaking `systemd-repart` and the only reason to do this is so that we create identical disks, I don't know if it matters.  If we're breaking compatibility, like if running repart on our disks doesn't work because of this, then it's fine to consider the change.  But even then, filling up those gaps just makes more sense to me.

supakeen:
> I was thinking about this again this morning and I'm wondering _why_ we need this change.
> 
> The description on this PR and the previous one it links to talk about being compatible with `systemd-repart` and what it expects. Commit [88301ec](https://github.com/osbuild/image-builder/commit/88301ec9ef17f9d2b5fc56f58f5ca06841dbf132) also talks about having the last partition be exactly the expected size. Now, I realise there's a lot of cases where we might not want to grow the last partition to fill the space on the disk (we should talk about that at some point find a nice way to expose that option to users). But if we _are_ going to grow it to fill the space, why leave a gap? At the end of the day, the size of the disk is the same in both cases. With the change in this PR, the only difference is that you have about 16 KiB less space. It's not a waste or anything, but since we know exactly the size of the disk and all the partitions, why leave the tiny gap there? It feels a bit odd to me. Like if I was creating partitions on a disk with some interactive tool and, while I was creating one last partition, asked the tool to make that partition take up the rest of the (usable) space, I'd probably be confused if the partitioning tool I was using was leaving, what looked like, an arbitrary number of bytes unused (which I realise is what `systemd-repart` does by default).
> 
> If this isn't breaking `systemd-repart` and the only reason to do this is so that we create identical disks, I don't know if it matters. If we're breaking compatibility, like if running repart on our disks doesn't work because of this, then it's fine to consider the change. But even then, filling up those gaps just makes more sense to me.

Running repart on our disks indeed doesn't work; you can build any of our disk images and then run  `systemd-repart` on the booted system to see the error; but it comes down to that it wants to resize the root partition and it can't because there are 0 bytes available. This is due to how we make the root partition slightly *smaller* and then put the footer after it.

If you then rebuild the disk image with the appropriate option that's available (e.g: https://github.com/teamsbc/artifacts/commit/937012fb8bbe624687bbfc7491e3404d0923c244) repart will *not* complain (but it will grow the partition to footer).

This is especially relevant for images that run systemd-repart in their initramfs or as part of their boot sequence.

If you don't want to build an image then here are two downloads:

1. https://artifacts.teamsbc.net/main/2026.07.10/45/x86_64/index.html (without footer alignment).
2. https://artifacts.teamsbc.net/main/2026.07.16/45/x86_64/index.html (with footer alignment).

That you can boot to see the difference. Specifically on the 2026.07.10 image:

image

You can see the shipped repart configuration here: https://github.com/teamsbc/packages/tree/main/common/teamsbc-config.

And here is what it did for the 2026.07.16 image:

image

Note the growth to the footer, which is interesting because when we did that ourselves it's invalid but also harmless.

It does make it feel a bit funky.

supakeen:
Going to draft temporarily, I want to investigate the bit why `systemd-repart` grows things by 1004K; it's not harmful, but I just need to know first.

achilleas-k:
It looks like systemd-repart is aligning to a 4 KiB grain, so it'll grow to fill our space left for a 1 MiB grain and only leave 7 sectors at the end.

supakeen:
> It looks like systemd-repart is aligning to a 4 KiB grain, so it'll grow to fill our space left for a 1 MiB grain and only leave 7 sectors at the end.

Yep, that's exactly what's happening I think as well. So for full compatibility we *might* want to change our grain size; or we might not and accept that repart will grow the partition very slightly if enabled inside an image?

I'll adjust the grain size to 4096 for `teamsbc` to validate this, if I recall correctly changing the grain size *also* needs the absolute start offset (that's why I had a commit for it) as otherwise we start our partitions at grain size instead of first_lba (which would be 2048 blocks) and sfdisk errors out see [1] for details on that.

[1]: https://github.com/osbuild/image-builder/pull/2510/changes/776118de2310e252bc043e3477aa13b0d1e76d54

So this PR will make our partition tables compatible; systemd-repart will slightly *grow* the last partition but this is harmless.

We could also go all the way and make the minimum 1 MiB absolute start offset the default; but that only matters *if* a custom grain size is passed.

See: https://github.com/teamsbc/artifacts/commit/8fcb3b8035a386e445f313f845c1028fc59a353d

achilleas-k:
systemd-repart supports setting the grain size, but I only see it as a command line option [1].  It also says
> Setting this to "1M" ensures proper alignment for modern storage devices even after small fixed-size partitions such as a verity signature partition.

which to me sounds like we're better off keeping our grain at 1 MiB and configuring systemd appropriately.

[1] https://www.freedesktop.org/software/systemd/man/latest/systemd-repart.html#--grain-size=BYTES

supakeen:
> systemd-repart supports setting the grain size, but I only see it as a command line option [1]. It also says
> 
> > Setting this to "1M" ensures proper alignment for modern storage devices even after small fixed-size partitions such as a verity signature partition.
> 
> which to me sounds like we're better off keeping our grain at 1 MiB and configuring systemd appropriately.
> 
> [1] https://www.freedesktop.org/software/systemd/man/latest/systemd-repart.html#--grain-size=BYTES

Right, but that would mean we need to PR systemd itself for its service file (or Fedora downstream for its service file); or more appropriately `systemd` itself since users will *always* have to pass that make it the default for `systemd-repart`?

---

I don't think that's a blocker for this PR?

supakeen:
Thinking about it more I'm going to close this PR. repart compatibility can be achieved by setting the compatibility flags on our partition tables.

In the future I'd like to instead have `systemd-repart` be one of the partitioning tools we can use; which guarantees compatibility without us having to change our defaults.

achilleas-k:
> > systemd-repart supports setting the grain size, but I only see it as a command line option [1]. It also says
> > > Setting this to "1M" ensures proper alignment for modern storage devices even after small fixed-size partitions such as a verity signature partition.
> > 
> > 
> > which to me sounds like we're better off keeping our grain at 1 MiB and configuring systemd appropriately.
> > [1] https://www.freedesktop.org/software/systemd/man/latest/systemd-repart.html#--grain-size=BYTES
> 
> Right, but that would mean we need to PR systemd itself for its service file (or Fedora downstream for its service file); or more appropriately `systemd` itself since users will _always_ have to pass that make it the default for `systemd-repart`?
> 
> I don't think that's a blocker for this PR?

I was thinking we could add a drop-in that changes the repart Exec line to add the grain, but then manual runs wont be affected.  Then again, someone running things manually can set any grain they want and doesn't need to care about _our_ defaults.

I think the goal here for us should primarily be to create partition layouts that make sense and don't cause issues with other tooling.  Based on that, this PR is valid as it stops repart from trying to resize the partition to fit its calculated padding (grain alignment).

Beyond that, I don't know if I want to change the grain size globally.  The "proper alignment for modern storage devices" tells me we had it right to begin with for physical sector alignment (it also seems to be the default for `sfdisk`[^1]).  If we want some images to have 4K aligned partitions, we can make that configurable for repart compatibility, but I wouldn't change it globally to match systemd, especially given that it somewhat recommends using 1M too.

[^1]: Funnily enough, `sfdisk(8)` says this about the `grain` option: Specify  minimal  size  in  bytes used to calculate partitions alignment. The default is 1MiB and it’s strongly recommended to use the default. Do not modify this variable if you’re not sure.

supakeen:
Re-opened this PR after some discussion with @achilleas-k (who will update here).

achilleas-k:
For images that enable systemd-repart I see three options here, from worst to best:
1. systemd-repart errors out on boot: current scenario.
2. systemd-repart resizes a tiny bit on first boot: result of this PR.
3. systemd-repart does nothing on first boot.

All cases assume booting a disk image created by IB without growing the disk itself and with systemd-repart enabled and configured to grow root.

I think 2 is better than 1.  I also think we can achieve 3 in one of two ways: either using a 4 KiB grain for images that we have configured to run repart on boot, which we can do easily, or modify the `systemd-repart` unit to set grain to 1 MiB, which is a bit fiddly, but also doable.

Both options 2 and 3 require that we do the same last-partition-end-alignment though, so let's merge this PR.

achilleas-k (approved):

avitova (approved):
Thanks!

PR #2563, opened 2026-07-28
imgtestlib: configure setuptools package discovery

thozza:
Configure setuptools to discover packages under `test/scripts` so `imgtestlib` installs with this project instead of only being reachable through pytest `pythonpath` settings in local test runs. Enable downstream projects to import and reuse `vm.py` for image boot tests without copying helper modules into each repository.

I would like to try using this for boot testing the RHEL EKS AMI images (without forking the scripts).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,9 @@ dependencies = [
   "scp==0.15.0",
 ]
 
+[tool.setuptools.packages.find]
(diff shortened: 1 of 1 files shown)


supakeen (approved):

achilleas-k (approved):
Neato

PR #2565, opened 2026-07-29
Simplify early manifest generation code and enable custom seeds for bootc-based images

achilleas-k:
Another small PR towards simplifying and deduplicating the code that handles image-builder and bootc-image-builder.  This still doesn't actually deduplicate anything, but it simplifies the manifest generation code in `cmd/image-builder` a bit.  It also makes one functional change: The `--seed` option had no effect when generating bootc manifests and that's now been fixed.


brlane-rht (approved):
Looks Good!

supakeen (approved):

PR #2569, opened 2026-07-30
test: remove iot-bootable-container image type

avitova:
Remove iot-bootable-container from the Fedora test config list so CI no longer schedules a job for it, as it fails.

Relevant: HMS-1129


achilleas-k:
Problem is that we have checks that make sure there are no untested image types, so we'd have to work around that as well.

supakeen:
@achilleas-k Would it not be acceptable to drop the image type entirely? I don't think it is being produced anywhere?

achilleas-k:
> @achilleas-k Would it not be acceptable to drop the image type entirely? I don't think it is being produced anywhere?

Yes, let's drop it.  We can bring it back if we need to.
I initially wanted it as a POC for building bootc base images and to have something we build ourselves as an input artifact for bootc-based disk images and ISOs.  But it needs more work than we have time for now, so let's kill it.

avitova:
Thanks for your comments! I removed the image type completely. :)

achilleas-k (approved):

supakeen (approved):

PR #2571, opened 2026-07-30
manifest: accept human-readable image sizes

### LINKED ISSUE #2570: Accept a `datasizes.Size` for `--image-size`
`image-builder` has an `--image-size` argument that takes a `uint` in bytes for the size of the image. It'd be nicer if we accept a `datasizes.Size` here so users can pass `1GiB` and such.

tomatotomata:
Fixes #2570

## Summary

- let `manifest --image-size` use the existing `datasizes.Size` parser
- accept both raw byte counts and values with data-size units
- cover text marshaling and command flag parsing for both input forms

## Testing

- `go test ./pkg/datasizes`
- `GOOS=linux CGO_ENABLED=0 go test ./cmd/image-builder` *(cannot complete on this Windows host because `github.com/proglottis/gpgme` has no non-cgo build; the command package requires the project's Linux build dependencies)*

### DIFF
--- a/pkg/datasizes/size.go
+++ b/pkg/datasizes/size.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	"github.com/osbuild/image-builder/internal/common"
 )
@@ -17,6 +18,19 @@ type Size uint64
 // it is strictly equivalent to uint64(Size(1))
 func (si Size) Uint64() uint64 {
 	return uint64(si)
+}
+
+func (si *Size) UnmarshalText(data []byte) error {
+	value, err := Parse(string(data))
+	if err != nil {
+		return err
+	}
+	*si = Size(value)
+	return nil
+}
+
+func (si Size) MarshalText() ([]byte, error) {
+	return []byte(strconv.FormatUint(si.Uint64(), 10)), nil
 }
 
 func (si *Size) UnmarshalTOML(data interface{}) error {
--- a/pkg/datasizes/size.go
+++ b/pkg/datasizes/size.go
@@ -23,7 +23,7 @@ func (si Size) Uint64() uint64 {
 func (si *Size) UnmarshalText(data []byte) error {
 	value, err := Parse(string(data))
 	if err != nil {
-		return err
+		return fmt.Errorf("error decoding size: %w", err)
 	}
 	*si = Size(value)
 	return nil
--- a/cmd/image-builder/cmd.go
+++ b/cmd/image-builder/cmd.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/osbuild/image-builder/internal/olog"
+	"github.com/osbuild/image-builder/pkg/datasizes"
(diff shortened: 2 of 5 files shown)


lucasgarfield:
Huh, a robot?

image

pkg/datasizes/size.go
```diff
@@ -19,6 +20,19 @@ func (si Size) Uint64() uint64 {
 	return uint64(si)
 }
 
+func (si *Size) UnmarshalText(data []byte) error {
+	value, err := Parse(string(data))
+	if err != nil {
+		return err
```
supakeen: The other unmarshallers wrap this is a new error (see for example line 39), let's do the same here.

supakeen (commented):
Code works, thank you. One request for change in the review comments just for consistency.

It might be a bit harder but I'd like a commit message as well instead of just a subject line. It can include the usage (since users need to quote values-with-spaces).

---

Not 100% necessary in my opinion but maybe a manifest smoke test in `main_test.go` might make sense to ensure things don't blow up when passing this into manifest generation?

supakeen:
> Huh, a robot?
> image

Robots are taking over.

tomatotomata:
I took the consistency request as adding a fuller command description, not only the one-line subject. Commit 8dda9ec46 adds Long and an explicit quoted image-size 1 GiB example, with a small help-text regression test. The existing parser and flag tests stay unchanged; let me know what you think.

supakeen:
> I took the consistency request as adding a fuller command description, not only the one-line subject. Commit [8dda9ec](https://github.com/osbuild/image-builder/commit/8dda9ec4691c01f7ac09ce01aa03c5381483b437) adds Long and an explicit quoted image-size 1 GiB example, with a small help-text regression test. The existing parser and flag tests stay unchanged; let me know what you think.

You didn't address the error wrapping comment: https://github.com/osbuild/image-builder/pull/2571#discussion_r3688412066 which is probably more important :)

pkg/datasizes/size.go
tomatotomata: Fixed this in 2023bcb0d. UnmarshalText now wraps Parse errors with `error decoding size: ...`, matching the JSON and TOML paths, and TestSizeUnmarshalTextUnhappy locks in the context. go test ./pkg/datasizes passes. I left the unrelated worktree changes untouched. Let me know what you think.

supakeen (dismissed):
Thank you, looks good now. I'd prefer if you squash the fix commit(s) into the first commit but I won't block on that.

lzap (dismissed):
I am tired counting figures, love it.

supakeen:
Needs a rebase @tomatotomata.

tomatotomata:
I rebased the branch onto the current `main` and pushed `5db48e118`. The focused parser and command tests still pass locally, and `git diff --check` is clean. The full command-package test remains Linux-only from this Windows host because the local checkout cannot build `gpgme` and the Unix `statfs` path. The new CI run is in progress. Let me know what you think once it finishes.

supakeen (approved):

lzap (approved):

PR #2573, opened 2026-08-01
cmd/image-builder: drop `images` from version

supakeen:
It no longer makes sense to show the version of the `images` library as it no longer exists. This executable version is already the version of the 'library'.


avitova (approved):

brlane-rht (approved):
Ack!

PR #2578, opened 2026-08-03
Add 'raw-xz' image type to bootc-generic

rmsc:
This allows building compressed raw images from a bootc reference.


supakeen:
Mmm, my review queue was weird. You can undraft this if you want so we can review?

rmsc:
@supakeen thanks, but I put this in draft mode because the final image was not being compressed, even if the extension was .xz. My investigation seemed to suggest some changes to osbuild were needed too. I've since switched to a pure bootc install, and never got around to this..

I don't think I'll be able to finish this, so I'm closing. Feel free to pick this up if you think it's useful.

PR #2582, opened 2026-08-05
Start container with repos support for legacy ISO only [HMS-11156]

achilleas-k:
When generating the manifest for the legacy anaconda ISO, start the container (or the build container, when different from the base) with repos support, meaning it has network access and the secrets are bind mounted from the host.  In all other cases, we now start the container with no network and without explicitly bind mounting `/run/secrets/`.  Note that secrets are still implicitly bind mounted by podman in its default configuration when running image-builder or bootc-image-builder outside a container.

### DIFF
--- a/pkg/bootc/resolver.go
+++ b/pkg/bootc/resolver.go
@@ -56,7 +56,7 @@ type Container struct {
 	id        string
 	root      string
 	arch      string
-	extraOpts []string
+	storeOpts []string
 }
 
 // New creates a new running container from the given image reference.
@@ -65,7 +65,7 @@ type Container struct {
 // - --net host is used to make networking work in a nested container
 // - /run/secrets is mounted from the host to make sure RHSM credentials are available
 func NewContainer(ref string) (*Container, error) {
-	extraOpts := []string{}
+	var storeOpts []string
 	if isRootless, _ := isPodmanRootless(); isRootless {
 		// When running bc-i-b In a rootless container, its typically the case that /var/lib/containers/storage
 		// is a bind-mount of ~/.local/share/containers/storage, and we can't use this directly with podman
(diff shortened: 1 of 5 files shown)


achilleas-k:
We don't test with RHEL bootc containers in CI, but I did a lot of manual local testing for this.  No manifests were affected by the change.  Testing included:
- CentOS bootc disk images and installers.
- CentOS bootc disk images and installers with separate build containers.
- Same as above with RHEL containers.

brlane-rht (dismissed):
Looks good

pkg/bootc/resolver.go
```diff
@@ -74,83 +100,78 @@ func NewContainer(ref string) (*Container, error) {
...
 
-	args = append(args, extraOpts...)
+		// Mount the secrets directory only if it exists
+		if _, err := os.Stat(secretDir); err == nil {
+			args = append(args, "--volume", secretVolume)
+		}
+	}
 
-	args = append(args, ref, "infinity")
+	args = append(args, cnt.storeOpts...)
```
supakeen: This appends `cnt.storeOpts` a second time (the first time was at line 109. That's probably a bug? I don't know if it's harmless.

pkg/bootc/resolver.go
```diff
@@ -74,83 +100,78 @@ func NewContainer(ref string) (*Container, error) {
...
+	cnt.id = strings.TrimSpace(string(output))
 	// Ensure that the container is stopped when this function errors
 	defer func() {
 		if err != nil {
-			if stopErr := c.Stop(); stopErr != nil {
+			if stopErr := cnt.Stop(); stopErr != nil {
 				err = fmt.Errorf("%w\nstopping the container failed too: %s", err, stopErr)
 			}
-			c = nil
+			cnt = nil
```
supakeen: This isn't necessary anymore now that `start` is a method right? Previously it was stack local but now you're trying to set the object this method is attached to to `nil`?

supakeen (changes requested):

pkg/bootc/resolver.go
achilleas-k: Oops.  Thanks for catching that.  Probably a rebase artifact.

pkg/bootc/resolver.go
achilleas-k: Yup also good catch.

supakeen (approved):

brlane-rht (approved):
ack!

PR #2583, opened 2026-08-10
distrodefs: Add aliases to Fedora for GCP (HMS-11182)

tkoscieln:
This should enable building Fedora through service as CRC's API uses `gcp` which is something that the destrodefs do not recognize. Fix it by adding aliases similar to guest-image and others.

### DIFF
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -2759,6 +2759,7 @@ image_types:
 
   "cloud-gce": &cloud_gce
     <<: *cloud_base
+    name_aliases: ["gcp", "gce"]
     filename: "image.tar.gz"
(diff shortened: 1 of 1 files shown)


supakeen (commented):
Are we actually going to be building Fedora through the service and if so; which variant? We might need to verify parity work.

Also needs a decision if we want to support multiple variants or not. There are GCE images for Cloud (and Server I believe?).

If we're going to be building Fedora Cloud we need buy in from the Cloud SIG IMHO if we call it Fedora Cloud.

tkoscieln:
To add a context - this PR is not about enabling official support for Fedora, this is aimed at internal usage for CI only and is just a bugfix.

supakeen (approved):
Thanks, that makes sense.

achilleas-k (approved):

PR #2586, opened 2026-08-11
osbuild/rpm: implement generic env  

supakeen:
In [1] I landed changes to support setting arbitrary environment variables during RPM stage execution. This implements settings them in the wrapper in `image-builder`.
                                                
[1]: https://github.com/osbuild/osbuild/pull/2524

---

Note that I didn't implement the newly added kernel install environment variables; we have no use for them (yet) so I left the struct as is.


achilleas-k (dismissed):

pkg/osbuild/rpm_stage_test.go
```diff
@@ -103,6 +110,15 @@ func TestRPMStageOptionsClone(t *testing.T) {
 				clone.KernelInstallEnv.BootRoot = "modified"
 				assert.NotEqual(t, tt.opts.KernelInstallEnv.BootRoot, clone.KernelInstallEnv.BootRoot)
 			}
+
+			// Verify deep copy of GenericEnv map
+			if len(clone.GenericEnv) > 0 {
```
avitova, line 115: I believe this condition is not needed, and if the len was 0, ranging over it would not do any iterations, right?

avitova (dismissed):
lgtm, failures do ont seem to be related, but let's see after the rerun

achilleas-k (approved):

avitova (approved):

PR #2587, opened 2026-08-11
experimental: `image-version`, `image-id`

supakeen:
Allow use of the experimental flags to set `image-version` and `image-id`. We don't have a good idea yet how to pass these values otherwise and introducing command line flags is too early at this point.
    
This PR is based on top of [1].

[1]: https://github.com/osbuild/image-builder/pull/2586


supakeen:
Note that this is experimental, I can envision that these properties would actually be filled into the distro object instead so we can template them in definitions and such (for example; to set the right label that includes the image_version for sysupdate enabled partitions).

supakeen:
Closing this PR since it will be incorporated into another one.

PR #2589, opened 2026-08-11
anaconda_installer: platform build packages

supakeen:
Include the platform build packages in the Anaconda Installer pipeline generator. Previously we didn't do this and this leads to missing packages in the build (`grub2-ppc64le-modules` on ppc64le, and `s390utils-base` on `s390x`) causing builds to fail.

---

The first commit here deduplicates the packages we add to the build pipeline, this has the same end result as giving duplicates to the depsolver but makes the manifest checksums more stable when a package is added twice.

This is why the second commit indeed only shows `ppc64le` and `s390x` manifests changing but the first commit shows changing a lot of other manifests because duplicates are now gone there.

---

This commit should unblock us on Koji builds for these architectures, but also allows to (in the future) get rid of more of the code-based injections of build packages.


brlane-rht (approved):
Thanks!

supakeen:
```
Image build successful: fedora-44-everything-network-installer-s390x/fedora-44-everything-network-installer-s390x.iso

real    15m32.902s
user    0m16.716s
sys     0m2.742s(edited)
```

```
Image build successful: fedora-44-everything-network-installer-ppc64le/fedora-44-everything-network-installer-ppc64le.iso

real    17m37.034s
user    0m31.599s
sys     0m6.530s
```

With the RPMs from this PR.

croissanne (approved):
lgtm but I wonder if it's worth opening an issue to go through some of the other ones, like the coreos installer etc.?

Issue #2557, opened 2026-07-27
customizations.hostname does not apply to `image-installer` (HMS-11121)


secnum:
- image-builder version : 52.1
- osbuild version : 174
- Image type : image-installer
- Target OS : RHEL 8.10
- osbuild host : RHEL 10.2
- Partial blueprint : 

> [customizations]
> hostname = "baseimage"

This option does not produce the expected result. The installed system stays with the default name: localhost.

secnum:
Workaround  : 

```
[customizations.installer.kickstart]
contents = """
network --hostname=example.com
"""
```

Same behavior on an up to date Fedora 44 as osbuild host.

Issue #2481, opened 2026-07-02
Customizations.timezone not supported in fedora-44/server-network-installer


wt:
See [here](https://osbuild.org/docs/user-guide/image-descriptions/fedora-44/server-network-installer/) for the descriptions of the `server-network-installer` targets on fedora 44. The profile does not support options like timezone. There are many other customizations that are available in `pxe-tar-xz` that might be useful in `server-network-installer` also, e.g. `fips`, `openscap`, `firewall`, etc.

supakeen:
To be clear; the request here is to have `customizations.timezone` be translated to a kickstart option, and *not* to apply it to the installer environment itself?

wt:
> To be clear; the request here is to have `customizations.timezone` be translated to a kickstart option, and *not* to apply it to the installer environment itself?

That is correct. Ideally have the settings apply to the installed system. Hostname is another that probably makes sense in a similar way.

brlane-rht:
server-network-installer is the equivalent of the traditional boot.iso, if you want to customize the installed system you should be using a kickstart with it.

supakeen:
> server-network-installer is the equivalent of the traditional boot.iso, if you want to customize the installed system you should be using a kickstart with it.

The request here is though; why not generate that kickstart in `image-builder` from the customizations in the blueprint. We already do so for (some) other customizations; for example users.

All that would require is plumbing it into `kickstart.New()` in `pkg/customizations/kickstart.go`.

brlane-rht:
Personally I find the mixing of blueprint and kickstart confusing. In my way of thinking the blueprint is for customizing the image you are building (the netinst iso), and kickstarts are for the system being installed. We've gone far enough down this path that I'm not exactly sure what applies where without looking at the code.

wt:
@brlane-rht I think that this is a reasonable take.

However, I think there are reasons to consider implementing the customizations. The kickstart file contents are a blob to image-builder. There isn't a lot of validation that image-builder can do to the kickstart. Customizations can have validations that cause an error during image-builder building the image, that may provide a faster dev-test-release cycle for building images since you won't have to boot and image and try to use it before knowing if the image is correct.

If that is the direction you want to go, it might be nice to do two things to the docs.

1. Don't say that certain customizations land in the kickstart file. See [here](https://osbuild.org/docs/user-guide/blueprint-reference/#installer:~:text=Blueprint%20customizations%20that%20match%20the%20kickstart) for example.
[...]

PR #2590, opened 2026-08-11
image-installer: apply kernel.append to installed system via kickstart (HMS-11181)

swthakur:
## Summary
- Fixes `customizations.kernel.append` for classic `image-installer` ISOs.
- Previously, these options were applied to the installer ISO boot cmdline.
- Now they are applied to the installed system via kickstart `bootloader --append`.
- Installer-only unattended ISO options remain on the ISO boot cmdline.

## Why
Blueprint kernel append is intended for the installed system bootloader.
For `image-installer`, the OS payload is a liveimg tar (no disk partition table), so
`org.osbuild.kernel-cmdline` is not emitted for the payload. Kickstart is the correct path.

## Changes
- `pkg/distro/generic/images.go`: copy kernel options to `Kickstart.KernelOptionsAppend`, clear them from `InstallerCustomizations.KernelOptionsAppend`.
- `pkg/manifest/anaconda_installer_iso_tree.go`: emit `bootloader --append` in `makeKickstartStages`.

Related to #2175 (previous attempt, closed unmerged).

### DIFF
--- a/pkg/distro/generic/images.go
+++ b/pkg/distro/generic/images.go
@@ -929,12 +929,24 @@ func imageInstallerImage(t *imageType,
 	img.Kickstart.Keyboard = img.OSCustomizations.Keyboard
 	img.Kickstart.Timezone = &img.OSCustomizations.Timezone
 
+	// customizations.kernel.append (and image-type defaults) belong on the
+	// *installed* system. For image-installer the OS payload is a tar with no
+	// partition table, so org.osbuild.kernel-cmdline is never emitted for it.
+	// Pass the options through kickstart `bootloader --append` instead.
+	img.Kickstart.KernelOptionsAppend = append(img.Kickstart.KernelOptionsAppend, kernelOptions(t, customizations)...)
+
 	img.ExtraBasePackages = packageSets[installerPkgsKey]
 
 	img.InstallerCustomizations, err = installerCustomizations(t, bp.Customizations, options)
 	if err != nil {
 		return nil, err
 	}
+
+	// installerCustomizations() currently copies OS kernel options into
+	// InstallerCustomizations.KernelOptionsAppend, which anaconda_tar_installer
+	// then places on the ISO GRUB cmdline. Clear those so blueprint/OS kargs
+	// are not applied to the installer media itself.
+	img.InstallerCustomizations.KernelOptionsAppend = nil
 
 	img.ISOCustomizations, err = isoCustomizations(t, bp.Customizations)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


supakeen (commented):
So as I said elsewhere this needs some discussion. First: the implementation code-wise is mostly fine.

However this change would break our backwards compatibility guarantees that we give. I agree that it might make more sense that `kernel.append` applies to the to-be installed system and that perhaps we should have (if we don't already) an `installer.kernel.append` that only affects the installer.

But because the current behavior affects the installer users might be relying on it. Thus taking it away is a problem as it would break *other* users if we don't offer an alternative way of doing it, document it as a breaking change, and make sure people *know* about it.

Personally I'd go with:

1. `kernel.append` affects the *installed* system.
2. `installer.kernel.append` affects the *installer*.

Both in a single implementation, but I don't know how we handle the breaking of previous workflows correctly.

cc @achilleas-k / @ondrejbudai for your input on this 🙂

supakeen:
Also related (in the same area of certain customizations not affecting the installer or the installed system): #2481, #2557.

swthakur:
Thanks — agreed on avoiding a silent breaking change; I previously raised a broader installer-customization RFE (RHEL-191772).
That RFE is about customizing the Anaconda installer environment itself (packages/files/drivers/boot menu), separate from installed-system customizations.

Your proposal fits as a focused first step:
- `customizations.kernel.append`  installed system
- `customizations.installer.kernel.append`  installer ISO boot cmdline

achilleas-k:
I see two options here:
1. Break compatibility and make it obvious.
2. Hold off on the change until we do a larger customization migration.

The problem with 2 is I can't give any guarantees about when it might happen.
The problem with 1 is I don't have a good idea how to make it obvious.

I'm leaning more towards 1, but only if we can find a way to make users aware of the change so they're not surprised by the changing behaviour.  We should definitely put it in release notes and docs, but I don't expect that will reach many people.  Would a big `WARNING` be enough?

supakeen:
> I see two options here:
> 
>     1. Break compatibility and make it obvious.
> 
>     2. Hold off on the change until we do a larger customization migration.
> 
> 
> The problem with 2 is I can't give any guarantees about when it might happen. The problem with 1 is I don't have a good idea how to make it obvious.
> 
> I'm leaning more towards 1, but only if we can find a way to make users aware of the change so they're not surprised by the changing behaviour. We should definitely put it in release notes and docs, but I don't expect that will reach many people. Would a big `WARNING` be enough?

Personally I think breaking compatibility here is OK; we really do need a better way of communicating this to users though otherwise we'll be emitting warnings to infinity.

PR #2592, opened 2026-08-12
image-builder: reuse progress bars for upload

croissanne:
Using the proxy reader and the ability to restart progress bars, the
    progress bars can be used for uploads as well.

### DIFF
--- a/pkg/progress/proxy_reader.go
+++ b/pkg/progress/proxy_reader.go
@@ -1,0 +1,42 @@
+package progress
+
+import (
+	"io"
+)
+
+type Reader struct {
+	io.Reader
+	pb    ProgressBar
+	done  int
+	total int
+}
+
+func NewProxyReader(r io.Reader, total int, pb ProgressBar) (*Reader, error) {
+	err := pb.SetProgress(0, "Uploading", 0, total)
+	if err != nil {
+		return nil, err
+	}
+	return &Reader{
+		Reader: r,
+		pb:     pb,
+		done:   0,
+		total:  total,
+	}, nil
+}
+
+func (r *Reader) Read(p []byte) (int, error) {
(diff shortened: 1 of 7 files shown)


achilleas-k (dismissed):

pkg/progress/proxy_reader.go
```diff
@@ -0,0 +1,43 @@
...
+
+type Reader struct {
+	io.Reader
+	pb    ProgressBar
+	done  int
+	total int
+}
+
+func NewProxyReader(r io.Reader, total int, pb ProgressBar) (*Reader, error) {
+	pb.SetMessagef("Uploading %d bytes", total)
```
lzap: For image uploads, this number will be massive. You could use our unit convertor, or just simply chose at least MiBs... but this is good enough for the first version.

lzap (approved):
Haven't tested this.

pkg/progress/proxy_reader.go
croissanne: hmm good point, I'll look at this again and attach some screenshots so people have a better idea of the term bar

croissanne:
Alright, I changed it a bit to make the bar identical to the upload bar (which is just the default config with bytes true)

old:

progress-original

new:

progress-new

lzap (dismissed):
Nice

croissanne:
I removed the kinda useless message underneath
image

achilleas-k (dismissed):
Neat

achilleas-k:
We have unit tests that check for specific parts of the progress bar that changed now it seems.

croissanne:
> We have unit tests that check for specific parts of the progress bar that changed now it seems.

ye, fixed now

achilleas-k (approved):
Neat again

supakeen (approved):

PR #2594, opened 2026-08-13
test: mock host arch for package search

supakeen:
Mock the host architecture for package search. When running tests on riscv64 the distribution used in the test does *not* have available repositories which causes the test to fail.

Mock it.


lzap (approved):

brlane-rht (approved):
Ack!

PR #2597, opened 2026-08-14
fedora: drop explicit `anaconda-widgets`

supakeen:
We explicitly require `anaconda-widgets` but `anaconda` uses rich requires to select between `anaconda-gui` and `anaconda-webui`.

`anaconda-gui` requires `anaconda-widgets` so the solver sees `anaconda-widgets` in the transaction and finds it can solve more dependencies by selecting `-gui` instead of `-webui`.

Drop the explicit listing of `anaconda-widgets` to ensure rich dependency resolution works.

I've applied this to all versions instead of version gating it; it's theoretically a bug to have this in the list and shouldn't affect the actual output artifacts.


KKoukiou (approved):

brlane-rht (approved):
Ack!

lzap (approved):

PR #2601, opened 2026-08-17
ci: fake podman calls

supakeen:
Actually executing podman fails for likely reasons related to [1]; trying to work around it is very annoying. Mock things instead so we don't execute actual podman when not necessary.

[1]: https://github.com/actions/runner-images/issues/14473


supakeen:
Nope; this just introduces (or uncovers) *other* failures in the same tests.

PR #2602, opened 2026-08-18
Fix the container resolver tests

achilleas-k:
Separating this from #2599 to focus on fixing the resolver tests without needing to do all the rebuilds once it's done.

### DIFF
--- a/pkg/bootc/resolver_test.go
+++ b/pkg/bootc/resolver_test.go
@@ -175,6 +175,7 @@ func makeFakePodman(t *testing.T, content string) {
 	err := os.WriteFile(filepath.Join(tmpdir, "podman"), []byte(content), 0755)
 	assert.NoError(t, err)
 }
+
 func TestNewFakedUnhappy(t *testing.T) {
 	if os.Geteuid() != 0 {
 		t.Skip("skipping test; not running as root")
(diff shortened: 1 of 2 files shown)


pkg/bootc/resolver_test.go
```diff
@@ -192,8 +193,10 @@ exec /usr/bin/podman "$@"
 `
 	makeFakePodman(t, fakePodman)
 	_, err := bootc.NewContainer(testingImage)
-	assert.ErrorContains(t, err, fmt.Sprintf("mounting %s container failed: ", testingImage))
-	assert.ErrorContains(t, err, "stderr:\nforced-crash")
+	// Exact error message matching fails in GitHub Actions because of changes
+	// in the podman package in the ubuntu runner. Checking only that it
+	// failed.
+	assert.Error(t, err)
```
schuellerf, line 199: LGTM - but why not downgrade the runner for now? would this also be a solution?

schuellerf (approved):
But I won't block on that

croissanne (approved):

pkg/bootc/resolver_test.go
achilleas-k: Afaik, pinning isn't possible with the native GH runners.
https://github.com/actions/runner-images/issues/13034

achilleas-k: This is (probably) the issue we're dealing with: https://github.com/actions/runner-images/issues/14569

PR #2603, opened 2026-08-19
Run privileged unit tests on gitlab

achilleas-k:
Some of our unit tests require root and/or access to the host container storage.  These become tricky to run in GitHub actions because of a combination of a couple of reasons.  The Ubuntu runners were recently updated and seem to be having a lot of issues with podman.  Running them in the Fedora container that we use for the other unit tests has issues with using the container storage.

Run these tests in GitLab where we have better control of the environment.  This also much more closely matches production environments.

---

This is a follow-up from #2602.  The tests there succeeded, but I've been seeing some intermittent failures on the resolver tests still, like this one https://github.com/osbuild/image-builder/actions/runs/32161286115/job/95790547360?pr=2600

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,11 +59,8 @@ jobs:
 
       - name: Run unit tests
         # Skip tests that resolve manifests from local container storage.
-        # They are tested separately (see below: requires root)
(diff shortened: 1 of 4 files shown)


supakeen (approved):

brlane-rht (approved):
Looks good!

PR #2606, opened 2026-08-19
pkg/container: fix overriding sourceSpec input val

avitova:
On failure, Resolve returns an empty Spec. Add assigned that value the same variable as the input argument, so the error wrap used a new value returned from a failed function instead of the image that failed to resolve.


supakeen (approved):
Nice catch.

achilleas-k (approved):
Oh!  Nice one!

PR #2607, opened 2026-08-20
Rewrite container handling internals to drop graphdriver and gpgme dependencies [RHEL-56367]

### LINKED ISSUE #2624: skopeo copy leaks the credentials to any user running `ps` during a build
This leaks the credentials to any user running `ps` during a build. Is there no other way of passing these?

_Originally posted by @supakeen in https://github.com/osbuild/image-builder/pull/2607#discussion_r3854153379_

achilleas-k:
This PR rewrites container-related things in order to drop our dependency on graphdrivers and gpgme.  It finishes the work I started in #2346, where I changed the container resolver internals to use `skopeo`.

Changes in this PR, in order:
1. Rewrite testregistry to use github.com/distribution/distribution.  This simplifies the implementation of the internal container test registry significantly, offloading all the work to distribution/distribution.  It also gives us a more fully-featured test registry, so we can test pushing containers as well.
2. Test `osbuild-upload-container` with the new test registry.  The upload code wasn't being tested at all before, so any testing here is an improvement.
3. Rewrite `osbuild-upload-container` to use `skopeo copy` internally.

With these rewrites, we rely on `skopeo` for almost all of our interactions with container registries, which means we don't depend on any of the filesystem libraries or gpgme.


pkg/container/client.go
```diff
@@ -253,90 +231,81 @@ func (cl *Client) SkipTLSVerify() {
...
 
-	retryOpts := retry.RetryOptions{
-		MaxRetry: cl.MaxRetries,
+	if authfile := cl.GetAuthFilePath(); authfile != "" {
+		cmd.Args = append(cmd.Args, fmt.Sprintf("--authfile=%s", authfile))
 	}
 
-	var manifestDigest digest.Digest
+	if dockerAuth := cl.sysCtx.DockerAuthConfig; dockerAuth != nil {
+		cmd.Args = append(cmd.Args, fmt.Sprintf("--dest-creds=%s:%s", dockerAuth.Username, dockerAuth.Password))
```
supakeen: This leaks the credentials to any user running `ps` during a build. Is there no other way of passing these?

pkg/container/client.go
```diff
@@ -19,13 +19,8 @@ import (
 	_ "github.com/containers/image/v5/oci/layout"
```
supakeen: Do we still need these? I think they were used as side-effects for transports that are no longer being used.

pkg/container/client.go
```diff
@@ -253,90 +231,81 @@ func (cl *Client) SkipTLSVerify() {
...
-	destRef, err := docker.NewReference(target)
-	if err != nil {
-		return "", err
+	if tls := cl.GetTLSVerify(); tls != nil && !*tls {
+		cmd.Args = append(cmd.Args, "--dest-tls-verify=false")
 	}
 
-	retryOpts := retry.RetryOptions{
-		MaxRetry: cl.MaxRetries,
+	if authfile := cl.GetAuthFilePath(); authfile != "" {
```
supakeen, line 339: I think this is never empty since it has a fallback? Also what happens if it passes a non-existent file (e.g. it's fallback path?).

supakeen: I guess this was pre-existing it's just a new call site.

pkg/container/client.go
achilleas-k: I had the same thought but the only reason we have the user:pass option is because the osbuild-upload-container supports `--username` and `--password` and in that case, the original call leaks them to begin with.  Maybe we should get rid of that entirely and only support auth-file.

pkg/container/client.go
achilleas-k: Yeah I didn't want to change any functionality during this refactoring.
Also, it can still be set to nothing explicitly using `SetAuthFilePath("")`.

pkg/container/client.go
achilleas-k: No, you're right.  Thanks.

achilleas-k: Dropped.

pkg/container/client.go
supakeen: Sounds OK.

pkg/container/client.go
supakeen: Let's turn this into an issue here.

achilleas-k: https://github.com/osbuild/image-builder/issues/2624

supakeen (dismissed):

pkg/container/client.go
lucasgarfield: This seems to also cause a leak in the osbuild-composer workers when uploading container images. They call (*Client).UploadImage() at osbuild-composer/cmd/osbuild-worker/jobimpl-osbuild.go:1319 which results in the same argv leak under discussion here. 

Given that this is a new leak affecting the library API we may want to reconsider deferring to the future. I think it would make sense to update this pull request to only support auth-file.

lucasgarfield (changes requested):
Overall looks great!

Only thing I caught is that the skopeo argv leak is a bit worse than we thought. I'd request that we address that somehow (if we choose to defer, that's fine, but we should be intentional about it) before approval.

I would also be curious to hear why we decided to go with distribution/distribution for the registry. Did you consider google/go-containerregistry/pkg/registry? I have no strong opinion and can see good arguments for both, just curious about you decided. distribution/distribution has more features (which I'm not sure we're taking advantage of... yet?) but also more dependencies.

pkg/container/client.go
achilleas-k: Oh we do use it!  Ok then, I'll take care of it in this PR.

lucasgarfield (approved):
Looks good! Leak is fixed and I'm happy you ended up liking go-containerregistry/pkg/registry... the main thing that made me interested in it is that I'm now always on the lookout to minimize transitive dependencies. 😄

supakeen (approved):

PR #2608, opened 2026-08-21
test: drop the disable-nm-cloud-setup config [RHEL-112357]

achilleas-k:
We added this a long time ago to work around a bug in cloud-init that
was causing the service to fail on startup.  This issue has been
resolved and we can drop the workaround.

See RHEL-112357


croissanne (approved):
neatttttttttt

supakeen (approved):

test/config-list.json
```diff
@@ -696,10 +696,10 @@
...
       "distros": [
         "rhel-8*",
-        "rhel-9.6",
-        "rhel-9.7",
-        "rhel-9.8",
-        "rhel-9.9",
+        "rhel-9*",
+        "rhel-9*",
+        "rhel-9*",
+        "rhel-9*",
```
achilleas-k, line 702: What am I doing here?

PR #2613, opened 2026-08-24
pkg/rhsm: use Katello CA as default on Satellite hosts

lzap:
On Satellite-registered hosts, content is served through the Satellite content proxy whose TLS certificate is signed by the Katello CA. The default `redhat-uep.pem` only contains the Red Hat CDN CA and cannot validate these certificates, causing TLS failures when fetching packages from RHSM-backed repositories.

When `/etc/rhsm/ca/katello-server-ca.pem` exists (installed by the Satellite registration process), use it as the default CA. This single-point fix works transparently across all download backends (curl, librepo, dnf) since they all read the default CA from the same `Subscriptions` class. When a repo file explicitly sets `sslcacert`, that value still takes precedence.

A hybrid approach combining the RHSM CA and the system CA bundle was rejected because the three download backends configure CA certs differently (librepo has no `LRO_SSLCAPATH`, dnf does not expose `sslcapath`), making a reliable cross-backend setup fragile and hard to test.

Mirrors osbuild PR https://github.com/osbuild/osbuild/pull/2537.

Jira: https://redhat.atlassian.net/browse/SAT-37848

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/rhsm/secrets.go
+++ b/pkg/rhsm/secrets.go
@@ -42,6 +42,25 @@ type ConsumerSecrets struct {
 	ConsumerCert string
 }
 
+// defaultCACert picks the CA certificate to use when a repo does not set
+// sslcacert explicitly. On Satellite-registered hosts content is served through
+// the Satellite content proxy whose TLS certificate is signed by the Katello CA;
+// the default redhat-uep.pem only holds the Red Hat CDN CA and cannot validate
+// it. When katello-server-ca.pem is present (installed by Satellite
+// registration) it contains the CA chain needed for the content proxy, so prefer
+// it; otherwise fall back to redhat-uep.pem. A repo's explicit sslcacert still
+// takes precedence over this default (see GetSecretsForBaseurl).
+// katelloCACert is a variable so tests can point it at a temp file; production
+// code never reassigns it.
+var katelloCACert = "/etc/rhsm/ca/katello-server-ca.pem"
+
+func defaultCACert() string {
+	if _, err := os.Stat(katelloCACert); err == nil {
+		return katelloCACert
+	}
+	return "/etc/rhsm/ca/redhat-uep.pem"
+}
+
 func getRHSMSecrets() (*RHSMSecrets, error) {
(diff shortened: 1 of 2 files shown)


pkg/rhsm/secrets.go
```diff
@@ -42,6 +42,25 @@ type ConsumerSecrets struct {
...
+// takes precedence over this default (see GetSecretsForBaseurl).
+// katelloCACert is a variable so tests can point it at a temp file; production
+// code never reassigns it.
+var katelloCACert = "/etc/rhsm/ca/katello-server-ca.pem"
+
+func defaultCACert() string {
+	if _, err := os.Stat(katelloCACert); err == nil {
+		return katelloCACert
+	}
+	return "/etc/rhsm/ca/redhat-uep.pem"
```
supakeen, line 61: Maybe make both consts and swap between them? I dunno, stylistic only.

pkg/rhsm/secrets_test.go
```diff
@@ -132,6 +134,23 @@ func TestGetSecretsForBaseurlFallback(t *testing.T) {
...
+func TestDefaultCACert(t *testing.T) {
+	orig := katelloCACert
+	t.Cleanup(func() { katelloCACert = orig })
+
+	// Katello CA absent: fall back to the Red Hat CDN CA.
+	katelloCACert = filepath.Join(t.TempDir(), "does-not-exist.pem")
+	assert.Equal(t, "/etc/rhsm/ca/redhat-uep.pem", defaultCACert())
+
+	// Katello CA present: prefer it.
+	present := filepath.Join(t.TempDir(), "katello-server-ca.pem")
```
supakeen, line 148: It doesn't matter but why an additional `present` var here, can't it just assign directly to `katelloCACert`?

supakeen (approved):

lucasgarfield (approved):
Looks good to me. I would add to the merge queue but will wait for you to choose whether or not to address @supakeen's small issues, @lzap.

PR #2614, opened 2026-08-24
many: `riscv64` support for installer(s) (HMS-11278, HMS-11279, HSM-11280)

supakeen:
Initial support for `riscv64` ISO builds. Sets up repositories for Fedora 44 and 45, enables the bootloaders in our code, sets up a platform for it and then as special sauce copies all DTBs into the ISO to make manual selection easier.

---

- [x] #2619


achilleas-k:
Needs a rebase.

I think it'd be good to start treating riscv64 like we do s390x and ppc64le.  Let's add the upstream repos to `test/data/repositories/` as well so we get manifest checksums to make these changes visible.  I don't know if we also want to have manifest creation in GitLab CI.  Maybe we can do that after we added rpmrepo snapshots.

supakeen:
> Needs a rebase.
> 
> I think it'd be good to start treating riscv64 like we do s390x and ppc64le. Let's add the upstream repos to `test/data/repositories/` as well so we get manifest checksums to make these changes visible. I don't know if we also want to have manifest creation in GitLab CI. Maybe we can do that after we added rpmrepo snapshots.

These are follow-ups (and separate tickets we created in the JIRA epic), I don't think we want all of that in this PR?

(rebased)

achilleas-k:
I thought it'd be nice if the checksums were added first so we can see the changes here, but it's not a big issue.

achilleas-k (approved):

PR #2617, opened 2026-08-25
Implement rootless container introspection

### LINKED ISSUE #2568: Rootless container inspection (HMS-11220)
Currently our `osinfo` module requires root as we want to mount the container filesystem and then access it. We could do the mount in an unshare but that requires re-execing ourselves which might be fraught with details.

The 'simpler' solution is to rewrite our `osinfo` module to instead use `podman exec` with userland filesystem utilities to read files (as that's all we do).

alexlarsson:
This makes container introspection (osinfo) work when run as non-root, by using unprivileged podman and looking in the per-user podman container storage.

Fixes https://github.com/osbuild/image-builder/issues/2568

This means something like `image-builder manifest --bootc-default-fs ext4 --bootc-ref quay.io/fedora/fedora-bootc:44 qcow2` now works as a user. 

NOTE; `image-builder build`, doesn't yet work as it needs some additional work to have osbuild look into the right container store.

### DIFF
--- a/pkg/bootc/unsharefs.go
+++ b/pkg/bootc/unsharefs.go
@@ -1,0 +1,240 @@
+package bootc
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/fs"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const fieldSep = "\x1f"
+
+// podmanUnshareFS is an fs.FS that accesses a container or image
+// mount that exists inside the rootless "podman unshare" mount
+// namespace.
+type podmanUnshareFS struct {
+	root      string
+	argPrefix []string
+}
+
+// Ensure we implement the related ifaces
+var _ fs.FS = podmanUnshareFS{}
+var _ fs.ReadFileFS = podmanUnshareFS{}
+var _ fs.StatFS = podmanUnshareFS{}
+var _ fs.ReadDirFS = podmanUnshareFS{}
+
+func newPodmanUnshareFS(root string) podmanUnshareFS {
+	return podmanUnshareFS{root: root, argPrefix: []string{"podman", "unshare"}}
+}
+
+func (fsys podmanUnshareFS) run(args ...string) ([]byte, error) {
+	argv := append(append([]string{}, fsys.argPrefix...), args...)
+	/* #nosec G204 */
+	cmd := exec.Command(argv[0], argv[1:]...)
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("%w\nstderr:\n%s", err, stderr.String())
+	}
+	return stdout.Bytes(), nil
+}
+
+func (fsys podmanUnshareFS) fullPath(name string) string {
+	return filepath.Join(fsys.root, name)
+}
+
+func (fsys podmanUnshareFS) exists(name string) bool {
+	_, err := fsys.run("test", "-e", fsys.fullPath(name))
+	return err == nil
+}
+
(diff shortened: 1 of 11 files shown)


pkg/container/client.go
```diff
@@ -113,7 +113,28 @@ type Client struct {
...
+}
+
+func defaultStorePaths() (graphRoot, runRoot string) {
+	if os.Geteuid() == 0 {
+		return "/var/lib/containers/storage", "/run/containers/storage"
+	}
+
+	dataHome := os.Getenv("XDG_DATA_HOME")
+	if dataHome == "" {
+		dataHome = filepath.Join(os.Getenv("HOME"), ".local", "share")
```
supakeen: What if both are unset, then we end up with a relative path. Maybe that should blow up instead?

pkg/bootc/unsharefs.go
```diff
@@ -0,0 +1,171 @@
...
+		if len(f) != 2 {
+			continue
+		}
+		entries = append(entries, dirEntry{name: f[1], isDir: f[0] == "d"})
+	}
+	sort.Slice(entries, func(i, j int) bool { return entries[i].Name() < entries[j].Name() })
+	return entries, nil
+}
+
+func (fsys podmanUnshareFS) Open(name string) (fs.File, error) {
```
supakeen, line 125: I think that we're supposed to be able to call `Open()` on a directory but that would fail with this implementation right?

pkg/bootc/unsharefs.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (fsys podmanUnshareFS) exists(name string) bool {
+	_, err := fsys.run("test", "-e", fsys.fullPath(name))
+	return err == nil
+}
+
+// Ensure os.IsNotExist() works on the errors
+func (fsys podmanUnshareFS) wrapErr(op, name string, err error) error {
+	if !fsys.exists(name) {
```
supakeen, line 61: I don't have a better idea but this triggers a second subprocess for each error case.

alexlarsson: The alternative is an "exists" subprocess for *each* operation, doesn't seem better...

pkg/container/client.go
alexlarsson: It seems very unlikely that HOME is not set. I just made it return the global store in that case.

supakeen (dismissed):
The previous PR was supposed to be an approval since I had only nitpicky small things but apparently I had hit the wrong checkbox.

pkg/bootc/unsharefs.go
alexlarsson: I fixed this

supakeen (dismissed):
Thank you for fixing my nits however unlikely they are to be triggered.

supakeen (approved):

pkg/container/client.go
```diff
@@ -113,7 +113,41 @@ type Client struct {
...
+func getXdgDataDir() string {
+	dataHome := os.Getenv("XDG_DATA_HOME")
+	if dataHome == "" {
+		home := os.Getenv("HOME")
+		if home != "" {
+			dataHome = filepath.Join(home, ".local", "share")
+		}
+	}
+	return dataHome
+}
```
avitova, line 129:
```golang
func getXdgDataDir() string {
	if dataHome := os.Getenv("XDG_DATA_HOME"); dataHome != "" {
		return dataHome
	}
	if home := os.Getenv("HOME"); home != "" {
		return filepath.Join(home, ".local", "share")
	}
	return ""
}
```
Feels a bit more readable to me, but I like early reaturns overall, and it's probably a question of preference.

avitova (approved):
Just a preference/readability thing, but I don't want to block this, let's get this in.

pkg/container/client.go
alexlarsson: I agree, but lets not block this.

achilleas-k (approved):
Great work.  Thank you.

LGTM

PR #2618, opened 2026-08-25
many: support `kernel.version` customization

supakeen:
Support optionally pinning the kernel version through the kernel customization. This can be useful when debugging with older kernel packages or when dealing with certain NVIDIA drivers.

---

Will need https://github.com/osbuild/blueprint/pull/55 to land first so draft.

### DIFF
--- a/cmd/check-host-config/check/kernel.go
+++ b/cmd/check-host-config/check/kernel.go
@@ -34,6 +34,19 @@ func kernelCheck(meta *Metadata, config *buildconfig.BuildConfig) error {
 		log.Printf("Kernel name check passed: %s is installed\n", expected.Name)
 	}
 
+	if expected.Version != "" {
+		stdout, _, _, err := ExecString("rpm", "-q", "--queryformat", "%{VERSION}-%{RELEASE}", expected.Name)
+		if err != nil {
+			return Fail("failed to query kernel version:", err)
+		}
+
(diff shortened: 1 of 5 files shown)


cmd/check-host-config/check/kernel.go
```diff
@@ -34,6 +34,19 @@ func kernelCheck(meta *Metadata, config *buildconfig.BuildConfig) error {
...
+		if err != nil {
+			return Fail("failed to query kernel version:", err)
+		}
+
+		if !strings.HasPrefix(stdout, expected.Version) {
+			return Fail("kernel version mismatch: expected", expected.Version, "got", stdout)
+		}
+
+		log.Printf("Kernel version check passed: %s matches %s\n", stdout, expected.Version)
+	}
```
achilleas-k, line 48: Thinking about how to test this:
We could have a test config with a custom repo pointing to an older snapshot with an older kernel and pin it in the config.  We could do it with an older distro version, something with long term support (RHEL 9.6 probably), so we don't need to change it any time soon.

supakeen: Yea this sounds like something that should work. I was going to initially maybe not provide a test config for this yet on the other hand if it's straightforward to test maybe we should.

achilleas-k (approved):
Let's follow up with a test config.

PR #2621, opened 2026-08-26
conditions: add `not_arch`

supakeen:
Support a `not_arch` condition similar to `not_distro`, this is useful for this [1] PR to exclude firmware packages only on `s390x`.

[1]: https://github.com/osbuild/image-builder/pull/2620


avitova (commented):
Although I do not like the Eval function, this addition looks good.
Why are we using the `match = match && ...`, when we can just return `false` as soon as `match == false`?

avitova (approved):

achilleas-k (approved):

PR #2627, opened 2026-08-26
many: default extension changes

supakeen:
Change the default filenames for the container and vagrant image types. This changes what `image-builder` sees as the extension and thus renames files differently.

Previously a `libvirt-vagrant` image would be called something like `centos-10-x86_64.box` and with this change it would be called `centos-10-x86_64.vagrant-libvirt.box` which is the actually correct name.

We need these names changed otherwise `pungi` can't discern between different productmd types.

---

This is realistically speaking a breaking change for users who rely on the filenames being a certain way for the `container`, `container-minimal`, `vagrant-libvirt`, and `vagrant-virtualbox` image types. I realize this. Practically, we have no alternative or at least; no viable alternative and need to do this to unbreak tooling.

---

See also: https://forge.fedoraproject.org/pungi/pungi/issues/1923


thozza (approved):
LGTM.

avitova (approved):

Issue #2566, opened 2026-07-30
Deployment variant configurations for bootc containers (HMS-11199)


supakeen:
Currently `image-builder` sources configuration from `/usr/lib/image-builder/bootc`. Specifically it will read `disk.yaml` (for partition table information), and/or `iso.yaml` (for ISO configuration).

These files partially define *how* `image-builder` will convert a given bootable container into an artifact.

There is a wish from (amongst others) the CoreOS team to be able to select how things are converted, not based on a different input container, but based on a runtime option. As an example for `s390x` secure execution the partition table needs to include verity partitions but we can imagine other usecases such as a `btrfs` or `ext4` variant for Fedora images. These variants would be user-selectable.

The idea is that we consider any files found in `/usr/lib/image-builder/bootc` the 'default' variant. By that we mean 'no variant selected'.

Variants can be defined in `/usr/lib/image-builder/bootc/variant.d/` each one is a directory and the variant name is the directory name, for example: `/usr/lib/image-builder/bootc/variant.d/secure-execution` containing a `disk.yaml` with specific partition layout for `secure-execution`. The person building the bootable container can decide which variants are included where (and on which architectures) through the use of `$TARGETARCH` or any logic they want to use.

Another example of this would be to have a `rpi3` and a `rpi4+` variant where the former is MBR-partitioned and the latter uses GPT-partitioning.

1. `image-builder (build|manifest)` would gain a new command line option tentatively named `--bootc-variant` which takes a string and modifies the path we look at for the configuration. Handling errors if it doesn't exist.
2. `image-builder bootc inspect` would also get a `--variant` argument to select the variant to inspect for.
3. `image-builder bootc variants` would list the available variants (could live elsewhere too?).

It's up to the publishers of the containers to make people aware of variant usage. A full end-to-end example would be:

```
$ image-builder bootc variants --ref quay.io/fedora/fedora-bootc:latest
btrfs
plain
$ image-builder build --bootc-ref quay.io/fedora/fedora-bootc:latest --bootc-variant plain
# ...
```

cgwalters:
One overall tension is that I think bigger picture we want to support going from container -> disk image without necessarily making that a "bootc" image.

If we're extending things here, I would actually argue for `/usr/lib/image-builder/disk.yaml` and `/usr/lib/image-builder/variant.d/`. As far as I know, bootc is not involved really in any way in this.

[...]

PR #2636, opened 2026-09-01
many: `bootc` deployment variants

supakeen:
Implementation of #2566; I've left the path to be under `/usr/lib/image-builder/bootc` as I don't want to change the path *again* and we don't have a use case for package based trees yet.

Commits should be self explanatory see the issue for the reasoning.


pkg/bib/osinfo/osinfo.go
```diff
@@ -203,24 +204,36 @@ type diskYAML struct {
...
-		f, err := fsys.Open(p)
-		if err != nil {
-			if os.IsNotExist(err) {
-				continue
+
+		// When a variant is selected, try the variant path first,
+		// then fall back to the default path.
+		paths := []string{path.Join(prefixPath, "disk.yaml")}
+		if variant != "" {
+			paths = []string{
```
lucasgarfield: **[bug]** When a variant is selected, `readDiskYaml` (and the same pattern in `readISOYaml` at line 261) searches `[prefix/variant.d/<name>/disk.yaml, prefix/disk.yaml]` inside each `searchPaths` entry before moving on. That means a default file in `usr/lib/image-builder/bootc` wins over a variant file that only exists under `usr/lib/bootc-image-builder/variant.d/<name>/`. `ListVariants` unions names from every prefix, and `variantExists` returns true if the directory exists in any prefix, so `bootc variants` can list a name that `--bootc-variant` then silently does not apply. Example: default `usr/lib/image-builder/bootc/disk.yaml` plus variant-only `usr/lib/bootc-image-builder/variant.d/btrfs/disk.yaml` loads the default table for `--bootc-variant btrfs`.

**Suggestion:** Search every prefix for the variant file first, then fall back to the default file across prefixes. Extract a shared helper used by both `readDiskYaml` and `readISOYaml`, and add a test with default yaml in the first search path and variant yaml only in the second.

cmd/image-builder/main.go
```diff
@@ -158,6 +162,46 @@ func cmdBootcInspect(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func cmdBootcVariants(cmd *cobra.Command, args []string) error {
```
lucasgarfield: **[suggestion]** The new CLI surfaces are effectively untested. `cmdBootcVariants` calls `bootc.ResolveBootcVariants` directly (no analog of `bootcResolveInfo`), so format handling, missing `--ref`, and error paths cannot be unit-tested the way `bootc inspect` is. Existing inspect tests were only signature-updated and never assert that `--variant` is forwarded (see `cmd/image-builder/bootc_test.go:25`). The new `cannot use --bootc-variant without --bootc-ref` check in `getImage` is also untested, unlike the similar `--distro`/`--bootc-ref` guard.

**Suggestion:** Add a `bootcResolveVariants` var and mock, covering text/json/yaml/unsupported format for `bootc variants`. Assert inspect mocks receive the `--variant` value (including empty). Add a `manifest`/`build` test that `--bootc-variant` without `--bootc-ref` returns the new error.

doc/01-usage.md
```diff
@@ -329,6 +329,36 @@ $ image-builder bootc inspect --ref quay.io/centos-bootc/centos:stream10 --forma
...
+
+```console
+$ image-builder bootc variants --ref quay.io/fedora/fedora-bootc:latest
+btrfs
+plain
+```
+
+The output format can be changed with `--format`. Available formats are `text` (default), `json`, and `yaml`.
+
+#### `bootc-variant`
```
lucasgarfield: **[suggestion]** `--bootc-variant` is a `build`/`manifest` flag, but it is documented under `## image-builder bootc` / `### variants` rather than with the other `--bootc-*` flags under `## image-builder build` / `### bootc` (around the `--bootc-ref` / `--bootc-build-ref` / `--bootc-default-fs` section). Readers of the build docs will not see how to select a variant.

**Suggestion:** Document `--bootc-variant` next to the other `--bootc-*` build flags, and keep a short pointer from `bootc variants`.

pkg/bib/osinfo/osinfo.go
```diff
@@ -203,24 +204,36 @@ type diskYAML struct {
...
+func readDiskYaml(fsys fs.FS, variant string) (*diskYAML, error) {
 	for _, prefixPath := range searchPaths {
 		var disk diskYAML
-		p := path.Join(prefixPath, "disk.yaml")
-		f, err := fsys.Open(p)
-		if err != nil {
-			if os.IsNotExist(err) {
-				continue
+
+		// When a variant is selected, try the variant path first,
```
lucasgarfield: **[suggestion]** The comment restates the following `paths` assignment ("try the variant path first, then fall back to the default path") without explaining why fallback is per-file or how it interacts with `searchPaths`. `readISOYaml` has the same logic and no comment, so the two helpers have drifted.

**Suggestion:** Drop the comment, or replace it with a one-liner on the intended global order (variant files in all prefixes, then defaults) once that order is fixed.

pkg/bib/osinfo/osinfo.go
```diff
@@ -203,24 +204,36 @@ type diskYAML struct {
 	PartitionTable     *disk.PartitionTable        `json:"partition_table" yaml:"partition_table"`
 }
 
-func readDiskYaml(fsys fs.FS) (*diskYAML, error) {
+func readDiskYaml(fsys fs.FS, variant string) (*diskYAML, error) {
```
lucasgarfield: Right now a default `disk.yaml` in the `image-builder` prefix takes precedence over a variant `disk.yaml` in the `bootc-image-builder` prefix. This feels like a bug, since it means that `bootc variants` can list name(s) that `--bootc-variant` won't actually end up using.

e.g. imagine you have 

/usr/lib/image-builder/bootc/disk.yaml
and
/usr/lib/bootc-image-builder/variant.d/btrfs/disk.yaml

It would return as soon as it found /usr/lib/image-builder/bootc/disk.yaml, ignoring the btrfs variant.

Should be an easy fix – just invert the nesting so we loop over [variant, default] on the outside and searchPaths on the inside.

pkg/bib/osinfo/osinfo.go
```diff
@@ -240,24 +253,34 @@ type isoYAML struct {
 	} `json:"grub2" yaml:"grub2"`
 }
 
-func readISOYaml(fsys fs.FS) (*isoYAML, error) {
+func readISOYaml(fsys fs.FS, variant string) (*isoYAML, error) {
```
lucasgarfield: Same issue as readDiskYaml above.

lucasgarfield (changes requested):
Pretty easy for me to follow so I suppose that means it's pretty nice. 😅 

Only one real concern.

I could imagine a bit more test coverage but am not yet up to speed on the repo's conventions so I'm not asking for extra coverage - I'd be curious for you to comment on it, though, just for my own information.

pkg/bib/osinfo/osinfo.go
supakeen: I don't agree with the solution. What you describe is indeed the behavior but I think mixing two different prefixes and trying to merge them could maybe be wrong-er here. That said: we document that the new prefix takes precedence over the old one [1] so it isn't super weird that if a `disk.yaml` exists in that prefix that it wins over anything in the compatibility prefix.

What we *should* (probably) start doing is to emit a warning when we detect a container that has files in the old prefix path. I've filed [2] for this.

As for how to solve this with the above in mind I think there are three options:

1. Leave as-is.
2. Don't use both prefixes for any variant handling *only* `/usr/lib/image-builder`.
3. Adjust `ListVariants` to not continue to the compatibility prefix when `/usr/lib/image-builder` exists.

What has your preference? For me it's the the last one and we could probably apply it to all cases where we check both prefixes. 

[1]: https://osbuild.org/docs/developer-guide/projects/image-builder/advanced/bootc/sources-of-configuration/#filesystem
[2]: https://github.com/osbuild/image-builder/issues/2647

supakeen: By the way, I initially thought that merging the variants across prefixes was nice (it even has a test!) but your comment made me reconsider 🙂.

supakeen: #2648 for the warnings. #2649 for never falling back in general.

achilleas-k: I think to push for the deprecation of the bootc-image-builder path, we should stop adding features to it.  Anyone who wants to use this new variant thing should use the image-builder paths.

So my suggestion would be:
- When listing prefixes, only look in `/usr/lib/image-builder/variants.d/`.  Drop the `bootc` even.
- When a variant is named (during manifest creation or build), go straight to `/usr/lib/image-builder/variants.d` and fail if `$variantname/` doesn't exist.  Don't fall back to anything.
- When no variant is specified, look for `disk.yaml` files in the following order and stop at the first one found:
    1. `/usr/lib/image-builder/disk.yaml`
    2. `/usr/lib/image-builder/bootc/disk.yaml`
    3. `/usr/lib/bootc-image-builder/disk.yaml`
- If 2 or 3 are used, emit a warning deprecating the paths and directing users to 1.

Let's start cleaning up some of this mess and standardising.  The variants and overrides might, in the future, generalise beyond bootc-based images, so let's not keep namespacing it with `bootc/`.

Also, in _every_ case, print the path of the override being used before it's loaded.

cmd/image-builder/cmd.go
```diff
@@ -207,6 +221,7 @@ func setupManifestCmd() (*cobra.Command, error) {
 	manifestCmd.Flags().String("ostree-parent", "", `OSTREE parent`)
 	manifestCmd.Flags().String("ostree-url", "", `OSTREE url`)
 	manifestCmd.Flags().String("bootc-ref", "", `bootc container ref`)
+	manifestCmd.Flags().String("bootc-variant", "", `deployment variant to use for the bootc container`)
```
achilleas-k, line 224: Make it plain `--variant`.  No need to namespace it under bootc.  We could be generalising this in the future.

pkg/bib/osinfo/osinfo.go
supakeen: This sounds OK, my only concern here is the removal of `bootc` in the path and the argument name. We have no current use case or requests for this in package-based land and I don't want to imply these things work there.

achilleas-k: > my only concern here is the removal of `bootc` in the path and the argument name

Yeah, fair.  That's the only one I didn't feel too strongly about and expected I might get push back.  Let's leave it there then :)

lucasgarfield: I like Achilleas' suggestions. In particular I like the idea of failing instead of falling back. I can easily imagine a situation where someone names their variant disk.yml instead of disk.yaml (or makes any kind of typo)... failing would also guard against this.

For what it's worth... from Simon's original alternatives, if I had to choose, I would also choose option 3.

lucasgarfield: No strong opinion on keeping or removing `bootc` in the path, though.

supakeen:
Could someone take a look at this one? I've rebased it due to conflicts just now.

achilleas-k (approved):
Skimmed the commits and re-reviewed the changes since last time.  Sorry I left this hanging.

LGTM!

lucasgarfield (approved):
lgtm

PR #2637, opened 2026-09-01
Enable bootc-image-builder ISO tests on aarch64 [HMS-11073]

achilleas-k:
**bootc-image-builder: update registry ref**

We use a clone of the registry container, pushed to our own ghcr.io
namespace, so that we don't hit dockerhub rate limits when running too
many tests.  The new ref has been updated with the latest amd64 and
arm64 registry containers.  The old one under bootc-image-builder only
had a copy of the amd64 container which was causing issues with the ISO
tests on aarch64.

---

**ci: run test_build_iso on aarch64**

With the registry container now having both architectures, the tests
should be ok now.

---

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1224,6 +1224,25 @@ fail:
     - init
 
 
+"bootc-image-builder/test/test_build_iso.py [aarch64]":
+  stage: test
+  extends: .terraform
+  variables:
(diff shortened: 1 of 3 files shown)


lucasgarfield (approved):
I checked and can confirm that the test is running and passing in Gitlab.

avitova (approved):
Thanks!

PR #2644, opened 2026-09-02
bootc-image-builder: boot AMI in AWS after cross-arch building [HMS-11073]

achilleas-k:
**bootc-image-builder: rename qemu-cross test case to ami-cross**

The test case was already building a raw image, which is identical to
the ami.  Explicitly building an "ami" (as opposed to "raw") takes care
of uploading to AWS.

---

**bootc-image-builder: boot AMI in AWS after cross-arch building**

Boot testing the artifact of a cross-arch build with emulation is slow
and unreliable.  For cross-arch tests, build and upload AMIs and test
them on machines that match the image architecture.

---

### DIFF
--- a/bootc-image-builder/test/test_build_cross.py
+++ b/bootc-image-builder/test/test_build_cross.py
@@ -1,7 +1,9 @@
 import platform
 
+import imgtestlib as testlib
 import pytest
 
+import testutil
 # pylint: disable=unused-import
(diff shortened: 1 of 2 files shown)


brlane-rht (approved):
Looks good!

ondrejbudai (approved):

Issue #2647, opened 2026-09-02
Handle prefixes in containers differently


supakeen:
When reading sources of configuration [1] we document that `/usr/lib/image-builder` takes precedence over `/usr/lib/bootc-image-builder` but we don't define what happens when content exists in both.

Let's start emitting warnings when the old prefix exists at all.

Separately, change things such that if the new prefix exists we should not look in the old prefix at all. Short-circuit directly and don't accidentally mix things.

[1]: https://osbuild.org/docs/developer-guide/projects/image-builder/advanced/bootc/sources-of-configuration/#filesystem

achilleas-k:
This is done. Closing. Thank you, Simon.

PR #2649, opened 2026-09-02
osinfo: never fall back if new prefix exists

supakeen:
When both the new and old prefix exist things get confusing. Instead of falling back *don't*. When the new prefix exists we don't even look at the old prefix at all.

---

See #2647


achilleas-k (approved):

brlane-rht (approved):
Looks Good!

PR #2650, opened 2026-09-03
pkg/crypt: add support for yescrypt hashed passwords

lfauster:
Detect yescript hashes via the `$y$` prefix.
Before, if a yescript hash was provided, it got re-hashed with sha-512, leading to broken login credentials.


supakeen (approved):
Might be OK to drop the comment, I think by now we've surpassed `lorax`'s initial implementation but that's such a nitpick that you don't need to address it on my behalf 🙂

lucasgarfield (approved):
Nice - this is your first pull request on the project, isn't it? Unless you sneakily did one before you officially started? 😄

PR #2651, opened 2026-09-03
cmd/image-builder: only run findDistro once

croissanne:
`findDistro` should only be run once per command, when using it as an input for `getOneImage`. Once an `imagefilter.Result` is created, the distro can be accessed through the `distro.ImageType` interface.

`getOneImage` is called for `describe`, `pkgsearch`, `manifest` and `build`.


supakeen (approved):

achilleas-k (approved):
Very nice!

PR #2655, opened 2026-09-04
gce: ensure disk images are GiB-aligned

### LINKED ISSUE #2413: gce image type: disk.raw not guaranteed to be GiB-aligned, causing GCP import failures (HMS-11022)
**Problem**

GCP requires that disk.raw files are aligned to a GiB boundary (a multiple of 1,073,741,824 bytes). The gce image type currently produces a disk.raw that may not satisfy this requirement, depending on the requested image size and filesystem layout. This causes GCP import to fail.

**Root cause**

In pkg/distro/generic/bootc_imagetype.go, genPartitionTableFsCust() uses imageSize derived from ImageTypeYAML.DefaultSize (or basept.Size) directly, without rounding up to the nearest GiB:

```
  // pkg/distro/generic/bootc_imagetype.go:918
  imageSize := t.ImageTypeYAML.DefaultSize
  if basept.Size != 0 {
      imageSize = basept.Size
  }
  return disk.NewPartitionTable(basept, fsCustomizations, imageSize, ...)

```
  The gce entry in data/distrodefs/bootc-generic/imagetypes.yaml (lines 148–152) has no alignment constraint:

```
  "gce":
    <<: *raw_image_type
    filename: "image.tar.gz"
    mime_type: "application/gzip"
    exports: ["gce"]
```

**Proposed fix**

**Option A (recommended) — align at partition table generation time**

  1. Add a GiBAlignedDisk bool field to ImageTypeYAML in pkg/distro/defs/loader.go:
  GiBAlignedDisk bool `yaml:"gib_aligned_disk"`
  2. Set it in data/distrodefs/bootc-generic/imagetypes.yaml:
  "gce":
    <<: *raw_image_type
    filename: "image.tar.gz"
    mime_type: "application/gzip"
    exports: ["gce"]
    gib_aligned_disk: true
  3. In genPartitionTableFsCust(), round imageSize up before calling NewPartitionTable:
  if t.ImageTypeYAML.GiBAlignedDisk {
      imageSize = (imageSize + datasizes.GiB - 1) &^ (datasizes.GiB - 1)
  }

**Option B — resize in the GCE pipeline**

  Add a truncate/pad stage in pkg/image/gce.go after the disk image is built and before the tar step. More invasive and happens after the fact rather than sizing correctly up front.

Why Option A

  It ensures disk.raw is sized correctly from the start — no post-processing or repackaging needed. The datasizes.GiB constant (1 << 30) is already available in   pkg/datasizes/constants.go.

supakeen:
GCP requires disk files to be aligned to a GiB boundary. Add a disk_image_gib_aligned flag to ImageTypeYAML and enable it for all GCE image type definitions.

Closes: #2413

### DIFF
--- a/data/distrodefs/bootc-generic/imagetypes.yaml
+++ b/data/distrodefs/bootc-generic/imagetypes.yaml
@@ -153,6 +153,7 @@ image_types:
     filename: "image.tar.gz"
     mime_type: "application/gzip"
     exports: ["gce"]
+    disk_image_gib_aligned: true
 
   ova:
     <<: *raw_image_type
--- a/data/distrodefs/eln-11/disk.yaml
+++ b/data/distrodefs/eln-11/disk.yaml
@@ -820,6 +820,7 @@ image_types:
     exports: ["archive"]
     bootable: true
     default_size: "20 GiB"
+    disk_image_gib_aligned: true
     required_partition_sizes: *default_required_dir_sizes
     platforms:
       - <<: *x86_64_uefi_platform
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -2765,6 +2765,7 @@ image_types:
     filename: "image.tar.gz"
     mime_type: "application/gzip"
     exports: ["archive"]
+    disk_image_gib_aligned: true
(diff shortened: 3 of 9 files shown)


avitova (approved):
I'd probably name it with something like `_required`. If I understood the reasoning correctly.

achilleas-k (approved):

PR #2657, opened 2026-09-04
bootc: explicit home directories

### LINKED ISSUE #2596: Incorrect user home directory location (HMS-11218)
When building a raw bootc-based image and adding a user via blueprint customization the home directory is being created in `/ostree/deploy/default/var/home/user`, and not in `/var/home/user`. The user itself is created and I can log in with an error about a missing home directory

The blueprint used:

```toml
[[customizations.user]]
name = "user"
password = "123456"
groups = ["wheel"]
```

Build log: [image-builder.log](https://github.com/user-attachments/files/31045528/image-builder.log)

supakeen:
When useradd runs with --root against an ostree deployment, it follows the symlink chain (`/home` -> `var/home` -> `../../var`) and puts the deployment path (/ostree/deploy/default/var/home) in /etc/passwd instead of the runtime /var/home path.

Set explicit home directories for users before passing them to the users stage instead.

Closes: #2596


achilleas-k (approved):
Sweet!

PR #2659, opened 2026-09-04
el9/10: parity bits

supakeen:
Previously produced container for CentOS 9 and 10 did not include weak dependencies. Including them pulls in a whole chain of packages (see [1]). Turning it off again seems uncontroversial so let's do that.

[1]: https://github.com/osbuild/image-builder/issues/2658

---

Also add the `install_langs` macro during build time to make sure we don't include additional languages.

---

Also *persist* the `install_langs` macro in the built image.


achilleas-k (dismissed):
LGTM

pkg/osbuild/rpm_macros_stage.go
achilleas-k, line 1: Oh wow, this stage is ancient.  How did we not have it until now?

achilleas-k (approved):
> Allow configuring RPM macros to be persisted into an image. This is *slightly* awkward since the `_install_langs` macro can now be set in two places one that only affects build time, and one that only affects the built image.
>
> We can consolidate that later.

I assume we're in a hurry with these?  Let's make sure to clean it up, it's a bit weird I agree.

pkg/osbuild/rpm_macros_stage.go
supakeen: We never actually needed it but I'm happy it exists 🙂

brlane-rht (approved):
Looks good!

PR #2660, opened 2026-09-04
Change s3_touch() to tag objects and re-enable the function on cache hits [HMS-11071]

### LINKED ISSUE #2388: Updating the modification time of images in the CI cache takes too long (HMS-11071)
When the CI build cache gets a hit, it updates the timetamp of ("touches") the objects so they don't get garbage collected while they're still relevant.  To do this though, we need to copy an S3 object over itself, which can take a while for large files (e.g. 20 GiB images).

The images are needed by osbuild's CI to run image info on images built by older versions and verify that the info for an image hasn't changed when the same manifest is rebuilt into another image.  If we stop touching these objects, osbuild's CI will fail.

One idea is to stop saving images in the cache altogether and instead store the image info alongside the build information and the manifest.  This would save on storage and transfer costs and might even speed up osbuild's CI a bit since it wont have to download the image itself and generate its image info.

This conflicts with another idea / plan we had with @supakeen however for reusing images to boot test them when they've already been built but not successfully boot tested.

CC @thozza for thoughts on the osbuild tests.

achilleas-k:
Instead of adding metadata, which requires copying a file over itself and can take time, tag an object with a DeleteAfter date that the cloud-cleaner reads to determine if an object should be deleted.  Unlike the previous method of modifying objects, this does not change the mtime.  However, the action is a single (fast) API call.

See also https://github.com/osbuild/cloud-cleaner/pull/59

### DIFF
--- a/test/scripts/imgtestlib/cache.py
+++ b/test/scripts/imgtestlib/cache.py
@@ -1,7 +1,7 @@
 import json
 import os
 import subprocess as sp
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import List, Optional
 
 from .build import (gen_build_name, get_manifest_id, read_build_info,
@@ -10,8 +10,11 @@ from .gitlab import log_section
 from .run import runcmd, runcmd_nc
 from .testenv import get_ci_runner_for, get_osbuild_commit
 
-S3_BUCKET = "s3://" + os.environ.get("AWS_BUCKET", "images-ci-cache")
+S3_BUCKET_NAME = os.environ.get("AWS_BUCKET", "images-ci-cache")
+S3_BUCKET = "s3://" + S3_BUCKET_NAME
 S3_PREFIX = "images/builds"
+
+DELETE_AFTER_TD = timedelta(days=14)
 
 
 # pylint: disable=too-many-arguments,too-many-positional-arguments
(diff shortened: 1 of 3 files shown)


test/scripts/imgtestlib/cache.py
```diff
@@ -109,16 +112,21 @@ def gen_build_info_s3_dir_path(distro=None, arch=None, manifest_id=None, osbuild
...
+    Update the DeleteAfter tag of a path in S3 to the current time + 14 days (DELETE_AFTER_TD). The tag is used by the
+    cloud cleaner to garbage-collect old images. Only the info.json of a prefix containing an image needs to be tagged.
     """
-    # NOTE: we can make this async - updating the objects can take a few seconds and can be done in parallel
-    s3url = gen_build_info_s3_dir_path(distro, arch, manifest_id, osbuild_ref, runner_distro)
+    s3path = os.path.join(S3_PREFIX,
+                          gen_build_info_dir_path_prefix(distro, arch, manifest_id, osbuild_ref, runner_distro),
+                          "info.json")
+    s3url = os.path.join(S3_BUCKET, s3path)
     # the exact key and value don't matter, but let's add the current datetime to make it a bit more meaningful
```
supakeen: They do matter now 😉.

test/scripts/imgtestlib/core.py
```diff
@@ -258,11 +258,9 @@ def filter_builds(manifests, distro=None, arch=None, skip_ostree_pull=True):
         if check_for_build(manifest_fname, build_request, data["manifest"], build_info_dir, errors):
             build_requests.append(build_request)
         else:
-            # The specific build configuration exists in the cache and wont be rebuilt. Update the file timestamps to
+            # The specific build configuration exists in the cache and wont be rebuilt. Update the DeleteAfter tag to
             # keep them fresh in the cache.
-            # touch_s3(distro, arch, manifest_id)
-            # NOTE(2026-06-18): Disabling this temporarily since it slows down the filtering process by a lot
-            print("WARNING: timestamp updating has been temporarily disabled", file=sys.stderr)
+            touch_s3(distro, arch, manifest_id)
```
supakeen, line 263: If `touch_s3` fails here it aborts everything and no other manifests get touched. That might be fine or not?

achilleas-k: True.  But that was always the case (and it never caused issues).  I could make it fail with a warning if think it's a good idea.

test/scripts/imgtestlib/cache.py
achilleas-k: Good catch.  Fixed! :)

test/scripts/imgtestlib/core.py
achilleas-k: I think it's fine for a whole CI job to fail if any step fails, now that I think about it.  I'd rather not discover that this isn't working after things start getting garbage collected mysteriously.

supakeen (approved):

brlane-rht (approved):
Excellent idea!

PR #2665, opened 2026-09-06
pkg/osbuild: import `tree-delta`, `dd`, `os-release` stages

supakeen:
Import the `org.osbuild.tree-delta`, `.dd`, and `.os-release` stages which were added to `osbuild` over the recent releases.


achilleas-k (commented):
Looks good but has the same issue as https://github.com/osbuild/image-builder/pull/2664#discussion_r3948088105.  Maybe we can merge the two PRs since they're both quite straightforward and share the same general purpose.

avitova (approved):
Yeah, this makes sense to put into one PR, I think. Thank you!

achilleas-k (approved):
Ship it!

PR #2667, opened 2026-09-06
pkg/manifest: erofs pipeline generator

supakeen:
Introduce an erofs pipeline generator that takes an input and turns it into an erofs image.

### DIFF
--- a/pkg/manifest/erofs.go
+++ b/pkg/manifest/erofs.go
@@ -1,0 +1,53 @@
+package manifest
+
+import (
+	"github.com/osbuild/image-builder/pkg/artifact"
+	"github.com/osbuild/image-builder/pkg/osbuild"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/manifest/erofs.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (p *Erofs) serialize() (osbuild.Pipeline, error) {
+	pipeline, err := p.Base.serialize()
+	if err != nil {
+		return osbuild.Pipeline{}, err
+	}
+
+	erofsOptions := osbuild.ErofsStageOptions{
+		Filename: p.filename,
+	}
```
supakeen, line 40: Really simple for now, plumbing through more erofs options can be done later :)

achilleas-k (approved):

PR #2671, opened 2026-09-07
Build cache based on export pipeline (HMS-11233)

avitova:
Use the image type's export pipeline when computing/validating cache entries instead of always using the last pipeline in the manifest.


test/scripts/imgtestlib/core.py
```diff
@@ -230,17 +255,22 @@ def filter_builds(manifests, distro=None, arch=None, skip_ostree_pull=True):
...
 
     errors: list[str] = []
-    for manifest_fname, data in manifests.items():
-        manifest_id = data["id"]
-        data = data.get("data")
-        build_request = data["build-request"]
+    for manifest_fname, entry in manifests.items():
+        manifest_id = entry["id"]
+        manifest_data = entry["data"]
+        build_request = manifest_data["build-request"]
```
achilleas-k, line 254: Good change.  That `data` overwrite was a nasty bug waiting to happen.

test/scripts/imgtestlib/build.py
```diff
@@ -155,11 +174,49 @@ def write_build_info(build_path: str, data: Dict):
...
+    manifest_body = osbuild_manifest_body(manifest_data)
+    return {pipeline["name"] for pipeline in manifest_body.get("pipelines", [])}
+
+
+def resolve_export_pipeline(manifest_data, image_type: Optional[str] = None) -> Optional[str]:
+    build_request = manifest_data.get("build-request")
+    if isinstance(build_request, dict):
+        export_pipeline = build_request.get("export-pipeline")
+        if export_pipeline:
+            return export_pipeline
```
achilleas-k: This can be collapsed to:
```py
export_pipeline = manifest_data.get("build-request", {}).get("export-pipeline")
```

test/scripts/imgtestlib/core.py
```diff
@@ -194,6 +200,25 @@ def check_for_build(manifest_fname, build_request, manifest_data, build_info_dir
     else:
         print("  No PR/branch info available")
 
+    req_img_type = build_request["image-type"]
+    cached_img_type = dl_config.get("image-type")
+    if cached_img_type != req_img_type:
+        print(f"  Cached image-type {cached_img_type!r} does not match requested {req_img_type!r}")
+        print("  Adding config to build pipeline.")
+        return True
```
achilleas-k: This is unnecessary I think.  The image type name shouldn't matter and if it changes without affecting the manifest itself (say, if we rename an image type or two image types are identical), it shouldn't do a build.

achilleas-k (dismissed):
Nice work, thank you.

I'm going to approve this since it appears to be working but there's something missing here.  I'm looking at the files in the cache from the S3 bucket and I'm seeing a lot of `info.json` files with `"export-pipeline": null`.  I suppose this means that the `resolve_export_pipeline()` call in `upload_results()` is reaching the final fallback state where it returns `None`.  It's not causing problems because the `check_for_build()` function only looks at export pipeline names if it's set in the build info.  I'd like to see if we could (in a follow-up) change things to always have the export pipeline names when we need them and always include them in the build info.  But for now it's fine as it is.

And a small note for the future: @supakeen started work towards having multiple exports.  We might have to generalise this solution at some point.  It might be as simple as putting all the export pipelines in the build request and combining them to form the manifest ID (e.g. the ID of a manifest with multiple exports could be a concatenation or hash of the pipeline IDs), we'll see.

LGTM!

achilleas-k (approved):

supakeen (dismissed):

achilleas-k:
> RHOS is getting really annoying

#2689

achilleas-k:
Another rebase for the KVM runners.

achilleas-k:
I removed it from the queue.  I noticed that it was rebuilding everything on the merge queue even though it shouldn't.  I think the export-pipeline check isn't totally correct and it cause every image to be added to the config to be built.

achilleas-k:
FYI, if you have AWS credentials with access to the bucket (under the 438669297788 account), you can verify the behaviour quite easily with something like
```
./test/scripts/generate-build-config --distro centos-10 --arch x86_64 config.yaml
```

If I run this on the branch, I see messages like this
```
  Manifest centos_10-x86_64-container-empty.json was successfully built in commit 72e60262a714f74943defb24b5cb1b15612a6521
  https://github.com/osbuild/images/commit/72e60262a714f74943defb24b5cb1b15612a6521
  This commit was on a merge queue: gh-readonly-queue/main/pr-2671-5809a0e36f75edda6010490281b5856095241238
  Cached build info missing export-pipeline
  Adding config to build pipeline.
```

Which tells me that the image was built in the pipeline for the merge queue for this PR, but the export pipeline wasn't saved to the build info when it was done.

I thought I had verified this already but maybe I missed something.

achilleas-k (changes requested):
I'm requesting changes here because if we merge it I think we're going to have caching issues.  Something's not 100% right with the logic here I think.

test/scripts/imgtestlib/build.py
```diff
@@ -156,11 +175,47 @@ def write_build_info(build_path: str, data: Dict):
...
+    if export_pipeline:
+        return export_pipeline
+    if image_type == "gce":
+        pipeline_names = manifest_pipeline_names(manifest_data)
+        for name in ("archive", "gce"):
+            if name in pipeline_names:
+                return name
+    if image_type:
+        return EXPORT_PIPELINE_BY_IMAGE_TYPE.get(image_type)
+    return None
```
achilleas-k: I think the source of the problem is this.  This gets called by the `upload_results()` function and if we return null in some cases, then we end up writing that to the `info.json` for a build.  Then if a build request includes the export pipeline, it will never match and the image will always be built.

achilleas-k: My suggestion would be to:
- Always have export pipelines as part of the image definition, so that the build request from `gen-manifests` always includes an `export-pipeline` field.
- Never try to guess the pipeline name but always take it from the build request and write it to info.json directly from the build request.

achilleas-k (approved):
Nice work.  I've confirmed that this works as expected now.
Needing to pass the export-pipeline around isn't the prettiest solution, but it works.

Thank you.  LGTM!

PR #2673, opened 2026-09-08
test: enable fedora boot.iso (HMS-11283)

supakeen:
An old(er) bug prevented us from testing Fedora boot.iso-style media. Let's see if the bug still applies and fix it if it does.

### DIFF
--- a/test/scripts/imgtestlib/boot.py
+++ b/test/scripts/imgtestlib/boot.py
@@ -599,7 +599,8 @@ def boot_image(search_path, build_config_path, keep_booted=False):
         boot_qemu(arch, image_path, build_config_path, keep_booted=keep_booted)
     elif image_type in ("image-installer", "minimal-installer"):
         boot_qemu_iso(arch, image_path, build_config_path)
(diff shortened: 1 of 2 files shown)


Schutzfile
```diff
@@ -1,12 +1,12 @@
 {
   "common": {
-    "rngseed": 2026070700,
+    "rngseed": 2026090800,
```
achilleas-k: This shouldn't be necessary.  The change in `can_boot_test()` should cause the relevant images to be rebuilt so they can be boot tested now that it's supported.

achilleas-k: Here's the relevant code that decides what gets built in each run: https://github.com/osbuild/image-builder/blob/ecc5a324adfafe338af9bfc70351dbdb94d58099/test/scripts/imgtestlib/core.py#L197-L208

supakeen: Oh, this isn't what happened. I amended my commit with a seed bump after it didn't test the installers but maybe I looked wrong.

supakeen: Reverted the seed change.

achilleas-k (approved):
Very nice!

PR #2674, opened 2026-09-08
rpmmd: Honor ignore_ssl when loading repositories

ondrejbudai:
I tried --force-repo-dir and I needed to ignore SSL, but it didn't work. Turns out we silently drop the field when loading repos. This commit fixes that.


supakeen (approved):
Nice find. Thanks. Did you by any chance check if this was the case anywhere else?

pkg/rpmmd/repository.go
```diff
@@ -22,7 +22,7 @@ type repository struct {
 	GPGKey         string   `json:"gpgkey,omitempty"`
 	GPGKeys        []string `json:"gpgkeys,omitempty"`
 	CheckGPG       bool     `json:"check_gpg,omitempty"`
-	IgnoreSSL      bool     `json:"ignore_ssl,omitempty"`
+	IgnoreSSL      *bool    `json:"ignore_ssl,omitempty"`
```
achilleas-k, line 25: Since the default (`null`) case is `false`, do we need this?

achilleas-k (approved):

PR #2675, opened 2026-09-09
Merge pkg/distro/generic/ into pkg/distro/defs/

achilleas-k:
This is a simple code-moving PR in preparation for bigger changes in the way we load and handle image definitions.

---

This commit moves everything from pkg/distro/generic/ into pkg/distro/defs/.  Import paths have been adjusted throughout the project.

The genManifest() function in bootc_test.go was rewritten to not depend on manifestgen since that creates an import cycle.

The purpose of this change is to isolate any handling of image definition loading and mentions of YAML to one package, pkg/distro/defs/.  In the end, loading functionality and all types initialised from the distro definitions will be defined in pkg/distro/defs/.  YAML-specific types will be private and convert to internal types for handling.  As much as possible, the properties of these types will be immutable.

The GitLab CI config has been updated to test the new paths when explicitly tested.


supakeen (dismissed):
Nice cleanup.

supakeen (approved):

PR #2676, opened 2026-09-10
many: cross-arch bootstrapping from packages (HMS-11277)

supakeen:
Historically we've used pre-built containers for cross-arch bootstrapping. This PR allows for cross-arch bootstrapping through RPMs instead. When a distribution definitions defines a set of packages for a given arch it is preferred over the container.

Initially this only swaps over the riscv64 bootstrap as they're the most problematic and a good area to test this. If you want to test properly it's best to rebase this PR on top of #2614a and cross-building an `everything-network-installer` for example.


achilleas-k (approved):
Thanks!  LGTM

PR #2677, opened 2026-09-10
cmd/image-builder: support templating output directory and artifact names (HMS-11286)

supakeen:
Allow using Go templates (and use them internally) on the `output-name` and `output-dir` fields. This will make life easier in the future when we have multiple artifacts and generally make things a bit easier for users who want nested outputs or such.


cmd/image-builder/main.go
```diff
@@ -379,7 +435,8 @@ func getImage(cmd *cobra.Command, args []string) (*imagefilter.Result, error) {
 		}
 	}
 	if len(img.ImgType.Exports()) > 1 {
-		return nil, fmt.Errorf("image %q has multiple exports: this is current unsupport: please report this as a bug", basenameFor(img, ""))
+		name, _ := basenameFor(img, "")
+		return nil, fmt.Errorf("image %q has multiple exports: this is current unsupport: please report this as a bug", name)
```
achilleas-k: "currently unsupported"

cmd/image-builder/main.go
```diff
@@ -65,9 +66,59 @@ func defaultCacheDir() string {
 	return cacheDirForUid(os.Getuid())
 }
 
+type outputTmplData struct {
+	Distro       string
+	Arch         string
+	ImageType    string
+	Stem         string
```
achilleas-k, line 73: Not sure about the name "stem".  Is that the name of the output artifact?  I think we need to find a more obvious name for it.

achilleas-k (changes requested):
Nice work.  Looks good overall, but I'm not sure about the "stem".

Also, error message language needs a fix.

cmd/image-builder/main.go
supakeen: It's the stem of the output filename that's defined in the definitions which is what we previously used for the default name (hence it comes back in the template). Initially I had tried `Filename`, `ImageName`, but didn't like those either.

cmd/image-builder/main.go
supakeen: Fixed.

achilleas-k:
> maybe we _don't_ want to do that since we only want what's relevant for output filenames in here?

I think so.  In my mind, distro, arch, image-type, blueprint name, and export pipeline name are enough, but I'm sure people will want to get more creative at some point.  Maybe we can stick to those first and then expand as needed if we ever get requests for more.

cmd/image-builder/main.go
achilleas-k: `Filename` would make sense but then that gets confusing because you are, in fact, _setting the filename_ with these components.  `ArtifactName` maybe?

supakeen: We had a chat about this and decided to go with a nested struct that allows for easier expansion in the future.

achilleas-k (approved):
Ship it!

PR #2679, opened 2026-09-10
Add rhel10 & fedora image-installer for s390x & ppc64 architectures (HMS-10747)

lfauster:
Adds image-installer to the fedora and rhel10 distro definitions.
Some packages needed to be moved around a bit, as s390x does not use grub2 but zipl as bootloader. Also, both ppc64 and s390x do not use EFI.
I also used the opportunity to de-duplicate some redundant grub2/efi related packages.

The s390x image installer uncovered what I think is a bug in anaconda:
Anaconda installs the bootloader (in s390x's case zipl) before it creates the BLS entries, which leads to installation failure of zipl, as it depends on them.

The s390x image installer was tested on a beaker s390x machine with a patched anaconda [1].
The ppc64 images were built on a beaker ppc64 machine, and the installation was - for reasons - tested with qemu emulation on a x86_64 machine; .

[1] https://github.com/lfauster/anaconda/tree/zipl-fix-s390x-liveimg


supakeen (approved):
Thank you; this looks good! Feel free to undraft, it can go in even with the potential Anaconda bug still existing.

I also noticed in the commits here that checksums changed where I feel they /shouldn't/ have changed. I've filed https://github.com/osbuild/image-builder/issues/2680 to fix that up.

---

As a bonus, it might be nice to also include these in `eln-11` since that will become `rhel-11` in the future.

data/distrodefs/fedora/imagetypes.yaml
```diff
@@ -72,11 +72,9 @@
       - "dejavu-sans-fonts"
       - "dejavu-sans-mono-fonts"
       - "device-mapper-persistent-data"
-      - "dmidecode"
```
achilleas-k, line 75: The checksum changes here (in the first commit) appear to be from `dmidecode` being duplicated in the original manifest and it's now only appearing once.  Real manifests are unaffected.

data/distrodefs/rhel-10/imagetypes.yaml
```diff
@@ -416,23 +416,28 @@
...
     conditions:
+      "grub common packages for the anaconda boot pkgset":
+        when:
+          # grub2 is used on all architectures, except s390x
+          not_arch: "s390x"
+        append:
+          include:
+            - "grub2-tools"
+            - "grub2-tools-extra"
+            - "grub2-tools-minimal"
```
achilleas-k, line 427: Same issue as the fedora changes.  Duplicate packages are reduced (`dmidecode`, `efibootmgr`, `grub2-tools`, `grub2-tools-extra`, `grub2-tools-minimal`) but real manifests are unaffected.

achilleas-k (approved):
Great stuff.  Thank you.  LGTM!

PR #2684, opened 2026-09-14
README: swap image-builder - images READMEs [HMS-10586]

achilleas-k:
Let's keep the images README for a bit.  I'd like to turn it into a developer document.


avitova (approved):
Makes sense

supakeen (approved):

PR #2685, opened 2026-09-14
distro/defs: drop the ImageTypeYAML property from defs.imageType

achilleas-k:
The newImageTypeFrom() function now initialises the imageType with fully static properties and doesn't rely on the YAML backend.

Currently, this means that more work is done up front that may not be necessary.  Image types are initialised and dynamic properties are resolved on startup rather than on-demand.  However, future plans are for image types to be initialised dynamically in their entirety.  This change makes the creation and usage of the defs.imageType type simpler.

The ImageTypeYAML is now not visible outside of the defs package where it serves the loader.

The bootcImageType still relies on embedding ImageTypeYAML for some methods.  I'm planning on doing the same to the bootcImageType in a follow-up.

The first commit was originally #2675.


supakeen (approved):
This ... is extremely clean for so many changes even if mostly mechanical.

PR #2686, opened 2026-09-15
fedora: use LVM GPT UUID for server

supakeen:
Fedora Server erroneously sets a UAPI DPS specification root UUID on its partition tables. It should be the LVM UUID instead. Not doing this confuses some LVM tooling (or so I've been told).


avitova (approved):
Nice, thank you:)

PR #2687, opened 2026-09-15
bootc: aarch64 bootc-generic-iso

supakeen:
The `bootc-generic-iso` hardcoded grub2's i386-pc platform regardless of the architecture being built on, or for. Instead default to UEFI only and switch to hybrid only when we're on x86.


brlane-rht (approved):
Looks good!

PR #2688, opened 2026-09-15
rhsm: support cross major version builds

lucasgarfield:
Wildcard the RHEL major path segment (/rhel8, /rhel9, /rhel10) when
matching subscription baseurls, so a host subscribed to rhelX can
download content for rhelY. Mirrors the matching in osbuild's rhsm.py.

See corresponding osbuild PR: https://github.com/osbuild/osbuild/pull/2565


lucasgarfield:
@supakeen we discussed adding documentation but when I went to write it I couldn't decide what to add... do we want to add specific instructions for using Satellite? Including cross major builds? But whatever we would write for cross major builds on Satellite also applies to the CDN, so perhaps just a section for cross major builds in general? But given that it "just works" now (or it should...) do we need to really document it? I'm going to sleep on it, but would be curious to hear what you think!

lucasgarfield:
Draft for now. I want to test against a real Satellite before marking ready for review.

lucasgarfield:
I have verified that building RHEL 9.9 on a RHEL 10.2 host using a Satellite with RHEL 9 and 10 repos works.

image
image
image

supakeen (approved):
From our chats the diff does what you've told us it should do and sounds like a logical thing to do. Obviously there's the minor possibility that a build host is using different subscription certificates for different majors but that can be added later if we must.

My SUPER nitpicky comment would be that the commit message says 'so that a host subscribed to rhelX can download content for rhelY' that isn't what this does (though it's what it wants to achieve). The change is basically 'use the first RHEL subscription certificate you find no matter the major'.

PR #2691, opened 2026-09-16
doc: many many bits

supakeen:
Document a whole bunch of stuff that was previously undocumented and is probably really nice for users to know.


doc/01-usage.md
```diff
@@ -280,6 +394,105 @@ $ image-builder describe --arch aarch64 minimal-raw-xz
...
+The `upload` command uploads a previously built image to a cloud provider. The `--to` flag selects the target cloud. When using `build`, upload flags can be passed directly and the upload happens automatically after the build completes.
+
+```console
+$ image-builder upload --to aws --aws-region us-east-1 --aws-bucket my-bucket --aws-ami-name my-image fedora-43-server-ami-x86_64/fedora-43-server-ami-x86_64.raw
+# ...
+```
+
+The architecture is detected from the image filename when possible. Use `--arch` to override this. The output format can be changed with `--format` (yaml, json).
+
+### `--to aws`
```
schuellerf, line 407: Usually the HTML anchors are auto-derived from the title… let's see if this causes problems :-D 
(just a note - I'll keep an eye open if this works…)

schuellerf (dismissed):
This documents so many very cool features - thank you!

doc/01-usage.md
```diff
@@ -209,6 +281,48 @@ $ sudo image-builder build --bootc-ref quay.io/fedora/fedora-bootc:rawhide --boo
...
+Some values in a build are derived randomly (e.g. partition UUIDs). The `--seed` flag pins the random number generator to a fixed integer value, making builds more reproducible.
+
+```console
+$ sudo image-builder build --seed 42 --distro fedora-43 server-qcow2
+# ...
+```
+
+### Cloud upload
+
+The `build` command supports all [`upload` flags](#image-builder-upload) directly, allowing you to build and upload in a single step. The target cloud is defined by the image type (e.g. `server-ami` uploads to AWS).
```
lfauster, line 318: Should this here (and down below) say `generic-ami` (or the `ami` alias) instead of `server-ami`? I could be wrong, but I think `server-ami` does not exist as image type.

Same probably true for `server-vhd` and `server-openstack`.

supakeen: Yea, I used a (nonexistent) example image type since image types are different for all distributions, not a concrete one. Maybe I *should*.

supakeen:
I'm going to drop `.Pipeline.ExportName` from this documentation commit; I want to keep it undocumented as it's only used for an experimental feature that hasn't even been PR'ed yet and I'm not yet certain it will stay.

Since that means I'm squashing anyhow I'll also address @lfauster's remark.

supakeen:
> I'm going to drop `.Pipeline.ExportName` from this documentation commit; I want to keep it undocumented as it's only used for an experimental feature that hasn't even been PR'ed yet and I'm not yet certain it will stay.
> 
> Since that means I'm squashing anyhow I'll also address @lfauster's remark.

Done.

doc/01-usage.md
```diff
@@ -126,6 +126,48 @@ WARNING: using experimental cross-architecture building to build "s390x"
...
+$ ls *.osbuild-manifest.json
+fedora-43-server-qcow2-x86_64.osbuild-manifest.json
+```
+
+`--with-sbom` places an [SPDX](https://spdx.dev/) Software Bill of Materials document alongside the image.
+
+```console
+$ sudo image-builder build --with-sbom --distro fedora-43 server-qcow2
+# ...
+```
```
achilleas-k, line 149: No info on the naming of the sbom?

doc/01-usage.md
```diff
@@ -251,6 +241,21 @@ $ sudo image-builder build --bootc-ref quay.io/fedora/fedora-bootc:rawhide --boo
...
 ```
 
+### Cross-architecture builds
+
+> [!WARNING]
+> Cross-architecture building is an experimental feature.
+
+When passed `--arch` `image-builder` will try to build for a different architecture. Not all image types are available for all architectures.
+
+Cross-architecture builds are much slower than building on native hardware. However, if no native hardware is available they might be an acceptable compromise.
```
achilleas-k, line 251: I realise this is a pre-existing block but the language here is a bit awkward.  "When passed ..." doesn't flow very nicely.
How about:

> The architecture is automatically selected to match the host when building an image. This can be overridden with the `--arch` option. Building an image for a foreign architecture (i.e. an architecture different from the host machine) is much slower than building natively.

I think the part about it being an acceptable compromise is unnecessary in these kinds of docs, but we can keep it if you like it.

achilleas-k (approved):
Nice work.  Thank you.  LGTM.

Some comments below but wont block.

doc/01-usage.md
supakeen: Let's do follow-ups on these.

doc/01-usage.md
supakeen: Let's do follow-ups on these.

PR #2692, opened 2026-09-16
distrodefs/rhel-9: drop net.ifnames=0 from OCI images

achilleas-k:
In most of our cloud images, we're stuck with `net.ifnames=0` on RHEL 9 even though it's not recommended because we want to maintain compatibility during the major RHEL 9 lifecycle.  OCI images however are new, which means we can avoid this legacy config in this particular instance.

DRAFT: We're still waiting for approval to rebuild the OCI images before the release.  I'm opening this PR pre-emptively so that it's ready to merge when we get it.  It's possible this will not be merged and we will need to keep the option in OCI RHEL 9 images as well.


ondrejbudai (approved):

achilleas-k:
Decision: We wont be changing this now after all.  OCI will keep the `net.ifnames=0` option along with the other cloud images on RHEL 9.

PR #2699, opened 2026-09-17
cmd/image-builder: support multiple exports (HMS-11287)

supakeen:
If `.Exports()` were to ever return multiple exports instead of a single one then we should be writing out all of them. There are currently no image types that define multiple exports but they could.

This PR gives us the ability to export multiple pipelines from an image, while still using a single output template (both for `output-directory` and `output-name`) *and* being pluggable to other things wanting to select exports in the future.

The last bit is important here as while additional exports defined on an image type are *always* exported I do want the option to let users export 'extras' optionally.


supakeen:
You can test this locally with something like:

```
€ git diff
diff --git a/data/distrodefs/fedora/imagetypes.yaml b/data/distrodefs/fedora/imagetypes.yaml
index c58f13ff4..94275f450 100644
--- a/data/distrodefs/fedora/imagetypes.yaml
+++ b/data/distrodefs/fedora/imagetypes.yaml
@@ -1931,7 +1931,7 @@ image_types:
     bootable: true
     default_size: "2 GiB"
     image_func: "disk"
-    exports: ["xz"]
+    exports: ["xz", "image"]
     platforms:
       - <<: *x86_64_uefi_platform
€ go build ./cmd/image-builder 
€ sudo ./image-builder build --distro fedora-44 minimal-raw-xz
# ...
Image build successful: fedora-44-minimal-raw-xz-x86_64/fedora-44-minimal-raw-xz-x86_64.raw.xz
€ ls -larth fedora-44-minimal-raw-xz-x86_64 
total 2.1G
drwxr-xr-x. 20 user user 4.0K Sep 17 14:03 ..
-rw-r--r--.  1 root root 5.2G Sep 17 14:07 fedora-44-minimal-raw-xz-image-x86_64.img
-rw-r--r--.  1 root root 657M Sep 17 14:07 fedora-44-minimal-raw-xz-x86_64.raw.xz
drwxr-xr-x.  2 root root 4.0K Sep 17 14:08 .

```

supakeen:
I'm going to close this PR. Instead we're going to go with an explicit `extras` approach as it'd be super beneficial to have additional metadata for these exports.

Issue #2662, opened 2026-09-06
Experimental support for "systemd hermetic"` style systems


supakeen:
**TBD, this ticket is a work in progress and does not yet contain all the work (I'll fill out this ticket but it's good to have a ticket to attach PRs to).**

This ticket is to track the ability for `image-builder` to produce systems according to [1]. There isn't really a good name for this style of system so I'm going to call it "systemd hermetic" or such. It's a good idea to read that blog post to get a general idea of how the parts should fit together.

Our default shipped definitions will likely not include any systems built in this style for a little while but my personal TeamSBC Remix [2] is using these features as they become available and the testbed for this.

This ticket does not track work on the `systemd`-side nor the `osbuild`-side. For that there's a tracking ticket in TeamSBC [3] that tracks work on all its upstreams including `image-builder`, `systemd`, and Fedora.

The intent here is to be able define a Fedora-based system that builds into the necessary artifacts *without* doing things "out-of-packages". This means there is work necessary both in Fedora packaging and in `image-builder`.

Many of the things listed here are also things that are relevant to other image based systems (`ostree`, `bootc`, etc), or just in general, of interest to people building certain artifacts. In this issue I keep things coming from the "systemd" angle but relevant tickets for other systems are linked.

---

## `IMAGE_ID` and `IMAGE_VERSION` in `/etc/os-release`

When `systemd-sysupdate` is used it sources the version for the updates from `IMAGE_VERSION` in `os-release`. We should be able to write it at build time. The approach taken here is that the RPM that owns the `os-release` file (usually the relevant `-release` package) writes these additional variables when they are set in the environment so `image-builder` only has to set them while RPM installs things.

For an example implementation of a `-release` RPM that implements this see [5].

- [x] https://github.com/osbuild/image-builder/pull/2586
- [ ] #2663.

## `sysupdate` components

TBD

### Partition Exports

TBD

- [x] https://github.com/osbuild/image-builder/pull/2665
- [ ] https://github.com/osbuild/image-builder/pull/2677

### `sysext`s and `confext`s

`sysext`s (system extensions) and `confext`s (configuration extensions) are concepts in `systemd` that are overlays on a running system; they can be enabled, disabled, and updated and are bound to a given target system. This is not a full package manager since there's no inter-sysext relationships.

For this we need to be able to generate a tree (and package it up into an erofs, squashfs, disk image, etc) that is a delta to the tree for an image and write an `os-release` file into it that (partially) matches the image produced.

[...]

Issue #2632, opened 2026-08-31
separate raw images instead of single raw image with partitions (HMS-11261)


gmankab:
hi!

currently image-builder produces single `disk.raw` file with partitions inside

in [pocketblue](https://github.com/pocketblue/pocketblue) we use [split-partitions.sh](https://github.com/pocketblue/pocketblue/blob/705fc3febe2ae75f8281e1d33a989f76f9cf306d/tools/postprocess/split-partitions.sh) script which uses dd to extract partitions into separate `.raw` files

i was wondering if there are any plans to support exporting partitions directly as separate `.raw` files from image-builder

thanks for considering it!

supakeen:
Hey, yes there are plans and I have a local branch for it. We mostly need to figure out an OK user interface for defining which partitions to split.

gmankab:
@supakeen, something like this looks nice to me https://github.com/gmankab/pocketblue/commit/ed5ac7

supakeen:
> [@supakeen](https://github.com/supakeen), something like this looks nice to me [gmankab/pocketblue@ed5ac7](https://github.com/gmankab/pocketblue/commit/ed5ac7)

Very similar to what I have at the moment:

```yaml
extras:    
  split_parts:    
    - mountpoint: "/usr"    
      filename: "usr.raw.xz"    
      compression: "xz"    
```

Where I'm doing an "extras" for a given image type in the YAML (since I have more extras than just split partitions). I assume this is being done for the same reasons I have (`systemd-sysupdate`)?.

gmankab:
this looks nice for me too

onesaladleaf:
> I assume this is being done for the same reasons I have (systemd-sysupdate)?

Our usecase is that we need all partitions to be split so that they can be individually flashed to existing partitions on a device via fastboot

supakeen:
> > I assume this is being done for the same reasons I have (systemd-sysupdate)?
> 
> Our usecase is that we need all partitions to be split so that they can be individually flashed to existing partitions on a device via fastboot

Thanks, makes sense. I'll reference this ticket when I open my PR and/or see if I can split out the relevant commits if the other parts take longer.

gmankab:
hello, any news on this feature?

supakeen:
> hello, any news on this feature?

[...]

PR #2700, opened 2026-09-18
many: extras (user-selectable multiple explicit exports) (HMS-11288)

supakeen:
This PR introduces the concept of 'extras' to `image-builder` and its definitions. `extras` are specifically-typed additional artifacts with metadata that are part of a single image definition.

Some examples of extras:

1. Partitions to be exported separately from the main artifacts disk image (for example `/usr`, `/`, etc).
2. `sysexts` which are built against the main artifacts `os` pipeline.
3. Files to be copied out of the main artifacts `os` pipeline.
4. Copying out the UKI.
5. Pipelines as long as they are explicitly listed as an extra for a given image type.

This PR has gone through a lot of cycles and time mostly due to the fact that we wanted to be generic (and kinda implicit) by allowing arbitrary pipeline exports. In the end we always needed explicitness (multiple pipeline exports still need to be listed in YAML) and more explicitness allows for more metadata, hence the addition of an `extras`.

Extras pipelines are *always* in the manifest and not conditionally inserted; I consider them to be part of the build and thus they should be able to be built everywhere the manifest goes.

This PR comes with two concrete implementations of extras. `sysext` (#2662) and `partition` (#2632) but really they're just in a supporting role here for the idiom of extras.

I'm still not *especially* fond of some things but I think it's largely in a good enough state to land and iterate on without it blocking any potential use cases or breaking current ones.

Note: extras can practically export multiple pipelines, current ones do not. If we were to do that and those exports would have the same extension then we need to be careful as we can't currently disambiguate in the `output-name` template. We can introduce that once we cross that bridge. It's fine as long as we have, for example, a single extra export two pipelines: one that creates `partition.raw` and one that creates a `partition.raw.sig` or such.

---

A quick example of how this would look like. Firstly available extras are listed in the output of the `describe` command. The format is an 'extra reference' which is `$type:$name`.

```
€ image-builder describe qcow2
# ...
extras:
  - sysext:nginx
  - sysext:podman
  - partition:usr
# ...
```

When building the image type *normally* we get what we expect, however when we build we can enable additional extras to be produced. (In the future we will probably include either globbing for extras an 'all extras' flag).

```
€ image-builder build
    --with-extra sysext:nginx
    --with-extra sysext:podman
    --with-extra partition:usr
  qcow2
# ...
Image build successful: teamsbc-46-makalu-virt-x86_64/teamsbc-46-makalu-virt-x86_64.raw.xz
Image build successful: teamsbc-46-makalu-virt-x86_64/teamsbc-46-makalu-virt-nginx-x86_64.erofs
Image build successful: teamsbc-46-makalu-virt-x86_64/teamsbc-46-makalu-virt-podman-x86_64.erofs
Image build successful: teamsbc-46-makalu-virt-x86_64/teamsbc-46-makalu-virt-usr-x86_64.raw.xz
€ ls -larth teamsbc-46-makalu-virt-x86_64 
total 710M
drwxr-xr-x. 9 user user 4.0K Sep 18 08:42 ..
-rw-r--r--. 1 root root 472K Sep 18 08:42 teamsbc-46-makalu-virt-x86_64.osbuild-manifest.json
-rw-r--r--. 1 root root 270M Sep 18 08:45 teamsbc-46-makalu-virt-usr-x86_64.raw.xz
-rw-r--r--. 1 root root 351M Sep 18 08:45 teamsbc-46-makalu-virt-x86_64.raw.xz
-rw-r--r--. 1 root root 3.6M Sep 18 08:48 teamsbc-46-makalu-virt-nginx-x86_64.erofs
-rw-r--r--. 1 root root  86M Sep 18 08:50 teamsbc-46-makalu-virt-podman-x86_64.erofs
-rw-r--r--. 1 root root 114K Sep 18 08:50 teamsbc-46-makalu-virt-x86_64.buildlog
```

---

This doesn't yet implement extras for bootable container based systems. However this will be a follow-up. We'll do that by allowing bootable containers to supply an `extras.yaml` file in the well-known locations we use for other sources of configuration: https://osbuild.org/docs/developer-guide/projects/image-builder/advanced/bootc/sources-of-configuration/#filesystem


supakeen:
Though I didn't want this PR to be too much about the concrete extras themselves (more the concept of extras) we're bound to have some ideas/questions for them.

---

How should the partition splitting work for volume-based mountpoints (`btrfs`, `lvm`). I think it'd be OK to export the partition that the volume is *on* and I have a commit I can squash in that does just that.

Right now in this PR it fails if people define a `partition` extra that refers to a btrfs subvolume or such.

supakeen:
> Though I didn't want this PR to be too much about the concrete extras themselves (more the concept of extras) we're bound to have some ideas/questions for them.
> 
> How should the partition splitting work for volume-based mountpoints (`btrfs`, `lvm`). I think it'd be OK to export the partition that the volume is _on_ and I have a commit I can squash in that does just that.
> 
> Right now in this PR it fails if people define a `partition` extra that refers to a btrfs subvolume or such.

I've pushed my commit that does this (it's the last one).

cmd/image-builder/cmd.go
```diff
@@ -276,6 +276,7 @@ func setupBuildCmd() (*cobra.Command, error) {
 		SilenceUsage: true,
 		Args:         cobra.ExactArgs(1),
 	}
+	buildCmd.Flags().StringArray("with-extra", nil, `include an extra by type:name in the build (e.g. --with-extra sysext:nginx)`)
```
achilleas-k, line 279: I think instead of an example it would be more useful to mention where one can find the supported extras.  Pointing users towards using `describe` to get a list of supported extra names would be better.

achilleas-k (approved):
Nice work on all of this.  I'm very happy with how the new pipelines look.  Still not entirely happy with the way we define these, but we're not locked into that so we can see how it feels going forward and iterate.

Thank you.

LGTM!

Regarding the last commit (exporting volumes from inside containers): I agree that the behaviour in the PR is probably the best workaround for btrfs subvolumes, but I think for LVM logical volumes, it's possible to `dd` the logical volume separately instead of the whole partition (volume group), isn't it?

cmd/image-builder/cmd.go
supakeen: Agreed, but let's not back it out over this 🙂

PR #2701, opened 2026-09-18
Enable CUI OpenSCAP profile for RHEL 10 (HMS-11244)

kingsleyzissou:
## Summary

RHEL 10 currently rejects blueprints that request the CUI OpenSCAP profile because the profile is missing from its distro definition allowlist. Add the profile to the RHEL 10 allowlist so it can be used once supported by the SCAP Security Guide.


supakeen (approved):

lucasgarfield (approved):

kingsleyzissou:
Not sure why the linter is failing. This also doesn't even touch any go code

achilleas-k:
> Not sure why the linter is failing. This also doesn't even touch any go code

The PR's merge base is quite old (fe117f53b0bff262a6d45f2c27eb2707303f6f29).  Not sure exactly how that results in the linter failing, but it is complaining about missing a dependency that we've now dropped, so I'm sure that's related.

Clicking rebase.

achilleas-k (approved):

PR #2702, opened 2026-09-19
bootc: extras (partitions) (HMS-11288)

supakeen:
Note, this PR is rebased on top of #2700 which might still change so I'm keeping it as draft. Any discussion about the non-`bootc` parts and the general idea happens there.

---

This is the implementation of the "extras" from #2700 for bootable container based images. It introduces a new `extras.yaml` in the well-known configuration location, and wires up the necessary helpers and manifest generation for those exports to work.

These extras allow for the production of coupled artifacts when building an image. Specifically in this PR the ability to export partitions separately from the main disk image being produced. This is useful for for example PocketBlue (#2632) where for their Android devices they want the ESP and root partitions separately for flashing.

Usage would look like (for a container that has an `extras.yaml` looking like):

```yaml
partitions:
  usr:
    mountpoint: "/usr"
```

```console
$ image-builder build raw
    --bootc-ref localhost/my-container:latest
    --with-extra partition:usr
```

We only support `partition`-type extras for bootable containers (at least initially) as the `sysext`-type extras are tied to the package sets being installed, in the future we can probably inspect the container for installed packages; or default to always redo the entire transaction but I didn't want to make that decision for this PR.


supakeen:
The #2700 branch grew a commit to work with mountpoints inside container type filesystems so I've rebased this one.

supakeen:
This PR has been rebased on top of #2700 and I've merged the variant support in as well so it's ready to be undrafted.

pkg/distro/defs/bootc.go
```diff
@@ -169,6 +169,16 @@ func NewBootcWithLoader(loader *Loader, name string, cinfo *bootc.Info) (*BootcD
...
+		if osInfo.ExtrasInfo.Partitions != nil && imgTypeYaml.Extras.Partitions == nil {
+			imgTypeYaml.Extras.Partitions = make(map[string]partitionDef)
+		}
+		for name, p := range osInfo.ExtrasInfo.Partitions {
+			imgTypeYaml.Extras.Partitions[name] = partitionDef{
+				Mountpoint:  p.Mountpoint,
+				Filename:    p.Filename,
+				Compression: p.Compression,
+			}
+		}
```
lfauster, line 181: If I read it right, we're adding the partition extra here onto every bootc image type. But the actual pipeline is only created for disk image types in `manifestForDisk`, which I suppose makes sense. 
If that's the case, could that cause problems somewhere?
Should the combination of partition extra and non-disk types maybe throw an error?

lfauster:
LGTM, with one small nitty question/finding.

pkg/distro/defs/bootc.go
supakeen: It's a good question, we don't really have a nice way of scoping down extras only to specific image types at the moment. Some of the *other* extras might in the future work for non-disk image types.

We don't have this issue on the non-`bootc` side of things as there they are defined directly on the image type but in the `bootc` world the container can be any image type.

For me personally this isn't a big enough issue to include a bunch of filtering (containers that can be turned into ISOs are already quite specific and probably wouldn't include an `extras.yaml` (or `disk.yaml`) [1]) but let's see how others feel about this too (cc @ondrejbudai / @achilleas-k).

We could opt for the simple approach for now as well and deny *any* extras in `manifestForIso`.

[1]: https://osbuild.org/docs/developer-guide/projects/image-builder/advanced/bootc/isos/#generic

brlane-rht (approved):
Looks good.

